PR #312, opened 2014-04-30
Changes to better support accessibility via ARIA

### LINKED ISSUE #304: ARIA landmarks are not quite correct
In a PFWG Editor's meeting today, we reviewed the aria landmarks that ReSpec is inserting into documents.  There are some additional landmarks that could be automatically generated (e.g., a role of navigation on the table of contents), and some that are not quite right (e.g. a role of heading on an h\* element is not strictly necessary).

Definition lists can have role of definition on the defining text, linked to the things they define - in conjunction with dfn elements anyway.  Also, inline dfn elements can have their enclosing element be used as a definition.

halindrome:
These changes should close #304


marcoscaceres:
Closing this old PR

PR #359, opened 2014-11-11
Implemented support for dfn aliases and scoping

halindrome:
Added support for data-title='foo|bar|bat' style aliases for
definitions.
Added support for data-local-title='s|p|t' style local aliases for
definitions.
Added support for data-dfn-type to specify the scope of a
definition or group of definitions.
Added support for export and noexport to flag that a definition should
be made available outside of the document being generated.

Note that these changes are all backward compatible.  Current documents
should see no changes in their output.

If you start using the new data-dfn-type to define the type of a
definition or of a reference, it will cause the typeName to be included
in the ID / IDREF.  Otherwise there are no current effects.

### DIFF
--- a/examples/getusermedia.html
+++ b/examples/getusermedia.html
@@ -4,7 +4,6 @@
 <head>
   <meta name="generator" content=
   "HTML Tidy for HTML5 (experimental) for Linux https://github.com/w3c/tidy-html5/tree/68a9e74">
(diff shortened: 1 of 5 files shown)


marcoscaceres:
@halindrome we don't need this PR anymore, right?

halindrome:
Overcome by events.

PR #363, opened 2014-11-26
Fixed test warnings

halindrome:
When run with TRACE=1, many many tests emitted errors or warnings
that could color the results of testing or result in false
positives.  As it turned out, none of them actually did but it was
still worth fixing. Now the TRACE output does not contain spurious
warnings.


marcoscaceres:
@halindrome, can we close this PR from last year too?

halindrome:
I guessw we can close this.  If I care enough I will do it again.

PR #373, opened 2014-12-13
Add the ability to process contiguous WebIDL blocks

### LINKED ISSUE #334: noLegacyStyle removes member documentation entirely
Compare http://whatwg.github.io/serial/?noLegacyStyle=false#h2_serialoptions-dictionary with http://whatwg.github.io/serial/?noLegacyStyle=true#h2_serialoptions-dictionary. The former has documentation for the attributes, while the latter doesn't.

jyasskin:
@darobin

Resolves #334, which also has more discussion of the idea.

### DIFF
--- a/js/core/dfn.js
+++ b/js/core/dfn.js
@@ -1,6 +1,6 @@
 
 // Module core/dfn
-// Finds all <dfn> elements, gives them IDs, and populates conf.definitionMap to identify them.
+// Finds all <dfn> elements and populates conf.definitionMap to identify them.
 define(
     [],
     function () {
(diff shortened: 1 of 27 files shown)


jyasskin:
Here's a version that lets spec authors write something closer to Anolis, although I haven't implemented it as systematically as I'd like. It's missing some abbreviations and disambiguation by type, and it lower-cases IDs, while I think Anolis may leave them CamelCased.

darobin:
Thanks a *lot* @jyasskin!

I'm leaving off on a trip, but I think it would be great to make a release with this addition as well as @halindrome's PR.

Issue #389, opened 2015-01-14
Permalinks get messed up "content" attribute if heading text has special characters


michael-n-cooper:
Example: https://rawgit.com/w3c/coga/master/user-research/index.html

There are headings in this document that have apostrophes in them, e.g., "Description of Alzheimer's". Heading text is used to populate various attributes including id and title. In most cases, the apostrophe is converted to a hyphen, so for instance, you get id="h-description-of-alzheimer-s". However, for the permalink there is a that has messed up output for this attribute which messes up parsers:

content="3.4.2.2 Description of Less Common (Non-Alzheimer" s)'=""

which is invalid markup. What should happen is:

content="3.4.2.2 Description of Less Common (Non-Alzheimer's)"

(preferred), or

content="3.4.2.2 Description of Less Common (Non-Alzheimer-s)"

if using the same "any character I don't know becomes a hyphen" algorithm.

I think this might be in the RDFa part of Respec, but not sure, I'm not that wise about output.

michael-n-cooper:
Sorry for submitting empty description on first attempt, I pressed "enter" after entering the title and it submitted rather than move to the description field as I expected.

Example: https://rawgit.com/w3c/coga/master/user-research/index.html

There are headings in this document that have apostrophes in them, e.g., "Description of Alzheimer's". Heading text is used to populate various attributes including id and title. In most cases, the apostrophe is converted to a hyphen, so for instance, you get id="h-description-of-alzheimer-s". However, for the permalink there is a <span> that has messed up output for this attribute which messes up parsers:

content="3.4.2.2 Description of Less Common (Non-Alzheimer" s)'=""

which is invalid markup. What should happen is:

content="3.4.2.2 Description of Less Common (Non-Alzheimer's)"

(preferred), or

content="3.4.2.2 Description of Less Common (Non-Alzheimer-s)"

if using the same "any character I don't know becomes a hyphen" algorithm.

I think this might be in the RDFa part of Respec, but not sure, I'm not that wise about output.

halindrome:
I will look at this shortly.

darobin:
This was fixed by #391.

PR #391, opened 2015-01-14
Ensure that apostrophes in headers are escaped

halindrome:
Issue #389 reported that the content attribute for permalinks was
being incorrectly populated with embedded apostrophes.  This change
escapes those.

### DIFF
--- a/js/w3c/permalinks.js
+++ b/js/w3c/permalinks.js
@@ -53,9 +53,11 @@ define(
                                 if (!conf.permalinkEdge) {
                                     content += "&nbsp;";
                                 }
+                                var ctext = $item.text().replace("'", "&apos;") ;
+
                                 content += "<a href='#"+resourceID+"' " + urlprop + "aria-label='Permalink for "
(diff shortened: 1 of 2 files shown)


darobin:
Is there a good reason why this content is generated using string concatenation instead of jQuery? It would be a lot safer (and easier to read too).

halindrome:
No.  No reason.  It just was.  I think what I hear you suggesting is creating DOM nodes via jQuery and then inserting them into the document.  Is that what you mean?

darobin:
Yes, I really think it's best practice except in a few well-controlled cases. But using concatenation along with user-provided content is likely a bad idea!

halindrome:
Sure.  Obviously it has been this way for a while, but I did it initially so... I will effect repairs.

halindrome:
@darobin are you okay with this now?

darobin:
Apart from the minor nit with replace() this LGTM. Feel free to merge after the correction.

Issue #314, opened 2014-06-07
How do I get specref references when using respec2html.js


silviapfeiffer:
When I create my spec using phantomjs, it's missing the references:

  $ phantomjs respec/tools/respec2html.js index.html Overview.html

When I just load the page, the respec references are included.

I wasn't able to find documentation of how to get them when using respec2html.js .

dontcallmedom:
not yet; I'm trying to get an instance of specrefs hosted on https://x.w3.org, but this is still in progress

dontcallmedom:
hmm... this was fixed long ago, closing

PR #392, opened 2015-01-23
use labs.w3.org url for specrefs; see #314

dontcallmedom:


darobin:
I presume that's a proxy to the real thing?

dontcallmedom:
it is a proxy; at some point, we might want to replicate the whole service, but I think a proxy is good enough for now.

PR #401, opened 2015-02-15
Update ReSpec to heroku URL.

tobie:
As nodejitsu's been acquihired, they're closing down their service in about half a year, so went ahead and mirrored this on heroku until I figure out how to properly send HTTPS over a full domain name.


marcoscaceres:
LGTM

Issue #361, opened 2014-11-25
Support maplike declarations


ddorwin:
ReSpec fails with many errors and warnings when using the new maplike declaration (http://heycam.github.io/webidl/#idl-maplike) in an interface. The same is true for setlike.

For example:
  `<dt>readonly maplike &ltlong, DOMString&gt</dt>`
  `<dt>maplike &ltlong, DOMString&gt</dt>`

At a minimum, the above syntax should add the declaration in the interface's IDL. A bonus would be to provide information about the additional methods.

yutakahirano:
http://webaudio.github.io/web-midi-api/ plans to use maplike, so I'm trying add the functionality.

ddorwin:
@yutakahirano, Thank you for adding this!

While trying to use it, I noticed that a subsection named "Maplike" with the text from the `<dd>` element is added. The tests at https://github.com/w3c/respec/blob/develop/tests/spec/core/webidl.html demonstrate this. That text should probably be ignored or inserted into the main section.

I would also be nice to add something like the following from the EME spec:
>The readonly maplike declaration adds `entries`, `forEach`, `get`, `has`, `keys`, `values`, `@@iterator` methods and a size getter.

A nit: In the tests mentioned above, the referenced "hoge" and "fuga" are not used in the actual example.

yutakahirano:
Thank you for the feedback.

> That text should probably be ignored or...

I need it, because I want to describe the key value of a maplike<DOMString, V>, as in the example. Just DOMString is not enough.

> ...inserted into the main section.

I created a [branch] (https://github.com/yutakahirano/respec/tree/maplike-update) to update the description. What do you think about it?

> I would also be nice to add something like the following from the EME spec:

>> The readonly maplike declaration adds entries, forEach, get, has, keys, values, @@iterator methods and a size getter.

I added the description in the branch but I'm not sure if it should be inserted automatically or by hand (by each spec author).

ddorwin:
Thanks. The branch looks good, though I'd have to see the output to be sure.

I think it makes sense to automatically add the text since it will always be accurate and doing so ensures specs are consistent and correct. Are there cases where you think the author might want to modify it?

yutakahirano:
> I think it makes sense to automatically add the text since it will always be accurate and doing so ensures specs are consistent and correct. Are there cases where you think the author might want to modify it?

OK, then let's generate the description.
[...]

PR #403, opened 2015-02-23
Cleanup write as html

yutakahirano:
WebIDLProcessor.prototype.writeAsHTML is very long. This change split the function, to make it easy to read and modify.
This is a preliminary change for https://github.com/w3c/respec/issues/361.

### DIFF
--- a/js/core/webidl-oldschool.js
+++ b/js/core/webidl-oldschool.js
@@ -678,6 +678,214 @@ define(
                 return { type: matched[1], parameter: matched[2] };
             },
 
+            writeInterfaceAsHTML: function (obj) {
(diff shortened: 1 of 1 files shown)


yutakahirano:
@darobin, can you take a look at this change?

I have another [branch] (https://github.com/yutakahirano/respec/tree/maplike) depending on this change ([diff] (https://github.com/yutakahirano/respec/compare/yutakahirano:cleanup-write-as-html...maplike)) to introduce maplike. I would create a pull-request from that branch once this change is accepted or rejected.

Thanks,

darobin:
Thanks!

PR #408, opened 2015-03-03
Add maplike support.

yutakahirano:
This change add support for "[readonly] maplike<K, V>;" in an interface as
the grammar is recently added to WebIDL. #361


darobin:
Great, thanks! Making a release now.

darobin:
v3.2.44 now supports this!

PR #417, opened 2015-03-28
Add UI to show a list of definitions.

gkellogg:
This is a simplistic implementation to add a UI component to show a list of definitions from the document.

I suspect that there will be scrolling issues in the modal, and it would be nice to link to the source of the definition, but this likely requires some magic I'm not familiar with to get out of the model and relocate to the definition.

I've often lost track of all of the definitions in a document, or need a way to find definitions in some other companion document for which I might not be an editor, so having a UI to provide a quick reference to definitions may be more generally useful. Of course, there could also be a "Table of Definitions" which could be added to the document, similar to a table of figures or the table of contents, but this would not be commonly used, as it's mostly useful for spec writers, not end users.

### DIFF
--- a/js/profile-w3c-common.js
+++ b/js/profile-w3c-common.js
@@ -58,6 +58,7 @@ define([
             domReady(function () {
                 ui.addCommand("Save Snapshot", "ui/save-html", "Ctrl+Shift+Alt+S");
                 ui.addCommand("About ReSpec", "ui/about-respec", "Ctrl+Shift+Alt+A");
+                ui.addCommand("Definition List", "ui/dfn-list", "Ctrl+Shift+Alt+D");
                 ui.addCommand("Search Specref DB", "ui/search-specref", "Ctrl+Shift+Alt+space");
                 runner.runAll(args);
             });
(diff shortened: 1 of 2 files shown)


darobin:
This is a good starting point but I'm not sure it is mature enough to warrant inclusion.

The `definitionMap` holds elements that ought to have an ID. You can use that to link.  See code in `core/dfn` and `core/link-to-dfn` to see how they're used. You can then listen to clicks on the links and when clicked simply call `closeModal()` to dismiss the modal.

Scrolling should just work: the modal normally limits itself to not grow beyond the fold, and then overflows inside its own boundaries. Just try searching for something with lots of hits for instance to see the effect.

PS: you should change the comment at the top as well!

gkellogg:
Thanks. I did use the dfn content to get a link before, but didn't do try to catch and close the modal. I'll gove that a go.

gkellogg:
Thanks @darobin, I made the suggested changes; simpler than I had realized on the first pass.

gkellogg:
(CI Failure seems completely unrelated to this commit, FWIW).

Issue #419, opened 2015-04-01
fix-headers incorrectly messes with non-section headers


halindrome:
If a document has a bunch of h* elements within a section, where these elements are not ALL the same (e.g., the section heading is an h3 and all of the headings within the section are h4), this module will incorrectly match on the subheadings and tweak them.  Basically, the CSS selector is naive. Thanks @ianbjacobs for pointing this out.

darobin:
The issue here is that you shouldn't use `hN` that aren't section titles. I guess ReSpec should actually flag that as an error.

halindrome:
I disagree strongly.  There are lots of reasons to use headers.

marcoscaceres:
> I disagree strongly. There are lots of reasons to use headers.

Can you provide some examples?

halindrome:
Sure.  But more importantly, ReSpec shouldn't be proscriptive.  If an author wants to use <h5> to indicate that something is a header, why would we prevent them?

In today's crisis, the Web Payments interest group has a number of use cases.  The major headings that go into the ToC are of course hN elements and ReSpec correctly numbers them.  Within these use cases there are sub items that are NOT "sections". However, the editors have chosen to identify the various sub-cases with hN elements for the headers.  

This resulted in styling challenges because ReSpec was munging the headers randomly.

See https://dvcs.w3.org/hg/webpayments/raw-file/default/latest/use-cases/index.html#cryptocurrency-payment-bitcoin-ripple

darobin:
What is the header a header of if not of a section? Using headers without sections only relies on the outline algorithm, which really, really isn't great. ReSpec is prescriptive whenever there's a good practice.

halindrome:
I am confused.  What outline algorithm? 

And regardless of what authors should be doing, the behavior ReSpec currently exhibits is clearly incorrect.  It shouldn't mess with some headers and not with others.  That doesn't make any sense.

darobin:
The algorithm that outlines an HTML document.

It's only processing those headers differently because they are misplaced. If you put your abstract at the end it will also do strange things!

halindrome:
Not sure how that is relevant.  But again, regardless, I think that respec should either touch ALL of the headers within a section or only the first one, but not some random number that is based upon what hN elements the author happened to use in the source.  The fix I have provided has respec adjust the level of the first item, which is clearly what the code was trying to do originally.

msporny:
[...]

PR #420, opened 2015-04-01
Changed header numbering logic

halindrome:
Only the first heading within a section should be
renumbered to match depth.

Also, fix-headers.js and structure.js were doing the same
thing to the H elements.  I removed the duplicate code from
structure.js since it was redundant.

This fixes Issue #419.


marcoscaceres:
@halindrome what do you want to do about this one?

halindrome:
I think it should be merged.  @darobin disagrees because headers without a section are somehow bad.

marcoscaceres:
closing this, not because I don't ❤️ ... but because bit rot.

PR #421, opened 2015-04-02
Feature/notoc

### LINKED ISSUE #419: fix-headers incorrectly messes with non-section headers

gkellogg:
This adds support for the class "notoc" on a section to cause the section to be omitted from the TOC.
Fixes #419.

### DIFF
--- a/js/core/structure.js
+++ b/js/core/structure.js
@@ -28,9 +28,10 @@ define(
                 var $ul = $("<ul class='toc'></ul>");
                 for (var i = 0; i < $secs.length; i++) {
                     var $sec = $($secs[i], doc)
-                    ,   isIntro = $sec.hasClass("introductory") || $sec.hasClass("notoc")
+                    ,   isIntro = $sec.hasClass("introductory")
+                    ,   noToc = $sec.hasClass("notoc")
                     ;
-                    if (!$sec.children().length) continue;
+                    if (!$sec.children().length || noToc) continue;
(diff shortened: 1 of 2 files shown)


halindrome:
I don't mind this, nor that it was merged.  However, for a11y purposes it may be the case that headers that are NOT in the ToC are also headers that should not be 'landmarks' from an ARIA perspective.  I will raise this with the PFWG since they are our experts on the topic.  @michael-n-cooper can you please put this on the agenda?

michael-n-cooper:
My personal opinion is these sections should still be landmarks even if 
it the author prefers not to include them in the TOC. But that's just a 
personal opinion. I have put this on the agenda for the next PF editors 
meeting. Michael

halindrome:
Thanks. On a related note, we probably shouldn't be using headers for the
role / state /property definitions in aria core since these are not really
sections... Also for the editors to hash out.

michael-n-cooper:
PF editors discussed this and said don´t exclude sections from being landmarks just because they have been exluded from TOC. I think that means no change from current behaviour. Tangent issues are on future PF editors agenda but probably don't need to be tracked in this issue.

PR #428, opened 2015-04-22
Support required dictionary members in contiguous-idl mode.

jyasskin:
This imports webidl2.js from darobin/webidl2.js#25, which hasn't been merged yet. We should probably wait to merge this PR until the current PRs from that repo have been accepted, and then re-import webidl2.js from there.

Also note @dontcallmedom's https://github.com/dontcallmedom/respec/commit/3bbe54f9adac0d72b962c836908f64901c80e9b9, which fixes this in oldschool mode. Since oldschool mode isn't deprecated yet, we should probably take both.


darobin:
I just shipped a new build. Thanks to you both!

PR #430, opened 2015-04-28
allow to filter tests run based on spec name

dontcallmedom:
e.g. node ./tests/headless.js webidl will only run tests whose spec name contains "webidl"


halindrome:
LGTM.  However, it would be *nice* if it were possible to select multiple specs - sometimes I put new tests in more than one.

marcoscaceres:
The filtering code in this could be improved, but I think it's ok.

PR #432, opened 2015-04-29
Avoid cloning nodes in issues and notes.

jyasskin:
Cloning the nodes breaks references in conf.definitionMap, so the <dfn>s that
link-to-dfn.js assigns IDs weren't the ones that actually wind up in the result
document.


marcoscaceres:
LGTM

PR #433, opened 2015-05-01
Improve linking to and around WebIDL terms

jyasskin:
* `[link-for]` used to break non-IDL links in its scope.
* It was difficult to link to members when their interface wasn't immediately obvious. Now `<a>Interface.member</a>` works.
* Now links point to declarations if no definition is available.
* Any `<code>` definition now causes its links to also be `<code>`.

This builds on #432.

### DIFF
--- a/js/core/link-to-dfn.js
+++ b/js/core/link-to-dfn.js
@@ -34,7 +34,12 @@ define(
                     var title = $ant.dfnTitle();
                     var link_for = ($ant.attr("for") || $ant.closest("[link-for]").attr("link-for") || "").toLowerCase();
                     if (titles[title] && titles[title][link_for]) {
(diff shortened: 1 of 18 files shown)


marcoscaceres:
@jyasskin, sorry, can you rebase please.

darobin:
Yup, #432 was merged but this still can't merge automatically. @jyasskin do you mind fixing?

jyasskin:
Yep, sorry for the delay. I should get to this today. I may also have found a bug in it, which I need to investigate before merging.

jyasskin:
Here's the rebase with 2 bug fixes.

darobin:
Looks excellent, thanks!

PR #435, opened 2015-05-03
Use promises to yield between phases

jyasskin:
This makes browser profiles more meaningful, since each plugin doesn't appear to include the entire runtime of the next plugin.


js/core/base-runner.js
```diff
@@ -98,43 +98,49 @@ define(
...
                 respecEvents.pub("start", "core/base-runner");
 
+                if (respecConfig.preProcess) {
+                    for (var i = 0; i < respecConfig.preProcess.length; i++) {
+                        try { respecConfig.preProcess[i].apply(this); }
+                        catch (e) { respecEvents.pub("error", e); }
+                    }
+                }
+
+                var pipeline = Promise.resolve();
```
marcoscaceres: We are sure all users of ReSpec are using Promise capable browsers, right? I have not idea if Internet Explorer latest stable supports them... if yes, let's merge this. We can always ask on the mailing list.

marcoscaceres: (or does "Promise.min" include some kind of pollyfill?)

marcoscaceres: NM, I see the commit message answered my question :)

Issue #436, opened 2015-05-06
There should be a way to create an Editor's Note


halindrome:
Right now there are Note, Issue, and Warning classes.  There should be an ednote class that inserts an editor's note into the document.

gkellogg:
+1, I've often thought of this.

Also, when the annotations WG comes along, allowing generalized annotations of editors drafts, perhaps with hooks into GitHub issues list? (just dreaming now)

halindrome:
Yeah...  I like that.  Next time!

On Wed, May 6, 2015 at 2:34 PM, Gregg Kellogg <notifications@github.com>
wrote:

> +1, I've often thought of this.
>
> Also, when the annotations WG comes along, allowing generalized
> annotations of editors drafts, perhaps with hooks into GitHub issues list?
> (just dreaming now)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/w3c/respec/issues/436#issuecomment-99582017>.
>



-- 
Shane McCarron
halindrome@gmail.com

halindrome:
This was addressed in pull request #437

PR #437, opened 2015-05-07
Added support for class ednote

halindrome:
Creates an Editor's Note that is styled like a Note.

Addresses issue #436


marcoscaceres:
LGTM.

PR #438, opened 2015-05-11
Fixed header pollution from permalinks

halindrome:
hN elements were having permalink data included within them.  This made
them less accessible for people with screen readers.

In addition, the permalink label / title was such that it was difficult
to use assistive technology and keyboard shortcuts to navigate the links
to find a specific section.

The changes in this branch address the issues raised in #425

### DIFF
--- a/js/w3c/permalinks.js
+++ b/js/w3c/permalinks.js
@@ -45,16 +45,18 @@ define(
                             // if we still have resourceID
                             if (resourceID != null) {
                                 // we have an id.  add a permalink
+                                $item.addClass("permalink");
+                                $item.wrap("<div></div>") ;
                                 // right after the h* element
-                                var theNode = $("<span></span>");
+                                var theNode = $("<p></p>");
                                 theNode.attr('class', 'permalink');
                                 if (conf.doRDFa) theNode.attr('typeof', 'bookmark');
                                 var ctext = $item.text();
                                 var el = $("<a></a>");
                                 el.attr({
                                     href:         '#' + resourceID,
-                                    'aria-label': 'Permalink for ' + ctext,
-                                    title:        'Permalink for ' + ctext });
+                                    'aria-label': ctext + " permalink",
+                                    title:        ctext + " permalink" });
                                 if (conf.doRDFa) el.attr('property', 'url');
                                 var sym = $("<span></span>");
                                 if (conf.doRDFa) {
@@ -66,14 +68,7 @@ define(
                                 el.append(sym);
                                 theNode.append(el);
 
-                                // if this is not being put at
-                                // page edge, then separate it
-                                // from the heading with a
-                                // non-breaking space
-                                if (!conf.permalinkEdge) {
-                                   $item.append("&nbsp;");
-                                }
-                                $item.append(theNode);
+                                $item.parent().append(theNode);
                             }
                         }
                     });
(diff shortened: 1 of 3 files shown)


marcoscaceres:
needs rebase. Can I be a pain and ask for either a screenshot or an example of the markup after this patch is applied? Just to get a better idea of the change.

halindrome:
That's not unreasonable at all.  Basically, we are converting the hN so that it has a permalink associated with it but they are on the same display line in the spec, so we are changing:

`<hN>My heading</hN>`

to:

`<div><hN class='permalink'>My heading</hN><p class='permalink'>permaLink stuff</p></div>`

Then there is a style definition for permalink that does the display:inline stuff.

darobin:
That breaks assumptions about `hN` being a direct child of a `section`. Bikeshed specs have solved this differently.

halindrome:
I don't believe we hace any such assumptions.  Nor does the HTML outline algorithm seem to require this.  What am I missing?

halindrome:
Oh - unless you mean assumptions that ReSpec itself might make.  The permalink module runs very late in the sequence, so that shouldn't be an issue for ReSpec.

halindrome:
Actually, I looked into it.  You are correct that, for example, the id-headers.js component expects the immediate parent if an hN element to be a section element.  Which is a little silly, but whatever.  

I don't know what bikeshed might have done, but CSS Cascading and Inheritance Level 4, for example, embeds their permalinks within the hN elements.  Which is EXACTLY what we are trying to avoid. 

Can you point me to an example of a solution that does it some other way?

halindrome:
I had some PFWG people look at the bikeshed approach, and I got the following responses:

> Shane, I only looked at the surface of this with a screen reader ... I didn't dig any deeper; not sure if you wanted that.
> The css spec has the permalink outside the heading, which is good.
> The css spec has accessible lable of "section" on every permalink, so that is not good.
> I like what you have done so that all permalinks have a unique accessible label -- conforms to wcag that way too.

I then asked for some clarification about how the CSS spec was being "rendered" by the AT and got:

> I found that I had Firefox 31 running when I looked before. But, as I test this time, both ff 31 and 38 are behaving very similarly with JAWS -- the permalink is being read as part of the heading.
>
> Before, I didn't look with IE. In IE, the permalink is part of the heading, and very ugly. JAWS sees it as, for example,
> WD-css-cascade-4-20150421/#cascading
> I can't copy it from the virtual buffer in IE; I have to get it from speech history. I think this may be an effect of the css.
>
> So, this rendering is all around much worse than what we have now in aria spec.

I am going to file some issues against bikeshed related to this.  But in short I don't think their approach is something that is going to address the A11Y issues with permalinks in ReSpec.  

I would be completely open to any suggestion that gets the permalink content outside of the header, but still renders on the same 'line' for visual users and keeps the hN element as an immediate child of a section element.  @darobin do you have any suggestions for how to do that?

halindrome:
Ping @darobin - see the commend above.  Do you have a suggestion?  Otherwise I would just like to merge this as I think it solves the commentor's problem.

darobin:
I would like to avoid using a `div`, there really are assumptions littered in various places that assume that's not the case. (Silly or not, it'll break stuff.)

Could this not work as a span that (in the tree) is after the `hN` but is positioned properly with negative margin?

marcoscaceres:
Closing this, not because I don't ❤️ ... but because bit rot.

PR #445, opened 2015-06-04
L10N support + Simplified Chinese

darobin:
Support localising string that are produced by the system, include Chinese (Simplified) translations, and start making use of them.

We have Chinese users for this.

### DIFF
--- a/js/w3c/l10n.js
+++ b/js/w3c/l10n.js
@@ -19,19 +19,29 @@ define(
                 ,   toc:                        "Table of Contents"
                 ,   note:                       "Note"
                 ,   fig:                        "Fig. "
+                ,   bug_tracker:                "Bug tracker:"
+                ,   file_a_bug:                 "file a bug"
+                ,   open_bugs:                  "open bugs"
+                ,   open_parens:                "("
+                ,   close_parens:               ")"
             }
             ,   zh: {
                     this_version:               "本版本："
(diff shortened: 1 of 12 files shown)


marcoscaceres:
This looks good, @darobin.

js/w3c/l10n.js
```diff
@@ -0,0 +1,60 @@
...
+                ,   abstract:                   "摘要"
+                ,   sotd:                       "关于本文档"
+                ,   status_at_publication:      "本章节描述了本文档的发布状态。其它更新版本可能会覆盖本文档。W3C的文档列 表和最新版本可通过<a href='http://www.w3.org/TR/'>W3C技术报告</a>索引访问。"
+                ,   toc:                        "内容大纲"
+                ,   note:                       "注"
+                ,   fig:                        "圖"
+                ,   bug_tracker:                "错误跟踪："
+                ,   file_a_bug:                 "反馈错误"
+                ,   open_bugs:                  "修正中的错误"
+                ,   open_parens:                "（"
```
marcoscaceres: Why are extra spaces needed for zh?

darobin: They're not extra spaces :) It's a full-width parenthesis. It just looks like there's an extra space. Using the locale's correct punctuation is required.

marcoscaceres: ok, I wasn't sure. Thanks for clarifying :)

PR #447, opened 2015-06-08
This module simplifies acknowledging contributions made through GitHub.

tobie:
Fetches names of contributors from github and uses them to fill
in the content of elements with key identifiers:
`#gh-commenters`: people having contributed comments to issues.
`#gh-contributors`: people whose PR have been merged.
Spec editors get filtered out automatically.

/cc @marcoscaceres

### DIFF
--- a/js/core/contrib.js
+++ b/js/core/contrib.js
@@ -71,8 +71,6 @@ define(
                 }
 
                 msg.pub("start", "core/contrib");
-                var $commenters = $(doc).find("#gh-commenters")[0];
-                var $contributors = $(doc).find("#gh-contributors")[0];
                 if (conf.githubAPI && ($commenters || $contributors)) {
                     github.fetch(conf.githubAPI).then(function (json) {
                         return $.when(
@@ -90,6 +88,8 @@ define(
                         });
                     }).then(theEnd);
                 } else {
+                var $commenters = doc.querySelector("#gh-commenters");
+                var $contributors = doc.querySelector("#gh-contributors");
                     theEnd();
                 }
             }
--- a/js/core/contrib.js
+++ b/js/core/contrib.js
@@ -1,0 +1,98 @@
+// Module core/contrib
+// Fetches names of contributors from github and uses them to fill
+// in the content of elements with key identifiers:
+// #gh-commenters: people having contributed comments to issues.
+// #gh-contributors: people whose PR have been merged.
+// Spec editors get filtered out automatically.
+
+define(
+    ["github"],
+    function (github) {
+        return {
+            run: function (conf, doc, cb, msg) {
+                function theEnd () {
+                    msg.pub("end", "core/contrib");
+                    cb();
+                }
+                
+                function prop(prop) {
+                    return function (o) {
+                        return o[prop];
+                    };
+                }
+                
+                function slice(args) {
+                    return Array.prototype.slice.call(args, 0)
+                }
+
+                function findUsers() {
+                    var users = {};
+                    slice(arguments).forEach(function (things) {
+                        things.forEach(function (thing) {
+                            if (thing.user) {
+                                users[thing.user.url] = true;
+                            }
+                        });
+                    });
+                    return Object.keys(users);
+                }  
+
+                function join(things) {
+                    if (!things.length) {
+                        return "";
+                    }
+                    things = things.slice(0);
+                    var last = things.pop();
+                    var length = things.length;
+                    if (length === 0) {
+                        return last;
+                    }
+                    if (length === 1) {
+                        return things[0] + " and " + last;
+                    }
+                    return things.join(", ") + ", and " + last;
+                }
+
+                function toHTML(urls, editors, element) {
+                    return $.when.apply($, urls.map(function (url) {
+                        return github.fetch(url);
+                    })).then(function () {
+                        var names = slice(arguments).map(function (user) {
+                            user = user[0];
+                            return user.name || user.login;
+                        }).filter(function (name) {
+                            return editors.indexOf(name) < 0;
+                        });
+                        names.sort(function (a, b) {
+                            return a.toLowerCase().localeCompare(b.toLowerCase());
+                        });
+                        $(element).html(join(names)).attr("id", null);
+                    });
+                }
+
+                msg.pub("start", "core/contrib");
+                var $commenters = $(doc).find("#gh-commenters")[0];
+                var $contributors = $(doc).find("#gh-contributors")[0];
+                if (conf.githubAPI && ($commenters || $contributors)) {
+                    github.fetch(conf.githubAPI).then(function (json) {
+                        return $.when(
+                            github.fetchIndex(json.issues_url),
+                            github.fetchIndex(json.issue_comment_url),
+                            github.fetchIndex(json.contributors_url)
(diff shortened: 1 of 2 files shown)


js/core/contrib.js
```diff
@@ -0,0 +1,98 @@
...
+                        return github.fetch(url);
+                    })).then(function () {
+                        var names = slice(arguments).map(function (user) {
+                            user = user[0];
+                            return user.name || user.login;
+                        }).filter(function (name) {
+                            return editors.indexOf(name) < 0;
+                        });
+                        names.sort(function (a, b) {
+                            return a.toLowerCase().localeCompare(b.toLowerCase());
```
marcoscaceres: nice touch with localeCompare :)

js/core/contrib.js
```diff
@@ -0,0 +1,98 @@
...
+                        });
+                        $(element).html(join(names)).attr("id", null);
+                    });
+                }
+
+                msg.pub("start", "core/contrib");
+                var $commenters = $(doc).find("#gh-commenters")[0];
```
marcoscaceres: Maybe we should stop relying on jQuery here? a simple querySelector would do.

js/core/contrib.js
```diff
@@ -0,0 +1,98 @@
...
+                var $contributors = $(doc).find("#gh-contributors")[0];
```
marcoscaceres: as here.

js/core/contrib.js
tobie: MDN copy and paste. You should know my coding style by now.

js/core/contrib.js
```diff
@@ -0,0 +1,98 @@
...
+                if (conf.githubAPI && ($commenters || $contributors)) {
+                    github.fetch(conf.githubAPI).then(function (json) {
+                        return $.when(
+                            github.fetchIndex(json.issues_url),
+                            github.fetchIndex(json.issue_comment_url),
+                            github.fetchIndex(json.contributors_url)
+                        ).then(function(issues, comments, contributors) {
```
marcoscaceres: Whoa! The API gives you back more than one single value in the returned promise????

js/core/contrib.js
```diff
@@ -0,0 +1,98 @@
...
+                            var editors = respecConfig.editors.map(prop("name"));
+                            var commenters = findUsers(issues, comments);
+                            contributors = contributors.map(prop("url"));
+                            return $.when(
+                                toHTML(commenters, editors, $commenters),
+                                toHTML(contributors, editors, $contributors)
+                            );
+                        });
+                    }).then(theEnd);
```
marcoscaceres: no .catch() needed?

marcoscaceres: You should 
`return;`
here. and remove the `} else {` and the just call `theEnd();`

js/core/contrib.js
```diff
@@ -0,0 +1,98 @@
...
+                    };
+                }
+                
+                function slice(args) {
+                    return Array.prototype.slice.call(args, 0)
+                }
+
+                function findUsers() {
+                    var users = {};
+                    slice(arguments).forEach(function (things) {
```
marcoscaceres: Nit: this nested .forEach bothers me, tbh. These kinds of functions are supposed to be chained, not nested. It probably means that the first forEach is not what you want here?

marcoscaceres:
Generally, this looks really good @tobie! just a couple of nits.

js/core/contrib.js
tobie: JQuery promises are all kinds of weird.

js/core/contrib.js
tobie: Indeed.

js/core/contrib.js
tobie: Ideally you'd need `flatten()` here, followed by `uniq`. But JS doesn't have either, so I'm not sure what else to do here.

js/core/contrib.js
tobie: Yeah I got lazy and forgot how to report errors in ReSpec. Should do.

darobin:
+1

tobie:
@marcoscaceres: I don't know how to address or what to do about your concerns re nested `forEach` loops, nor, tbh, what those concerns precisely are ;). Other than that, this is ready to merge as far as I'm concerned.

darobin:
I don't understand the `forEach` thing either — if you gentlemen agree on the nitpicks please go ahead and merge!

marcoscaceres:
:ship: IT!!!!

marcoscaceres:
> I don't know how to address or what to do about your concerns re nested forEach loops, nor, tbh, what those concerns precisely are ;)

The concern is just nested loops are ugly and fragile: consider, if we wanted to then `.map` or `.filter` inside there later we would end up with a confusing pyramid of nested scopes. That's why you really want atomic functional operations: 
```
let result = list
   .map(item => ...)
   .filter(item => ...)
   .map(item => ...)
   .reduce(item => ...)
```

Rather than (contrived example and not a reflection of your code!!!, but you get the idea):

```JS
list.forEach( thing => {
   things.forEach( thing => {
      thing.map( 
          thing => filter()
      })   
  })
)
```
It's a nit, tho - so no worry.

tobie:
Heh, I know you're not out to get at my coding style, so don't worry. I'm just genuinely trying to figure out how you'd write this differently in a way that doesn't suck.

Here's what it would look like using only the functional array tools available:

```js
function findUsers() {
    return slice(arguments).reduce(function(output, thing) {
        output.push.apply(output, thing);
        return output;
    }, []).filter(function (thing) {
        return thing.user;
    }).map(function(thing) {
        return thing.user.url;
    }).reduce(function(output, url) {
        if (output.indexOf(url) < 0) output.push(url);
        return output;
    }, []);
}  
```

I tend to find that style a slightly dreadful use of `reduce` but YMMW.

Of course, in PrototypeJS that would have been written as:

```js
function findUsers() {
    return slice(arguments).flatten().pluck("user").compact().pluck("url").uniq();
}  
```

…which is nice, sweet and expressive.

LoDash isn't bad either:

```js
function findUsers() {
    return _(arguments).flatten().pluck("user").compact().pluck("url").uniq().value();
} 
```

I can't believe these things aren't in ES proper at this point.

tobie:
ES6 makes the first example look a little less bad:

```js
function findUsers() {
    return slice(arguments)
        .reduce((output, x) => {
            output.push.apply(output, x);
            return output;
        }, [])
        .filter(x => x.user)
        .map(x => x.user.url)
        .reduce((output, url) => {
            if (output.indexOf(url) < 0) output.push(url);
            return output;
        , []);
}
```

But still. Those `.reduce()` calls… ew!

marcoscaceres:
Argh, yeah... so depressing not having `.pluck()` and friends in ES. You know... we have JQuery already, why not include LoDash? Makes a lot of stuff way more legible.

darobin:
I'm not sure that lodash would gain us that much. I think that switching to Browserify+ES6 using Babel would get us a lot of clarity (and possibly the ability to ditch HBS), I'd focus more on that.

tobie:
HBS?

darobin:
It's shipped :rocket:
HBS=Handlebars

PR #449, opened 2015-06-16
Add support for iterable to oldschool IDL

mwatson2:
Add support for iterable to oldschool IDL
Also add missing tests for maplike


darobin:
Great, thanks! Shipping it now.

PR #456, opened 2015-07-16
Makes getDfnTitles work when called again

halindrome:
Some add-on code (used by PFWG and WPIG currently) call this utility
method on dfn elements again to get the title.  If there are aliases
for a dfn, these are moved into data-lt during the definition
processing phase, so a subsequent call will return the wrong list in
that case.  This fixes that (and improves the tests so we know it
works).

### DIFF
--- a/js/core/utils.js
+++ b/js/core/utils.js
@@ -34,11 +34,19 @@ define(
         // if args.isDefinition is true, then the element is a definition, not a 
         // reference to a definition.  Any @title or @lt will be replaced with
         // @data-lt to be consistent with Bikeshed / Shepherd.
+        //
+        // This method now *prefers* the data-lt attribute for the list of 
+        // titles.  That attribute is added by this method to dfn elements, so 
+        // subsequent calls to this method will return the data-lt based list.
         $.fn.getDfnTitles = function ( args ) {
             var titles = [];
             var theAttr = "";
             var titleString = ""; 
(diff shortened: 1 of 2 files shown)


halindrome:
Note that this change means anyone who uses `<a title="lala">the term</a>` is going to get a warning to switch over to using @lt instead.  Is that what you had in mind @darobin ?

js/core/utils.js
```diff
@@ -34,18 +34,32 @@ define(
...
+                theAttr = "data-lt";
             }
             else if (this.attr("lt")) {
                 titleString = this.attr("lt");
                 theAttr = "lt";
             }
+            else if (this.attr("title")) {
+                titleString = this.attr("title");
+                theAttr = "title";
+                respecEvents.pub("warn", "Using deprecated attribute @title to specify link title for '" + this.text() + "': Change to @lt");
```
marcoscaceres: can you point to respec documentation URL instead? At least there we can examples, etc.

marcoscaceres:
This LGTM. Will give it a little bit for @darobin to respond.

js/core/utils.js
halindrome: Hmm.  Let me put together a pull request with better documentation in there first so the URI can be specific.

sideshowbarker:
This change should be reverted. See #466

Issue #458, opened 2015-07-21
Member-SUBM specStatus for Member Submissions somewhat borked


tidoust:
The documentation mentions a "Member-SUBM" specStatus but that status does not really do the job:

1. Member Submissions usually have completely different copyrights. ReSpec uses the default W3C copyright for the spec, as explained in the documentation. However, setting `additionalCopyrightHolders` will not replace the whole copyright, and `overrideCopyright` is not documented (on purpose?)
2. The right "Member Submission" logo is not automatically added. This can of course easily be achieved thanks to the `logos` setting.
3. ReSpec complains about the document being "on track but no previous version" and adds an empty "Previous Version" in the links section.
3. The SoTD boilerplate is not the right one although note it is not really possible to generate the "right" boilerplate since the links it contains include a link to the Team Comment that cannot be known a priori.

For info, the Member Submission template with the SoTD boilerplate is at:
http://www.w3.org/2005/07/pubrules?uimode=filter&docstatus=mem-subm#sample-template

I am not sure whether this issue should be treated as a request to change the code of ReSpec or as a request to document the limitations. A Member Submission does not really need to be iterated over once submitted, so it's not such a big deal if some manual editing needs to take place right before the submission happens.

darobin:
Does #459 fix all the issues you had? If so, please close this. Thanks!

tidoust:
All issues fixed for me, thanks!

PR #459, opened 2015-07-21
Improved Member-SUBM and Team-SUBM spec status support

tidoust:
This is a proposed fix for #458.

When the spec status is "Member-SUBM" or "Team-SUBM", ReSpec expected a
"Previous version" link that does not mean anything in this context, did not
set the proper companion logo next to the W3C logo and did not include the
right Status of This Document boilerplate.

Please note that the links in the SoTD boilerplate for Member submissions still
contain "@@@". That is normal since these links are set by W3C Team right before
the submissions is accepted, and cannot be computed a priori.

ReSpec still conserves the default copyright in that case, but users simply
have to use "overrideCopyright" to override that setting.

This commit contains a few regression tests.

### DIFF
--- a/js/w3c/headers.js
+++ b/js/w3c/headers.js
@@ -296,6 +296,9 @@ define(
                 conf.publishHumanDate = utils.humanDate(conf.publishDate);
                 conf.isNoTrack = $.inArray(conf.specStatus, this.noTrackStatus) >= 0;
                 conf.isRecTrack = conf.noRecTrack ? false : $.inArray(conf.specStatus, this.recTrackStatus) >= 0;
+                conf.isMemberSubmission = conf.specStatus === "Member-SUBM";
+                conf.isTeamSubmission = conf.specStatus === "Team-SUBM";
+                conf.isSubmission = conf.isMemberSubmission || conf.isTeamSubmission;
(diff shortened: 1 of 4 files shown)


js/w3c/headers.js
```diff
@@ -296,6 +296,9 @@ define(
                 conf.publishHumanDate = utils.humanDate(conf.publishDate);
                 conf.isNoTrack = $.inArray(conf.specStatus, this.noTrackStatus) >= 0;
                 conf.isRecTrack = conf.noRecTrack ? false : $.inArray(conf.specStatus, this.recTrackStatus) >= 0;
+                conf.isMemberSubmission = conf.specStatus === "Member-SUBM";
```
halindrome: This looks good, but I noticed that in the status2rdf and status2maturity mapping tables elsewhere in this document there is no mapping for these two types of document.  The rdf will require an update to  http://www.w3.org/2001/02pd/rec54, which may not be in the offing right now.  But what maturity should these be classified as?  A special class (SUBMISSION)?  Or "NOTE"?

tidoust: That's a good question. I see "fromSubmission" in http://www.w3.org/2001/02pd/rec54 that applies to NOTE, so the intent was perhaps to classify Member and Team Submissions as Notes. It makes sense from a maturity perspective at least.

PR #468, opened 2015-07-31
Change to prefer data-lt for referring to link-titles

halindrome:
Addresses concerns from the user community about the use of the 'lt' attribute because it does 
not validate.

### DIFF
--- a/js/core/utils.js
+++ b/js/core/utils.js
@@ -32,7 +32,7 @@ define(
         // actual title of a <dfn> element (but can apply to other as well).
         // 
         // if args.isDefinition is true, then the element is a definition, not a 
-        // reference to a definition.  Any @title or @lt will be replaced with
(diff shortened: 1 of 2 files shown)


sideshowbarker:
:sunflower: @halindrome cheers

sideshowbarker:
> I'm still 50/50 is we should ban lt outright because it's non-conforming. @halindrome, @tobie, @darobin?

IMHO you should still allow it (and document it). People who want to opt-in to using `lt` should be able to make that choice.

P.S. Being non-conforming with regard to document-conformance requirements in the HTML spec shouldn’t be considered absolutely bad. It’s a very different kind of conformance than browser conformance to UA requirements in the spec. Sometimes individual Web developers/authors have their own good reasons for choosing to use some non-conforming HTML markup for some particular case.

halindrome:
I agree completely. Which is why I left the support in. I did not leave the
documentation in.  I can update that.

Issue #464, opened 2015-07-30
contiguous-idl should use "lt" attributes instead of deprecated "title" for dfn/a


inexorabletash:
Re: https://github.com/w3c/respec/pull/373

It looks like webidl-contiguous.js is still looking for/generating `title` attributes in `<dfn>` and `<a>` elements, rather than the newfangled `lt` or `data-lt`. Since `title` is deprecated, this produces warnings.

cc: @jyasskin

PR #469, opened 2015-07-31
Changed webidl generation to use data-lt attribute

halindrome:
This should address issue #464, but I don't have any specs with
webidl in them to verify that!  Can someone please take a close look?

### DIFF
--- a/js/core/webidl-contiguous.js
+++ b/js/core/webidl-contiguous.js
@@ -119,7 +119,7 @@ define(
                 if (obj.dfn) {
                     var result = "<a for='" + Handlebars.Utils.escapeExpression(obj.linkFor || "") + "'";
                     if (obj.name) {
-                        result += " title='" + Handlebars.Utils.escapeExpression(obj.name) + "'";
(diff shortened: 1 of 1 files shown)


marcoscaceres:
@jyasskin can you please take a look?

js/core/webidl-contiguous.js
```diff
@@ -720,7 +720,7 @@ define(
                     // Found it: update the definition to specify its [for] and title.
```
jyasskin: Update the comments too.

jyasskin:
There's also a use of `@data-title` in `run()`. Can you double-check whether that needs to change?

Other than that and the comments, this looks good to me. Thanks.

inexorabletash:
@marcoscaceres - any chance for progress here?

marcoscaceres:
Merging this, if there are issues with `@data-title`, can fix them in a new bug. We should find out pretty quickly.

halindrome:
Fwiw I looked at this and data-title is being used for something else.
Shouldn't be a problem.

marcoscaceres:
Thanks for checking that @halindrome

PR #471, opened 2015-07-31
Fixed needing double def of term when data-lt is used (closes #470)

### LINKED ISSUE #470: Double definition of term when data-lt is used
when using `data-lt`, ReSpec should grab the text content from the <dfn> and use that as the first item in the `data-lt. Otherwise, you end up with,  E.g.,:

```HTML
<dfn data-lt="installed|installing">installed</dfn>
```

marcoscaceres:


marcoscaceres:
Just noticed there was already a `normalizeString` equiv. called `norm`.

marcoscaceres:
* added additional test for when dfn is normalized to the empty string
* removed use of `.startsWith()` as apparently PhantomJS that runs on TravisCI is still in the dark ages?

darobin:
@marcoscaceres PhantomJS is pretty much always in the dark ages...

marcoscaceres:
Maybe in a year or so we can move to https://slimerjs.org/  ... though I see they are also behind :( this is why we can't have nice things.

darobin:
We should give jsdom a new shot now that they've fixed their performance issues with ReSpec. Also, I'm told Phantom 2 is actually getting decent(er).

PR #484, opened 2015-08-31
Include config information in generated output (closes #478)

### LINKED ISSUE #478: Adding a <script> element with the config information to the generated output
If a post-processor wants to do something with the generated HTML file, it would become very useful if the original config would be available in the file in a machine readable way. After all, many information (document status, short name, list of editors, etc) are available; currently the only way to get to those information is to extract them essentially scraping the content. An easy way would be to generate a `<script>` element to the file containing a JSON serialization of the config data. 

(My use case is a Python script I am finalizing for the generation of an EPUB3 version of a TR document.)

marcoscaceres:


marcoscaceres:
@tobie or @halindrome, r?

tests/spec/core/include-config-spec.js
```diff
@@ -0,0 +1,52 @@
...
+        function(rsdoc) {
+          doc = rsdoc;
+        });
+    });
+    waitsFor(function() {
+      return doc;
+    }, MAXOUT);
+    runs(function() {
+      var $script = $("#respecConfig", doc);
+      expect($script[0] instanceof HTMLScriptElement).toEqual(true);
```
tobie: That's testing the implementation of the DOM in that user agent rather than the output (which is what you should be after).

I'd go for tagName here.

tests/spec/core/include-config-spec.js
```diff
@@ -0,0 +1,52 @@
...
+      expect($script.attr("type")).toEqual("application/json");
+      flushIframes();
+    });
+  });
+  it("should have the same content for the config and the script's text", function() {
+    var doc;
+    runs(function() {
+      makeRSDoc({
+          config: basicConfig,
+          body: $("")
```
tobie: I have no idea what that does. It's jQuery magic. I hate jQuery magic. Especially in tests.

tests/spec/core/include-config-spec.js
```diff
@@ -0,0 +1,52 @@
+describe("Core — Include config as JSON", function() {
+  var MAXOUT = 5000,
+    basicConfig = {
+      editors: [{
+        name: "Robin Berjon"
```
tobie: image

js/core/include-config.js
```diff
@@ -0,0 +1,20 @@
...
+
+define(
+  [],
+  function() {
+    return {
+      run: function(conf, doc, cb, msg) {
+        msg.pub("start", "core/include-config");
+        var script = doc.createElement("script");
+        script.id = "respecConfig";
+        script.innerHTML = JSON.stringify(conf, null, 2);
```
tobie: Any reason for using `innerHTML` over `textContent`?

js/core/include-config.js
```diff
@@ -0,0 +1,20 @@
...
+// Inject's the document's configuration into the head as JSON.
+
+define(
+  [],
+  function() {
+    return {
+      run: function(conf, doc, cb, msg) {
+        msg.pub("start", "core/include-config");
+        var script = doc.createElement("script");
+        script.id = "respecConfig";
```
tobie: Please choose a different id for this element as it'll end up in the global scope thanks to the ridiculous legacy feature that DOM ids are globals. I have plenty of script which test for `window.respecConfig` to decide whether or not client-side rendering is going on. That'll break them.

js/core/include-config.js
```diff
@@ -0,0 +1,20 @@
...
+        script.innerHTML = JSON.stringify(conf, null, 2);
+        script.type = "application/json";
```
tobie: Is the above validator friendly? I suppose so, but just want to make sure, here.

tests/spec/core/include-config-spec.js
marcoscaceres: good call.

tests/spec/core/include-config-spec.js
marcoscaceres: Oh, I just assumed it created an empty body... but you are right, who knows what voodoo jQuery might do with "". I think this is what happens: https://www.youtube.com/watch?v=g_VtIkF8pDY

tests/spec/core/include-config-spec.js
marcoscaceres: so much.

js/core/include-config.js
marcoscaceres: No, fixed.

js/core/include-config.js
marcoscaceres: will change to `respecFinalConfig`

js/core/include-config.js
marcoscaceres: Confirmed that it's valid in validator.nu.

marcoscaceres:
Thanks for the great feedback @tobie! Will push the changes up now.

tobie:
LGTM

PR #490, opened 2015-09-09
using .innerHTML makes JSON reappear (closes #478)

### LINKED ISSUE #478: Adding a <script> element with the config information to the generated output

marcoscaceres:
For whatever reason, .innerText means that when .toString() is called within the output function, the inner text doesn't show up. However, using .innerHTML means that it does show up.

### DIFF
--- a/js/core/base-runner.js
+++ b/js/core/base-runner.js
@@ -23,6 +23,16 @@
     ,   embedded = (top !== self)
     ;
     if (!("respecConfig" in window)) window.respecConfig = {};
+    // clone the initial user configuration.
+    try{
+        if(Object.assign){
+            respecConfig.initialUserConfig = Object.assign({}, respecConfig);
+        } else {
+            respecConfig.initialUserConfig = JSON.parse(JSON.stringify(respecConfig));
+        }
+    }catch(err){
+        respecConfig.initialUserConfig = {};
+    }
     GLOBAL.respecEvents = {
         pub:    function (topic) {
             var args = Array.prototype.slice.call(arguments);
--- a/js/core/include-config.js
+++ b/js/core/include-config.js
@@ -8,7 +8,7 @@ define(
       run: function(conf, doc, cb, msg) {
         msg.pub("start", "core/include-config");
         var script = doc.createElement("script");
-        script.id = "respecFinalConfig";
+        script.id = "initialUserConfig";
         var confFilter = function(key, val) {
           // DefinitionMap contains array of DOM elements that aren't serializable
           // we replace them by their id
@@ -21,7 +21,7 @@ define(
           }
           return val;
         }
-        script.innerText = JSON.stringify(conf, confFilter, 2);
+        script.innerHTML = JSON.stringify(conf.initialUserConfig, confFilter, 2);
         script.type = "application/json";
         doc.head.appendChild(script);
         msg.pub("end", "core/include-config");
--- a/tests/spec/core/include-config-spec.js
+++ b/tests/spec/core/include-config-spec.js
@@ -1,11 +1,11 @@
 describe("Core — Include config as JSON", function() {
-  var MAXOUT = 5000,
-    basicConfig = {
-      editors: [{
-        name: "Robin Berjon"
-      }],
-      specStatus: "WD"
-    };
+  var MAXOUT = 5000;
+  var basicConfig = {
+    editors: [{
+      name: "Robin Berjon"
+    }],
+    specStatus: "WD"
+  };
   it("should have a script tag with the correct attributes", function() {
     var doc;
     runs(function() {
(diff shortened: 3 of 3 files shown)


marcoscaceres:
Tested in Safari, Chrome, and FF on Mac.

marcoscaceres:
@iherman, I'm still not comfortable with this. Maybe it's better if you tell us exactly which fields you actually need for what you are doing?

iherman:
@marcoscaceres I am not sure what you ask. Are you asking what I would use the respec config from my code if it was there?

If so: *at this moment* it would give me an easy way of finding, e.g., the publication date, the short name, the list of editors and authors, the document status and probably other things. All these information are in the initial respec setup but are thrown away; I must scrape the HTML output to find them. (Eg, find the latest URL relying on a particular class name, extract from the URL format whether it is a REC, or WD, or whatever, and extract the publication date.) I am sure other tools that may want to postprocess the HTML file may find other usage. 

(If it was *only* for my tool then I wouldn't have raised this feature request. After all, I have already gone through the trouble of doing it, although it is not future proof as it relies on a particular, and undocumented, structure of the final HTML.)

marcoscaceres:
> On Sep 9, 2015, at 1:11 AM, Ivan Herman <notifications@github.com> wrote:
> 
> @marcoscaceres I am not sure what you ask. Are you asking what I would use the respec config from my code if it was there?
> 
> If so: at this moment it would give me an easy way of finding, e.g., the publication date, the short name, the list of editors and authors, the document status and probably other things.
> 

Ok, in addition to the above, I want to know the list that makes up "probably other things" (I can make it "all use-setable options, as per our documentation). We are vomiting a lot of ReSpec's internals into the JSON, and I would  prefer to only spit out the properties you + others need.

iherman:
@marcoscaceres I would say that anything that the user can set should be visible to the output. We should not come up with a list of "visible" properties imho.

marcoscaceres:
Sent from my iPhone

> On Sep 9, 2015, at 1:26 AM, Ivan Herman <notifications@github.com> wrote:
> 
> @marcoscaceres I would say that anything that the user can set should be visible to the output. We should not come up with a list of "visible" properties imho.
> 

The problem is that ReSpec abuses the user's configuration to also configure its own things. What I will do is snapshot the initial configuration (before ReSpec starts changing it) and return that. That should give you what you need without exposing ReSpec's internal config. Sound good?

iherman:
> On 09 Sep 2015, at 07:34 , Marcos Caceres <notifications@github.com> wrote:
> 
> 
> 
> Sent from my iPhone
> 
> > On Sep 9, 2015, at 1:26 AM, Ivan Herman <notifications@github.com> wrote:
> >
> > @marcoscaceres I would say that anything that the user can set should be visible to the output. We should not come up with a list of "visible" properties imho.
> >
> 
> The problem is that ReSpec abuses the user's configuration to also configure its own things. What I will do is snapshot the initial configuration (before ReSpec starts changing it) and return that. That should give you what you need without exposing ReSpec's internal config. Sound good?

That sounds perfect. Thanks!

js/core/include-config.js
```diff
@@ -21,7 +21,7 @@ define(
           }
           return val;
         }
-        script.innerText = JSON.stringify(conf, confFilter, 2);
+        script.innerHTML = JSON.stringify(conf.initialUserConfig, confFilter, 2);
```
marcoscaceres: Note, innerText doesn't work... it doesn't output when respec calls .innerHTML on the root node.

marcoscaceres:
@tobie r?

js/core/base-runner.js
```diff
@@ -23,6 +23,16 @@
     ,   embedded = (top !== self)
     ;
     if (!("respecConfig" in window)) window.respecConfig = {};
+    // clone the initial user configuration.
+    try{
+        if(Object.assign){
```
tobie: Style.

tobie: Not sure what the convention us but iirc ReSpec has ws all over the place.

js/core/include-config.js
tobie: I don't understand why this wouldn't work, but frontend dev is full of mysteries.

tobie:
LGTM

js/core/base-runner.js
marcoscaceres: going to fix all that in #493.

marcoscaceres:
@tobie thanks for the review + time!

PR #494, opened 2015-09-10
Make invalid attrs to data-* attrs (closes #483)

### LINKED ISSUE #483: "for" attributes used for WebIDL references remain in generated snapshot
The "for" attributes that one needs to use to reference a particular interface in WebIDL prose remain in the generated snapshot. This means that the resulting spec does not pass markup validation, which in turns prevents publication to /TR/ space for W3C specs.

The "for" attributes should rather be dropped from the generated spec. The same thing should happen to "link-for" and "dfn-for" attributes.

I noticed the problem while trying to publish the Presentation API spec [1], which makes use of such references. Using spec-generator to produce the snapshot [2] (manual generation yields the same result) produces a spec that does not pass markup validation [3].

Note that I'm very likely going to update the Presentation API spec to add some post-processing code that drops the problematic attributes so the links below may not show errors for long. The bug still exists.

[1] http://w3c.github.io/presentation-api/
[2] https://labs.w3.org/spec-generator/?type=respec&url=https://w3c.github.io/presentation-api/index.html?specStatus=WD;shortName=presentation-api
[3] https://validator.w3.org/nu/?doc=https%3A%2F%2Flabs.w3.org%2Fspec-generator%2F%3Ftype%3Drespec%26url%3Dhttps%3A%2F%2Fw3c.github.io%2Fpresentation-api%2Findex.html%3FspecStatus%3DWD%3BshortName%3Dpresentation-api

marcoscaceres:


marcoscaceres:
@tobie R?

js/ui/save-html.js
```diff
@@ -9,6 +9,22 @@ define(
         var cleanup = function (rootEl) {
             $(".removeOnSave", rootEl).remove();
             utils.removeReSpec(rootEl);
+           function attrToDataAttr(name){
```
marcoscaceres: I went all hipster functional programming here :dancer:

js/ui/save-html.js
```diff
@@ -9,6 +9,22 @@ define(
...
+               return function(elem){
+                   var value = elem.getAttribute(name);
+                   elem.removeAttribute(name);
+                   elem.setAttribute("data-" + name, value);
+               }
+           }
+
+           var forList = rootEl.querySelectorAll("[for]");
```
tobie: Is `[foo]` a valid selector? Don't you have to use `*[foo]`?

js/ui/save-html.js
tobie: <3

js/ui/save-html.js
```diff
@@ -9,6 +9,22 @@ define(
...
+           Array.prototype.forEach.call(forList, attrToDataAttr("for"));
```
tobie: I tend to favor: 

```js
var _forEach = Array.prototype.forEach;
_forEach.call(…);
```
but YMMV.

js/ui/save-html.js
marcoscaceres: > On Sep 10, 2015, at 3:04 AM, Tobie Langel <notifications@github.com> wrote:
> 
> In js/ui/save-html.js:
> 
> > @@ -9,6 +9,22 @@ define(
> >          var cleanup = function (rootEl) {
> >              $(".removeOnSave", rootEl).remove();
> >              utils.removeReSpec(rootEl);
> > +           function attrToDataAttr(name){
> > +               return function(elem){
> > +                   var value = elem.getAttribute(name);
> > +                   elem.removeAttribute(name);
> > +                   elem.setAttribute("data-" + name, value);
> > +               }
> > +           }
> > +
> > +           var forList = rootEl.querySelectorAll("[for]");
> Is [foo] a valid selector? Don't you have to use *[foo]?
> 

Seems to work fine without "*", but *[foo] is more clear.

tobie:
LGTM outside of the affore-mentioned selector syntax (possible) issue.

js/ui/save-html.js
tobie: Yeah, I'm not sure what the spec says + might be tainted from my pre-qSA days when this generally broke.

marcoscaceres:
Moved the proposed code to the bottom of link-to-dfn.js instead, as that is the place where all the `for=` magic happens (so it makes more sense to just do the cleanup there).

PR #496, opened 2015-09-12
Wrap content in main element (closes #479)

### LINKED ISSUE #479: Encapsulating the content into a <main> element?
I generate an EPUB3 version of the TR document produced by respec. Unfortunately, there are discrepancies among epub readers; one is that they artificially set zero padding on the body element. (This is the case with Kobo or with the iBook readers of Apple.) This means the logo on the header area is overwritten by the content.

One way of getting around that, which I believe would be useful for other post-processing facilities, is to encapsulate the whole content of the document in a `<main>` element (which would also make sense logically, so to say). That means the styling of the post-processed content could become easy.

marcoscaceres:
* if there is no main, wraps everything in a main. 
* if a main or [role=main] exists, it won't wrap. 
   - it will use main the main content to generate the ToC

### DIFF
--- a/js/core/structure.js
+++ b/js/core/structure.js
@@ -95,8 +95,9 @@ define(
 
                 // makeTOC
                 if (!conf.noTOC) {
-                    var $ul = makeTOCAtLevel($("body", doc), doc, [0], 1, conf);
-                    if (!$ul) return;
+                    var main = doc.querySelector("main, *[role=main]");
+                    var $ul = makeTOCAtLevel($(main || doc.body), doc, [0], 1, conf);
+                    if (!$ul) return finish();
                     var $sec = $("<section id='toc'/>").append("<h2 class='introductory'>" + conf.l10n.toc + "</h2>")
                                                        .append($ul)
                     ,   $ref = $("#toc", doc), replace = false;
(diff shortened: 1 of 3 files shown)


plehegar:
With $("body > section").wrapAll("<main></main>"), it is really worth adding a main element if everything in the body element gets into main anyway? The only element which seems excluded from main is the div.head, right?

stevefaulkner:
On 9 November 2015 at 16:14, Philippe Le Hegaret <notifications@github.com>
wrote:

> it is really worth adding a main element if everything in the body element
> gets into main anyway? The only element which seems excluded from main is
> the div.head, right?


if there is a header, then yes its worth adding <main> for navigation and
semantic purposes

--

Regards

SteveF
Current Standards Work @W3C
<http://www.paciellogroup.com/blog/2015/03/current-standards-work-at-w3c/>

stevefaulkner:
related - " if everything in the body element gets into main anyway?" why
would everything in the body be wrapped in a main?


--

Regards

SteveF
Current Standards Work @W3C
<http://www.paciellogroup.com/blog/2015/03/current-standards-work-at-w3c/>

halindrome:
I continue to feel that only the main sections should be wrapped in a 'main' element (or have role='main' put on a wrapping div).  The appendices are by definition ancillary but supporting content.  However, since the only roles that might be appropriate for such content are 'note' and 'complementary', and neither of those are *quite* right... I would be comfortable wrapping all of the sections in a role='main' or main element.

Issue #395, opened 2015-01-23
Warn that webidl-oldschool is deprecated


jyasskin:
Forked from https://github.com/w3c/respec/pull/373#discussion_r22034180. This should probably wait until webidl-contiguous has seen a bit of use and is stable.

marcoscaceres:
As I've already had to clean up one spec using old IDL, I think it's time to start showing a real warning.

marcoscaceres:
ReSpec now warns.

PR #497, opened 2015-09-12
Warn that webidl-oldschool is deprecated (#395)

marcoscaceres:
Before merging: 

* [ ] Get feedback from spec-prod@w3c.org.


plehegar:
btw, I'm not done switching the webperf doc to continuous IDL but I'm ok with a warning.

PR #487, opened 2015-09-04
Add ability to save as EPUB as a service

iherman:

PR #500, opened 2015-09-14
Fix save-ui for respec2html

dontcallmedom:
#487 broke respec2html.js because phantom apparently got confused with function binding (not sure why); this fixes this.

Incidentally, it would be great if running respec2html.js was part of the test suite.


marcoscaceres:
Released in 3.2.72

PR #502, opened 2015-09-17
Export config fixup

marcoscaceres:
* exporting the intial config was having issues if the userConfig was before or after the respec.js file. This hopefully fixes that now.


marcoscaceres:
@tobie r?

tobie:
LGTM though a test would have been welcomed for this case (not sure if feasible).

marcoscaceres:
Not to worry, there is test already from the previous attempt :)

tobie:
:+1:

Issue #501, opened 2015-09-14
Run respec2html.js was part of the test suite.


marcoscaceres:
@marcoscaceres keeps breaking respec2html.js because it's not tested as part of the our test suite. We should either ban Marcos or fix the test suite.

marcoscaceres:
@dontcallmedom thoughts on how we do this... I don't have any direct experience with respec2html.js.

dontcallmedom:
my suggestion (and sorry I can't afford to provide a patch for this at the moment):
* prepare a set of respec-valid documents that loads respec from within the repo; ideally, they would exercise as much of respec as possible (e.g. a bunch of webidl)
* add to the "test" script in package.json a rule to run "./tools/respec2html.js -e -w " on these documents
* a bonus point might to be to run an HTML5 validator on the output of that command — that way, we would detect when respec starts generating invalid HTML5; this might be better done in Travis CI though (see [how I do it e.g. for the WebRTC spec](https://github.com/w3c/webrtc-pc/blob/master/.travis.yml))

(I don't know if the stuff in tests/specifications was supposed to be used for this kind of test? maybe that wouldn't be a bad place to put these new "respec-valid" documents)

PR #516, opened 2015-11-04
Fix bug in editor name check

dontcallmedom:
Introduced in 21efc4e521d30b90752cbb99919af176c1e3d2b1 when switching from $.each to forEach

Another case were #501 would have helped :)


marcoscaceres:
Released in 3.2.77

PR #533, opened 2015-12-04
WebIDL: setID removes digits from the id

deniak:
The WebIDL support creates `id` from the name of the attributes. If these attributes contain digits in their names, [`setID`](https://github.com/w3c/respec/blob/develop/js/core/webidl-oldschool.js#L121) removes them.
This is a problem for specs like [Vehicule Data](https://w3c.github.io/automotive/vehicle_data/data_spec.html). For instance:
```
    readonly        attribute float?          batteryTempT0;
    readonly        attribute float?          batteryVoltageT120;
    readonly        attribute float?          batteryTempT120;
    readonly        attribute float?          batteryVoltageESD30;
    readonly        attribute float?          batteryTempESD30;
    readonly        attribute float?          batteryVoltageESD450;
    readonly        attribute float?          batteryTempESD450;
    readonly        attribute float?          batteryVoltageESD870;
    readonly        attribute float?          batteryTempESD870;
    readonly        attribute float?          batteryVoltageESD1290;
    readonly        attribute float?          batteryTempESD1290;
    readonly        attribute float?          batteryVoltageESD1710;
    readonly        attribute float?          batteryTempESD1710;
```
resulting in duplicate `id`.

That PR aims at keeping the digits.

### DIFF
--- a/js/core/webidl-oldschool.js
+++ b/js/core/webidl-oldschool.js
@@ -118,7 +118,7 @@ define(
         WebIDLProcessor.prototype = {
             setID:  function (obj, match) {
                 obj.id = match;
-                obj.refId = obj.id.replace(/[^a-zA-Z_\-]/g, "");
+                obj.refId = obj.id.replace(/[^a-zA-Z0-9_\-]/g, "");
                 obj.unescapedId = (obj.id[0] == "_" ? obj.id.slice(1) : obj.id);
             }
         ,   nullable:   function (obj, type) {
--- a/js/core/webidl-oldschool.js
+++ b/js/core/webidl-oldschool.js
@@ -118,7 +118,7 @@ define(
         WebIDLProcessor.prototype = {
             setID:  function (obj, match) {
                 obj.id = match;
-                obj.refId = obj.id.replace(/[^a-zA-Z_\-]/g, "").replace(/^[0-9\-]*/, "");
+                obj.refId = obj.id.replace(/[^a-zA-Z0-9_\-]/g, "").replace(/^[0-9\-]*/, "");
                 obj.unescapedId = (obj.id[0] == "_" ? obj.id.slice(1) : obj.id);
             }
         ,   nullable:   function (obj, type) {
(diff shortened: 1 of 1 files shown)


marcoscaceres:
As an aside: that API looks broken. Putting the value into the name of the attribute???

js/core/webidl-oldschool.js
```diff
@@ -118,7 +118,7 @@ define(
         WebIDLProcessor.prototype = {
             setID:  function (obj, match) {
                 obj.id = match;
-                obj.refId = obj.id.replace(/[^a-zA-Z_\-]/g, "");
+                obj.refId = obj.id.replace(/[^a-zA-Z0-9_\-]/g, "");
```
dontcallmedom: I don't think this is quite right yet; the id needs to not start with a number. I guess
`obj.refId = obj.id.replace(/^[^a-zA-Z_]*[^a-zA-Z0-9_\-]*/, "");`

deniak: Good point! I don't see a good way to do it with only once `replace` so I'm just using 2.

tguild:
Agree with @dontcallmedom refinement.  Thanks @deniak 

@marcoscaceres that isn't value of the attribute but distinguishing characteristic of the attribute.

You want for example the temperature of the battery of the car as it starts and has been running for awhile.

[[batteryTempT0 of type float, readonly , nullable
    MUST return battery temperature at engine start time = 0
batteryTempT120 of type float, readonly , nullable
    MUST return battery temperature at engine start time = 120]]

plehegar:
I concur with @marcoscaceres : this API would benefit from a review from webapps imho. The current design is very unusual.

plehegar:
To be clear, this is still a bug to be fixed in respec in any case.

tguild:
Review would be most welcome. After a read I can coordinate a call with editors and chairs if that would be helpful.

tripu:
The change LGTM.

plehegar:
LGTM as well

marcoscaceres:
Will try to check it today.

plehegar:
Merged since it does look fine.

marcoscaceres:
Thanks @plehegar.

PR #538, opened 2015-12-08
Added a -exclude-scripts to respec2html

plehegar:
typical use would be "-exclude-script https://www.w3.org/scripts/"
See also 
 https://github.com/w3c/spec-generator/issues/21

### DIFF
--- a/tools/respec2html.js
+++ b/tools/respec2html.js
@@ -10,6 +10,7 @@ var page = require("webpage").create()
 ,   timer
 ,   reportErrors = false
 ,   reportWarnings = false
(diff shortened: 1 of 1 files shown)


plehegar:
any taker for this one?

dontcallmedom:
https://github.com/w3c/respec/issues/541 should do the trick, although it forces to use imported libraries as if they were not loaded; e.g. for the MathJax case, the call to MathJax will create a ReferenceError if it's not tested for undefined beforehand.

marcoscaceres:
Can close this request, right?

PR #541, opened 2015-12-17
Add --exclude-script to respec2html

dontcallmedom:
close #538

### DIFF
--- a/tools/respec2html.js
+++ b/tools/respec2html.js
@@ -45,12 +45,15 @@ var source = args[1]
 
 
 if (args.length < 2 || args.length > 4) {
-    var usage = "Usage:\n   phantomjs --ssl-protocol=any respec2html.js [-e] [-w] respec-source [html-output] [timeout]\n" +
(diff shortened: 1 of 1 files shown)


tools/respec2html.js
```diff
@@ -30,6 +31,13 @@ if (args.indexOf("-w") !== -1) {
     reportWarnings = true;
 }
 
+if (args.indexOf("--exclude-script") !== -1) {
+    var idx = args.indexOf("--exclude-script");
+    var values = args.splice(idx, 2);
```
marcoscaceres: Are you certain you want "splice" here (instead of `slice`)? Splice modifies the args array, which is not great IMO (i.e., we should keep inputs immutable).

marcoscaceres:
LGTM

tools/respec2html.js
dontcallmedom: I've re-used here the approach taken in the two previous argument parsing blocks.

plehegar:
Doesn't this only permit one exclusion instead of a list of exclusions?

dontcallmedom:
it filters any scripts whose source starts with the url passed in parameter; I could support several of those, but that doesn't seem required yet

plehegar:
:+1:

PR #543, opened 2016-01-04
[Urgent] Avoid assigning property to a string object (non-strict compatible)

dontcallmedom:
Breaks WebIDL Contiguous mode in Chrome/Edge
cf https://github.com/w3c/presentation-api/issues/232#issuecomment-168742148

### DIFF
--- a/js/core/webidl-contiguous.js
+++ b/js/core/webidl-contiguous.js
@@ -640,11 +640,12 @@ define(
 
                     case "enum":
                         name = defn.name;
-                        defn.values = defn.values.filter(function (v) { return v.type === undefined;})
-                        .map(function(v) {
-                            return { toString: function() {return v;},
-                                     dfn: findDfn(name, v, definitionMap, msg)
-                                   };
+                        defn.values.forEach(function(v,i) {
+                            if (v.type === undefined) {
+                                defn.values[i] = { toString: function() {return v;},
+                                                   dfn: findDfn(name, v, definitionMap, msg)
+                                                 };
+                            }
                         });
                         defn.idlId = "idl-def-" + name.toLowerCase();
                         break;
--- a/js/core/webidl-contiguous.js
+++ b/js/core/webidl-contiguous.js
@@ -641,8 +641,10 @@ define(
                     case "enum":
                         name = defn.name;
                         defn.values.filter(function (v) { return v.type === undefined;})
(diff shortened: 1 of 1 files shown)


js/core/webidl-contiguous.js
```diff
@@ -641,8 +641,10 @@ define(
                     case "enum":
                         name = defn.name;
                         defn.values.filter(function (v) { return v.type === undefined;})
-                                   .forEach(function(v) {
-                            v.dfn = findDfn(name, v, definitionMap, msg);
+                        .map(function(v) {
```
tidoust: Won't "map" return a new array? Where is that array stored?
I would expect something like defn.values = defn.values.filter().map(function(v) {...});

dontcallmedom: argh, you're right

js/core/webidl-contiguous.js
```diff
@@ -640,9 +640,12 @@ define(
 
                     case "enum":
                         name = defn.name;
-                        defn.values.filter(function (v) { return v.type === undefined;})
-                                   .forEach(function(v) {
-                            v.dfn = findDfn(name, v, definitionMap, msg);
+                        defn.values.forEach(function(v,i) {
+                            if (v.type === undefined) {
```
marcoscaceres: I don't understand why we can't filter here? 

Shouldn't this just:
```JS
defn.values
  .filter(function(item){return typeof item === "undefined"})
  // create the objects
  .map(function(item){ return { 
    toString: function() {return item;},
    dfn: findDfn(name, item, definitionMap, msg),
  })
  // reduce them into the defn
  .reduce(function(collector, next, index){  
       collector.values[i] = next;
       return collector;
   },  defn);
```

Or am I missing something? Is it that it needs to use the original index?

marcoscaceres: (if I got the above wrong, and we just need a hot fix, then quickly then just merge ... this bit is not particularly complicated in the overall scheme of things)

dontcallmedom: we need to keep the items that have .type different from undefined in defn.values; filtering them out was what was causing the previous Travis failure.

Given your note above, I'll merge this PR; hopefully someone can make a release out of it.

halindrome:
I will push it

Issue #546, opened 2016-01-08
Dealing with overloaded operations


dontcallmedom:
If an interface uses overloaded operations in WebIDL contiguous, all the IDL declarations with only link to one definition (the first with a matching name), making it impossible to link to definitions specific to the overloaded version(s) of the operation.

PR #548, opened 2016-01-08
Avoid multiple dfn errors by prefering dfn tags

dontcallmedom:


tidoust:
As said on IRC, the update seems benign but arbitrary. I wonder how we can end up in a "normal" situation where there are multiple definitions for the same term, with the explicit <dfn> one being the one to choose. I would rather assume that finding multiple definitions is always a bad sign, and that the user should be alerted about that as is the case today.

Could you clarify (possibly in a comment) what case you're trying to capture here?

dontcallmedom:
so the WebIDL contiguous code generates `<span>` that are recognized as definitions in the definitionMap; when there are multiple such definitions (e.g. for overloaded operations), this triggers a "multiple dfn" error.

Note that preferring dfn to other type of definitions is in-line with what is done in js/core/link-to-dfn.js:
```
                            // We want <dfn> definitions to take precedence over
                            // definitions from WebIDL. WebIDL definitions wind
                            // up as <span>s instead of <dfn>.
```

tidoust:
Hmm, but if a spec has two contiguous code that generate `<span>` that are recognized as definitions in the definitionMap and not other definition, your code will filter both of them and report that the term is not defined, whereas it should report that there are multiple definitions for the term, no?

dontcallmedom:
hmm... you're right; also, it looks like the bug I was trying to solve is already solved by my other PR #546 ; let's close this for now.

PR #549, opened 2016-01-08
Avoid duplicating ids on multiple partials

### LINKED ISSUE #545: Dealing with multiple partials of the same interface in a doc
WebIDL contiguous generates duplicate ids in documents where several fragments of IDL define a partial of the same interface.

For instance
```
partial interface Foo {
  attribute DOMString bar;
};
partial interface Foo {
  attribute DOMString quuz;
};
``` 
will have the `idl-def-foo` id duplicated.

dontcallmedom:
Close #545


dontcallmedom:
merged after r+ from @tidoust

PR #553, opened 2016-01-11
Run respec2html on example files in test suite

dontcallmedom:


dontcallmedom:
merging after r+ from @tidoust

PR #558, opened 2016-02-09
Align new TR design with latest requirements

plehegar:
- removed width = content-width from viewport
- add fixup.js

### DIFF
--- a/js/w3c/fixup.js
+++ b/js/w3c/fixup.js
@@ -1,0 +1,40 @@
+
+// Module w3c/fixup
+// Inserts a link to the appropriate JS Extension for the W3C Spec Style Sheet.
+// CONFIGURATION
+//  - specStatus: the short code for the specification's maturity level or type (required)
+
+define(
+    ["core/utils"],
+    function (utils) {
+        return {
+            run:    function (conf, doc, cb, msg) {
+                msg.pub("start", "w3c/fixup");
+                if (!conf.specStatus) msg.pub("error", "Configuration 'specStatus' is not set, required for w3c/fixup");
+                var statStyle = conf.specStatus;
+                if (statStyle === "FPWD"    ||
+                    statStyle === "LC"      ||
+                    statStyle === "WD-NOTE" ||
+                    statStyle === "LC-NOTE") statStyle = "WD";
+                if (statStyle === "FPWD-NOTE") statStyle = "WG-NOTE";
+                if (statStyle === "finding" || statStyle === "draft-finding") statStyle = "base";
+                var fixup = "null";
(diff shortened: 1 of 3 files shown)


js/w3c/fixup.js
```diff
@@ -0,0 +1,40 @@
...
+                msg.pub("start", "w3c/fixup");
+                if (!conf.specStatus) msg.pub("error", "Configuration 'specStatus' is not set, required for w3c/fixup");
+                var statStyle = conf.specStatus;
+                if (statStyle === "FPWD"    ||
+                    statStyle === "LC"      ||
+                    statStyle === "WD-NOTE" ||
+                    statStyle === "LC-NOTE") statStyle = "WD";
+                if (statStyle === "FPWD-NOTE") statStyle = "WG-NOTE";
+                if (statStyle === "finding" || statStyle === "draft-finding") statStyle = "base";
+                var fixup = "null";
```
marcoscaceres: Note, there is a bug here. it is set to the string "null", but later you check if it's `null`.

js/w3c/fixup.js
```diff
@@ -0,0 +1,40 @@
...
+
+define(
+    ["core/utils"],
+    function (utils) {
+        return {
+            run:    function (conf, doc, cb, msg) {
+                msg.pub("start", "w3c/fixup");
+                if (!conf.specStatus) msg.pub("error", "Configuration 'specStatus' is not set, required for w3c/fixup");
+                var statStyle = conf.specStatus;
+                if (statStyle === "FPWD"    ||
```
marcoscaceres: Also, `statStyle` is never actually used? so all this code is redudant?

marcoscaceres: ah, that's the next big `if()`... this all seems sub-optimal. This should be handled by the `js/w3c/style.js` module.

marcoscaceres: ... which this module duplicates from.

marcoscaceres:
@plehegar, @halindrome, I'm not comfortable with this module and feel it should be backed out. I can see that it was kinda slapped together quickly and I don't think it's release quality (e.g., duplicates code, lacks tests, doesn't adhere to new coding conventions). 

Apart from using the "experimental" style, it seems that updating `js/w3c/style.js` would actually be the right thing to do here.

marcoscaceres:
Another question: is this module time sensitive?

halindrome:
@plehegar asked me to to a release with this module yesterday, so I pushed it.  TBH I didn't even review it for quality.  I had the impression from IRC that it was needed right then so I just did it.  I don't mind backing it out, but I want a reaction from @plehegar first.

PR #565, opened 2016-02-15
Made timeout of respec2html.js 20 secs (closes #564)

### LINKED ISSUE #564: CI - Travis keeps timing out
At some point, a 10 second timeout was added to the tests. This is causing Travis to timeout and thrown an error. 

Anyone know when the timeout was added?

marcoscaceres:


marcoscaceres:
@dontcallmedom r? Do you see any issues with making this 20 secs? Seems to make travis happy again.

marcoscaceres:
Merging this to avoid the timeout errors. At first glance, I can't see any harm in merging. @dontcallmedom, when you see this, let me know if you think a better value should be used.

dontcallmedom:
the only drawback is that when something goes wrong in respec execution, you'll have to wait 20s to know; but that's probably no big deal; if it is, we can always make the timeout configurable in respec2hml, and use 20 as the timeout value in the test suite.

That said, I'm curious as to why it would take 20s to run respec2html.js; do you have a pointer on which spec caused the timeout?

PR #568, opened 2016-02-16
Teach ReSpec's references about publishers (closes #518)

### LINKED ISSUE #518: References section is formatted incorrectly
W3C style manual on formatting references: http://www.w3.org/2001/06/manual/#ref-section

Respec docs say localBiblio should be formatted like this: http://www.w3.org/respec/ref.html#localbiblio

I see two issues:
1. Respec lists authors before the title, style manual says title should come first.
2. Respec ignores the publisher field.

marcoscaceres:
Publishers now show up when available in a bibliographical reference, after the author like so... 

screenshot 2016-02-16 20 08 24

### DIFF
--- a/js/core/biblio.js
+++ b/js/core/biblio.js
@@ -44,6 +44,14 @@ define(
                 if (ref.etAl) output += " et al";
                 output += ". ";
             }
+            if(ref.publisher){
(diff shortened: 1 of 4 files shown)


plehegar:
:+1:

PR #578, opened 2016-02-29
Moved fixup.js to end

halindrome:
in order to comply with pubrules as per @plehegar


marcoscaceres:
Beat me to it!:) will make the final changes when I get to work in around 2 hours.

PR #586, opened 2016-03-02
Fix issue where callback was expected where promise is used

marcoscaceres:


marcoscaceres:
checked by @halindrome

PR #591, opened 2016-03-02
stop headless.js testing basic.built.html

marcoscaceres:
This is causing headless.js to fail unnecessarily.


marcoscaceres:
Merging one liner, which may fix a few things.

PR #588, opened 2016-03-02
Removed some conditional style changes

halindrome:
There were some conditional changes for the new styles that should now
be permanent.

PR #596, opened 2016-03-02
Always use a nav

plehegar:
We should always use nav#toc independently of the experimental style or not.


halindrome:
This is addressed in #588

plehegar:
it's in 13c0ce6ad5c7dde654b9b618c8935ab67142245f

Issue #597, opened 2016-03-02
jquery prototype is not extended with functions in utils


halindrome:
Many extension functions, both embedded in script elements and in external files, rely upon the '$' Object being extended by the $.fn methods in the utils.js module.  This does not appear to be happening now.  So, for example, the following code in a method that was called using data-oninclude used to work and now fails:

<pre>
   $.each(base.querySelectorAll("dfn"), function(i, item) {
        var $t = $(item) ;
        var titles = $t.getDfnTitles();
        var n = $t.makeID("dfn", titles[0]);
        if (n) {
            termNames[n] = $t.parent() ;
        }
    });
</pre>

$t.getDfnTitles and $t.makeID both fail to resolve now.  I am attempting to run this down, but I suspect @marcoscaceres will know immediately where to look!

marcoscaceres:
Ah! I broke that, indeed :) adding things to the prototype of jquery is a terrible idea regardless, so will make a new helper object.

marcoscaceres:
This was fixed by release 106

PR #598, opened 2016-03-03
Make jquery-enhanced get used in right places places

marcoscaceres:


marcoscaceres:
@halindrome, this should fix #597. Please review.

halindrome:
This does not appear fixed to me.  Use case: data-include is used to bring in an external file.  data-oninclude is used to transform the content of that file.  The function called uses jQuery constructs and the functions in core/jquery-enhanced are not available.  Not sure what I am missing.  If you want an example try http://pandora.aptest.com/w3c/aria-current/aria/aria.html - It references a version of respec that is built using this branch.

marcoscaceres:
The simplest solution is just to include jquery-enhanced everywhere.

marcoscaceres:
Will do that now.

halindrome:
Just checking - do you think this is working now??

marcoscaceres:
Not yet. Hopefully Monday. Found some other broken things that are blocking this one.

marcoscaceres:
Ok, I think this is working... but need to break it up into smaller commits.

marcoscaceres:
Today was a sad day - I swear i had it working and then it blew up. Getting close tho.

marcoscaceres:
Ok, going to see what can be salvaged from here.... closing this PR, as #623 now includes at lot of this.

Issue #291, opened 2013-12-20
noRecTrack should add Note-specific boilerplate


darobin:
"The group does not expect this document to become a W3C Recommendation. It is expected to become a W3C Note."

marcoscaceres:
Closed via #601

PR #601, opened 2016-03-03
Bug291 no rec track should add note

marcoscaceres:
Review please :smile:


halindrome:
Did someone ask for this change?  I don't remember a bug report about it.

marcoscaceres:
It's in the 3.2 milestone (see #291). Robin filed it. I'm working my way through the 3.2 list.

halindrome:
Oh!  Then sure.  It LGTM.  However, if we could fix the critical path problem with data-oninclude first that would be super helpful.

marcoscaceres:
Will try to do the data-oninclude today.

PR #602, opened 2016-03-03
Try to figure out why Chrome keeps failing

marcoscaceres:
* disabling firefox, which is working fine...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,4 +22,4 @@ before_script:
 script:
   - npm run pretest
   - npm run test:headless
-  - karma start --single-run --reporters progress karma.conf.js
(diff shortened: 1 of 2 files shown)


marcoscaceres:
Ok, it was running both browsers at the same time that was causing the issues. 

@halindrome r?

marcoscaceres:
I also switched to Mocha style reporting because it's :rainbow: pretty.

marcoscaceres:
I'm merging this one because: 
 * it will fix all my other prs
 * it's a tiny change. 
 * it doesn't affect end-users - only maintainers. 

However, a post approval would be great.

halindrome:
Seems to have helped in my dev environment.

PR #579, opened 2016-02-29
Changed CGBG style references to new repo

halindrome:
This change assumes that the CG and BG stylesheets and logos are integrated with StyleSheets/TR/2016 as per my pull request w3c/tr-design#70

PR #606, opened 2016-03-03
Code to switch CG and BG reports to new styles

halindrome:
The requisite styles and logos are already in the
new location - this just enables their use.  

This supercedes PR #579


marcoscaceres:
Awesome! R+ from me.

PR #610, opened 2016-03-04
feat: (tools) Creates a JSON list of tests

marcoscaceres:
Want to use this so we stop trying to manually maintain the list of tests in the SpecRunner.html.


marcoscaceres:
hmmmm.... maybe we should later write a little local test server using express, that routes to a list of these files.

marcoscaceres:
Mergin this, as a I really need it now and it doesn't affect end users.

PR #611, opened 2016-03-04
Added testing for data-oninclude

halindrome:
Small script that does some post-processing on the included content.

Also fixed a bug in the original tests - it was looking for data-include
on a &lt;p&gt; element instead of on a &lt;div&gt;.

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -68,6 +68,10 @@ module.exports = function(config) {
         pattern: "./node_modules/webidl2/lib/*.js",
         included: false,
         served: true,
+      }, {
+        pattern: "builds/respec-w3c-common.js",
+        included: false,
+        served: true,
       },
       "tests/spec/SpecHelper.js",
       "tests/test-main.js",
@@ -85,6 +89,7 @@ module.exports = function(config) {
       "/tests/": "/base/tests/",
       "/spec/": "/base/tests/spec/",
       "/node_modules/": "/base/node_modules/",
+      "/builds/": "/base/builds/",
     },
 
     // preprocess matching files before serving them to the browser
--- a/tests/spec/core/data-include-spec.js
+++ b/tests/spec/core/data-include-spec.js
@@ -14,7 +14,7 @@ describe("Core — Data Include", function() {
       expect($sec.find("dfn").text()).toEqual("INCLUDED");
       expect($sec.find("div > p").length).toEqual(1);
       expect($sec.find("div").attr("data-include")).toBeFalsy();
-      expect($sec.find("dfn").attr("id")).toEqual("dfn-included");
+      expect($sec.find("dfn").attr("data-testing")).toEqual("itWorked");
     };
     var ops = {
       config: makeBasicConfig(),
--- a/tests/spec/core/data-include-spec.js
+++ b/tests/spec/core/data-include-spec.js
@@ -11,9 +11,10 @@ describe("Core — Data Include", function() {
     var theTest = function(doc) {
       var $sec = $("#includes", doc);
       expect($sec.find("p").length).toEqual(1);
-      expect($sec.find("p").text()).toEqual("INCLUDED");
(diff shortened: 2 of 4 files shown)


marcoscaceres:
LGTM. Merge when ready.

halindrome:
Weirdly, this test does not fail.  I sort of want to know why before I
merge it.

On Fri, Mar 4, 2016 at 9:09 PM, Marcos Caceres <notifications@github.com>
wrote:

> LGTM. Merge when ready.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/w3c/respec/pull/611#issuecomment-192556992>.
>



-- 
Shane McCarron
halindrome@gmail.com

marcoscaceres:
yeah, was wondering about that...

halindrome:
Interesting - the new test was passing when referencing the source form of respec. Referencing the BUILT form of respec gets different answers.  That's horrifying.

karma.conf.js
```diff
@@ -68,6 +68,10 @@ module.exports = function(config) {
         pattern: "./node_modules/webidl2/lib/*.js",
         included: false,
         served: true,
+      }, {
```
marcoscaceres: We shouldn't include the built version. That kinda defeats the purpose (making it extremely difficult to debug any problems). If there are differences between the two versions, we should fix those.

karma.conf.js
```diff
@@ -85,6 +89,7 @@ module.exports = function(config) {
       "/tests/": "/base/tests/",
       "/spec/": "/base/tests/spec/",
       "/node_modules/": "/base/node_modules/",
+      "/builds/": "/base/builds/",
```
marcoscaceres: As above

halindrome: Hmm - maybe I don't understand the karma testing strategy.  It builds the
built version as a first step.  Why does it do that if it is not going to
test it.  Also, while I agree that testing a static version is different
than testing the dynamic version, we don't /ship/ the dynamic version.  If
they test differently it is definitely a problem, but it is also a problem
if we do all our testing against something we don't ship.

What am I missing?

On Sat, Mar 5, 2016 at 3:11 AM, Marcos Caceres <notifications@github.com>
wrote:

> In karma.conf.js
> <https://github.com/w3c/respec/pull/611#discussion_r55118571>:
>
> > @@ -85,6 +89,7 @@ module.exports = function(config) {
> >        "/tests/": "/base/tests/",
> >        "/spec/": "/base/tests/spec/",
> >        "/node_modules/": "/base/node_modules/",
> > +      "/builds/": "/base/builds/",
>
> As above
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/w3c/respec/pull/611/files#r55118571>.
>



-- 
Shane McCarron
halindrome@gmail.com

marcoscaceres: > Hmm - maybe I don't understand the karma testing strategy. It builds the built version as a first step. 

Sorry, let me clarify. That's not Karma, that's `npm test`. "npm test" is our battery of tests, one of which is running Karma. 

> Why does it do that if it is not going to test it.

It does test the build, by running it through PhantomJS. 

>  Also, while I agree that testing a static version is different than testing the dynamic version, we don't /ship/ the dynamic version.

Sure, but trying to test the generated version makes it hard for us to debug... however, as we now have source maps (only recently added), we **could** switch to just testing the built version (*terms and conditions apply).  

>  If they test differently it is definitely a problem, but it is also a problem if we do all our testing against something we don't ship. What am I missing?

Not much. We can try switching to the built one, but we need to make sure the source maps actually work. It doesn't help us debug things if we see errors like "ReSpec threw an error on line 1 character 11232" or an uglified code-base.

marcoscaceres: *  of an uglified code-base

marcoscaceres:
I've integrated the test into a local branch that includes the fix. Hoping to send it up in a few hours! @halindrome, thanks for putting the additional testing together here.

PR #620, opened 2016-03-08
Fix (best-practices-spec.js): Unessesary use of jQuery

marcoscaceres:


halindrome:
Wow - I had no idea that feature was even in there.  It is covered in the documentation even.  The change seems fine, although I don't know why it is necessary.  We use jQuery all over the place in tests.

marcoscaceres:
I'm trying slowly to kill our dependence on jquery (and non standard cruft) where possible. We don't need it and it slows everything down. Once I kill it completely, we can look at moving to a virtual DOM to make rendering fast. Killing it completely will take me a while (2+ months), but will be worth it.

halindrome:
I disagree.  Unless you are talking about just the tests.  Many many people have extended ReSpec in their work, and those extensions likely rely upon jQuery but do not include it themselves - because we have it in our environment.  We can't remove it.

marcoscaceres:
It would be a transitional thing, naturally. I've already started experimenting with jquery slim, and it works fine (just selector engine). At least our core modules should not rely on it. It doesn't give us anything nowadays (it's 2016, for Pete's sake!) that we can't do with actual browser standards. If we do find missing stuff, I know just the place to get those things added to the web platform;)

marcoscaceres:
@halindrome, ok. I've been giving this a lot more thought - and, despite my personal dislike of jQuery and its hidden magic, you are right that it's heavily relied upon. 

Let's do this: 

 * [ ] include jQuery as a global dependency. 
 * [ ] inject the Enhanced jQuery early.  
 * [ ] Systematically work towards eradicating it from our own code. 

Then, any new features are discouraged from using jQuery unless there is a **very good reason**:tm:  to do so.

halindrome:
I think that makes good sense.  And someday, over a beer, you can explain to me why you find it so distasteful.

marcoscaceres:
basically... 
 
 * (:beer:) it does things that were required back in/before 2010 (e.g., amazing CSS Selectors, browser compat), but are now supported natively ... Old IE support. These are no longer a problem. Thus, it longer buys us much. 
 * (:beer:) Using .append() etc. is really slow. We should only change the DOM once, right before the turn of the even loop.  
 * (:beer:) Promises they use are not compatible with real promises. 
 * (:beer:) It adds to the weight of ReSpec.
 * (:beer:) It adds indirection, which means when you are debugging, you have to skip over whatever it is doing.

halindrome:
I agree with most of the above.  Modern jquery short circuits those requests to the native browser support if available (I thought).  But yes, their Promises are broken and debugging is a pain.  There is a map somewhere I thought, but I have never tried to use it.

Anyway, I agree with the direction.

marcoscaceres:
Sent a new PR instead: #623

PR #621, opened 2016-03-08
Fix (SpecRunner.html): style-spec was being tested twice

marcoscaceres:


marcoscaceres:
One liner copy/pasta, doesn't require review.

PR #623, opened 2016-03-09
Globalize jquery enhanced

marcoscaceres:
See commit list... I'm going to break this up into several PRs where I can.


js/core/jquery-enhanced.js
```diff
@@ -0,0 +1,173 @@
+define(["jquery","core/utils"], function($, utils) {
```
marcoscaceres: This is a straight copy and paste from core/utils (where this code used to live). This code does not require review.

marcoscaceres:
Actually, I'm just going to annotate the code. As it's not as many changes as it looks.

js/core/utils.js
```diff
@@ -5,175 +5,8 @@
...
-                    for (var i = 0, len = node.childNodes.length; i < len; ++i) getTextNodes(node.childNodes[i]);
-                }
-            }
-            getTextNodes(this[0]);
-            return textNodes;
-        };
-
-
+    [],
+    function () {
```
marcoscaceres: This is where I ripped out the additions to JQuery from.

tests/SpecRunner.html
```diff
@@ -10,51 +10,33 @@
...
-    <script src="spec/w3c/rdfa-spec.js"></script>
-    <script src="spec/w3c/style-spec.js"></script>
-
-    <!-- SOURCE -->
-    <!-- none currently -->
+    <script src='/node_modules/requirejs/require.js'></script>
+    <script src="/node_modules/whatwg-fetch/fetch.js"></script>
     <script>
-     jasmine.DEFAULT_TIMEOUT_INTERVAL = 1220000;
+    "use strict";
```
marcoscaceres: Here, I use fetch to get the "testFile.json" list to test against.

tests/spec/core/jquery-enhanced-spec.js
```diff
@@ -0,0 +1,134 @@
+"use strict";
```
marcoscaceres: These tests used to live in utils-spec.js, they don't require review.

marcoscaceres:
Ok, finished annotations... ready for review.

halindrome:
I *am* reviewing this. Trying to be careful though.

marcoscaceres:
@halindrome really appreciate your time, thoroughness, help, and support! Your ideas are vastly improving what we are doing here.

halindrome:
Question - do you think this version fixes the general problem that our utils.js methods are not exposed to jquery extended objects when they are used in extensions? 'Cause my testing shows that getDfnTitles, for example, is not a method on a jquery extended object in a JS file loaded by a respec-enabled page.

marcoscaceres:
@halindrome, it should fix it. It literally just puts the extended jQuery into the window object before any of the plugins run.

js/core/jquery-enhanced.js
```diff
@@ -0,0 +1,173 @@
...
+      if (node.nodeType === 1 && excl[node.localName.toLowerCase()]) return;
+      if (node.nodeType === 3) textNodes.push(node);
+      else {
+        for (var i = 0, len = node.childNodes.length; i < len; ++i) getTextNodes(node.childNodes[i]);
+      }
+    }
+    getTextNodes(this[0]);
+    return textNodes;
+  };
+  window.$ = $;
```
marcoscaceres: @halindrome here be the magic!!!!! :star2:

marcoscaceres:
@halindrome, see: https://github.com/w3c/respec/pull/623/files#diff-2d019cf669f32febfab70c89006e39caR171

marcoscaceres:
An extension can also `define(["jquery-enhanced"])` ... I even think that is they ask for just "jquery", they will also get the enhanced, as the enhanced one is just enhancing the jQuery singleton.

halindrome:
I think we are talking past one another.  Before we started messing with this, tne utils.js module pushed a bunch of methods into the jquery function prototype so that they were all added to any jquery objects that are later created.  So, when data-oninclude is defined that method will be called and jquery already has the extensions that permit easy integration with respec.  That is not currently happening in this branch.  

Extension scripts don't use require, and in any event there are tons of them out there.  We can't ask them to all go change something in order to start working again.

halindrome:
Just to be clear - these extensions don't ASK for anything.  They are just snippets of JS that get run within the context of ReSpec because it just works.

marcoscaceres:
> Just to be clear - these extensions don't ASK for anything. They are just snippets of JS that get run within the context of ReSpec because it just works.

Understood. They don't need to ask. It will just work* (<small>terms and conditions apply</small>).

* so long as they at least wait for `isRespecReady` or use the `respecEvents` thing.

halindrome:
Well - in the case where I am testing my method is called from the data-oninclude handler and the methods are not htere.  Check out http://dev.spec-ops.io/webpayments-ig/VCTF/use-cases - it loads from this branch's build of respec.

marcoscaceres:
@halindrome, that version of ReSpec seems to be wrong (or didn't build?). It's missing this code. See, in particular, "profile-w3c-common.js". The code from this PR is not there?

halindrome:
yikes..... weird.  testing

marcoscaceres:
@halindrome, in particular: 
```
deps: ["core/jquery-enhanced", "Promise"]
```
is not there. And neither is the `core/jquery-enhanced.js` file.

marcoscaceres:
In other news! YAY! the sourcemap work just came in handy... in even other news, it's 4:30 am, of shit I better got to sleep! :+1:

marcoscaceres:
* oh

marcoscaceres:
Just to show it works for me locally (everything is there):

screenshot 2016-03-10 04 35 36

Did you do? `./tools/build-w3c-common.js` :zzz: :sleeping:

halindrome:
Okay - no idea why it had not built when I changed branches.  I do note that it is complaining about trying to load node_modules/promise-polyfill/Promise.js from the built version.  I will try adding that to collection of required components that need to be included in the build.  But yes, I can confirm that the object is extended!

Get some sleep.

marcoscaceres:
>  I do note that it is complaining about trying to load node_modules/promise-polyfill/Promise.js from the built version.

Can you please post a new bug with the error? That should only happen if it's missing from your `node_modules` dir, or something is not right in the builder. I'm not seeing that error locally here.

halindrome:
I will do.  I do see that there is a dependency on it in that core/jquer-enhanced requires Promise, but we never load hqyer-enhanced in the profile-w3c-common.js file

halindrome:
Oh - and otherwise LGTM.

marcoscaceres:
Confirmed... it is going to the network for Promise.js. My thought is to just remove it, as all the browsers we target support it: http://caniuse.com/#search=Promise

halindrome:
Huh.    Remember people use this for editors drafts and we have no control at all who revirews those.  Including people who use older  user agents.  ReSpec - it's not just for development any more. (tm)

marcoscaceres:
@halindrome, sure, but we have to cut off somewhere. Promises have been natively supported in browsers for over a year, and all browsers are now evergreen. It seems highly unlikely people reading specs (specially ed drafts) won't be running a browser that is at most 3 months old.

halindrome:
I am willing to try this - I just note for the record that it could bite us in the ass.  In particular with Windows users who have not moved to Windows 10 and for whatever reason use IE.

marcoscaceres:
Let's see what happens. If it's bad, we can quickly add the poly back.

halindrome:
Assuming this passes I am comfortable with you merging and pushing a new release.

marcoscaceres:
Ha! I think ./tests/headless.js just ruined the party.

halindrome:
you're right.  I had this problem the other day too.  UGGH

marcoscaceres:
Backed out the last commit.

marcoscaceres:
Ok, will merge this after it turns green and figure out why Promise is not being inserted tomorrow (unless you want to have a look). The `respecIsReady` promise is also ready to merge, I hope.

halindrome:
works for me.  I will play with it a little in a new (local) branch once it is all merged.

PR #626, opened 2016-03-09
Fix (headless.js): pipe child stdout and stderr to parent

marcoscaceres:
Reverts some stuff that was already in the original version which I accidentally took out. At least we will be able to see now what is going on.

### DIFF
--- a/tests/headless.js
+++ b/tests/headless.js
@@ -4,6 +4,7 @@
 "use strict";
 const fs = require("fs");
 const async = require("marcosc-async");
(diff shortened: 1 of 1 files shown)


marcoscaceres:
:tada:

halindrome:
Yeah - this would be nice.  LGTM.

It also explains why I was not getting an automatic build when I was running tests.

marcoscaceres:
I haven't figured out how to get it to stop onstderr, however. I'll need to do more reading.

PR #632, opened 2016-03-09
Added dependencies on Promise.

halindrome:
This should fix the loading issue with Promise polyfill.

### DIFF
--- a/js/core/base-runner.js
+++ b/js/core/base-runner.js
@@ -70,7 +70,7 @@ if (window.console) {
 
 
 define(
-    [],
+    ["core/respec-ready"],
     function () {
         return {
             runAll:    function (plugs) {
--- a/js/profile-w3c-common.js
+++ b/js/profile-w3c-common.js
@@ -4,7 +4,10 @@ require.config({
   shim: {
     shortcut: {
       exports: "shortcut"
-    }
+    },
+    Promise: {
+      exports: "Promise"
+    },
   },
   paths: {
     "handlebars": "/node_modules/handlebars/dist/handlebars",
@@ -13,9 +16,9 @@ require.config({
     "webidl2": "/node_modules/webidl2/lib/webidl2",
   },
   deps: [
-    "core/jquery-enhanced",
-    "core/respec-ready",
+    "jquery",
     "Promise",
+    "core/jquery-enhanced",
   ],
 });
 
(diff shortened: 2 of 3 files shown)


halindrome:
BTW I tested this in IE 10 and it worked as expected (no errors on a missing Promise)

halindrome:
However, it DOES throw a console error that core/respec-ready is not found.  So that isn't getting bundled somehow.

halindrome:
Oh - and the develop branch does that too.  So it's not my bug ;-) @marcoscaceres

halindrome:
@marcoscaceres r?

js/profile-w3c-common.js
```diff
@@ -14,8 +14,7 @@ require.config({
   },
   deps: [
     "core/jquery-enhanced",
-    "core/respec-ready",
-    "Promise",
+    "core/respec-ready"
```
marcoscaceres: I think we still want to list "Promise" as a dependency, if only to signal that that is its role.

js/core/base-runner.js
```diff
@@ -70,7 +70,7 @@ if (window.console) {
 
 
 define(
-    [],
+    [ "Promise" ],
```
marcoscaceres: I think the problem is that `profile-w3c-common.js` needs to define that it needs Promise, not particularly this module (as it's any module that can use it, and because Promise is attached to the global scope). 

I'm going to review the RequireJS docs about what hackery the Optimizer does, and why it's not being packaged properly... but I suspect it's just the above. 

Will report back.

marcoscaceres:
Ok, so yes... Optimizer does simple static analysis to figure out what to include. So, moving "Promise" to the define() array should fix this... trying that out.

marcoscaceres:
(I found a whole bunch of other cool stuff we can use too, which will make the builder much cleaner and awesome ... who would have thought reading the manual would be useful!!! )

marcoscaceres:
Amended the commit, based on what I found from the RequireJS docs. Testing it locally, works well... let's see if Travis likes it.

marcoscaceres:
It passes! :tada:

marcoscaceres:
Let's get this merged and ship a new version of ReSpec!!! woot Woot!!!

halindrome:
LGTM.  I tried the shim addition.  I also read the documentation for RequireJS.  It has a lot of features.

PR #638, opened 2016-03-10
Changed logo path to use new style location

halindrome:
Should probably be automatically switchable with the use
of useExperimentalStyles but that's not todays problem.

This should fix issues with passing pubrules.


marcoscaceres:
LGTM. Merge when ready

marcoscaceres:
Release script should be working fine again. Just run:

./tools/release.js

PR #641, opened 2016-03-10
Javascript errors shouldn't appear in the output

plehegar:
Otherwise spec-generator can't generate properly the clipboard-apis spec.

### DIFF
--- a/tools/respec2html.js
+++ b/tools/respec2html.js
@@ -80,6 +80,10 @@ page.onConsoleMessage = function(msg) {
   }
 };
 
+page.onError = function(msg) {
+  errors.push(msg);
+};
+
 page.open(source, function(status) {
   if (status !== "success") {
     console.error("Unable to access ReSpec source file.");


halindrome:
Umm - I don't *disagree* but... what javascript errors?

marcoscaceres:
@plehegar, this seems fine - but I also want a little bit more info before we merge.

plehegar:
http://w3c.github.io/clipboard-apis/
 generates Javascript errors on the console, because it embeds tests. I assume those are taken and run separately.

spec-generator uses respec2html to convert a respec doc into a TR ready one. See the result on clipboard-apis:
 https://labs.w3.org/spec-generator/?type=respec&url=http%3A%2F%2Fw3c.github.io%2Fclipboard-apis%2Findex.html%3FspecStatus%3DWD%3BshortName%3Dclipboard-apis

so the change is meant to mute those errors so that they don't end up in the output.

halindrome:
Okay, I get it!  Thanks.  BTW why does the spec have javascript errors in it?

Issue #647, opened 2016-03-12
W3C not serving source maps


marcoscaceres:
Source maps don't seem to be getting pushed to the w3c server - so don't get served. Will need W3C help here.

dontcallmedom:
can you say more about which additional files need to be pushed?

marcoscaceres:
@dontcallmedom builds/profile-w3c-common.build.js.map

dontcallmedom:
https://www.w3.org/Tools/respec/profile-w3c-common.build.js.map now proxies to http://w3c.github.io/respec/builds/profile-w3c-common.build.js.map — let me know if that matches what you needed or not.

marcoscaceres:
@dontcallmedom, sorry, I made a mistake :( The filename I was after was actually "respec-w3c-common.build.js.map"

dontcallmedom:
https://www.w3.org/Tools/respec/respec-w3c-common.build.js.map is up and running

marcoscaceres:
Yay! Thank you!

PR #648, opened 2016-03-14
Fix (builds/): remove profile-w3c-common.build.js.map, as it's not used (#647).

marcoscaceres:


marcoscaceres:
This just removes a file I accidentally added at some point. It doesn't require review.

PR #649, opened 2016-03-14
Fix (dfn.js): Remove object's prototype (closes #582)

### LINKED ISSUE #582: <dfn id="x" dfn-for="y">constructor</dfn> causes JS error
Presumably someone should be using a `Map` or `Object.create(null)` instead of an object, as otherwise `o.constructor === Object`.

Haven't created a tiny repro case but ran into this working on the custom elements spec.

marcoscaceres:


marcoscaceres:
r?

halindrome:
There's some libraries that use the definitionMap.  Have you done testing with, for example, the definition popup from the respec pill button?

marcoscaceres:
Checked the pill button, works fine. Other uses of this thing are as if it was a map, so shouldn't be any problem.

marcoscaceres:
* I mean, the pill button definition list.

PR #655, opened 2016-03-15
Fix (w3c/style.js): initial-scale should be 1 (closes #654)

### LINKED ISSUE #654: Pubrules doesn't like viewport
The latest respec inserts the following viewport meta tag:

<meta content="initial-scale=1.0, shrink-to-fit=no, width=device-width" name="viewport">

Pubrules expects:

<meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">

The primary difference seems to be initial-scale=1 vs initial-scale=1.0; I don't know if pubrules also cares about the order of things in the content attribute, nor do I know if there is a processing difference between 1 and 1.0. I just know that the webmaster won't let me publish unless there is a clean pubrules report. Again I can't point to a sample URI because I'm manually fixing this in order to pass pubrules and be allowed to publish.

marcoscaceres:


halindrome:
LGTM.  although in my opinion pubrules is just being stupid.

marcoscaceres:
It's hard to know... I think that name/value pair is some Apple proprietary thing.

PR #659, opened 2016-03-15
Fixes to release.js

marcoscaceres:


marcoscaceres:
r? ... we should probably just create a needs review label.

tools/release.js
```diff
@@ -175,18 +189,34 @@ function getBranchState() {
   });
 }
 
+function getCurrentBranch(){
+  return async.task(function*(){
+     const branch = yield git(`rev-parse --abbrev-ref HEAD`);
```
halindrome: Clever

halindrome:
LGTM

tools/release.js
marcoscaceres: Credit to stack overflow for that one:)

Issue #653, opened 2016-03-14
Says Note but Pubrules rejects


michael-n-cooper:
The latest Respec, at time of this filing, inserts the following text into the Status:

It is expected to become a <abbr title="World Wide Web Consortium">W3C</abbr> Note.

Pubrules rejects this as and declares the entire patent policy paragraph invalid (section 10). I can't point to a URI to demonstrate this because I am manually removing the offending text in order to pass pubrules and be able to publish. The version of Pubrules is at https://www.w3.org/2005/07/pubrules?uimode=checker_full&uri= and raises this error on a doc with noRecTrack: true in the respec config.

I previously said this error had gone away after a hard refresh.

https://lists.w3.org/Archives/Public/spec-prod/2016JanMar/0062.html

That turns out to be incorrect. I must have already reverted my staging doc to the known-working version of respec when I asserted that. This issue is still present in the current version of Respec.

halindrome:
I feel like this is a change @marcoscaceres pushed recently.  If so, can he please back it out and then test a note against pubrules?

marcoscaceres:
I'll investigate. @michael-n-cooper, in the future (or if you are around), please paste in your ReSpec config. It helps us greatly with debugging.

michael-n-cooper:
Here's the respec config. And BTW the behaviour is observed on 
https://rawgit.com/w3c/aria/Mar_2016_Heartbeats/practices/aria-practices.html 
though that URI will only be valid temporarily, and is not one that can 
be checked in pubrules directly. The respec config from that doc is (the 
last two lines use respec extension functions so will be unfamiliar):


var respecConfig = { // Embed RDFa data in the output. doRDFa: '1.1', 
includePermalinks: true, permalinkEdge: true, permalinkHide: false, // 
Specification status (e.g., WD, LC, NOTE, etc.). If in doubt use ED. 
specStatus: "WD", // crEnd: "2012-04-30", // perEnd: "2013-07-23", 
publishDate: "2016-03-17", noRecTrack: true, diffTool: 
"http://www.aptest.com/standards/htmldiff/htmldiff.pl", // The 
specifications short name, as in http://www.w3.org/TR/short-name/ 
shortName: "wai-aria-practices-1.1", // If you wish the publication date 
to be other than today, // set publishDate. // publishDate: 
"2009-08-06", copyrightStart: "2015", // If there is a previously 
published draft, uncomment this // and set its YYYY-MM-DD date and its 
[...]

PR #661, opened 2016-03-15
Put note text back the way it was.

halindrome:
To fix issue #653.  This puts everything back the way it was.  'cause notes dont pass pubrules now.

### DIFF
--- a/js/w3c/templates/sotd.html
+++ b/js/w3c/templates/sotd.html
@@ -134,6 +134,9 @@
               href='http://www.w3.org/Consortium/Patent-Policy-20040205/'>5 February 2004 W3C Patent
               Policy</a>.
             {{/unless}}
+            {{#if recNotExpected}}
+              The group does not expect this document to become a W3C Recommendation.
+            {{/if}}
             {{#unless isIGNote}}
               {{#if multipleWGs}}
                 W3C maintains {{{wgPatentHTML}}}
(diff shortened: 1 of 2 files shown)


marcoscaceres:
Can I kindly ask that PRs not be merged without review. We shouldn't hot fix stuff as it just introduces more cruft. It is possible for users to self-fix the generated versions, while we figure out appropriate solutions to bugs.

halindrome:
You are of course correct.  In my defense, I should never have agreed to the original change in the first place.  In fact, when I went and looked at the PR there were other things in there I wouldn't have agreed to had I actually been paying attention when I looked at it.  That's my fault.

marcoscaceres:
> On 16 Mar 2016, at 7:08 PM, Shane McCarron <notifications@github.com> wrote:
> 
> You are of course correct. In my defense, I should never have agreed to the original change in the first place. In fact, when I went and looked at the PR there were other things in there I wouldn't have agreed to had I actually been paying attention when I looked at it. That's my fault.
> 

I wouldn't say it's your fault, but the process is there for a reason (and I strongly believe it works): Marcos, or any of the maintainers, will on occasion propose less than ideal PRs. Each PR requires as much consideration from the reviewer as the code the PR adds/removes. Having said that, we will on occasion make mistakes. And that's totally ok. We have the luxury here that, in an organization where it takes on average 5 years to get to REC, a delay of a day or two fixing a bug or merging a PR is not a big deal. 

So, it's totally ok to say, "I don't like this... I need to sleep on it... Have you maybe thought of doing it this way... Etc."

Issue #531, opened 2015-12-04
Skip specific scripts while generating snapshots


deniak:
Is there a configuration option to skip specific scripts while a snapshot is being generated?

For instance, the [Web Audio API specification](http://webaudio.github.io/web-audio-api/) relies on MathJax and we want to avoid executing that lib when we generate a snapshot.

```HTML
<pre>
              $$
                \left(\frac{\max(d, d_{ref})}{d_{ref}}\right)^{-f}
              $$
</pre>
```
becomes
```HTML
[...]

PR #678, opened 2016-04-01
Use nightmarejs

marcoscaceres:
@tripu @dontcallmedom, this uses NightmareJS instead of phantomjs to generate the spec. Would appreciate some help making sure I didn't break anything... and that all your requirements are met.


dontcallmedom:
from what I can see (but haven't tested), this removes the ability to detect the presence of errors and warnings from respec2html — is that on purpose? I use that feature extensively in my group repos to make sure we don't introduce errors as we update our specs.

(note that I won't be around much the next ~10 days; I would appreciate if this wasn't merged until I'm back; alternatively, @vivienlacourba might be available to verify that the new respec2html.js can be compatible with our usage in the w3c/webrtc-respec-ci

marcoscaceres:
> from what I can see (but haven't tested), this removes the ability to detect the presence of errors and warnings from respec2html — is that on purpose?

Only warnings. Errors are still caught and are fatal. See:

```JS
  nightmare.on("console", function(type, message){
    if(type === "error"){
      const errorMessage = `❌ ReSpec error: ${colors.debug(message)}`;
      console.error(colors.red(errorMessage));
      nightmare.proc.kill();
      process.exit(1);
      return;
    }
  });
```

marcoscaceres:
Just to be clear, warnings are non-fatal because they don't affect spec generation. 

I'm also working towards eliminating runtime errors being reported to end-users (instead, they just go to console.error(), where they belong). Runtime errors are a concern of the testing framework and the `develop` branch... eventually, runtime errors could be pinged back to developers as github issues.

marcoscaceres:
I just removed the presentation API commits from this PR, hopefully making it easier to review.

marcoscaceres:
@dontcallmedom, I'm happy to have warnings routed to stderr, but not cause processing to halt?

marcoscaceres:
@dontcallmedom, if you get time, let me know what other things I should check. Happy to do that while you are away. This PR blocks a bunch of other stuff I'm cooking up, as it all uses "new" :tm: JS stuff that Phantomjs can't handle... omg, so excited about being able to use iterators like it's 2014!!!!!!1!!one!!!

dontcallmedom:
could I have a command line switch to make warnings fatal? that's what "-w" did with the phantomjs version.

halindrome:
I would also like a way to make warnings fatal.

package.json
```diff
@@ -41,6 +42,7 @@
     "mocha": "^2.4.5",
     "moment": "^2.12.0",
     "mozilla-download": "^1.1.1",
+    "nightmare": "^2.2.0",
     "phantomjs-prebuilt": "^2.1.5",
```
dontcallmedom: shouldn't this go? or is phantom still used elsewhere?

marcoscaceres:
@dontcallmedom added -w and -e back in :)

package.json
marcoscaceres: I think you are right. Squeee!!! Will double check.

marcoscaceres:
@halindrome @dontcallmedom, ok, sending an update that makes warnings fatal - but defaults to false.

package.json
marcoscaceres: Ok, removed :dancer: :dancer: :dancer: :dancers:!

tools/respec2html.js
```diff
@@ -1,141 +1,150 @@
...
-if (args.indexOf("-w") !== -1) {
-  args.splice(args.indexOf("-w"), 1);
-  reportWarnings = true;
-}
-
-if (args.indexOf("--delay") !== -1) {
-  delay = args.splice(args.indexOf("--delay"), 2)[1];
-}
-
-if (args.indexOf("--exclude-script") !== -1) {
```
dontcallmedom: looks like you didn't replicate this feature — it is needed for the spec generator

tools/respec2html.js
```diff
@@ -1,141 +1,150 @@
...
-          });
-        });
+  const Nightmare = require("nightmare");
+  const nightmare = new Nightmare({
+    show: false,
+    timeout: parsedArgs.timeout,
+  });
+  const URLParser = require("url");
+  const url = URLParser.parse(parsedArgs.src).href
+    .concat(`?cache-bust=${Math.random()}`);
```
dontcallmedom: does this work only for http urls? the current script works with local files

dontcallmedom:
I tried running `node tools/respec2html.js --src  http://w3c.github.io/webrtc-pc/webrtc.html`, but got an error `Cannot read property 'dispatchEvent' of null`.

tools/respec2html.js
marcoscaceres: Nightmare can't intercept network events, I think. I don't think it's worth supporting doing that, tbh.

dontcallmedom: hmmm... the trouble is that when e.g. a spec imports MathJax, it doesn't want to be saved with the MathJax export, but with the MathJax script ready to run on execution.
There are probably other ways to solve this than network interception, but we need it before we can make the switch.

dontcallmedom: (I assume your comment was related to "--exclude-script", not loading from HTTP only)

marcoscaceres:
@dontcallmedom, wrote:
> but got an error Cannot read property 'dispatchEvent' of null.

Ah, it's because that document needs the updates to ui.js and save-as-html.js that are part of this PR.

dontcallmedom:
oh right, that makes sense

dontcallmedom:
hmm... thinking some more about it, it might be better to leave the non-required changes to ui.js and save-as-html.js out of this PR; this would make it easier to test, and the said changes can always be brought in later.

marcoscaceres:
> hmm... thinking some more about it, it might be better to leave the non-required changes to ui.js and save-as-html.js out of this PR; this would make it easier to test, and the said changes can always be brought in later.

Agree. Will split them out.

marcoscaceres:
@dontcallmedom, tested webrtc locally and it works ok.

tools/respec2html.js
marcoscaceres: It seems very strange that MathJax would cause issues. Which spec is using it?

tools/respec2html.js
marcoscaceres: @dontcallmedom, which scripts need to be excluded?

marcoscaceres:
Ok, I've split out the UI code and shipped it. That now allows us to test this script on current *real* specs. I've also updated how caching is handled (nicely :laughing:!)... just get's dumped into /tmp/ - before, it was aggressively caching resources, which was not great. 

The last outstanding thing relates to "--exclude-script". I'm still not clear why that is needed. I see it being used to exclude scripts from w3c.org in the spec-generator.

dontcallmedom:
The --exclude-script thing is what @deniak was describing in #531 ; see also #538 ; since I'm not available this week, maybe @plehegar can help describe the problem and give input on the alternatives.

marcoscaceres:
It seems that if it's only WebAudio that needs this, then we should work with @padenot to find a solution.

Paul, some context: we are trying to remove Phantomjs as a dependency for ReSpec (because Phantomjs runs an ancient Webkit and it's a maintenance burden), and instead move to using a Chromium-based solution. 

At the moment, Phantomjs is suppressing the Mathjax script by intercepting network requests - something we can't do with the new solution, unfortunately. IIUC, the rationale given is that MathJax produces invalid markup, so disabling it during spec generation avoids issues with PubRules.

A simple solution might be to only include the Mathjax script if `navigator.useragent` is not "respec2html" (which I can add dynamically). It would mean changing WebAudio to the following:

```JS
      window.addEventListener("load", function(){
        // don't run MathJax during spec generation
        if(navigator.userAgent === "respec2html"){
          return;
        }
       // Let's include Mathjax! 
        var url = "https://www.w3.org/scripts/MathJax/2.6.1/MathJax.js?config=TeX-AMS-MML_HTMLorMML";
        var mathjax = document.createElement('script');
        mathjax.src = url;
        new Promise(function(resolve, reject){
          // Safari doesn't support load event on scripts... it's 2016 :/
          var id = setInterval(function(){
            if(window.MathJax){
              clearInterval(id);
              resolve();
            }
          }, 100);  
          mathjax.onload = function(){
            clearInterval(id);     
            resolve();
          };
          mathjax.onerror = function(err){
            var error = new Error("MathJax failed to load. " + (err || "") );
            reject(error);
          };
          setTimeout(mathjax.onerror, 20000);
          document.body.appendChild(mathjax);
        }).then(function(){
          MathJax.Hub.Config({
              tex2jax: {
                 skipTags: ["script","noscript","style","textarea","code"]
              }
            });
        }).catch(function(error){
          console.error(error);
        });
    });
```

Works for me locally.

marcoscaceres:
(updated the above to use DOMContentLoaded instead, which would make it a little faster... note that that the spec generator doesn't load images, so it's fairly fast to fire the load event)

marcoscaceres:
Sent a PR to Web Audio: https://github.com/WebAudio/web-audio-api/pull/782

marcoscaceres:
@dontcallmedom, if you are back. I'm keen for us to continue our discussion towards merging this.

marcoscaceres:
(heh, I just realized that both "new firefox" and ReSpec will use Electron :stuck_out_tongue_winking_eye:)

dontcallmedom:
I'm a bit doubtful about requiring any spec that would use third-party scripts to include such a piece of code; could we have instead something more declarative?

I wish @plehegar could chime in too, since he was the one asking for that feature in the first place.

marcoscaceres:
Note, when I said "PR does exactly what..."  I was talking about the PR I made to the Web Audio spec - not this spec.

marcoscaceres:
Err.. Not this PR. I should just stop responding to comments after 1am💩. 💤

dontcallmedom:
By declarative, I meant a way for a spec author to mark `<script>` that should be left as is when saved by ReSpec (either via the generator, or via the button in the UI).

marcoscaceres:
The problem is not the script: it's what the script does to the DOM. When the script is included, it removes all the Math declarations. It's why it needs to be blocked (as Phantom currently does) or conditionally injected (as per my PR).

dontcallmedom:
I understand that; but can't ReSpec prevent the said scripts from running once they have been identified declaratively (and in the specific context of generating a to-be-saved document)?

marcoscaceres:
I tried declaring a script tag without a src, and tried using a "data-src" attr, but the browser thinks it's an inline script; so when you change data-src to just src nothing happens. That makes sense. So using a scrip tagt, at least, is out of the question. Even if we use something else, it's still going to require a dynamic script tag to be created.

marcoscaceres:
In an ideal world, we would just take the garbage markup dumped from Mathjax and somehow tell the validator to ignore it (in all cases, we are basically just hacking around the HTML validator requirements of PubRules): or better, get Mathjax to start outputting valid HTML. It's an open source project, after all.

dontcallmedom:
@marcoscaceres before putting this in a gh-pages release, I would suggest:
* giving a heads-up to spec-prod since it will likely break the set up of some respec users
* bumping the version of respec to 4.x since it's a breaking change in the API of the tool (so is the WebIDL oldschool deprecation that seems to have been merged as well)

It would also be good if we could sync on that release to avoid disruption in the automated publication system; I'm working on updating the spec generator to match the new tool, but the two releases will need to be sync'd as much as possible.

dontcallmedom:
@marcoscaceres could you extract the method that actually builds the generated output into an exported function? that would simplify integration in the spec generator.

anssiko:
The latest (6390363) `respec2html` from the `develop` branch works fine with MathJax.js loaded using `<script>` with no special trickery in my case:

`$ node respec2html.js -s https://w3c.github.io/mediacapture-depth/ -o snapshot.html`

Related: the HTML-CSS output processor of MathJax produces invalid markup that does not pass validation:
```
$ validate snapshot.html 
"snapshot.html":1203.721-1203.721: error: Element “nobr” not allowed as child of element “span” in this context. (Suppressing further errors from this subtree.)
```
This would cause PubRules to fail, and render the Echinda unusable. I reported this as a MathJax bug, see https://github.com/mathjax/MathJax/issues/1440. Let's see if we can get that fixed upstream.

dontcallmedom:
@anssiko I'm not very versed in what MathJax does, but it it does any user agent sniffing or feature detection, you need it to run at execution time rather than generation time.

marcoscaceres:
Indeed, see web audio spec 🙃

anssiko:
The `respec2html` output is pure browser-agnostic HTML and CSS and does not rely on sniffing or feature detection (apart form using some prefixed CSS properties) and does not contain any MathJax scripts (ReSpec cleans them out since they're `<script class="remove">`). I prepared an example `snapshot.html` for you, see:

Generated output:
https://rawgit.com/w3c/mediacapture-depth/mathjax-snapshot-example/snapshot.html#dfn-rules-to-convert-using-range-inverse
https://github.com/w3c/mediacapture-depth/blob/mathjax-snapshot-example/snapshot.html

ReSpec source:
https://github.com/w3c/mediacapture-depth/blob/mathjax-snapshot-example/index.html

The HTML generated by MathJax for the formulas is essentially a bunch of nested `<span>`s and a `<style>` block:

https://github.com/w3c/mediacapture-depth/blob/mathjax-snapshot-example/snapshot.html#L1203
https://github.com/w3c/mediacapture-depth/blob/mathjax-snapshot-example/snapshot.html#L15

As noted earlier, my only issue that remains is that validator fails with `Element nobr not allowed as child of element span in this context.`:

https://validator.w3.org/nu/?doc=https%3A%2F%2Frawgit.com%2Fw3c%2Fmediacapture-depth%2Fmathjax-snapshot-example%2Fsnapshot.html

For that, I opened a MathJax issue.

marcoscaceres:
Ok, so yeah - the problem is the validator issue. If Mathjax folks fix it, then you won't need the code from the web audio spec. However, if you do want to work around the validator you can use the web audio code.

anssiko:
I fixed the MathJax validation issues with a handcrafted [`make-valid.sh`][1]. I will run this script on the `respec2html` output before I push the spec to the GH repo that triggers Echidna.

This stopgap solution works for this particular spec since we don't publish on every commit, only when we push to a release branch.

[1]: https://github.com/w3c/mediacapture-depth/blob/mathjax-snapshot-example/make-valid.sh

marcoscaceres:
On April 14, 2016 at 11:15:24 PM, Anssi Kostiainen (notifications@github.com) wrote:
> I fixed the MathJax validation issues with a handcrafted [`make-valid.sh`][1]. I will  
> run this script on the `respec2html` output before I push the spec to the GH repo that triggers  
> Echidna.

Nice :) 

> This stopgap solution works for this particular spec since we don't publish on every  
> commit, only when we push to a release branch.
>  
> [1]: https://github.com/w3c/mediacapture-depth/blob/mathjax-snapshot-example/make-valid.sh  

Ok, at least that addresses the issue for now :)

anssiko:
I cherrypicked the script to `gh-pages` branch:
https://github.com/w3c/mediacapture-depth/blob/gh-pages/releases/make-valid.sh

I plan to keep it updated for the purposes of this spec as long as the upstream issue https://github.com/mathjax/MathJax/issues/1440 remains un addressed. I may add some further fixes if I discover other bugs. If someone feels like integrating such a hack into the Echidna toolchain, feel free to.

marcoscaceres:
@dontcallmedom wrote:
> bumping the version of respec to 4.x since it's a breaking change in the API of the tool (so is the WebIDL oldschool deprecation that seems to have been merged as well)

We haven't really used very good sem-versions in the past, and I've tied the milestones to different point releases. Going to 4.0 would mean I'd have to go an update all the milestones - or the milestones wouldn't match current ReSpec... in short, I would prefer we don't go to 4.0 and just keep riding the 3.2 release train.

marcoscaceres:
(we should start proper sem versioning in 4.0 tho)

anssiko:
Thanks to advise from @pkra I was able to fix the MathJax validation bug by using its CommonHTML output. Thus no need to hack around with `make-valid.sh` script anymore.

I attempted to publish with Echidna, failed with:
https://lists.w3.org/Archives/Public/public-tr-notifications/2016Apr/0024.html

Also

dontcallmedom:
@anssiko fwiw, the change discussed in this pull request hasn't landed yet (either in ReSpec or in Echidna).

@deniak and I looked into the failure, and it was (ironically) because the spec-generator was relying on a version of phantomjs that had issues with establishing an SSL connection with rawgit.com; https://github.com/w3c/spec-generator/commit/8eb969346fb083d83cc1ddcc773663b3f2738979 fixed it

anssiko:
Sorry for highjacking this issue -- on a related note, I line up to be a beta tester when you move this to production.

Continuing Echidna MathJax integration issues debugging in w3c/echidna#271.

anssiko:
Fixed the MathJax integration issues with mathjax-node, see https://github.com/w3c/echidna/issues/271#issuecomment-211390047

marcoscaceres:
@anssiko, great to hear!

PR #681, opened 2016-04-02
Feat (ui.js,save-html.js): Give UI elements ids

marcoscaceres:
This is part 1 of enabling nightmarejs. It gives certain elements IDs so we can target them more easily from nightmarejs. It also allows us to test live specs before fully migrating.


marcoscaceres:
Carrying over review from #678.

PR #688, opened 2016-04-12
Remove class 'toc-sidebar' from <body> when generating a snapshot

### LINKED ISSUE #644: Save snapshot leaves fixup.js generated content in the saved document
This results in two TOC toggles being inserted.  Arguably a bug in fixup.js, but we can address it here too by removing the node with ID toc-nav

deniak:
As mentioned in https://github.com/w3c/tr-design/pull/91#issuecomment-208467559, the generated snapshot shouldn't add the class 'toc-sidebar' to the `<body>`. It's added by JS only.
Shoud fix #644

### DIFF
--- a/js/ui/save-html.js
+++ b/js/ui/save-html.js
@@ -9,6 +9,7 @@ define(
         var cleanup = function (rootEl) {
             $(".removeOnSave", rootEl).remove();
             $("#toc-nav", rootEl).remove() ;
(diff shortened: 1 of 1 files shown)


marcoscaceres:
LGTM, the generated CI errors are unrelated. Need to fix those.

marcoscaceres:
Will ship soon...

PR #692, opened 2016-04-14
WIP(respec2html): Externalize fetch and write

marcoscaceres:
@dontcallmedom WDTY? 

```JS
const fetchAndWrite = require("./tools/respecDocWriter.js").fetchAndWrite;
const src = "https://w3c.github.io/manifest/";
const out = "path/to/out"; // or null for stdout
const whenToHalt = {
  haltOnError: true,
  haltOnWarn: true,
};
const timeout = 10000;
fetchAndWrite(src, out, whenToHalt, timeout).then(() => console.log("done") );
```

### DIFF
--- a/tools/respec2html.js
+++ b/tools/respec2html.js
@@ -4,7 +4,7 @@
 "use strict";
 const async = require("marcosc-async");
 const colors = require("colors");
-const os = require("os");
+const fetchAndWrite = require("./respecDocWriter").fetchAndWrite;
 colors.setTheme({
   data: "grey",
   debug: "cyan",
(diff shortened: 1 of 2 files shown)


marcoscaceres:
(updated example above to show it returns a promise)

marcoscaceres:
@dontcallmedom, have a play and let me know if it's ok for the w3c spec-generator.

dontcallmedom:
for the spec generator, instead of writing to a file it would be best if I could have the content itself (since I'm serving it back via http)

marcoscaceres:
Np, I'll return it with the promise. Otherwise, just set out to null and pipe it straight from process.stdout

dontcallmedom:
Hi @marcoscaceres I don't think respec2html.js has been updated to return the data directly yet, but I need to update the spec generator; since new respec has been released, this is kind of urgent now

marcoscaceres:
@dontcallmedom, correct. Let me do that now for you 🔧

PR #699, opened 2016-04-19
Fix(w3c/style): WG-Note CSS doesn't link (closes #698)

### LINKED ISSUE #698: WG-Note CSS doesn't link properly
With the 2016 stylechange, the url ending with /Note now returns a 300 Multiple choice.

marcoscaceres:


marcoscaceres:
@halindrome r?

This also fixes that silly requirement of the status be being case sensitive for style selection. That's just annoying.

js/w3c/style.js
```diff
@@ -73,15 +73,16 @@ define(
             case "LC-NOTE":
               styleFile += "WD";
               break;
+            case "WG-NOTE":
             case "FPWD-NOTE":
-              styleFile += "WG-NOTE";
+              styleFile += "WG-NOTE.css";
```
halindrome: Why did you add .css to this?  I thought they were using conneg to get the file we want.

marcoscaceres: > On 20 Apr 2016, at 8:31 AM, Shane McCarron <notifications@github.com> wrote:
> 
> In js/w3c/style.js:
> 
> >              case "FPWD-NOTE":
> > -              styleFile += "WG-NOTE";
> > +              styleFile += "WG-NOTE.css";
> Why did you add .css to this? I thought they were using conneg to get the file we want.
> 
It seems the w3c changed their minds about conneg: hence this change. Without the '.css', it no longer returns the style sheet, but a 300 page.

halindrome: Actually it works fine without the .css.  I just tried it.  What doesn't work is http: - it redirects to https.

marcoscaceres: How odd? This is what I get?

screenshot 2016-04-20 12 47 02

Why am I seeing something different to you?

halindrome: Hmm - yes, you are:

screenshot 2016-04-19 at 22 01 12

halindrome:
If I were you I would file a bug against tr-design

js/w3c/style.js
marcoscaceres: That's messed up. However, if the "common basename" is .css, then doesn't that make it the canonical one?

marcoscaceres:
> If I were you I would file a bug against tr-design

will do, but it it could be ages before the W3c fix stuff. I'd rather fix this for ourselves + .css is that canonical one regardless.

marcoscaceres:
@halindrome, filed https://github.com/w3c/tr-design/issues/96

marcoscaceres:
rebased on top of #701

halindrome:
OK.   LGTM.  I note, however, that ALL of these have a canonical version that ends in .css.  Why not be explicit on all of them?

marcoscaceres:
> On 20 Apr 2016, at 11:32 PM, Shane McCarron <notifications@github.com> wrote:
> 
> OK. LGTM. I note, however, that ALL of these have a canonical version that ends in .css. Why not be explicit on all of them?
> 
Valid point. The one I fixed was the only one that didn't work for me tho.

PR #700, opened 2016-04-19
Fix(ui): dont show till respecIsReady (closes #410)

### LINKED ISSUE #410: Do not enable ReSpec bubble and/or Save Snapshot until "RESPEC DONE!"
During a bibrefs outage, I found that I could Save Snapshot even though processing was incomplete. I only discovered this when diffing the saved file. When I investigated, I found that the first two RESPEC log messages were present but "RESPEC DONE!" was not. It appeared that processing was hung waiting for the bibrefs server, but everything appeared fine (unless you opened the console and knew to look for "RESPEC DONE!"

To help authors avoid incomplete versions, I propose not enabling the Save Snapshot option until processing is actually complete.

Note: If I recall correctly, the server request eventually timed out. However, I didn't take note of the resulting behavior. If it doesn't already, ReSpec should ensure that such processing step failures are always result in the red error count displayed with the ReSpec bubble.

marcoscaceres:
@halindrome r?


marcoscaceres:
Added tests..

js/core/css/respec2.css
```diff
@@ -108,6 +108,10 @@ table.simple {
     margin-bottom:  0;
 }
 
+.hidden {
```
halindrome: I don't think we can add a class with a name like that.  It is far too likely to collide with something our users have.  True, their thing is likely the same, but we need to be respectful of the hosted document's namespace.  I suggesting using respec-hidden

marcoscaceres: > On 19 Apr 2016, at 11:53 PM, Shane McCarron <notifications@github.com> wrote:
> 
> In js/core/css/respec2.css:
> 
> > @@ -108,6 +108,10 @@ table.simple {
> >      margin-bottom:  0;
> >  }
> >  
> > +.hidden {
> I don't think we can add a class with a name like that. It is far too likely to collide with something our users have. True, their thing is likely the same, but we need to be respectful of the hosted document's namespace. I suggesting using respec-hidden
> 
Good point! Will fix

halindrome:
Why are the tests failing though?

marcoscaceres:
> On 20 Apr 2016, at 8:39 AM, Shane McCarron <notifications@github.com> wrote:
> 
> Why are the tests failing though?
> 

Requirejs path :( should be a one line fix, but always takes a bunch of pain to sort out.

marcoscaceres:
Rebased on top of #701

halindrome:
LGTM after rebase.

PR #707, opened 2016-04-21
Feat(respecDocWriter): improvements (closes #706)

### LINKED ISSUE #706: Improvements to respecDocWriter
While updating the spec generator to use respecDocWriter, I noted the following issues:
* it doesn't seem to fail gracefully (i.e. with a promise rejection) when it fails to download the url (I got on a 404 URL: "Unhandled error in async function. [Error: .wait() timed out after 30000msec]")
* when out is null, it prints out the generated HTML on stdout; at least in my case, I only need it as a returned value, not on stdout

marcoscaceres:
* halt on 404
 * don't output when out is the emtpy string

### DIFF
--- a/tests/headless.js
+++ b/tests/headless.js
@@ -57,7 +57,7 @@ const runRespec2html = async(function* (server) {
 
   // Incrementally spawn processes and add them to process counter.
   const executables = sources.map((source) => {
-    let cmd = `node ./tools/respec2html.js -e --timeout 10 --src ${server}/examples/${source} > /dev/null`;
+    let cmd = `node ./tools/respec2html.js -e --timeout 10 --src ${server}/examples/${source} --out /dev/null`;
     return cmd;
   }).map(
     toExecutable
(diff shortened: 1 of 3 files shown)


marcoscaceres:
@dontcallmedom r? Merge if happy. Going to 💤

dontcallmedom:
testing this quickly, I got "Unhandled error in async function. [TypeError: Cannot read property 'code' of undefined]"; I'll try to get something more specific, but in the meantime, I can leave with #706 unfixed, but would like a new respec release today if that's at all possible

marcoscaceres:
Fixed the headless error

marcoscaceres:
@dontcallmedom, strange about that code error. `npm update` just in case?

dontcallmedom:
ok, the code error seems to gone after a npm update; but the code still doesn't handle 404 gracefully: "Unhandled error in async function. [Error: HTTP Error 404: ...]"

marcoscaceres:
@dontcallmedom, the ""Unhandled error in async function. [Error: HTTP Error 404: ...]"" error is actually a bug in Node.js's implementation of Generators.

PR #708, opened 2016-04-21
Feat(respecDocWriter): throw if not a ReSpec doc (closes #704)

### LINKED ISSUE #704: respec2html should error if not a ReSpec document
@tripu suggested we make respec2html stop if ReSpec is not found.... which... makes complete sense. Right now, it just times out leaving the end user feeling empty.

marcoscaceres:


marcoscaceres:
@tripu or @dontcallmedom, wdty?

marcoscaceres:
I'll make it look like this: 

screenshot 2016-04-22 02 02 32

marcoscaceres:
Argh! Need an emoji in there too!

tools/respecDocWriter.js
```diff
@@ -96,6 +96,19 @@ const tasks = {
         const msg = `HTTP Error ${response.code}: ${url}`;
         throw new Error(msg);
       }
+      const isRespecDoc = yield nightmare
+        .wait(function(){
+          return document.readyState === "complete";
+        }) 
+        .evaluate(function(){
+          const selector = "script[src='https://www.w3.org/Tools/respec/respec-w3c-common']";
```
dontcallmedom: can't you test the presence of a ReSpec global (assuming there is still such a thing)? 

Relying the the URL detection makes the tool more brittle (I know I've used it on locally hosted version of respec for instance)

marcoscaceres: I'll add that as first check. Problem is that I've seen people add ReSpec asycn. I'll add as first check:

```
          if(document.hasOwnProperty("respectIsReady")){
            return true;
          }
```

marcoscaceres:
Even better with moar emoji... even @nrooney would approve. 
 
screenshot 2016-04-22 02 57 30

Issue #581, opened 2016-03-01
Should update markdown 


marcoscaceres:
We have an old version of this: 
https://github.com/chjj/marked

PR #712, opened 2016-04-27
Just to put this on your radar

nickevansuk:
Hi Marcos,

Just to let you know that the new version of the marked library includes table support for markdown, which you can activate if you switch to gfm mode. It also works quite nicely with highlight.js which I've also added for markdown mode (though understand you already have google-code-prettify in there so maybe don't want to conflate the two for the W3C usecase).

Anyway here's the code - let me know if you want me to cherry pick some of the changes into a proper pull request for you. I still need to fix the unit tests to work with gfm so it's not ready as-is.

We're a new W3C CG and we're using Respec for our specs - so will continue to pull request changes we make in case they're useful. Let me know if this is of interest.

Great work on Respec by the way! Robin's left a solid legacy with this, so well done for carrying the torch!

Thanks,

Nick


marcoscaceres:
@nickevansuk, this is great to see. I had filed a bug to do this (https://github.com/w3c/respec/issues/581), but ran into difficulties on my first attempt - so it's great to see you got it working. 

Would love to coordinate on upgrading marked.js and the highlighter.

marcoscaceres:
Closing this for now - but looking forward to collaborating.

PR #718, opened 2016-04-28
Feat(global): use CDN for jQuery (closes #636)

### LINKED ISSUE #636: Use CDN for jquery
It might be smarter to load jQuery from a CDN. Given the high cache hit rate for jquery served from google, we should save significant time just using Google's CDN version. 

The trade off is that we lose control over the version, but could be worth it - specially as we look at reducing our reliance on jQuery.

marcoscaceres:
* jQuery now loaded securely from JQuery CDN with added SRI

### DIFF
--- a/js/profile-w3c-common.js
+++ b/js/profile-w3c-common.js
@@ -1,4 +1,13 @@
 "use strict";
+/**
+ * Sub resource integrity strings can be dynamically added to each node
+ * as RequireJS loads them from a CDN.
+ *
+ */
+var sriTable = {
+  "https://code.jquery.com/jquery-2.2.3.min.js": "sha256-a23g1Nt4dtEYOj7bR+vTu7+T8VP13humZFBJNIYoEJo="
+};
+
 // this is only set in a build, not at all in the dev environment
 require.config({
   shim: {
@@ -12,7 +21,7 @@ require.config({
   paths: {
     "fetch": "/node_modules/whatwg-fetch/fetch",
     "handlebars": "/node_modules/handlebars/dist/handlebars",
-    "jquery": "/node_modules/jquery/dist/jquery",
+    "jquery": "https://code.jquery.com/jquery-2.2.3.min",
     "Promise": "/node_modules/promise-polyfill/promise",
     "webidl2": "/node_modules/webidl2/lib/webidl2",
   },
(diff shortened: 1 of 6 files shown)


marcoscaceres:
@dontcallmedom or @tripu, will it be ok to load jQuery from their CDN?

dontcallmedom:
To make sure I parse this right, jquery is only retrieved from the CDN when building ReSpec, not when a ReSpec document is being rendered, right?

marcoscaceres:
@dontcallmedom, correct.

dontcallmedom:
as @halindrome points out in #636 I'm not quite sure what the advantage is (vs using npm); it adds a new point of failure in the build process, and doesn't gain us much.

The patch looks good otherwise, but I guess I would prefer to understand its goal better.

marcoscaceres:
@dontcallmedom, I'm trying to reduce the size of ReSpec itself. The advantage is in caching. Basically, this shaves almost 23% of the size of ReSpec: it goes from 389Kb to 303Kb.

dontcallmedom:
I'm confused then; this seems to imply that jquery *is* retrieved at rendering time (although the retrieval might be fast thanks to caching/CDN).

It may be a perf gain, but it's not even that clear; we would have to be using a popular version of jquery on a popular CDN, and hope that its caching is more robust in our user base than the caching of ReSpec as a whole (which for people that jumps from w3c specs to w3c specs may not be obvious)

Beyond the non-obvious benefit, I'm also somewhat reluctant because adding a new network dependency at rendering time creates more risks for breakage (e.g. in a Travis context).

marcoscaceres:
For instance, someone (i.e., me) going to:
https://w3c.github.io/manifest/

Would see a small performance gain. That is good. 

Someone going to:
https://www.w3.org/TR/appmanifest/

Would see no difference. 

> Beyond the non-obvious benefit, I'm also somewhat reluctant because adding a new network dependency at rendering time creates more risks for breakage (e.g. in a Travis context).

This is true. That is a risk.

marcoscaceres:
Ok, let's stick with current. I'm going to see if I can get a custom build of jQuery, which might also help.

marcoscaceres:
@dontcallmedom thanks for the time and feedback. Was quite helpful 💐

PR #719, opened 2016-04-28
Feat(package): run tests in Safari Tech Preview

marcoscaceres:
* Enables the ability to test on Safari Tech Preview if available on the system.


marcoscaceres:
This is mainly just for me - for running tests locally on MacOS with latest Safari.

PR #727, opened 2016-05-02
WIP (appveyor): seeing if this works...

marcoscaceres:


marcoscaceres:
Ok, so right now the best we can do wait for Edge to become testable. This, at least, takes the first steps.

PR #728, opened 2016-05-02
Fix (biblio): Edge doesn't support includes yet

marcoscaceres:


marcoscaceres:
@halindrome r?

halindrome:
Your fix is fine.  I wonder if there isn't a polyfill so you can just use it though.

marcoscaceres:
I'd like us to eventually drop all polyfills. We are really close, After APPLE WWDC, all browsers should support like 95% of what we use. By next year, hopefully 100%. 🙏

halindrome:
Okay.  I appreciate your goals.  I am old school.  I think we need to support the last 3 generations of user agents as long as people are using live respec for editors drafts.  Otherwise people who are trying to socialize their documents are going to run into trouble with the broader community.

Case in point.  The Web Payments activity has A LOT of players who are in the financial industry.  That industry is notoriously slow about updating their user agents.  Some of my financial customers only just dropped their requirement that I support IE 6.

marcoscaceres:
> On 3 May 2016, at 1:40 AM, Shane McCarron <notifications@github.com> wrote:
> 
> Okay. I appreciate your goals. I am old school. I think we need to support the last 3 generations of user agents as long as people are using live respec for editors drafts.
> 

I can't imagine anyone working on web technology not running the latest browsers (for security reasons - and basically because we are specifying the next generation Web). 

However, it would be great to get some numbers. 

> Otherwise people who are trying to socialize their documents are going to run into trouble with the broader community.
> 
I agree that generated output should be backwards compatible - but the source files are too hard to maintain. At least, I would not be interested in participating as a ReSpec maintainer given such a requirement. I use ReSpec as my "fun" project - and having to worry about old browsers would take all the fun out of it for me. 

> Case in point. The Web Payments activity has A LOT of players who are in the financial industry. That industry is notoriously slow about updating their user agents. Some of my financial customers only just dropped their requirement that I support IE 6.
> 
Sure - but Echidna could address that (I.e., don't share the source files): auto publish generated content. 

(I can't believe the people we trust with our financial system are running IE6 - it's 2008 all over again:( )

PR #732, opened 2016-05-03
Fix pre issue + update marked.js

marcoscaceres:


marcoscaceres:
@tobie, git blame says this is mostly your baby. Hopefully I didn't break it. Would you mind having a looksy.

js/core/markdown.js
```diff
@@ -39,16 +39,29 @@
...
 define(
-    ['core/marked'],
-    function (coreMarked) {
+    ['marked'],
+    function (marked) {
         marked.setOptions({
-            gfm: false,
-            pedantic: false,
-            sanitize: false
+            sanitize: false,
```
marcoscaceres: the ones I removed default to either true or false.

js/core/markdown.js
```diff
@@ -39,16 +39,29 @@
...
         });
 
+        function nonEmpty(item){
+            return item.trim();
+        }
+
+        // We need deduce (reduce) the minimum column size
```
marcoscaceres: This used to reduce.... need to update comment.

js/core/markdown.js
```diff
@@ -133,73 +146,102 @@ define(
...
+            //     -----------
+            //
+            //     <pre>
+            // if (w.thing) {
+            //   x++;
+            // }
+            // </pre>
+            // </div>
+            normalizePadding: function(text) {
+                // How much we should chop from the left.
```
marcoscaceres: These comments need updates... oops

js/core/markdown.js
```diff
@@ -133,73 +146,102 @@ define(
...
+                // Now that we know the left column size, we can chop that
+                // off everywhere, except inside pre-tags.
+                var inPre = false; // Are we in pre-element?
+                var firstPaddedLine = text.split("\n").filter(nonEmpty).shift();
+                var baseTrim = calculateLeftPad(firstPaddedLine);
+                var result = text
+                  .split("\n")
+                  .reduce(function (collector, item, index, array) {
```
marcoscaceres: items declared but not all used.

js/core/markdown.js
```diff
@@ -208,7 +250,12 @@ define(
                 $('.issue, .note, .req', doc).each(function() {
                     div.innerHTML = self.toHTML(this.innerHTML);
                     this.innerHTML = '';
-                    var node = div.firstChild;
+                    // Same element, don't nest
```
marcoscaceres: this probably needs a better explanation.

package.json
```diff
@@ -70,7 +70,9 @@
     "command-line-args": "^2.1.6",
     "command-line-usage": "^2.0.4",
     "fs-promise": "^0.5.0",
+    "highlight.js": "^9.3.0",
```
marcoscaceres: should remove this... do a different PR to enable it.

tobie:
Your best bet is to compare https://coremob.github.io/coremob-2012/ with both scripts.

marcoscaceres:
@tobie, ah, super helpful... some breakage 😢 
screenshot 2016-05-03 20 40 19

halindrome:
I am concerned about a comment you made in the spec-prod email about IDs on section elements no longer being provided.  I wonder about the A11Y impact of that on screen readers and navigation.  Also on other parts of ReSpec or plugins that might expect sections to have IDs.  @ljwatson can you provide any insight on the A11Y front?

marcoscaceres:
> Also on other parts of ReSpec or plugins that might expect sections to have IDs

Very few people use markdown, afaict.

halindrome:
I am sure you are right.  I had forgotten it was supported.

marcoscaceres:
Specs I've found - mostly by @igrigorik:
* https://w3c.github.io/sensors/usecases.html
* https://w3c.github.io/frame-timing/
* http://w3c.github.io/netinfo/
* https://w3c.github.io/preload/
* https://w3c.github.io/requestidlecallback/
* https://w3c.github.io/resource-hints/

marcoscaceres:
Found via:
https://github.com/search?l=html&p=1&q=org%3Aw3c+format+markdown&type=Code

Will check those tomorrow and see what I broke with Core Mob.

marcoscaceres:
hmm... seems I need to convert all HTML nodes except section into text, then parse them as markdown. should be easy to do... famous last words.

marcoscaceres:
Ok, this seems happy

marcoscaceres:
And it passes!!!! 💃 woot woot!

js/core/markdown.js
```diff
@@ -1,280 +1,232 @@
...
+ *       <h2>Title</h2>
+ *       <section>
+ *         <h3>Subtitle</h3>
+ *         <p>Here's some text.</p>
+ *       </section>
+ *       <section>
+ *         <h3>Another subtitle</h3>
+ *         <p>More text.</p>
+ *       </section>
+ *     </section>
```
tobie: The `//` to `/* ... */` style change makes it super hard to figure out if the content changed or not.

marcoscaceres:
I've checked the spec list also. They all seem to parse correctly.

marcoscaceres:
@tobie, maybe you could give it a once over? let me know if it looks ok to merge.

js/core/markdown.js
```diff
@@ -1,280 +1,232 @@
...
-//         <h3>Another subtitle</h3>
-//         <p>More text.</p>
-//       </section>
-//     </section>
-//
-
-define(
-    ['core/marked'],
-    function (coreMarked) {
-        marked.setOptions({
```
tobie: So there's something fishy going on here with regards to the global `marked` object and the module `coreMarked` object. My hunch is that I had reasons to do this. I'm not sure why I didn't document them, though. It could also just be a refactoring mistake that I failed to notice because I was then hitting the global marked object.

js/core/markdown.js
marcoscaceres: True. Sorry about that.

marcoscaceres:
@tobie I only did very minor refactor of `makeBuilder()`...I couldn't figure it out... was super fancy impressive tho :) The rest I rewrote and chopped out a lot of code.

js/core/markdown.js
marcoscaceres: Will check. Remember  I also updated marked... so there should not be a global (coreMarked).

tobie:
> I only did very minor refactor of makeBuilder()...I couldn't figure it out... was super fancy impressive tho :) The rest I rewrote and chopped out a lot of code.

I wrote this a while back, now, so memory fails me, but iirc, the reason for this is that the rest of the pipeline expects nested sections, so you basically have to recreate the nesting from the flat H1-6 hierarchy that the markdown parser provides.

js/core/utils.js
```diff
@@ -14,6 +14,61 @@ define(
...
+                        return regex.test(text);
+                    }
+                }
+                if (!text) {
+                    return "";
+                }
+                if (typeof text !== "string") {
+                    throw TypeError("Invalid input");
+                }
+                var hasOpeningPre = testRegex(/<pre\b[^>]*>/i);
```
tobie: Conceptually hostile to the idea of recreating your own mini HTML parser using regexes here given you're in a browser and can leverage the native one. Maybe that's the onlu sane solution, though.

js/core/markdown.js
```diff
@@ -1,280 +1,232 @@
...
+      msg.pub("start", "core/markdown");
+      if (conf.format === 'markdown') {
+        // We transplant the UI to do the markdown processing
+        var rsUI = doc.getElementById("respec-ui");
+        rsUI.remove();
+        // Marked expects markdown be flush against the left margin
+        // so we need to normalize the inner text of all block
+        // elements.
+        processSections(doc);
+        processIssuesNotesAndReqs(doc);
```
tobie: Curious as to why this suddenly needs to become uncommented. Has marked stopped parsing markdown in nested markup?

tests/spec/core/markdown-spec.js
```diff
@@ -8,14 +8,18 @@ describe("Core - Markdown", function() {
     var ops = {
       config: makeBasicConfig(),
       body: makeDefaultBody() +
-       "\nFoo\n===\n",
+       "\n\nFoo\n===\n",
```
tobie: Why are you fixing tests, here?

tests/spec/core/markdown-spec.js
```diff
@@ -41,69 +45,86 @@ describe("Core - Markdown", function() {
...
+      expect(doc.querySelector(".req h3")).toBeTruthy();
     }).then(done);
   });
 
   it("should remove left padding before processing markdown content", function(done) {
     var ops = {
       config: makeBasicConfig(),
       body: makeDefaultBody() +
-       "\n    Foo\n    ===\n",
+       "\n\n  Foo\n  ===\n    * list item 1\n    * list item 2\n      * nested list item",
```
tobie: Changing the test here seems weird.

I was testing for:

```md
    Foo
    ===

```

You're now testing for:

```md

  Foo
  ===
    * list item 1
    * list item 2
      * nested list item
```

which doesn't seem to test for the same thing.

tests/spec/core/markdown-spec.js
```diff
@@ -41,69 +45,86 @@ describe("Core - Markdown", function() {
...
+      expect(nestedLi.textContent).toEqual("nested list item");
     }).then(done);
   });
 
   it("should structure content in nested sections with appropriate titles", function(done) {
     var ops = {
       config: makeBasicConfig(),
       body: makeDefaultBody() +
-       "\nFoo\n===\n\nBar\n---\n\nBaz\n---\n\n### Foobar ###\n\n#### Foobaz ####\n\nZing\n---\n\n",
+       "\n\nFoo\n===\n\nBar\n---\n\nBaz\n---\n\n### Foobar ###\n\n#### Foobaz ####\n\nZing\n---\n\n",
```
tobie: Same here, why add this extra `\n`?

tobie:
Alright--done with the questions/comments here.

tests/spec/core/markdown-spec.js
marcoscaceres: This was not valid markdown (I was surprised too!). It causes the section above to be wrapped in a p element.

marcoscaceres:
Thanks @tobie! Great feedback. Will fix up!

I'm wanting to add language-based syntax highlighting next via highlight.js 

Moar goodies coming soon
🍬🍦🍭🍩🍪🍡🍢🍮🍧

tests/spec/core/markdown-spec.js
marcoscaceres: I screwed up the processing where I was chopping off too much off the front of strings. This test helped (me) make sure I didn't do that: hence the list inside the list. 

Tl;dr: I broke your code (which was fine); this is how I made sure I fixed it.😊

tests/spec/core/markdown-spec.js
marcoscaceres: As before. Maybe it's a bug in marked.js?, but the output is not what you would expect. You can try it directly in the marked() parser independently of respec.

tests/spec/core/markdown-spec.js
tobie: What bothers me is that this test was making sure that 4 spaced indented content was treated as a massive code block.

Given this is now a 2 space indent, you're no longer testing for the same thing.

tests/spec/core/markdown-spec.js
tobie: That's weird. Not how CommonMark works, see http://spec.commonmark.org/dingus/?text=%0A%0AFoo%0A%3D%3D%3D%0A and http://spec.commonmark.org/dingus/?text=Foo%0A%3D%3D%3D%0A.

tobie: Not able to reproduce this in http://www.javascriptoo.com/marked

tests/spec/core/markdown-spec.js
marcoscaceres: Ah! My bad. Will fix

tests/spec/core/markdown-spec.js
marcoscaceres: I'll post what I'm getting locally tomorrow.

js/core/utils.js
marcoscaceres: I'll give this some thought... I could work with a live DOM, and check if `.parentElement` is `<pre>`

js/core/markdown.js
marcoscaceres: Will add the comments back in. I'm actually going to flip the processing order too: 

1. parse the whole document.
1. parse the sections.
1. parse the notes, reqs, etc.... (and I can do the above in one go, don't as section and note/req parsing is identical)

tests/spec/core/markdown-spec.js
marcoscaceres: Ok, this is kinda f'ed up...  
```JS
// Test 1 - what we currently 
marked('<section id="abstract"><p>test abstract</p></section><section id="sotd">\n\nFoo\n===')
"<p><section id="abstract"><p>test abstract</p></section><section id="sotd"></p>
<h1 id="foo">Foo</h1>
"
// Test 2 - original
marked('<section id="abstract"><p>test abstract</p></section><section id="sotd">\nFoo\n===')
"<p><section id="abstract"><p>test abstract</p></section><section id="sotd"></p>
<h1 id="foo">Foo</h1>
"
```
So, you are correct that the original was fine (actually, both are fine).... but marked is spitting out garbage.

js/core/markdown.js
marcoscaceres: Checked. Marked no longer leaks into global scope.

marcoscaceres:
@tobie, I rewrote it to use a HTML parser as you suggested. You were right: using regex for parsing HTML a disgrace.

marcoscaceres:
* was a

PR #758, opened 2016-05-13
Fix(respec2.css): Override code highlighter background

marcoscaceres:
Context: the syntax highlighter uses ".hljs" as its CSS class - but the syntax highlighting themes all have their own custom backgrounds. This makes specs look ugly, so I'm forcing ReSpec to set the background to transparent.

### DIFF
--- a/js/core/css/respec2.css
+++ b/js/core/css/respec2.css
@@ -3,8 +3,13 @@
  * Robin Berjon - http://berjon.com/
  *****************************************************************/
 
+/* Override code highlighter background */
+.hljs {
+    background: transparent !important;
+}
+
 /* --- INLINES --- */
-em.rfc2119 { 
+em.rfc2119 {
     text-transform:     lowercase;
     font-variant:       small-caps;
     font-style:         normal;


marcoscaceres:
Puts in place the css override infrastructure for highlighter.

marcoscaceres:
@dontcallmedom, here is another small one, if you want to review :)

dontcallmedom:
Well, a tad hard to review without context :)

Why are you forcing all `pre` to a transparent background? any adverse effect it could have on specs which would use another background color?

(maybe the commit message should say more about the reason for the change)

marcoscaceres:
> Why are you forcing all pre to a transparent background? any adverse effect it could have on specs which would use another background color?

Ah, well spotted. I thought for some reason that that was in the original style sheet. I really just want to be overriding `.hljs`. Fixing.

marcoscaceres:
Added some context above

marcoscaceres:
Oki, fixed:)

dontcallmedom:
LGTM; feel free to merge once Travis completes

marcoscaceres:
Example of output... 

screenshot 2016-05-13 19 11 19

PR #759, opened 2016-05-13
Feat(markdown): teach markdown how to highlight code

marcoscaceres:
Here we actually start teaching the markdown parser how to highlight markdown. This means we can do cool things like triple back tick  + language detection, just like in Github.


marcoscaceres:
Merging to avoid bit rot. Post merge review still appreciated.

PR #762, opened 2016-05-16
Feat(save-html): remove XHTML 1.0 save support

marcoscaceres:
This change is [Reviewable](https://reviewable.io/reviews/w3c/respec/762)

### DIFF
--- a/js/ui/save-html.js
+++ b/js/ui/save-html.js
@@ -113,21 +113,10 @@ define(
             id: "respec-save-as-xhtml5",
                     fileName: "index.xhtml",
                     popupContent: function () {
-                        self.toXHTMLSource(5);
+                        self.toXHTMLSource();
                     },
                     title: "Save as XHTML5",
-                    url: this.htmlToDataURL(this.toXML(5)),
-                });
-
-                // XHTML 1.0
-                addButton({
-            id: "respec-save-as-xhtml",
-                    fileName: "index.xhtml",
-                    popupContent: function () {
-                        self.toXHTMLSource(1);
-                    },
-                    title: "Save as XHTML 1.0",
-                    url: this.htmlToDataURL(this.toXML(1)),
+                    url: this.htmlToDataURL(this.toXML()),
                 });
 
                 // ePub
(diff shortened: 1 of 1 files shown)


halindrome:
I don't think this is a good idea.  There are a number of specs I personally produced that use this feature, and if any of those need to be updated in the future (errata or whatever) it would be impossible.

marcoscaceres:
> On 16 May 2016, at 9:24 PM, Shane McCarron <notifications@github.com> wrote:
> 
> I don't think this is a good idea. There are a number of specs I personally produced that use this feature,
> 

Can I have a look? What's the reason you are still using it over xhtml5? 

> and if any of those need to be updated in the future (errata or whatever) it would be impossible.
> 

Well, you'd have to update the output... But not impossible. Question is, is any spec published in xhtml actually used nowadays or has been updated in the last 3 years?

halindrome:
Updated in the last 3 years?  Probably.  TBH it all runs together.  XHTML+RDFa, for example.  Although oddly I don't see a link to it in TR.  But that was March 2015.  

https://www.w3.org/2010/02/rdfa/drafts/2015/REC-xhtml-rdfa-20150317/Overview.xhtml

marcoscaceres:
> I appreciate that you feel there is a cost to having code around in that it needs maintenance, but no one has touched this part of the code in years.

My concern is more that it removes structural tags from HTML5. It also encourages the use of outdated technology (XHTML) which we are trying to remove from the Web. Both are bad, IMO.

marcoscaceres:
* that should be XHTML 1.0, as per the very old spec that defined it. I don't mind so much XHTML5, as it supports the new structuring elements.

marcoscaceres:
@halindrome, I'm still not clear why one would publish with XHTML 1.0 even for updated errata (instead of publishing with newer HTML). Can you help me understand?

marcoscaceres:
Rebased...

marcoscaceres:
@halindrome, ping

halindrome:
Because the specs in question are actually markup languages based upon XHTML 1.0.  So a version in that format is a thing that needs to be published.

marcoscaceres:
> Because the specs in question are actually markup languages based upon XHTML 1.0. So a version in that format is a thing that needs to be published.

Sorry, I'm honestly a bit confused. Those documents also published as PDF, PostScript, etc... so I'm confused as to what the relationship is between what the spec is published and RDFa itself. 

Is the expectation that the RDFa specifications themselves are machine processable in such a way that they describe the language somehow? Like... I could take an RDF parser and pull in the spec and then somehow know RDFa?

marcoscaceres:
Note, I added "what format" above.

halindrome:
XHTML+RDFa is a markup language based upon XHTML Modularization.  Part of the publication process for XHTML+RDFa (and other XHTML grammars) is that they are published using their grammar.  If you look at the top of XHTML+RDFa you will see there is a link to a version that validates as XHTML+RDFa.

https://www.w3.org/TR/xhtml-rdfa/

dontcallmedom:
Publishing XHTML specs as XHTML is a nice to have, not a requirement.

But more generally speaking, I don't think we should force respec to be backwards compatible with any document ever published using respec for ever; specific (and hypothetical) cases can be dealt by using a fixed version of respec known to be compatible, or by building/maintaining a separate profile, or by converting from HTML5 to other flavors with some other mechanism, etc.

marcoscaceres:
@halindrome maybe a compromise is to label the save button "(deprecated)"? That allows you to keep using it, but discourages others from using the old format.

halindrome:
I am open to that.  I am also open to hiding the option by default - but
that might confuse anyone who needs it.  Would require documentation.

On Fri, May 27, 2016 at 3:04 AM, Marcos Cáceres <notifications@github.com>
wrote:

> @halindrome <https://github.com/halindrome> maybe a compromise is to
> label the save button "(deprecated)"? That allows you to keep using it, but
> discourages others from using the old format.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/w3c/respec/pull/762#issuecomment-222085490>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/AAfx8Lal1EsntcMB1xg_qkNCFFd5ldF9ks5qFqVzgaJpZM4Ie_Jd>
> .
>



-- 
Shane McCarron
halindrome@gmail.com

marcoscaceres:
Ok, so the least painful option is just to label the option as deprecated. I'll send a new PR. We can revisit removing the code in a year or so, if no further publications are made in this format.

PR #766, opened 2016-05-17
Feat(utils): add makeOwnerSwapper()

marcoscaceres:
Utility function that allows swapping a Node from one document to another. Useful for various incoming improvements.

### DIFF
--- a/js/core/utils.js
+++ b/js/core/utils.js
@@ -14,6 +14,35 @@ define(
                 msg.pub("end", "core/utils");
                 cb();
             }
+        ,
+        /**
+         * Allows a node to be swapped into a different document at
+         * some insertion point(Element). This function is useful for
+         * opportunistic insertion of DOM nodes into a document, without
+         * first knowing if that is the final document where the node will
+         * reside.
+         *
+         * @param  {Node} node The node to be swapped.
+         * @return {Function} A function that takes a document and a new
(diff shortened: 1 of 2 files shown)


marcoscaceres:
r please?

marcoscaceres:
Merging, as it's causing me bit-rot and can't proceed without it. Would still appreciate a once over.

halindrome:
Sorry, I have been heads down parenting for a week or so.  Question: do you know the level of support for adoptNode.  I had not seen it before, and the DOM Recommendation has text about what to do if it is not supported (which, I suppose, we could just do.  It seems simple enough).

marcoscaceres:
> On 18 May 2016, at 10:45 PM, Shane McCarron <notifications@github.com> wrote:
> 
> Sorry, I have been heads down parenting for a week or so. Question: do you know the level of support for adoptNode.
> 

It's a pretty ancient API (like late 1990s), so will be well supported. At least, not had any problems in the browsers on my machine. It's been in Firefox since version 3. 

> I had not seen it before, and the DOM Recommendation has text about what to do if it is not supported (which, I suppose, we could just do. It seems simple enough).
> 
The importNode() stuff? It's probably not needed, tbh. At least, I haven't had any issues as the sub documents we create are all spawned from the same document.

halindrome:
Excellent.  Thanks

PR #767, opened 2016-05-17
Feat(hightlight): use new highlighter

marcoscaceres:
depends on #766, so won't pass until that one is merged. 

I've simplified highlighting a bit more. As we have been warning people about deprecated classes since 2013, I've removed that warning code related to "sh_css sh_html sh_javascript sh_javascript_dom sh_xml". Highlighting now happens by default, instead of requiring eds. to put "highlight" everywhere.   

Disabling highlight can be done with "nohighlight".


marcoscaceres:
Review please 🍨

marcoscaceres:
rebased.

marcoscaceres:
highlighting now happens after linking, so not to affect linking.

halindrome:
This seems fine.  I am concerned, as usual, about backward compatibility. In this case the change of requiring nohighlight means documents that formerly would NOT have had highlighting on some examples will now suddenly have it.  This may come as a surprise to their editors.

marcoscaceres:
> On 19 May 2016, at 10:37 PM, Shane McCarron <notifications@github.com> wrote:
> 
> This seems fine. I am concerned, as usual, about backward compatibility. In this case the change of requiring nohighlight means documents that formerly would NOT have had highlighting on some examples will now suddenly have it. This may come as a surprise to their editors.
> 

🦄☄✨I like to imagine them being delighted by whimsy✨☄🦄

But that's a good point. I'll email spec-prod.

marcoscaceres:
Email sent!

jimkont:
is there a way to overwrite / bypass the new highlighting? (I also commented [here](https://github.com/w3c/respec/commit/37e5a662f9b7c820142ab0cc873bfd8acc330bae#commitcomment-17609603)).

We have special css formatting for different `pre` classes and this new change overrides everything we did. See w3c.github.io/data-shapes/shacl/ for example where each `pre` box had a different color based on the example type

marcoscaceres:
@jimkont I tried with setting "noHighlightCSS", seems to work well for your case: http://w3c.github.io/data-shapes/shacl/?noHighlightCSS=true

jimkont:
works perfect! thanks @marcoscaceres

PR #769, opened 2016-05-17
Feat(global): implement proper PubSubHub

marcoscaceres:
This is a proper implementation of PubSubHub. The only interesting file is:

js/core/pubsubhub.js

The rest of the changes are replacing "msg" for pubsubhub.


marcoscaceres:
R please? 🐰

js/core/pubsubhub.js
```diff
@@ -0,0 +1,78 @@
...
+      }
+      var args = restParams.map(function toStructuredClonable(arg) {
+        var str = String(JSON.stringify(arg.stack || arg));
+        return str;
+      });
+      var msg = {
+        topic: topic,
+        args: args,
+      };
+      window.parent.postMessage(msg, "*");
```
marcoscaceres: Need to check why we lost the origin.

js/core/pubsubhub.js
```diff
@@ -0,0 +1,78 @@
...
+    },
+    unsub: function(opaque) { // opaque is whatever is returned by sub()
+      var callbacks = subscriptions.get(opaque.topic);
+      if (!callbacks || callbacks.indexOf(opaque.cb) === -1) {
+        console.warn("Already unsubscribed:", opaque.topic, opaque.cb);
+        return;
+      }
+      callbacks.splice(callbacks.indexOf(opaque.cb), 1);
+    }
+  };
```
marcoscaceres: I'd be good to have a "seen()" method, which returns a promise that resolves once a message is seen (e.g., "end-all"). That avoids race conditions for one-off events.

marcoscaceres:
Rebased

marcoscaceres:
Merging, as I can't proceed without this and it's been sitting there for a week.

PR #770, opened 2016-05-18
Fix (SpecHelper): reduce dependece on jQuery

marcoscaceres:
Reduces jQuery dependency to just a polyfill for .append()


marcoscaceres:
This is just a refactor of tests, which doesn't affect users.

PR #781, opened 2016-05-23
Feat(shiv): remove HTML5 shiv

marcoscaceres:
This change is [Reviewable](https://reviewable.io/reviews/w3c/respec/781)


halindrome:
Before we merge this I would like someone at W3C to say it is okay to no longer support IE8.  @plehegar ?

dontcallmedom:
Hmm… Removing support for IE8 in exported documents doesn't seem great to me. Pubrules only vaguely states
>  Note: Please consider how your content will render in browsers that do not support HTML5. 

so it's not a strict requirement per se, but I think that in the spirit of inclusiveness, and since there are likely many readers that are stuck with a browser on which they have no control, I would personally prefer if we kept the shiv.

halindrome:
I also prefer not to break compatibility with older browsers.

marcoscaceres:
Yeah, different sites report different numbers for usage. I'm happy to wait till IE8 drops below 1%. It should be in a few months if current trends continue:
https://www.netmarketshare.com/report.aspx?qprid=3&qpaf=&qpcustom=Microsoft+Internet+Explorer+8.0&qpcustomb=0

marcoscaceres:
Closing for now... will reopen in a year or so.

PR #790, opened 2016-05-26
Fix uglify

marcoscaceres:
This change is [Reviewable](https://reviewable.io/reviews/w3c/respec/790)


marcoscaceres:
Merging this, as it doesn't affect users and only relates to releasing.

PR #796, opened 2016-05-26
Add firstElementChild polyfill

travisleithead:
Edge browser is broken when loading (live) respect documents for the past few days. Found (one of?) the root causes--lack of ElementTraversal support on the document node. After a quick scan of the respect code, it looks like only `firstElementChild` is used on the document, so this is a minimal polyfill to fix it.

Alternatively, in the place this is used (line 135), you could replace `doc.firstElementChild.addEventListener` with `doc.documentElement.addEventListener` which would be semantically equivalent and better supported (and not need to add this polyfill). We've had this bug in Edge for awhile, but have not yet prioritized fixing it--sorry!

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/respec/796)

### DIFF
--- a/js/core/ui.js
+++ b/js/core/ui.js
@@ -9,6 +9,21 @@
 //  - make a release candidate that people can test
 //  - once we have something decent, merge, ship as 3.2.0
 "use strict";
+
+// Quick polyfill for browser broken by lack of ElementTraversal APIs on the document object.
+if (!document.firstElementChild) {
+    Object.defineProperty(Document.prototype, "firstElementChild", { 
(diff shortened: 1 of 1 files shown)


marcoscaceres:
Using .documentElement is a better solution, indeed. I will make that replacement.

marcoscaceres:
@travisleithead, I'm going to see if I can fix it without including the polyfill. But really appreciate you putting together the PR! I use firstElementChild elsewhere, so will reuse your code to look up the first element child where needed.

PR #799, opened 2016-05-27
Fix(ui,utils): firstElementChild not in Edge (closes #798)

### LINKED ISSUE #798: firstElementChild is affecting MS Edge
@travisleithead reported: 

> Edge browser is broken when loading (live) respect documents for the past few days. Found (one of?) the root causes--lack of ElementTraversal support on the document node. After a quick scan of the respect code, it looks like only firstElementChild is used on the document

marcoscaceres:
This change is [Reviewable](https://reviewable.io/reviews/w3c/respec/799)

### DIFF
--- a/js/core/utils.js
+++ b/js/core/utils.js
@@ -34,13 +34,35 @@ define(
                 return function(doc, insertionPoint) {
                     node.remove();
                     doc.adoptNode(node);
-                    if (insertionPoint.firstElementChild) {
-                        insertionPoint.insertBefore(node, insertionPoint.firstElementChild);
+                    var firstElementChild = this.findFirstElementChild(insertionPoint);
+                    if (firstElementChild) {
+                        insertionPoint.insertBefore(node, firstElementChild);
                         return;
                     }
                     insertionPoint.appendChild(node);
-                };
+                }.bind(this);
             }
+        /**
+         * Finds the first Element child, given a node. Provides support for
+         * Microsoft Edge's missing support of .firstElementChild.
+         *
+         * @param  {Node} node The node to be traversed.
+         * @return {Element}   The first Element in the child list.
+         */
+        ,   findFirstElementChild: function(node){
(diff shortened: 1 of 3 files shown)


marcoscaceres:
@halindrome r?

marcoscaceres:
@travisleithead, I don't have access to Edge, but I'm assuming this will work based on your previous PR.

halindrome:
:lgtm:

<details><summary><i>Previously, marcoscaceres (Marcos Cáceres) wrote…</i></summary>
> @travisleithead, I don't have access to Edge, but I'm assuming this will work based on your previous PR. 
</details>


---

Reviewed 2 of 3 files at r1, 1 of 1 files at r2.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/w3c/respec/799#-:-KImPTEHGmVohGCsfgew:1417088721)*

js/core/utils.js
```diff
@@ -34,13 +34,35 @@ define(
                 return function(doc, insertionPoint) {
                     node.remove();
                     doc.adoptNode(node);
-                    if (insertionPoint.firstElementChild) {
```
travisleithead: If `insertionPoint` is never a `Document` object, then this is unnecessary, as all the ElementTraversal methods are supported on Element types in Edge :-)

travisleithead:
Left a comment. It's only the document object itself that doesn't support the ElementTraversal APIs. So on Elements the extra complication is unnecessary.

travisleithead:
Otherwise should be fine.

marcoscaceres:
Ah, thanks for the clarification! I'll remove the wrapper and just use documentElement.

marcoscaceres:
Merging this for now, as I'm going away for the weekend. Will remove the dead code next week, but at least it will be functional in Edge.

travisleithead:
Thanks!

Issue #808, opened 2016-06-07
Deprecate diffTool


marcoscaceres:
We should consider deprecating `diffTool`, as the default one doesn't support HTTPS + I can't see anyone using anything but the default one: 
https://github.com/search?p=1&q=org%3Aw3c+diffTool&type=Code

It's also not very helpful when compared to just looking at git's history or commits on github.

halindrome:
It is used for producing diff-marked versions of specs for publication.  I will add a certificate to the default server right now.  Note: it is also available at the W3C - presumably on a secured server already.

PR #813, opened 2016-06-07
Changed default diff tool URI to use https

halindrome:
Changed the default URI for the diff tool to use https.

Addresses issue #808

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/respec/813)

### DIFF
--- a/js/ui/save-html.js
+++ b/js/ui/save-html.js
@@ -42,7 +42,7 @@ define(
         return {
             show:   function (ui, _conf, _doc) {
                 doc = _doc, conf = _conf;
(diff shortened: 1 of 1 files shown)


js/ui/save-html.js
```diff
@@ -42,7 +42,7 @@ define(
         return {
             show:   function (ui, _conf, _doc) {
                 doc = _doc, conf = _conf;
-                if (!conf.diffTool) conf.diffTool = "http://www5.aptest.com/standards/htmldiff/htmldiff.pl";
+                if (!conf.diffTool) conf.diffTool = "https://www5.aptest.com/standards/htmldiff/htmldiff.pl";
```
marcoscaceres: Awesome! thanks for updating your server to work correctly with TLS!

PR #822, opened 2016-06-09
Fix figures so more than one link can be made to figures

nickevansuk:
Behaviour can be reproduced with multiple links created to a figure: without this change only the first one renders.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/respec/822)

### DIFF
--- a/js/core/figures.js
+++ b/js/core/figures.js
@@ -50,7 +50,7 @@ define(
                         .prepend($("<span class='figno'>" + num + "</span>"))
                         .prepend(doc.createTextNode(conf.l10n.fig))
                     ;
-                    figMap[id] = $cap.contents().clone();
+                    figMap[id] = $cap.contents();
                     var $tofCap = $cap.clone();
                     $tofCap.find("a").renameElement("span").removeAttr("href");
                     tof.push($("<li class='tofline'><a class='tocxref' href='#" + id + "'></a></li>")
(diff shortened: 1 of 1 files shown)


halindrome:
:lgtm:

<details><summary><i>Previously, nickevansuk wrote…</i></summary>
> #### Fix figures so more than one link can be made to figures
> 
> Behaviour can be reproduced with multiple links created to a figure: without this change only the first one renders.
> 
> 
> 
</details>


---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/w3c/respec/822#-:-KJpLTWnCtYZdYePhSYD:1417088721)*

marcoscaceres:
@nickevansuk, code looks good. Can you help me think of a test that we can add? I've not experienced this myself, so would like to make sure we have something we can see and test against (to make sure we don't accidentally regress, etc).

marcoscaceres:
Going to merge this in to prevent it from rotting away. But we need a test for it.

r12a:
We need a similar fix for links to section headings. Should i raise a different issue for that, or should we fix it here?

Also, what's needed for a test?  Does someone need to create a small document and put it somewhere?

marcoscaceres:
> On 21 Jul 2016, at 6:55 PM, r12a <notifications@github.com> wrote:
> 
> We need a similar fix for links to section headings. Should i raise a different issue for that, or should we fix it here?
> 

Different issue please.
> Also, what's needed for a test?
> 
Please see tests/spec/ folder (find the equivalent spec for the .js file being changed). I can help write them, but just need to understand the code/problem. But take a look, it's pretty straight forward.

To run tests, you need to run:

npm run test:karma

> Does someone need to create a small document and put it somewhere?
> 

No. The doc can be generated dynamically.

PR #830, opened 2016-06-16
Hljs update

marcoscaceres:
hightlight.js is intended to be built and bundled. Previously, we were
using a pre-built version of hightlightjs. However, the bundle included
all the language options, which bloated ReSpec significantly.

This new version now downloads the hightlighter repo and builds a
custom highlighter specifically for ReSpec using HTML/XML, JS, CSS, HTTP and
markdown as options.

It also now includes highlight.js via a symlink to the generated file.

Additionally, I added an `npm run build` script, which builds both ReSpec
and the highlighter.

This change is [Reviewable](https://reviewable.io/reviews/w3c/respec/830)

### DIFF
--- a/js/core/base-runner.js
+++ b/js/core/base-runner.js
@@ -42,7 +42,7 @@ define(
                         try {
                             respecConfig.preProcess[i].apply(this); }
                         catch (e) {
-                            window.console.error(e);
+                            window.console.error(e.stack || JSON.stringify(e));
                         }
                     }
                 }
--- a/js/core/biblio.js
+++ b/js/core/biblio.js
@@ -183,7 +183,7 @@ define(
                     cb();
                     return;
                 }
-                var url = "https://labs.w3.org/specrefs/bibrefs?refs=" + externalRefs.join(",");
+                var url = "https://specref.herokuapp.com/bibrefs?refs=" + externalRefs.join(",");
                 fetch(url)
                     .then(function(response) {
                         return response.json();
--- a/js/core/css/github.css
+++ b/js/core/css/github.css
@@ -1,1 +1,1 @@
-../../../node_modules/highlightjs/styles/github.css
+../../../node_modules/highlight.js/src/styles/github.css
--- a/js/core/highlight.js
+++ b/js/core/highlight.js
@@ -6,8 +6,8 @@ define(
   [
     "core/pubsubhub",
     "core/utils",
-    "highlight",
-    "text!highlightStyles/github.css",
+    "deps/highlight",
+    "text!core/css/github.css",
   ],
   function(pubsubhub, utils, hljs, ghCss) {
     // Opportunistically insert the style into the head to reduce FOUC.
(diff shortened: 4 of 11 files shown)


marcoscaceres:
Added change description

halindrome:
do you mean that you are excluding all of the other languages that it might highlight?  'cause there are a number of others that it would be nice to have around still.  JSON etc.

halindrome:
I looked through the supported languages briefly.  I would recommend that in addition to the ones you already identified we incorporate http, json, sql, and xquery.  

I would also like to add a special one or two (e.g., json-ld and json schema), but I should do that via a PR on highlight that we then migrate in over time.

marcoscaceres:
Currently, it builds with: "xml javascript css http markdown". Happy to add a few more. Do you know which specs are using sql and xquery? I thought the xquery they have was a different thing, but might be mistaken.

marcoscaceres:
Nb to self: need to also update markdown.js with the final list of languages above.

halindrome:
Weird question.... why does markdown.js care which languages have
highlighting support?

On Thu, Jun 16, 2016 at 11:28 AM, Marcos Cáceres <notifications@github.com>
wrote:

> Nb to self: need to also update markdown.js with the final list of
> languages above.
>
> —
> You are receiving this because you commented.
> Reply to this email directly, view it on GitHub
> <https://github.com/w3c/respec/pull/830#issuecomment-226539243>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/AAfx8Ns5S38DAok-XoRgGilFYCnQScH4ks5qMXm5gaJpZM4I3AgM>
> .
>



-- 
Shane McCarron
halindrome@gmail.com

marcoscaceres:
> I would also like to add a special one or two (e.g., json-ld and json schema), but I should do that via a PR on highlight that we then migrate in over time.

Would be awesome to upstream these new ones to hljs itself.

halindrome:
> Currently, it builds with: "xml javascript css http markdown". Happy to add a few more. Do you know which specs are using sql and xquery? I thought the xquery they have was a different thing, but might be mistaken.

xquery supports xpath, which is used in specs such as annotation as a way of expressing how to find the location of content that is annotated.  I don't know that there are actually examples that are stand-alone xpath right now, but I assume they are around somewhere.  Other specs use xpath too.

sql... well, I was actually thinking sparql when I typed that.  I don't know that any specs use sql.  We do use sparql in a bunch of RDF specs.

So yeah - ignore the sql thing.

marcoscaceres:
> On 17 Jun 2016, at 2:29 AM, Shane McCarron <notifications@github.com> wrote:
> 
> Weird question.... why does markdown.js care which languages have
> highlighting support?

You can provide them as a list from which hljs can choose from. The language selector can get confused sometimes. If we know we only support a limited subset, we can tell the highlighter ahead of time. 

I guess in theory now it won't matter, because we are restricting the set down so it shouldn't pick things outside the range that it knows about... So maybe I'm being too cautious? I'm just guessing about how it works.

PR #832, opened 2016-06-17
chore(deps): move dependencies to js/deps (closes #633)

### LINKED ISSUE #633: Investigate sym-linking depedencies
According to the require RequireJS docs, it might be cleaner to have dependencies in a "js/lib" folder, and then load them relative. Might be worth looking into.

marcoscaceres:
Turns out #633 was a brilliant idea. It normalized a bunch of testing code and generally simplified a bunch of requirejs + karma stuff. No ugly /node_modules/ URLs, etc.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/respec/832)


marcoscaceres:
To be clear, all this PR does is symlink the dependencies from node_modules to the "js/deps" directory. Then everything that refers to "deps" just automagically gets the right files. This includes both the source files and the builder. No need for explicit path declarations anymore 🎉. No more requirejs pain 🎉.

marcoscaceres:
@halindrome, any feedback or concerns about this PR? Although it's just changing the path to resources, it touches at lot of files, so would like to merge ASAP to avoid conflicts with any incoming PRs.

halindrome:
I'm sick in bed, but it seems harmless enough.

marcoscaceres:
Get well soon: 🍲!

marcoscaceres:
@halindrome, after you git pull, please do an `npm install` after you update, or it won't build.

PR #845, opened 2016-06-27
feat(data-include): deprecate data-include-sync

marcoscaceres:


marcoscaceres:
The only spec that was relying on this has been killed.

PR #855, opened 2016-07-04
fix(tests): prevent FF from timing out

marcoscaceres:


marcoscaceres:
This small fix seems to help a bit. Ran it a couple of times now, and at least Firefox is now finishing correctly. The problem of Chrome's random test restarts remains a mystery... but a quick search of the Internets suggests that lots of people experience those random intermittent failures. 

That should hopefully improve as Karma improves.

marcoscaceres:
Merging, as this doesn't affect end-users at all.

PR #865, opened 2016-07-12
fix(utils): toESIterable jsdom compat (closes #861)

### LINKED ISSUE #861: Check on function in utils.toESIterable makes spec loading with jsdom fail
I'm trying to use jsdom to load and process specifications, some of them being using ReSpec. This works well with most specs, but used to crash from time to time because ReSpec uses `DOMParser` to normalize padding, which jsdom did not support... until [v9.4.0](https://github.com/tmpvar/jsdom/blob/master/Changelog.md#940).

The problem now is that the `toESIterable` function checks its argument and throws an Exception when it is not an instance of `Function`:
https://github.com/w3c/respec/blob/develop/js/core/utils.js#L161

Using jsdom to load the spec, the function does get called with a function, but not with an `instanceof Function`.

An example of spec that creates that problem is: `https://w3c.github.io/mediacapture-image/`

Minimal Node.js code that exhibits the problem:

```js
const jsdom = require('jsdom');
jsdom.env({
  url: 'https://w3c.github.io/mediacapture-image/',
  features: {
      FetchExternalResources: ['script'],
      ProcessExternalResources: ['script'],
      SkipExternalResources: false
  },
  done: (err, window) => (err ? console.error(err) : console.log('doc loaded')),
  virtualConsole: jsdom.createVirtualConsole().sendTo(console)
});
```

... which reports
```
TypeError: Expected a function
    at TypeError (native)
    at Object.r.toESIterable (https://www.w3.org/Tools/respec/respec-w3c-common.js:7:20245)
    at Object.r.normalizePadding (https://www.w3.org/Tools/respec/respec-w3c-common.js:7:21745)
    at s (https://www.w3.org/Tools/respec/respec-w3c-common.js:13:30280)
    at https://www.w3.org/Tools/respec/respec-w3c-common.js:13:30446
    at Array.map (native)
    at https://www.w3.org/Tools/respec/respec-w3c-common.js:13:30408
    at Object.run (https://www.w3.org/Tools/respec/respec-w3c-common.js:14:160)
    at https://www.w3.org/Tools/respec/respec-w3c-common.js:10:27542
    at https://www.w3.org/Tools/respec/respec-w3c-common.js:10:27506
```

(That still issues "doc loaded", which is normal since ReSpec processing actually happens after doc is loaded)

Dropping the check from the `toESIterable` function makes things work as expected. I confess I do not know if it is a bug in ReSpec that incorrectly assumes that all functions will be `instanceof Function` or a bug in jsdom that does not use the appropriate function to create the node iterator.

marcoscaceres:


marcoscaceres:
@halindrome, r?

marcoscaceres:
it's a one liner, so please merge if it looks ok.

halindrome:
LGTM - merged.

PR #866, opened 2016-07-12
feat(linter): Respec linter

marcoscaceres:
Simple linter with one rule, that checks for privacy and security sections. 

@halindrome, wdyt?


js/w3c/linter.js
```diff
@@ -0,0 +1,42 @@
...
+      .some(function(text) {
+        var privOrSecRegex = /(privacy|security)/igm;
+        var considerationsRegex = /(considerations)/igm;
+        return privOrSecRegex.test(text) && considerationsRegex.test(text);
+      });
+  }
+
+  return {
+    run: function(conf, doc, cb) {
+      if (!conf.lint) {
```
marcoscaceres: I've disabled it by default, so could be "beta tested".

halindrome:
I like the concept.  It should only be on things that are rec track. And maybe check for accessibility as well?  Are there other requirements from the pub rules checker we could easily vet for too?

marcoscaceres:
>  It should only be on things that are rec track. 

Good point! 

> And maybe check for accessibility as well?

Sure. Looking forward to hearing suggestions on what to check. 

> Are there other requirements from the pub rules checker we could easily vet for too?

Checking for http urls in links might be an easy one.

marcoscaceres:
Basically, any *URI prop in config should be https...

marcoscaceres:
Ok, added check for "unofficial" to bail out. And added HTTP checker. 🎉

marcoscaceres:
Here it is in action on the manifest spec: 

screenshot 2016-07-13 17 26 01

halindrome:
I am going to raise this with the A11Y community to get some feedback.  If you would prefer to merge this and create a new issue for the A11Y linting, I am okay with that.

Issue #882, opened 2016-07-20
Change fugly-looking styling in .parameters th and .exceptions th?


patrickhlauke:
A matter of personal taste, but the current styling of `.parameters th` looks strangely off/whimsical/fugly, with its light blue and text shadow (sticks out stylistically compared to most other respec styles). Possible to adopt the same-ish look as other tables, like `.simple th`? 

> ![respect-parameters](https://cloud.githubusercontent.com/assets/895831/16986899/0d10c006-4e81-11e6-8b06-8fecb9a57f19.png)

(example from https://w3c.github.io/pointerevents/#pointer-event-types)

additionally, the current style also makes this quite difficult to read/low contrast (related: https://github.com/w3c/respec/issues/877)

patrickhlauke:
ditto for the styling used for `.exceptions` tables (simulated below)

> ![respec-exceptions](https://cloud.githubusercontent.com/assets/895831/16987093/4c17ebf2-4e82-11e6-98fd-55ba13131483.png)

PR #883, opened 2016-07-20
increase contrast for .parameters th and .exceptions th text

patrickhlauke:
drops the text shadow, darkens the .parameters background, explicitly
sets .parameters th color to white and .exceptions th to inherit/black

closes https://github.com/w3c/respec/issues/882


patrickhlauke:
before/after illustration of this change:

> respect-parameters-fix

> respec-exceptions-fix

marcoscaceres:
LGTM.

PR #897, opened 2016-08-01
Update l10n.js

r12a:
Adds boilerplate text for Japanese documents.

Can be tested at https://w3c.github.io/jlreq/ja/

### DIFF
--- a/js/w3c/l10n.js
+++ b/js/w3c/l10n.js
@@ -67,6 +67,26 @@ define(
                 ,   open_parens:                "（"
                 ,   close_parens:               "）"
             }
+            ,   ja: {
+                    this_version:               "このバージョン："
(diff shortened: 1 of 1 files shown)


r12a:
I will seek additional translations for strings that need it.

marcoscaceres:
@r12a, this is a great addition. Thank you! Do you know which spec in particular will be using this?

r12a:
> Do you know which spec in particular will be using this?

@marcoscaceres  https://w3c.github.io/jlreq/ja/ which is incubating an update for https://www.w3.org/TR/jlreq

Turns out i received the remaining translations faster than expected.  I'll raise a new PR to add those later today.

marcoscaceres:
Excellent, thanks for contributing these strings!

PR #902, opened 2016-08-03
fix: generalize nodejs location (closes #899)

### LINKED ISSUE #899: Build scripts hardcode location of node
The scripts assume node is at /usr/local/bin/node.  This isn't always practical.  I think they should be changed to just assume it is in the path.

For example, right now I do builds on a few different platforms and rely upon nvm to allow me to choose the version of node that is in use.

marcoscaceres:


marcoscaceres:
@dontcallmedom r?

PR #903, opened 2016-08-03
Ensure respec2html doesn't cache old respec versions in long-running processes

dontcallmedom:
This has caused puzzling errors in the spec generator

credits to @deniak for identifying the bug and fix


marcoscaceres:
Published to npm

PR #905, opened 2016-08-04
fix: toc should be ol

marcoscaceres:
see: https://github.com/w3c/specberus/issues/427


marcoscaceres:
@tripu, this is all that needs to change, right? 

Looking at the example specs, it seems to be all good.

marcoscaceres:
Actually, needs a bit more work...

tripu:
I was just looking at it, @marcoscaceres.
Yes, I think it's basically replacing that `ul` with an `ol`.

PR #907, opened 2016-08-04
feat: use ol in ToC (closes #835)

### LINKED ISSUE #835: Changed TOC requirements
The TOC should be an `<ol class="toc">` inside the main navigation element (`<nav id="toc">`).

marcoscaceres:


marcoscaceres:
@halindrome r?

marcoscaceres:
removed a bit more jquery garbage too

js/core/structure.js
```diff
@@ -70,25 +70,25 @@ define(
...
         current.pop();
       }
-      return $ul;
+      return $ol;
     }
 
     return {
       run: function(conf, doc, cb) {
-        if (!conf.tocIntroductory) {
+        if ("tocIntroductory" in conf === false) {
```
halindrome: I dont mind, but I am curious why you prefer this idiom?  It seems inconsistent with the rest of the code.

halindrome:
Modulo comment above, LGTM.

js/core/structure.js
marcoscaceres: The problem with the old check is that it's checking for a falsy value, but then overwriting it irrespective of it being there (that's a bug - because if the value is there as null, 0, false, or undefined, it should be left alone). 

What the code needs to be checking is checking if the property was defined, and if not, setting it to the default of false (new code).

marcoscaceres:
I'm going to send this again, as I was playing around with Travis-CI and accidentally caused the last push not to run. Will carry over the review.

PR #913, opened 2016-08-09
More Japanese boilerplate text for l10n.js

r12a:
"File a bug" is still awaiting translation. In the meantime, here are the remaining translations needed.


marcoscaceres:
Thanks @r12a!

marcoscaceres:
(Will ship in 12 hours or so, when I get to work)

PR #916, opened 2016-08-12
feat: remove promise polyfill (close #751)

### LINKED ISSUE #751: Drop promise poly
It seems that the [target install base](http://caniuse.com/#search=Promise) no longer needs the promise poly. We should remove it.

marcoscaceres:


tools/builder.js
```diff
@@ -95,9 +95,6 @@ var Builder = {
         name: "profile-w3c-common",
         logLevel: 2, // Show uglify warnings and errors.
         deps: [
-          "core/jquery-enhanced",
```
marcoscaceres: These deps are now automatically included from profile-w3c-common.js (Require figures it out on its own).

marcoscaceres:
Merging, as I broke the test suite. Happy to get a post review.

PR #938, opened 2016-10-04
Allow configurable canonical URIs for documents

dontcallmedom:
Can be configured to be either the latest TR url (default), the editors draft url, or any http(s) URL


js/core/seo.js
```diff
@@ -0,0 +1,55 @@
...
+                                      "no shortName is set in configuration");
+                    }
+                    break;
+                default:
+                    if (conf.canonicalURI
+                        && conf.canonicalURI.match(/^https?:/)) {
+                        uri = conf.canonicalURI;
+                    } else if (conf.shortName) {
+                        uri = 'https://www.w3.org/TR/' + conf.shortName + '/';
+                    }
```
marcoscaceres: Set this as the default for uri at the top?

js/core/seo.js
```diff
@@ -0,0 +1,55 @@
...
+                }
+                if (uri) {
+                    var linkElem = document.createElement("link");
+                    linkElem.setAttribute("rel", "canonical");
+                    linkElem.setAttribute("href", uri);
+                    document.head.appendChild(linkElem);
+                }
+
+                // Publish warnings
+                warnings.map(function(warn) {
```
marcoscaceres: .forEach here, as we are not actually mapping.

js/core/seo.js
```diff
@@ -0,0 +1,55 @@
...
+                    if (conf.shortName) {
+                        uri = 'https://www.w3.org/TR/' + conf.shortName + '/';
+                    } else {
+                        warnings.push("Canonical URI set to TR, but " +
+                                      "no shortName is set in configuration");
+                    }
+                    break;
+                default:
+                    if (conf.canonicalURI
+                        && conf.canonicalURI.match(/^https?:/)) {
```
marcoscaceres: We might be able to use .startsWith or .test() here

js/core/seo.js
```diff
@@ -0,0 +1,55 @@
...
+                    } else if (conf.shortName) {
+                        uri = 'https://www.w3.org/TR/' + conf.shortName + '/';
+                    }
+                    break;
+                }
+                if (uri) {
+                    var linkElem = document.createElement("link");
+                    linkElem.setAttribute("rel", "canonical");
+                    linkElem.setAttribute("href", uri);
+                    document.head.appendChild(linkElem);
```
marcoscaceres: Probably want doc.head here

js/core/seo.js
```diff
@@ -0,0 +1,55 @@
...
+                    if (conf.canonicalURI
+                        && conf.canonicalURI.match(/^https?:/)) {
+                        uri = conf.canonicalURI;
+                    } else if (conf.shortName) {
+                        uri = 'https://www.w3.org/TR/' + conf.shortName + '/';
+                    }
+                    break;
+                }
+                if (uri) {
+                    var linkElem = document.createElement("link");
```
marcoscaceres: As below, use doc.createElement

js/core/seo.js
```diff
@@ -0,0 +1,55 @@
...
+                    }
+                    break;
+                default:
+                    if (conf.canonicalURI
+                        && conf.canonicalURI.match(/^https?:/)) {
+                        uri = conf.canonicalURI;
+                    } else if (conf.shortName) {
+                        uri = 'https://www.w3.org/TR/' + conf.shortName + '/';
+                    }
+                    break;
```
marcoscaceres: Don't need break here

js/core/seo.js
```diff
@@ -0,0 +1,55 @@
...
+// Module core/seo
+// Manages SEO information for documents
+// e.g. set the canonical URL for the document if configured
+
+define(
+    ["core/pubsubhub"],
+    function (pubsubhub) {
+        return {
+            run:    function (conf, doc, cb) {
+                var warnings = [];
```
marcoscaceres: It's probably just as effective to just emit the warnings as you go, instead of collecting them.

js/core/seo.js
```diff
@@ -0,0 +1,55 @@
...
+                        uri = 'https://www.w3.org/TR/' + conf.shortName + '/';
+                    } else {
+                        warnings.push("Canonical URI set to TR, but " +
+                                      "no shortName is set in configuration");
+                    }
+                    break;
+                default:
+                    if (conf.canonicalURI
+                        && conf.canonicalURI.match(/^https?:/)) {
+                        uri = conf.canonicalURI;
```
dontcallmedom: you mean if there is something in canonicalURI that is not "TR", "edDraft", nor starts with "http" or "https", right?

js/core/seo.js
```diff
@@ -0,0 +1,55 @@
+
+// Module core/seo
+// Manages SEO information for documents
+// e.g. set the canonical URL for the document if configured
+
+define(
+    ["core/pubsubhub"],
+    function (pubsubhub) {
+        return {
+            run:    function (conf, doc, cb) {
```
marcoscaceres: Nit: one space here. I've been trying to adhere to standard js for style. Same with other white space: max two spaces of indentation.

js/core/seo.js
dontcallmedom: good catch (& will fix), but note that this is copy & paste from w3c/linter.js (so might be worth fixing there as well)

js/core/seo.js
marcoscaceres: I was thinking you can also "new URL(uri).href" the value.

js/core/seo.js
marcoscaceres: Argh, yeah 😊

marcoscaceres:
As this defaults to a W3C URL, I've moved it to the w3c/ folder. I also changed the whitespace and fixed a couple of tests. Looks good to go once TravisCI says it's good.

marcoscaceres:
@dontcallmedom sorry, one last thing, can you kindly please document the feature in the Wiki. I'll get it started for you.

marcoscaceres:
Ok... https://github.com/w3c/respec/wiki/canonicalURI

Admittedly, having written the above and reflected on this... I'm not a huge fan of having special keywords to trigger the desired behavior. It might be better to only allow a URL for canonicalURI. That would make it consistent with the other `*URI` options. 

@halindrome opinions ™️ ?

marcoscaceres:
(it would also make it generic again... and could move back to "core").

js/w3c/seo.js
```diff
@@ -0,0 +1,53 @@
...
+            }
+            break;
+          default:
+            if (conf.canonicalURI) {
+              if (/^https?:/.test(conf.canonicalURI)) {
+                uri = conf.canonicalURI;
+              } else {
+                pubsubhub.pub("warn", "Canonical URI configured, " +
+                  "but the value does not start with " +
+                  "http: or https:");
```
marcoscaceres: We don't test for "http" here, so the warning is incorrect.

dontcallmedom: actually, we do test for "http" (/s?/), so the warning is (was) correct

marcoscaceres: oh, d'oh... my bad. Will revert

dontcallmedom:
re keywords vs URIs - I wondered about that too, but in general, I think it's best if we avoid repeating the same URIs in several places in the config, since that makes it too easy to forget one.

So my preferences would be (in order):
* keep both keywords and URIs in the same parameter
* use a canonicalRef (keyword) and canonicalURI to distinguish the two (but with the risk that they may contradict each other, which I guess would trigger a warning)
* use only canonicalRef

My wishes for this feature is that it does the right thing by default (point to TR/foo as canonical) without needing any configuration change; I think there are cases where pointing to the editors draft remain more appropriate, so I think that should be configurable; beyond that, it's mostly bells and whistles as far as I am concerned.

marcoscaceres:
I personally don't mind the repetition of the URLs, so I'll defer to @halindrome.  I wouldn't be a huge fan of adding an additional prop for this simple feature.

js/w3c/seo.js
marcoscaceres: About https... The google page about canonical URLs recommends TLS... so maybe it would be good to force TLS for this? It's also "good for the inter-a-webs".

dontcallmedom: that'd work for me; the test case would have to be updated (and this ultimately depends on how we resolve the keyword vs url discussion)

marcoscaceres: Ok, will wait for @halindrome. 

As an aside, the linter is also supposed to complain if it spots any properties that end with "URI" that are not HTTPs... another reason why we might not want keywords here.

halindrome:
Okay.... sheepish look.  I KNOW what canonical means.  What I don't know is what it means in this context or why we need this change as opposed to doing what we have always done.  Sorry.  I must be thicker than usual today.

dontcallmedom:
sorry @halindrome I should have documented my intent (probably should have first opened an issue rather than a pull request).

The main idea here is to have the latest version of a spec gets better search engine rankings than any random dated version of the specs; the canonical URI achieves this by always pointing to the latest version as being the "right" version (per ["Consolidating link signals for the duplicate or similar content"](https://support.google.com/webmasters/answer/139066?hl=en))

halindrome:
Oh! Well, I can see how this is a problem.  And I assume that W3C pubs is behind wanting this implemented?  If so, then I am sort of cool with just doing it.  Have it in there all the time as whatever the short name is, and then override it if you really want the github.io to be canonical for some reason.  This could get a little political...   What does W3C *want* to be the thing search engines find?  The stable documents or the bleeding edge?  W3C doesn't really do "living standards", that's more of a WhatWG thing.

marcoscaceres:
> On 6 Oct. 2016, at 8:58 am, Shane McCarron <notifications@github.com> wrote:
> 
> Oh! Well, I can see how this is a problem. And I assume that W3C pubs is behind wanting this implemented? If so, then I am sort of cool with just doing it. Have it in there all the time as whatever the short name is, and then override it if you really want the github.io to be canonical for some reason.
> 
Yep. Like if there is no echidna attached.
> This could get a little political... What does W3C want to be the thing search engines find?
> 

/TR/

> The stable documents or the bleeding edge? W3C doesn't really do "living standards", that's more of a WhatWG thing.
> 

The W3C do it too via Echidna: see the web manifest spec, for example.

halindrome:
Huh - I had no idea.  But even for "living" standards the W3C keeps them in /TR/ as static documents that are frequently updated?

marcoscaceres:
Yep. Works exactly the same way WHATWG. The static version is automatically published when a pull request is merged.

marcoscaceres:
@halindrome, to be more clear... you could have "Spec A" automatically being published on /TR/ while also having an Editor's draft living on GitHub pages: this can happen for specs where a lot of links pointing to gh-pages were handed out to people, but the spec has since moved to automatically being published on TR (as a living document). The is precisely the situation the Web Manifest is in - but I would prefer people always go to /TR/ via Google. 

The fundamental question is about the special keywords ("TR", and "edDraft"). Could we get your opinion on those and if you think we should keep them - or solely restrict this property to a HTTPS URL.

marcoscaceres:
@dontcallmedom given the discussion thus far, I'm making the following changes:

 1. instead of having a `var url`, I'm going to override `conf.canonicalURI` with the derived URL. 
 1. `canonicalURI` is always the result of normalization via `new URL().href`, based on document  location. That will catch malformed URLs and invalid.
 1. Allow the linter will check for TLS URLs, instead of having this file do it.    
 1. the keywords remain for now.

marcoscaceres:
(made some tweaks to comment above)

dontcallmedom:
re "what W3C pubs is behind", I'll note that I didn't check with anyone else in the W3C staff, so this has no specific blessing beyond my own. My perspective was that:
* this is not prevented by pubrules
* this seems to align with the W3C publication system goals
* this would be an improvement in SEO for ReSped documents

My idea was to bring this as a feature to ReSpec, bring attention to it in spec-prod for feedback, and possibly adoption by other spec editor tools and editors; maybe overtime this should become part of pubrules.

halindrome:
I guess my only final comment is that we should warn if a URL is not HTTPs but not restrict to it.  Other people use ReSpec for things outside of the W3C (or they did, I assume they still do) and we can't assume that everyone always has https available.  Sure, they should.  But they might not for whatever reason.

Anyway, ship it whatever way you think is best.  Anything to get the search results more reasonable.

marcoscaceres:
Thanks for input. Will ship it tomorrow. 

Google changed their Chrome distribution URL and its broken our TravisCI setup:( Trying to work around it.

marcoscaceres:
This is released as 5.1.0.

PR #937, opened 2016-09-23
Update to UI + base runner 

marcoscaceres:
I've reworked ReSpec's UI pill a bit... consolidated the CSS that was in code into the CSS file.  

Also, now that we have ES6 support in all browsers: 

 * Can simplify async code using generators. 
 * use some ES6 syntax for object literals, functions, etc.

screenshot 2016-09-23 19 51 33

PR #941, opened 2016-10-12
refactor: use uglify-harmony in build process

marcoscaceres:


marcoscaceres:
carrying over review from #937.

PR #943, opened 2016-10-12
feat(async.js): add marcosc-async as dep

marcoscaceres:


marcoscaceres:
Small piece from #937

PR #944, opened 2016-10-12
feat(ui): update ReSpec UI 

marcoscaceres:


halindrome:
Is there documentation on the exported UI?  I know that ARIA could use being able to extend the menu.

marcoscaceres:
@halindrome it's basically what is in:
https://github.com/w3c/respec/blob/develop/js/core/ui.js

We need to remove the rest of the jQuery `.css()` garbage in there (to respect2.css). Then we should have a fairly clean file to work with.

halindrome:
Sure...  also, why respect2.css instead of respec2.css ?

marcoscaceres:
> On 13 Oct. 2016, at 10:07 pm, Shane McCarron <notifications@github.com> wrote:
> 
> Sure... also, why respect2.css instead of respec2.css ?
> 
Typo on my part. We should drop the 2 also:)

PR #947, opened 2016-10-13
feat(.travis.yml): use yarn instead of npm for install

marcoscaceres:


marcoscaceres:
@halindrome, I would like to move to using [Yarn](http://yarnpkg.com). It's significantly faster when developing locally (cuts about 1.5mins off "install" times on updates for me).  

However, you will need to install it locally also with either: 

```
curl -o- -L https://yarnpkg.com/install.sh | bas
``` 

OR 

```
npm install -g yarn
```

marcoscaceres:
That should be `bash` not `bas`... fixed above.

halindrome:
I don't mind... how do you feel about node 6.8 ?

marcoscaceres:
> On 13 Oct. 2016, at 10:09 pm, Shane McCarron <notifications@github.com> wrote:
> 
> I don't mind... how do you feel about node 6.8 ?
> 

Node 7 is only days away 🍾 I'm excited and I don't know why! 🤓

PR #956, opened 2016-10-31
fix(save-html): buttonCSS variable was undefined

marcoscaceres:


marcoscaceres:
As this is a one liner, I'll merge it as a hot fix if everything passes.

PR #957, opened 2016-10-31
refactor(copydeps): convert to nodejs script (closes #954)

### LINKED ISSUE #954: Make copydeps.sh Windows compatible
Should be converted to a Node script instead of a shell script.

marcoscaceres:


marcoscaceres:
@nickmccurdy, please take a look - I made it all run async. If I understand correctly, path.normalize() should handle normalizing the "/" to "\" on Windows. Everything else is resolve relative, so, in theory, it should work(tm).

I've not tested in Windows yet tho 👍.

.npmignore
```diff
@@ -15,5 +15,5 @@ examples/
 karma.conf.js
 node_modules/
 tests/
-tools/copydeps.sh
+tools/copydeps.js
```
nickmccurdy: Is it intentional that this is now the only file in `tools` that is in the npmignore (ignoring the `.jshintrc`)?

nickmccurdy:
This works great on Windows! The rest of the install will fail because of the semicolons in other commands (see #959), but if you just run `npm run copydeps` it works. The only changes that git reports are the line endings in `js/deps`, and I think those can be ignored because of git's built in line ending conversion.

PR #959, opened 2016-10-31
Fixes for Windows testing

nickmccurdy:
This branch fixes some issues with running `npm install` and `npm test` on Windows. The test suite will run, but it fails on IE.

## Notes
- ~~#957 is merged into this pull request (it depends on the `copydeps.js` script, otherwise `npm install` fails on Windows). If you want to merge this, please merge #957 first.~~
- The behavior for `&&` on Windows and POSIX systems is different. Windows runs expressions on both sides unconditionally, while POSIX systems only run the second expression if the first succeeds. Hopefully this hack won't cause us problems since the commands should succeed, but if it does we should refactor the multi-part npm scripts into more JS files.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
     "karma-chrome-launcher": "^2.0.0",
     "karma-detect-browsers": "^2.1.0",
     "karma-firefox-launcher": "^1.0.0",
+    "karma-ie-launcher": "^1.0.0",
     "karma-jasmine": "^1.0.2",
     "karma-mocha": "^1.2.0",
     "karma-mocha-reporter": "^2.2.0",
--- a/package.json
+++ b/package.json
@@ -45,24 +45,24 @@
     "whatwg-fetch": "^1.0.0"
   },
   "scripts": {
-    "build": "npm run build:highlight; npm run build:respec-w3c-common",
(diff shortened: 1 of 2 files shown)


nickmccurdy:
@marcoscaceres I noticed you merged #957. I'll rebase/merge this onto develop now.

marcoscaceres:
Ok, #957 is in. Could you please rebase. Seems there is a small conflict. 

Awesome work @nickmccurdy!

marcoscaceres:
@nickevansuk, we don't support IE, only Edge - so that's ok :)

marcoscaceres:
Also, noting the different behavior of "&&". I don't have a vm to test on on this computer, but will try it out tomorrow. Any issues, we can do as you suggested and split things up as you suggested.

nickmccurdy:
> we don't support IE, only Edge - so that's ok :)

In that case, the project should probably be reconfigured to not run IE tests locally. Unfortunately it's not as simple as removing the `karma-ie-launcher` dependency, `karma-detect-browsers` will fail without it on Windows.

> I don't have a vm to test on on this computer, but will try it out tomorrow.

It runs for me on Windows 10, and I haven't switched back to *nix yet but I assume it works because Travis is passing.

PR #961, opened 2016-11-01
Use http-server in the start and server scripts

nickmccurdy:
This way, the development server script doesn't depend on Python (among other things, this makes it work out of the box on Windows). I also renamed it to `start` as recommended by npm, though there's still an alias for `server`.


marcoscaceres:
Thanks, been meaning to fix this for a while.

PR #973, opened 2016-11-16
Updated W3C translations link

plehegar:
Matching the expectation in
 https://github.com/w3c/specberus/blob/master/lib/rules/headers/translations.js


marcoscaceres:
rebasing...

PR #994, opened 2016-12-20
fix(profile-w3c-common): do includes before markdown proc (closes #993)

### LINKED ISSUE #993: Support markdown includes
It would seem that the `data-include` always imports HTML or text when the `data-include-format` attribute is used.

I think that when the global setting mandates the use of markdown (`var respecConfig={ format: "markdown" }`) includes should also be treated as such.

Alternatively (or in addition), `data-include-format="markdown"` could also be supported.

marcoscaceres:


marcoscaceres:
Ok, this totally works... but need to add proper tests... there are some limitation tho. Code blocks inside elements don't get processes correctly.

marcoscaceres:
The limitations are also in the current markdown processor... so, it's all good (should fix as separate bug).

marcoscaceres:
@halindrome, quick one for you 🤗

PR #1001, opened 2016-12-22
fix(webidl-contiguous): don't link special IDL things (closes #996)

### LINKED ISSUE #996: Exclude trying to link to maplike and setlike
ReSpec currently warns when `maplike` and `setlike` are not linked to definitions.

marcoscaceres:
Currently, we don't support linking to "maplike", "setlike", "stringifier"


marcoscaceres:
Some more small goodies... The IDL warnings are much better now :crossed_fingers:

marcoscaceres:
Merging this, as the current code is causing specs to fail.

PR #1008, opened 2016-12-30
feat(data-cite): add support for data-cite attribute

marcoscaceres:
This adds a new foundation feature to ReSpec, allowing authors to directly cite another SpecRef spec in inline text. This is currently very simple, on purpose... so we can build: 

```HTML
<!-- normative -->
It's in the <a data-cite="!DOM">DOM spec</a>.

<!-- informative -->
It's in the <a data-cite="HTML">HTML spec</a>.
```  

That automatically adds `href` attribute pointing to the DOM spec.  It also adds the reference to the References section in the right place.

### DIFF
--- a/js/core/data-cite.js
+++ b/js/core/data-cite.js
@@ -1,0 +1,69 @@
+/**
+ * Module core/data-cite
+ *
+ * Allows citing other specifications using
+ * anchor elements. Simply add "data-cite"
+ * and key of specification.
+ *
+ * This module simply adds the found key
+ * to either conf.normativeReferences
+ * or to conf.informativeReferences depending on
+ * if it starts with a ! or not.
+ *
+ * Usage:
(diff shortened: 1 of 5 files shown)


marcoscaceres:
So, once this goes in, we can start targeting particular "types" - like methods, etc. that will allow seamless cross linking between specs.

marcoscaceres:
Thinking about next steps... I think I want:

```html
<a data-cite=spec data-lt="xref in other spec" data-cite-type="">thing</a>
```

We then need a xref service tho.... it's not clear how BikeShed specs are currently producing their xref.json files, but we need to generate the same thing.

Then we need a central database for those with a RESTful API on the front.

marcoscaceres:
Merging as it's blocking my other stuff... will fix the async() issue on develop branch, as it's unrelated to this. Would still appreciate any feedback.

PR #1013, opened 2017-01-04
feat(biblio): cache references in indexedDB

marcoscaceres:
The performance improvements from this are just amazing.


marcoscaceres:
Added tests...

PR #1021, opened 2017-01-06
refactor: convert various modules to ES6

marcoscaceres:


marcoscaceres:
Some small refactoring to use ES6 modules and some new ES6 features. They are automatically converted to requestjs (AMD) modules by Babel. No significant changes - but kinda neat.

PR #1065, opened 2017-02-02
feat: significantly improve ARIA support

marcoscaceres:
* document now signals when busy
* pre elements that are getting changed now signal when busy
* ReSpec pill is still not great... but less crappy: DOM ordering is not correct for tabbing. Correct roles and labels applied. We also notify when things menus and dialogs open and close. 

It's far from perfect, but it's hopefully a huge improvement.

### DIFF
--- a/src/core/aria.js
+++ b/src/core/aria.js
@@ -1,0 +1,7 @@
+import { sub } from "core/pubsubhub";
+
+document.body.setAttribute("aria-busy", "true");
+
+sub("end-all", () => {
+  document.body.setAttribute("aria-busy", "false");
+}, { once: true })
--- a/src/core/highlight.js
+++ b/src/core/highlight.js
@@ -49,9 +49,13 @@ export async function run(conf, doc, cb) {
       doc.querySelectorAll("pre:not(.idl):not(.nohighlight),code.highlight")
     )
     .map(element => {
+      element.setAttribute("aria-busy", "true");
+      return element;
+    })
+    .map(element => {
       return new Promise((resolve, reject) => {
         if (element.textContent.trim() === "") {
-          return resolve(); // no work to do
+          return resolve(element); // no work to do
         }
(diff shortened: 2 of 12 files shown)


marcoscaceres:
Added description above noting the changes. @halindrome, please to review. Please let know if you want to make changes... we should merge core/aria with w3c/aria. It seems people might benefit in general from the w3c/aria stuff, as it's not W3C specific, AFAICT.

marcoscaceres:
The above should say "DOM ordering is **now** correct for tabbing".

src/core/aria.js
```diff
@@ -0,0 +1,7 @@
+import { sub } from "core/pubsubhub";
+
+document.body.setAttribute("aria-busy", "true");
```
halindrome: Actually, I think this should be aria-hidden.  Some browsers do not allow the entire webview area to be marked as busy.  We could try aria-hidden.  Alternately, we could wrap the entire content in a div and then mark that div as busy / hidden.

marcoscaceres:
Merging for now. We can use this as a base to improve on... at least it's not as bad as it was

src/core/highlight.js
```diff
@@ -48,18 +48,22 @@ export async function run(conf, doc, cb) {
...
+          return resolve(element); // no work to do
         }
         const msg = {
           action: "highlight",
           code: element.textContent,
           id: Math.random().toString(),
           languages: getLanguageHint(element.classList),
         };
-
+        element.setAttribute("aria-live", "polite");
```
majornista: This presents a significant accessibility problem for screen reader users trying to read a spec containing a lot of code examples, because every code sample is sent to the the screen reader buffer to announce at its earliest convenience. I've sometimes been unable to test a page with VoiceOver with a W3C spec open in another tab, because all I hear are source code examples from the spec being announced.

PR #1079, opened 2017-02-13
fix: Reduce FOUC (closes #326)

### LINKED ISSUE #326: stop FOUC
when respec loads doc often there is a flash of unstyled content, would be nice if it did not occur

marcoscaceres:


marcoscaceres:
this is probably as good as it will get without reducing the ReSpec payload (which is mostly jQuery). There are a few other perf improvements I can make from here - but this helps with all the thrashing.

PR #1096, opened 2017-02-17
Fix a CSS property

tripu:
cf [`https://labs.w3.org/echidna/api/status?id=c6ba4368-f490-434e-aa84-a5b4da0cc912`](https://labs.w3.org/echidna/api/status?id=c6ba4368-f490-434e-aa84-a5b4da0cc912)


marcoscaceres:
Well spotted. Copied from JS so missed that. Will release later today or early tomorrow.

Issue #1039, opened 2017-01-16
2017 Process Document Operational on 1 March 2017


marcoscaceres:
@plehegar, it's not clear what the requirements are here (apart from defaulting to it)... could you kindly let me know if there are additional boilerplate things?

marcoscaceres:
https://services.w3.org/xslt?xmlfile=https://www.w3.org/2005/08/01-transitions2017.html&xslfile=https://www.w3.org/2005/08/transitions2017.xsl

plehegar:
It's part of https://www.w3.org/pubrules/doc, except that it won't get updated until tomorrow.
The only change is regarding the Process reference in the SOTD. After March 2017, one must use:
`<p>This document is governed by the <a id="w3c_process_revision" href="https://www.w3.org/2017/Process-20170301/">1 March 2017 W3C Process Document</a>.</p>`

marcoscaceres:
Wfm

PR #1114, opened 2017-02-28
Use Process 2017

plehegar:
For #1039 

Must be deployed between March 1 and March 31.


plehegar:
Not sure why the test fails since none of my changes are related to the test...

marcoscaceres:
Restarted the build. Sometimes it fails for random reasons.

PR #1115, opened 2017-03-01
fix(core/ui): return early if element is not passed

marcoscaceres:


marcoscaceres:
Note, recent failures might be because of AWS going down. So jobs are not even starting. I got an email from TravisCI saying that they are having AWS issues.

PR #1130, opened 2017-03-04
Expand handlebars path arguments for build to run on Windows

tidoust:
I was trying to build ReSpec on my Windows 10 machine but generation of `templates.js` failed with some "File not found" error. Investigation led me to wonder where the wildchar paths passed as argument to Handlebars were expanded, which took me to:
https://github.com/wycats/handlebars.js/issues/1318#issue-210707785.

This pull request implements the solution suggested in: https://github.com/Geonovum/respec/issues/5#issuecomment-282966082

In particular, it uses the "glob" library to expand the paths and passes the result to Handlebars. Build now works on my Windows 10 machine. Hopefully, the change does not break anything on other systems.


marcoscaceres:
Heh, we came up with nearly identical solutions:

```JS
#!/usr/bin/env node
"use strict";
const glob = require("glob");
const path = require("path");
const promises = process.argv
  .concat()
  .splice(2)
  .map(val => {
      glob(val, { nonull: true }, (err, files) => {
        if (err) {
          return reject(err);
        }
        const fullPaths = files.map(file => path.resolve(process.cwd(), file));
        resolve(fullPaths);
      });
    });
  });

Promise
  .all(promises)
  .then(
    paths => paths.reduce((arr, path) => arr.concat(path), []).join(" ")
  )
  .then(files => console.log(files));
```

I was thinking of doing it async tho.  Is there any reason not to do glob async? 

Also, ESlint doesn't like process.write, because the output doesn't get format. Should we just use console.log(), given that it's equivalent, but formatted?

tidoust:
> I was thinking of doing it async tho. Is there any reason not to do glob async?

4 easy-to-follow lines instead of 20 harder-to-process lines of code? ;)
Async is fine, but given that it's a shell script, does async give us anything?

Looking at your code, isn't `path.resolve(process.cwd(), file)` going to be problem if the name of a folder in the path that leads to the ReSpec folder contains a whitespace? Also, I actually give the full arguments line to "expandpaths", including `-a` and `-f` parameters and they would end up being prefixed by the current working directory using your code. That may depend on how you want to pipe the result to Handlebars, though.

> Also, ESlint doesn't like process.write, because the output doesn't get format. Should we just use console.log(), given that it's equivalent, but formatted?

Well, I'm fine with that as well... although I note you do not want the output to be formatted here. It won't be because `console.log` is called with only one parameter, and because ReSpec folder hierarchy is very unlikely to ever include file names with `%` in any case (well, probability goes up a bit if you make the tool return absolute paths as you cannot control the path that leads to respec folder), but the fact is you just do not want formatting.

Anyway, it's all your call. My goal was to save you some work, but that's too late: you already worked on a solution! Feel free to drop the PR and go with your solution, or let me know if you'd rather have me update this PR to switch to async and use of `console.log`.

tidoust:
> I actually did it on when I filed the bug... but then couldn't figure out what xargs equivalent would be on Windows if using cmd.exe... I guess there is no way to do this in Windows without a unix subsystem.

`xargs` is simply working in `cmd.exe` on my computer so I very candidly assumed it now shipped with Windows... But that's not true. In my case, it actually comes bundled with [Git for Windows](https://git-for-windows.github.io/) and the `Git/usr/bin` folder must be in the PATH as a consequence.

I don't know of an easy way to achieve the same thing without `xargs`. The only possibility that comes to mind would be to write a JS script that both expand paths and call Handlebars at once.

marcoscaceres:
> I don't know of an easy way to achieve the same thing without xargs. The only possibility that comes to mind would be to write a JS script that both expand paths and call Handlebars at once.

Yeah, I thought about that too... I think we are ok to just require Git Bash on Windows... I think it will be insanity if the next Windows version doesn't ship with a native unix subsystem.

PR #1160, opened 2017-03-15
test against Firefox ESR

marcoscaceres:


marcoscaceres:
I can't figure out why ESR doesn't work. Will have to keep testing it locally.

marcoscaceres:
(someone else wants to investigate, go for it)

PR #1174, opened 2017-03-29
Fix for ampersands in markdown

nickevansuk:
I believe this same issue still exists for ampersands


marcoscaceres:
@nickevansuk, this looks good! Could you provide me with an example of what it is effecting? I'd like to add a test for it, at least to check that we don't regress this in the future.

PR #1179, opened 2017-04-07
Fix link for CEReactions extended attribute

dontcallmedom:
Defined in custom-elements, not HTML51

### DIFF
--- a/src/core/webidl.js
+++ b/src/core/webidl.js
@@ -234,7 +234,7 @@ function typeIsWhitespace(webIdlType) {
 }
 
 const extenedAttributesLinks = new Map([
-  ["CEReactions", "HTML51#cereactions"],
+  ["CEReactions", "!custom-elements#cereactions"],
   ["Clamp", "WEBIDL#Clamp"],
   ["Constructor", "WEBIDL#Constructor"],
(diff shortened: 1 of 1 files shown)


src/core/webidl.js
```diff
@@ -234,7 +234,7 @@ function typeIsWhitespace(webIdlType) {
 }
 
 const extenedAttributesLinks = new Map([
-  ["CEReactions", "HTML51#cereactions"],
+  ["CEReactions", "!custom-elements#cereactions"],
```
marcoscaceres: @domenic, sorry, what's the right spec to reference for CEReactions? I see it formally defined in WHATWG HTML without linking to the custom elements spec.

domenic: WHATWG HTML is correct.

dontcallmedom: (I don't mind either way, but chose `custom-elements` based on what was done in bdb2df1e6dad5256078e4bda9345c9bdf3fe0e2b)

domenic: Looking over that it seems like HTMLConstructor is also pointing to the wrong place.

marcoscaceres:
Sending as a separate PR https://github.com/w3c/respec/pull/1191.

PR #1219, opened 2017-05-17
Update l10n.js

chaals:
Use the whole word "Figure". 3 extra characters per picture isn't that expensive really.

### DIFF
--- a/src/core/l10n.js
+++ b/src/core/l10n.js
@@ -17,7 +17,7 @@ const base = {
   editors_note: "Editor's note",
   editors: "Editors:",
   example: "Example",
-  fig: "Fig. ",
+  fig: "Figure ",
   file_a_bug: "File a bug",
   info_references: "Informative references",
   issue_summary: "Issue Summary",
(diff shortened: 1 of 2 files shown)


src/core/l10n.js
```diff
@@ -17,7 +17,7 @@ const base = {
   editors_note: "Editor's note",
   editors: "Editors:",
   example: "Example",
-  fig: "Fig. ",
+  fig: "Figure ",
```
marcoscaceres: I don't disapprove, but what motivates this change?

chaals: I started using `figure`s in a specification, and the abbreviation seemed unhelpful.

Admittedly this is close to the bike shed...

marcoscaceres: As my screenreader also doesn't know how to expand the "fig." abbr, I'm inclined to support this change.

marcoscaceres:
Will release soon.

src/core/l10n.js
chaals: (You can probably teach your screenreader to recognise it, but I think that's pushing the work to the wrong place :) ).

Thanks, BTW

PR #1249, opened 2017-06-03
BREAKING CHANGE: remove js/tmpl.js - not used

marcoscaceres:


marcoscaceres:
Templates are compiled so this was not needed anymore.

PR #1326, opened 2017-07-19
Warn about href's that link to nonexistent id's

### LINKED ISSUE #1318: Warn when href points to undefined id
Wouldn't it be super nice if respec warned you when you have an `a[href="#that"]` but no `[id="that"]`? Or maybe I'm not doing referencing right and there is a feature for that (instead of defining `id` attribute explicitly)?

I find it very helpful that respec highlights links like `<a>Something</a>` when there is no `<dfn>Something</dfn>` in the document. Similarly, I wish it highlighted `<a href="#that">...</a>` when there is no `<anything id="that">...</anything>` in the document too!

futpib:
Fixes #1318

### DIFF
--- a/js/profile-w3c-common.js
+++ b/js/profile-w3c-common.js
@@ -78,6 +78,7 @@ define(
     "w3c/permalinks",
     "core/id-headers",
     "core/rdfa",
+    "core/lint-hrefs-ids",
     "w3c/aria",
     "core/location-hash",
     "ui/about-respec",
--- a/src/core/lint-hrefs-ids.js
+++ b/src/core/lint-hrefs-ids.js
@@ -1,0 +1,21 @@
+// Module core/lint-hrefs-ids
+// Warns about href's that do not have corresponding id's
+
+import { pub } from "core/pubsubhub";
+
+export const name = "core/lint-hrefs-ids";
+
+export function run(conf, doc, cb) {
+  console.log()
+  Array.from(doc.querySelectorAll('a[href]'))
+    .map(a => ({ a, href: a.getAttribute('href') }))
+    .filter(({ href }) => Boolean(href))
+    .filter(({ href }) => href.startsWith('#'))
+    .map(({ a, href }) => ({ a, href, id: href.slice(1) }))
+    .filter(({ id }) => Boolean(id))
+    .filter(({ id }) => !doc.getElementById(id))
+    .forEach(({ href, id }) => {
+      pub("warn", `Found link with href '${href}' but no element with id '${id}'`);
+    });
+  cb();
+}
--- a/src/core/lint-hrefs-ids.js
+++ b/src/core/lint-hrefs-ids.js
@@ -6,7 +6,6 @@ import { pub } from "core/pubsubhub";
 export const name = "core/lint-hrefs-ids";
 
 export function run(conf, doc, cb) {
-  console.log()
   Array.from(doc.querySelectorAll('a[href]'))
     .map(a => ({ a, href: a.getAttribute('href') }))
     .filter(({ href }) => Boolean(href))
--- a/src/core/lint-hrefs-ids.js
+++ b/src/core/lint-hrefs-ids.js
@@ -13,8 +13,14 @@ export function run(conf, doc, cb) {
     .map(({ a, href }) => ({ a, href, id: href.slice(1) }))
     .filter(({ id }) => Boolean(id))
     .filter(({ id }) => !doc.getElementById(id) && !doc.getElementById(decodeURIComponent(id)))
-    .forEach(({ href, id }) => {
-      pub("warn", `Found link with href '${href}' but no element with id '${id}'`);
+    .forEach(({ a, href, id }) => {
+      const msg = `Found an element with href '${href}' but no element with id '${id}'.`
+            + " See developer console for offending element.";
+
+      pub("warn", msg);
+      console.warn("Offending element:", a);
+      a.title = msg;
+      a.classList.add("respec-offending-element");
     });
   cb();
 }
--- a/src/core/lint-hrefs-ids.js
+++ b/src/core/lint-hrefs-ids.js
@@ -12,7 +12,7 @@ export function run(conf, doc, cb) {
     .filter(({ href }) => href.startsWith('#'))
     .map(({ a, href }) => ({ a, href, id: href.slice(1) }))
     .filter(({ id }) => Boolean(id))
-    .filter(({ id }) => !doc.getElementById(id))
+    .filter(({ id }) => !doc.getElementById(id) && !doc.getElementById(decodeURIComponent(id)))
     .forEach(({ href, id }) => {
       pub("warn", `Found link with href '${href}' but no element with id '${id}'`);
     });
--- a/src/core/ui.js
+++ b/src/core/ui.js
@@ -217,6 +217,12 @@ export const ui = {
   warning: function(msg) {
     errWarn(msg, warnings, "warning", "Warnings");
   },
+  getErrors: function() {
+    return errors.slice();
+  },
+  getWarnings: function() {
+    return warnings.slice();
+  },
   closeModal: function(owner) {
     if ($overlay)
       $overlay.fadeOut(200, function() {
--- a/tests/spec/core/lint-hrefs-ids-spec.js
+++ b/tests/spec/core/lint-hrefs-ids-spec.js
@@ -4,22 +4,30 @@ describe("Core — Lint href's matching id's", function() {
     flushIframes();
     done();
   });
+
   it("warns if can't match a fragment identifier to an id in the document", function(done) {
-    var ops = {
+    const ops = {
       config: makeBasicConfig(),
       body:
         makeDefaultBody() + `
           <section>
-            <a href='#test'>foo</a>
+            <a id='offender' href='#test'>foo</a>
           </section>
         `,
     };
+
     makeRSDoc(ops, function(doc) {
-      var ui = doc.defaultView.respecUI;
+      const ui = doc.defaultView.respecUI;
+
       expect(ui.getErrors()).toEqual([]);
       expect(ui.getWarnings().length).toEqual(1);
+
+      const offender = doc.getElementById("offender");
+
+      expect(offender.classList.contains("respec-offending-element")).toBe(true);
     }).then(done);
   });
+
   it("does not warn if fragment identifier matches an id in the document", function(done) {
     var ops = {
       config: makeBasicConfig(),
@@ -42,8 +50,10 @@ describe("Core — Lint href's matching id's", function() {
           </section>
         `,
     };
+
     makeRSDoc(ops, function(doc) {
-      var ui = doc.defaultView.respecUI;
+      const ui = doc.defaultView.respecUI;
+
       expect(ui.getErrors()).toEqual([]);
       expect(ui.getWarnings()).toEqual([]);
     }).then(done);
--- a/tests/spec/core/lint-hrefs-ids-spec.js
+++ b/tests/spec/core/lint-hrefs-ids-spec.js
@@ -4,40 +4,48 @@ describe("Core — Lint href's matching id's", function() {
     flushIframes();
     done();
   });
-  it("should warn about href pointing nowhere", function(done) {
+  it("warns if can't match a fragment identifier to an id in the document", function(done) {
     var ops = {
       config: makeBasicConfig(),
       body:
-        makeDefaultBody() +
-          "<section>" +
-          "<a href='#test'>foo</a>" +
-          "</section>",
+        makeDefaultBody() + `
+          <section>
+            <a href='#test'>foo</a>
+          </section>
+        `,
     };
     makeRSDoc(ops, function(doc) {
       var ui = doc.defaultView.respecUI;
-      expect(ui.getErrors().length).toEqual(0);
+      expect(ui.getErrors()).toEqual([]);
       expect(ui.getWarnings().length).toEqual(1);
     }).then(done);
   });
-  it("should not warn about href pointing somewhere", function(done) {
+  it("does not warn if fragment identifier matches an id in the document", function(done) {
     var ops = {
       config: makeBasicConfig(),
       body:
-        makeDefaultBody() +
-          "<section id='test'>" +
-          "<a href='#test'>bar</a>" +
-          "<a href='http://example.com/#test'>buz</a>" +
-          "<a href='http://example.com'>qaz</a>" +
-          "<a href='#'>fuzz</a>" +
-          "<a href=''>fuzz</a>" +
-          "<dfn>zus</dfn>" +
-          "<a>zus</a>" +
-          "</section>",
+        makeDefaultBody() + `
+          <section id='test'></section>
+          <section id='[[Type]]'></section>
+          <section id='%5B%5BValue%5D%5D'></section>
+          <section>
+            <a href='#test'>bar</a>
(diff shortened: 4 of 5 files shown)


js/profile-w3c-common.js
```diff
@@ -78,6 +78,7 @@ define(
     "w3c/permalinks",
     "core/id-headers",
     "core/rdfa",
+    "core/lint-hrefs-ids",
```
marcoscaceres: I wonder if this should just be imported directly in the w3c/linter.js? It doesn't make sense to run it here because this check should only occur after everything else finishes (and shouldn't block rendering).

src/core/lint-hrefs-ids.js
```diff
@@ -0,0 +1,21 @@
...
+
+import { pub } from "core/pubsubhub";
+
+export const name = "core/lint-hrefs-ids";
+
+export function run(conf, doc, cb) {
+  console.log()
+  Array.from(doc.querySelectorAll('a[href]'))
+    .map(a => ({ a, href: a.getAttribute('href') }))
+    .filter(({ href }) => Boolean(href))
```
marcoscaceres: combine the two filters into a single filter here. Also, do the filtering before the map, otherwise, we end up creating a bunch of extra objects that we don't need.

src/core/lint-hrefs-ids.js
```diff
@@ -0,0 +1,21 @@
...
+// Warns about href's that do not have corresponding id's
+
+import { pub } from "core/pubsubhub";
+
+export const name = "core/lint-hrefs-ids";
+
+export function run(conf, doc, cb) {
+  console.log()
+  Array.from(doc.querySelectorAll('a[href]'))
+    .map(a => ({ a, href: a.getAttribute('href') }))
```
marcoscaceres: You probably don't need this map at all, actually.

src/core/lint-hrefs-ids.js
```diff
@@ -0,0 +1,21 @@
...
+    .filter(({ href }) => Boolean(href))
+    .filter(({ href }) => href.startsWith('#'))
+    .map(({ a, href }) => ({ a, href, id: href.slice(1) }))
+    .filter(({ id }) => Boolean(id))
```
marcoscaceres: you don't need this filter here, as `getElementById()` returns a falsy value (null).

src/core/lint-hrefs-ids.js
```diff
@@ -0,0 +1,21 @@
+// Module core/lint-hrefs-ids
+// Warns about href's that do not have corresponding id's
+
+import { pub } from "core/pubsubhub";
+
+export const name = "core/lint-hrefs-ids";
+
+export function run(conf, doc, cb) {
```
marcoscaceres: let's call this `export function lint(doc){}` instead. 

Then when importing it in  w3c/linter.js
```JS
import {lint as fragmentLinter} from "core/lint-hrefs-ids";
```

And let's maybe move "core/lint-hrefs-ids.js" to "core/linters/check-fragments.js" or something that makes it clear what the rule is.

src/core/lint-hrefs-ids.js
```diff
@@ -0,0 +1,21 @@
...
+  console.log()
+  Array.from(doc.querySelectorAll('a[href]'))
+    .map(a => ({ a, href: a.getAttribute('href') }))
+    .filter(({ href }) => Boolean(href))
+    .filter(({ href }) => href.startsWith('#'))
+    .map(({ a, href }) => ({ a, href, id: href.slice(1) }))
```
marcoscaceres: You can probably get away with doing away with this one too... just keep reusing the `a` without mapping it (or destructure it as you go, but you don't need to map it).

src/core/lint-hrefs-ids.js
```diff
@@ -0,0 +1,21 @@
...
+    .filter(({ id }) => Boolean(id))
+    .filter(({ id }) => !doc.getElementById(id))
+    .forEach(({ href, id }) => {
+      pub("warn", `Found link with href '${href}' but no element with id '${id}'`);
```
marcoscaceres: here we actually do want to `console.warn(a)` in addition to `pub()`. 

Then the message is:  `Found `a` element with href '${href}' but no element with id '${id}'. See developer console for offending element.`

You can also do: `a.classList.add("respec-offending-element")`, which will add a red underline to the offending element.

tests/spec/core/lint-hrefs-ids-spec.js
```diff
@@ -0,0 +1,43 @@
+"use strict";
+describe("Core — Lint href's matching id's", function() {
+  afterAll(function(done) {
```
marcoscaceres: Nit: Would be great to use ES6 syntax for these tests.

tests/spec/core/lint-hrefs-ids-spec.js
```diff
@@ -0,0 +1,43 @@
...
+    flushIframes();
+    done();
+  });
+  it("should warn about href pointing nowhere", function(done) {
```
marcoscaceres: nit: should be "warns if can't match a fragment identifier to an id in the document"

tests/spec/core/lint-hrefs-ids-spec.js
```diff
@@ -0,0 +1,43 @@
...
+    var ops = {
+      config: makeBasicConfig(),
+      body:
+        makeDefaultBody() +
```
marcoscaceres: use template literal here. Makes tests easier to maintain.

src/core/ui.js
```diff
@@ -217,6 +217,12 @@ export const ui = {
   warning: function(msg) {
     errWarn(msg, warnings, "warning", "Warnings");
   },
+  getErrors: function() {
```
marcoscaceres: Not a huge fan of this, but I guess there is no better solution at the moment.

tests/spec/core/lint-hrefs-ids-spec.js
```diff
@@ -0,0 +1,43 @@
...
+          "<section>" +
+          "<a href='#test'>foo</a>" +
+          "</section>",
+    };
+    makeRSDoc(ops, function(doc) {
+      var ui = doc.defaultView.respecUI;
+      expect(ui.getErrors().length).toEqual(0);
```
marcoscaceres: this is going to be very fragile... I'm ok to leave it for now, but will definitely need to come up with a better way to subscribe to these events.

tests/spec/core/lint-hrefs-ids-spec.js
```diff
@@ -0,0 +1,43 @@
...
+      expect(ui.getWarnings().length).toEqual(1);
+    }).then(done);
+  });
+  it("should not warn about href pointing somewhere", function(done) {
+    var ops = {
+      config: makeBasicConfig(),
+      body:
+        makeDefaultBody() +
```
marcoscaceres: As above, use a template literal instead of string concatenation here.

src/core/lint-hrefs-ids.js
```diff
@@ -0,0 +1,21 @@
...
+
+export function run(conf, doc, cb) {
+  console.log()
+  Array.from(doc.querySelectorAll('a[href]'))
+    .map(a => ({ a, href: a.getAttribute('href') }))
+    .filter(({ href }) => Boolean(href))
+    .filter(({ href }) => href.startsWith('#'))
+    .map(({ a, href }) => ({ a, href, id: href.slice(1) }))
+    .filter(({ id }) => Boolean(id))
+    .filter(({ id }) => !doc.getElementById(id))
```
marcoscaceres: note that "id" coming from a href may be URLEncoded(), so you might need to `decodeURIComponent(id)` to match the `id`.

marcoscaceres: Should probably add a test for the above. I'm currently dealing with this in #1327.

marcoscaceres: see https://github.com/w3c/respec/pull/1329/files#diff-671809c9e7ab22675ffafb5ff8939406R11 for the kind of test we need for this.

src/core/lint-hrefs-ids.js
futpib: Without `.map` I will have to repeat `.getAttribute('href')` and `href.slice`. `a.href` (or `{ href }` destructuring) alone will not suffice because for some reason it return the complete URL when href attribute returns only the hash part. We could parse that URL, but really what's wrong with the maps, is it that big of a performance hit or something?

src/core/lint-hrefs-ids.js
futpib: Without `.map` I will have to repeat `.getAttribute('href')` and `href.slice`. `a.href` (or `{ href }` destructuring) alone will not suffice because for some reason it return the complete URL when href attribute returns only the hash part. We could parse that URL, but really what's wrong with the maps, is it that big of a performance hit or something?

src/core/lint-hrefs-ids.js
futpib: But falsy `getElementById()` means that the link is "offending". This `filter` currently only handles the case of an empty string, this could be done above with something like `href.startsWith('#') && href.length > 1` or `/^#.+/.test(href)` but I don't thing it's cleaner that way.

tests/spec/core/lint-hrefs-ids-spec.js
futpib: No luck with that, you can check out my attempts [here](https://github.com/futpib/respec/commit/8b60c98b88fdce620b144a10c4739f20400e0831). Here is the error I got:

```
  Core — Lint href's matching id's
    ✖ encountered a declaration exception
      Chrome 59.0.3071 (Linux 0.0.0)
    Error: it expects a function argument; received [object AsyncFunction]
        at Suite.describe (tests/spec/core/lint-hrefs-ids-spec.js:8:3)
        at tests/spec/core/lint-hrefs-ids-spec.js:3:1
```

src/core/lint-hrefs-ids.js
futpib: I see that linting is skipped if `specStatus` is `"unofficial"` [here](https://github.com/w3c/respec/blob/develop/src/w3c/linter.js#L46). But I'm working on documents totally unrelated to W3C with help of respec. Should this linting really work only for "official" docs? Can we make it also cover my "unofficial" use-case somehow?

marcoscaceres: I agree we should enable it generally. Let's do that as a separate bug tho.

src/core/lint-hrefs-ids.js
```diff
@@ -0,0 +1,26 @@
...
+    .map(({ a, href }) => ({ a, href, id: href.slice(1) }))
+    .filter(({ id }) => Boolean(id))
+    .filter(({ id }) => !doc.getElementById(id) && !doc.getElementById(decodeURIComponent(id)))
+    .forEach(({ a, href, id }) => {
+      const msg = `Found an element with href '${href}' but no element with id '${id}'.`
+            + " See developer console for offending element.";
+
+      pub("warn", msg);
+      console.warn("Offending element:", a);
+      a.title = msg;
```
marcoscaceres: Check here is the title is empty before setting it, as the author might be using it for something else. 

Alternatively, we could add it as a tool-tip:

```JS
a.dataset.toolTip = msg;
```

Then I can add some CSS and also make it accessible elsewhere.

src/core/lint-hrefs-ids.js
```diff
@@ -0,0 +1,26 @@
...
+// Warns about href's that do not have corresponding id's
+
+import { pub } from "core/pubsubhub";
+
+export const name = "core/lint-hrefs-ids";
+
+export function run(conf, doc, cb) {
+  Array.from(doc.querySelectorAll('a[href]'))
+    .map(a => ({ a, href: a.getAttribute('href') }))
+    .filter(({ href }) => Boolean(href))
```
marcoscaceres: The intent here is really to check if this is a local reference, so we should do that properly (below).  I also realized that the current code won't handle cases where the `a.href` is to the same document, but set using an absolute URL, like this:

```HTML
<a href="http://example.com/#foo">this is not handled at the moment</a>
```

Thus, we really want:

```JS
Array.from(document.querySelectorAll("a[href]"))
  .filter(
    ({ pathname, origin, hash }) =>
      hash && // also filters out just "#" 🎉
      origin === doc.location.origin &&
      doc.location.pathname === pathname
  )
  .filter(({ hash }) => !doc.getElementById(decodeURIComponent(hash.slice(1))))
  .forEach(a => {
     // The rest
   });
```

src/core/lint-hrefs-ids.js
marcoscaceres: See above, this is solved by the `a.hash`'s IDL attribute - as it returns an empty string for when set to "#".

src/core/lint-hrefs-ids.js
marcoscaceres: fixed typo in code above.

tests/spec/core/lint-hrefs-ids-spec.js
```diff
@@ -0,0 +1,61 @@
...
+            <a href='#test'>bar</a>
+            <a href='#[[Type]]'>qaz</a>
+            <a href='#%5B%5BType%5D%5D'>qaz</a>
+            <a href='#%5B%5BValue%5D%5D'>qaz</a>
+            <a href='http://example.com/#test'>buz</a>
+            <a href='http://example.com'>qaz</a>
+            <a href='#'>fuzz</a>
+            <a href=''>fuzz</a>
+            <dfn>zus</dfn>
+            <a>zus</a>
```
marcoscaceres: maybe add:
```
<script>
document.write(`<a href="${document.location.href + "#test"}">`)
</script>
```

marcoscaceres: fixed code in suggested test above.

src/core/lint-hrefs-ids.js
```diff
@@ -0,0 +1,26 @@
...
+    .filter(({ href }) => href.startsWith('#'))
+    .map(({ a, href }) => ({ a, href, id: href.slice(1) }))
+    .filter(({ id }) => Boolean(id))
+    .filter(({ id }) => !doc.getElementById(id) && !doc.getElementById(decodeURIComponent(id)))
```
futpib: Not really, the thing is that
```html
<section id='%5B%5BValue%5D%5D'></section>
<a href='#%5B%5BValue%5D%5D'></a>
```
works ok (at least in chromium), that's the reason for two checks (with and without decode) and also for this [test case](https://github.com/w3c/respec/pull/1326/files/975a2376a8a4d492efea2553151a46c234f9689f#diff-ba3e08b1240269a63e384c97af73839aR38).

marcoscaceres:
I snuck in a bit of time this weekend and rewrote the linter and rules:
https://github.com/w3c/respec/pull/1332/files#diff-4788eee77f13cbc3541c56afa9d08763

I've split out the w3c rules and the core rules. 

Will try to land it tomorrow. Need to hook up the tests. Then slotting this as a one of the default rule will be trivial.

marcoscaceres:
@futpib it it ok if I take this and add it to #1332?

futpib:
@marcoscaceres yes, no problem.

marcoscaceres:
Ok, we now have a [proper linter](https://github.com/w3c/respec/blob/develop/src/core/linter.js)... so will try to migrate this to that soon:

marcoscaceres:
Closing, will file a new bug for this that can use the new linter.

PR #1334, opened 2017-07-26
Adds "preview" status, for @tobie previews

marcoscaceres:
* makes specs unofficial when previewing

### DIFF
--- a/js/w3c/templates/cgbg-sotd.html
+++ b/js/w3c/templates/cgbg-sotd.html
@@ -1,4 +1,16 @@
 <h2>{{l10n.sotd}}</h2>
+{{#if isPreview}}
+  <details class="annoying-warning" open="">
+    <summary>This is a preview</summary>
+    <p>
+      Do not attempt to implement this version of the specification. Do not reference this
+      version as authoritative in any way.
+      {{#if edDraftURI}}
+        Instead, see <a href="{{edDraftURI}}">{{edDraftURI}}</a> for the Editor's draft.
+      {{/if}}
+    </p>
+  </details>
+{{/if}}
 <p>
   This specification was published by the <a href='{{wgURI}}'>{{wg}}</a>.
   It is not a W3C Standard nor is it on the W3C Standards Track.
--- a/js/w3c/templates/sotd.html
+++ b/js/w3c/templates/sotd.html
@@ -1,7 +1,7 @@
 <h2>{{l10n.sotd}}</h2>
 {{#if isPreview}}
   <details class="annoying-warning" open="">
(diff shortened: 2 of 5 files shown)


marcoscaceres:
@tobie, this now will set things to unofficial and add a warning if the spec status is "unofficial". When you generate the spec, can you please add `?specStatus=unofficial` to the URL.

tobie:
Might also be worth adding a link back to the PR, etc.

tobie:
OK, shipped and deployed this. LMK how it goes?

tobie:
Oh, wait! Do you want `specStatus=preview` or `specStatus=unofficial`?

marcoscaceres:
specStatus=preview

tobie:
Fixed.

tobie:
This seems to be breaking specs which now reference missing resources on w3.org, e.g.: https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/issue-1470-patch.html

marcoscaceres:
@tobie thanks for trying it out. Upon reflection, I think what I actually want here is only to signal that it's a preview (`?isPreview=true`), which will add the ugly red-annoying-box but not change the status.

It occurred to me me that Keeping the spec status is useful for checking snapshots against PubRules etc. I'll update the PR and ping you to change things again on your side (to append  `isPreview=true` on your end). Appreciate your time and help with this!

tobie:
Added isPreview=true.

marcoscaceres:
@tobie, I'll implement the branch thing separately. Might need to chat to you about it in IRC. There are a few things that could work, like: 

`?isPreview=true&github={"branch": "branch", repoURL: "https://github.com/w3c/repo"}`

Then I can key off `github` there to provide the link back, similar the WHATWG specs. However, I don't know how useful that is, as people don't generally browse branches (maybe I'm missing something).

tobie:
I was mostly just thinking about having a link back to the PR. Useful in case you land on it and want the broader context and/or the merge status.

tobie:
You can get that easily from the [config object](https://github.com/tobie/pr-preview#params-optional):
```json
{
    "params": {
        "prUrl": "{{ pull_request.html_url }}",
        "prNumber": "{{ pull_request.number }}"
    }
}
```

So it's trivial to add them to extra/default params in the URL.

marcoscaceres:
oh, nice! Ok, would need to add those to ReSpec - but would be trivial.

tobie:
Well, LMK which ones you'll need. :)

marcoscaceres:
@tobie will do! Will try to hack on it next week.

PR #1354, opened 2017-08-09
Fix bug in detecting wrong configuration for multiple WGs

dontcallmedom:
Right now, respec complains if all of wg, wgUri, and wgPatentUri are arrays, when it should complain only if some but not all are
Bug was introduced in 0b706ce8ce3b524ba442bedd154dfb056503e131


marcoscaceres:
Whoops 😬 nice catch! Thanks for bisecting it.

marcoscaceres:
Ok, 🚢'ed

PR #1359, opened 2017-08-12
incl. meta-description in seo.js (closes #1336)

### LINKED ISSUE #1336: Google getting wrong description
Because specs lack `<meta description>`, Google is getting confused and grabbing the wrong sentence:

screenshot 2017-07-27 16 51 53

We should consider adding `<meta description>`

dhvenema:


src/w3c/seo.js
```diff
@@ -50,5 +50,11 @@ export function run(conf, doc, cb) {
     linkElem.setAttribute("href", conf.canonicalURI);
     doc.head.appendChild(linkElem);
   }
+  var metaDescription = doc.querySelector("#abstract p:first-of-type")
```
marcoscaceres: As this is universally useful, we should make a new module ("core/seo").  Also, remember to only use "var" for global variables (basically never): always use `const` unless the variable is going to change , in which case you use "let". 

```JS
export async function run(conf, doc, cb) {
  const metaDescription = doc.querySelector("#abstract p:first-of-type");
  if (!metaDescription) {
    return; // no abstract, so abort early
  }
  // This is not critical, so let's continue other processing first:
  cb();
  await doc.respecIsReady;
  const metaElem = doc.createElement("meta");
  metaElem.name = "description";
  metaElem.content = metaDescription;
  doc.head.appendChild(metaElem);
}
```

marcoscaceres:
Replaced by https://github.com/w3c/respec/pull/1364

PR #1375, opened 2017-08-28
Use locally cited definitions when available for automatic links on W…

### LINKED ISSUE #1372: Use defined terms list for autolinks in WebIDL when available
ReSpec kindly links to terms it magically knows in WebIDL fragments. But given that (for better or for worse), there are different versions of different specs that can be used as references for some of these terms, it would be best if it would use instead terms that the spec itself defines and references when it does.

e.g. if a spec uses `[CEReactions]` but doesn't define it, then it would use the current default link.

if a spec does `<dfn data-cite="!custom-elements#CEReactions">[CEReactions]</dfn>`, then the WebIDL fragment should link to that `dfn` (or maybe to the link it creates, I'm not sure).

dontcallmedom:
…ebIDL types

close #1372

### DIFF
--- a/src/core/webidl.js
+++ b/src/core/webidl.js
@@ -1136,6 +1136,16 @@ export function run(conf, doc, cb) {
     $(doc).find("head link").first().before($("<style/>").text(css));
   }
 
+  // update standardTypes based on local definitions
+  for (let k of standardTypes.keys()) {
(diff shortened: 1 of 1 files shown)


dontcallmedom:
replaced by #1376

PR #1376, opened 2017-08-28
Use locally defined references / definitions when available for auto links

dontcallmedom:


marcoscaceres:
Thanks! Will review first thing tomorrow.

marcoscaceres:
If you could provide some test cases, that would be great. I'll rework the algorithms into functional style (more maintainable for me).

marcoscaceres:
Thanks! will try to get to this really soon... bit sick at the moment, and it's slowing me down 🤧

marcoscaceres:
@dontcallmedom so, I started working on this, and I know we had a brief discussion... and this is kinda independent of ReSpec... but now I'm really concerned by what this PR would enable. 

Can you help me understand why, and which specs, would want to reference either outdated or forked specs? This seems like a technical bandaid to a much larger problem (and one that could negatively impact/confuse developers, implementers, and people working on documentation).

dontcallmedom:
re which specs would refer specs different than the one "blessed by ReSpec", any spec that would have to comply with the current [W3C normative references guidance](https://www.w3.org/2013/09/normative-references) - in particular, specs that only need and want to reference specs that are at a more advanced Rec-stage status.

There are also cases where a given spec cannot get consensus for publication if it refers to HTML-LS vs W3C-HTML - I don't recall off the top of my mind how much of has been shared publicly, so I'll keep it at that level of description.

Generally speaking, I think it's great when ReSpec helps with default references, but I feel pretty strongly ReSpec should not try to impose a particular view of what a right reference is (the same way it doesn't prevent from using dated references from bibref).

marcoscaceres:
Yeah, it's sometimes hard to keep neutral. I guess it already defaults to, um, "the right thing"(tm). Thanks for clarifying. I'll send an updated PR soon...

marcoscaceres:
I might have figured out a different way of solving this (maybe)... if you define a `localBiblio`, you can either override and/or alias a different spec. 

```JS
localBiblio: {
  "WEBIDL": {
    "authors": [""],
    "date": "",
    "href": "http://www.w3.org/TR/test",
    "publisher": "W3C",
    "status": "",
    "title": "This is a test"
  },
  "HTML": {
    aliasOf: "HTML52"
  }
}
```

dontcallmedom:
clever! the only possibly significant downside is that it only works if the paths (i.e. possible sub-page and anchor) match across the override.

marcoscaceres:
The problem is that ReSpec assumes `CEReactions` lives in the HTML spec. Do you know if the W3C HTML fork will eventually get the custom elements definitions?

marcoscaceres:
Good point about the anchors... I don't know if those have been carried over to the w3c spec. I would assume so.

dontcallmedom:
re anchors, they definitely have NOT in HTML-LS vs W3C-HTML

re custom elements, see https://github.com/w3c/html/issues/955

marcoscaceres:
@dontcallmedom, I asked in w3c/html#955 for clarification on Custom Elements (and if the w3c will actually take up that work separately or if it will be  merged into W3C HTML). If it gets merged into HTML, then we might be able to work around this with the `localBiblio` hack.

dontcallmedom:
@marcoscaceres as alluded above, I don't think solving the custom-elements specific issue is enough to make the `localBiblio` hack work. To give a specific example, `EventHandler` is auto-linked to `!HTML#eventhandler`, the HTML 5.2 equivalent is `!HTML52/webappapis.html#event-handler` - and that's one of many examples (I have had to fix a bunch of these recently).

marcoscaceres:
@dontcallmedom, ok, np. Will work to make this happen then.

dontcallmedom:
@marcoscaceres any news on this? Thanks!

marcoscaceres:
Sorry, was hoping to get clarity on Custom Elements:  https://github.com/w3c/html/issues/955#issuecomment-326502756

dontcallmedom:
@marcoscaceres but again :), even if custom-element was fixed, this would not alleviate the more generic issue.

marcoscaceres:
@dontcallmedom I'm really sorry for the delay here... I'm trying my best to get to this 😭 - super bogged down with payment request test suite. I still want to rewrite the algorithms to follow the more functional style (unless you want to do it):I have a hard time deciphering the current ones. 

The other thing is just a nit... we use [prettier](https://github.com/prettier/prettier) for code formatting in ReSpec now.

marcoscaceres:
@dontcallmedom, I've decided to rewrite the IDL and linking parts of ReSpec. All the linking code is a rat's nest + jQuery. 

What I'm going to do is allow a configurable override for your use case - so you can override the default links:

```JS
const respecConfig = {
   webIDL: {
       // These get merged, overriding the default ones
       extenedAttributesLinks: [
         ["CEReactions", "W3CWebComponets#cereactions"],
      ],
   }
}
```

dontcallmedom:
cool, thanks; will that work also for auto-linked interface names ("standard types" in the current code)?

marcoscaceres:
I'll make it so one can override whatever things are hard coded - so that should include "standard types".

marcoscaceres:
Quick update - I've (locally) created a module called "core/xrefs", which serves as global store for cross-refs. It will be overridable with: `conf.xrefs = [[foo, "spec#bar"]]`; 

Will hopefully have everything landed before TPAC.

marcoscaceres:
(updated comment above)

marcoscaceres:
Closing this - still planning to improve this situation... sorry @dontcallmedom, will try to do better here :(

PR #1377, opened 2017-08-28
When autolinking WebIDL, use a normative reference

### LINKED ISSUE #1214: Making WebIDL reference normative
It seems ReSpec references the WebIDL spec informatively when it uses WebIDL syntax which seems weird. See https://heycam.github.io/webidl/#referencing.

dontcallmedom:
close #1214


marcoscaceres:
I'm still a little unsure about this :(.... the problem is that, if I'm writing a Note or an example, then I might want to non-normatively reference WebIDL. 

To trigger a normative reference for WebIDL, the Editor just needs to have "!WEBIDL" somewhere in the document.

dontcallmedom:
I'm fine either way; I thought I would file the PR since you seemed to have leaned toward this in the last discussion I read on the topic, but I agree that there is nothing wrong in expecting editors to make at least one normative ref to WebIDL (at least as long as something along of #1372 gets merged)

marcoscaceres:
Ok, let's pushed for #1372 and related PR for now.

PR #1383, opened 2017-09-25
feat(core/css/respec): remove RFC2119 styling (close #1380)

### LINKED ISSUE #1380: Readability of RFC 2119 terms
Hello,

I have been noticing several specifications published with ReSpec that render RFC2119 terms as [red lowercase text](https://github.com/fcrepo/fcrepo-specification/issues/195#issuecomment-323314368), rather than capitalized.  Unfortunately, certain individuals (including myself) are colorblind, and cannot distinguish these terms at all when rendered in this manner.

If the desired outcome is that is that RFC2119 terms should receive no emphasis at all (as suggested [here](https://github.com/w3c/payment-request/issues/377#issuecomment-281285655), then please remove the color as well.  Otherwise, please choose a style that emphasizes the terms by altering their shape (such as small caps, italics, underline, monospace, etc).  As it stands now, the default rendering behavior is not entirely inclusive of the population of readers who cannot easily distinguish colors.

To the extent that style alignment with the W3C is a [broader issue](https://github.com/w3c/respec/issues/748), if this is not the right venue in which to bring up this topic, please suggest which one would be most appropriate.

My own personal preference, as somebody who reads both W3C and IETF standards documents, is that small caps are ideal for RFC2119 terms since since visually consistent between the two, emphasizes their importance, and makes it easier to find the important/normative clauses of specs when coding implementations.  [Distracting](https://github.com/w3c/payment-request/issues/377) is can be good when using a spec as a reference document when doing work

Potentially interested parties:
@anarchivist (from #1109)
@azaroth42 (from f628857)
@domenic (from [payment-request/#377](https://github.com/w3c/payment-request/issues/377))

marcoscaceres:


halindrome:
I am cool with this, but I assume we want to want on merging until the base style sheet from the W3C is updated?  Or has that already happened?

marcoscaceres:
Yep - we should wait, tho merging this even now would adhere more closely to the w3c style manual, no?

Kimberly050174:
V

PR #1387, opened 2017-10-06
Remove special chars

### LINKED ISSUE #1353: Parentheses in IDs
I just noted that automatically generated IDs in respec do not remove parentheses from the ID value, resulting, for example, in URLs like this: 

> https://rawgit.com/w3c/wcag21/target-size_ISSUE-60/guidelines/#target-size-(no-exception)

Having parentheses is not only hard to type, but they are also quite uncommon in URLs. Email clients sometimes struggle with those characters when linking them, too, resulting in missing the last `)`, therefore breaking the link.

Can respec just ignore parentheses in IDs, resulting in links like:

> https://rawgit.com/w3c/wcag21/target-size_ISSUE-60/guidelines/#target-size-no-exception

I recognize that the HTML spec allows [any value for IDs](https://www.w3.org/TR/html/dom.html#the-id-attribute), but this is more a style question, and I hope it can be considered.

marcoscaceres:
* closes #1353


dontcallmedom:
I only glanced through the changes, but it looks like it only special-cases parenthesis, where I think it would make more sense to only keep alphanumeric characters plus a few punctuations. If not, we should at least further special-case &quot; since that created [validity bugs](https://github.com/w3c/respec/issues/1353#issuecomment-333763926)

marcoscaceres:
Good point. I’ll restrict to ascii and just lower-case everything.

marcoscaceres:
I added some error recovery to lessen the impact a bit. Legacy fragments generated by ReSpec get converted to the new format. Custom ones that an author might have included (e.g., `id="foo()"`) get left alone.

PR #1402, opened 2017-10-20
Node 8 required

tripu:
It's my understanding that the current release of Respec requires Node.js &ge; 8. Is that so?

This makes that explicit in `package.json`.


marcoscaceres:
Thanks! I should have done the changes in a bit after a week when 8 goes LTS.

PR #1457, opened 2018-01-24
fix: dont announce highlights (closes #1456)

### LINKED ISSUE #1456: Code samples should not be set to aria-live="polite" after core/highlight executes
This is a regression for a7394e3cd22513e873a706f6e3391d6fb6bff323, which presents a significant accessibility problem for screen reader users trying to read a spec containing a lot of code examples, because every code sample is sent to the the screen reader buffer to announce at its earliest convenience. 

I've sometimes been unable to test a page using VoiceOver with a W3C spec open in another tab, because all I hear are source code examples from the spec being announced.

marcoscaceres:


marcoscaceres:
@majornista, is this all that is needed?

marcoscaceres:
@majornista, good point, removed the one on body too.

PR #1460, opened 2018-01-31
Fix Nightmare API use

saschanaz:
Nightmare doesn't support `timeout` or `userData` arguments so this PR maps them correctly.


marcoscaceres:
I guess once this gets rebased Travis will be happy.

PR #1459, opened 2018-01-31
Add package-lock.json

saschanaz:
This PR adds npm5+ package-lock.json [which is recommended](https://docs.npmjs.com/files/package-lock.json) for better installation speed.

PR #1461, opened 2018-02-01
fix: roll back to uglify-es 3.2.2

### LINKED ISSUE #1458: Test failure: 'root' is deprecated, use 'global'
https://travis-ci.org/w3c/respec/jobs/335038255

Not sure what happened there...

saschanaz:
Fixes #1458 

It turns out that the recent update of uglify-es is the culprit.

(This includes #1459 so you will probably want to merge it first)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4198,6 +4198,32 @@
             "lru-cache": "2.7.3",
             "sigmund": "1.0.1"
           }
+        },
+        "uglify-js": {
+          "version": "2.4.24",
+          "resolved": "https://registry.npmjs.org/uglify-js/-/uglify-js-2.4.24.tgz",
+          "integrity": "sha1-+tV1XB4Vd2WLsG/5q25UjJW+vW4=",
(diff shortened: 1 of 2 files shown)


saschanaz:
Probably worth trying babel-minify, BTW.

marcoscaceres:
Nice find! 🕵️‍♂️

marcoscaceres:
@saschanaz, will babel-minify allow us to get rid of the dependency on uglify? I guess we would only run it when we generate the packaged version.

saschanaz:
Yes, we won't have to call babel and then uglify, we can just call babel-minify and that's it.

marcoscaceres:
Ok, but it would be a hard requirement that we can generate a concatenated, but *not* minified, version for development... otherwise, debugging is going to be extremely hard. Sourcemaps are very very broken across browsers, so definitely don't want to be reliant on them (been there, done that, never again 🤪).

PR #1470, opened 2018-02-02
bundled starter page

nileshgulia1:
Bundled starter page,Can we use parcel's dev server instead of `http server`.


saschanaz:
What's the merit when we do this?

marcoscaceres:
Agree, would be great to hear a little more as I’ve never used parcel.

nileshgulia1:
Actually, parcel doesn't need a configuration file like webpack, we just point it to a single file and let it do the magic, I thought it will be good if we start with `starter` page or `basic .html` which will be good for the contributers in future.

saschanaz:
1. It seems parcel uses `uglify-es@^3.2.1` which includes `3.3.9` that caused https://github.com/w3c/respec/issues/1458.
2. Using parcel probably means we don't have to directly depend on uglify. Can you modify `tools/builder.js` to use parcel instead of babel+uglify?

nileshgulia1:
@saschanaz  I have removed `Uglify.js` package and make parcel to build `builder.js` file.

saschanaz:
(Note that Travis test won't run until the conflicts are resolved)

nileshgulia1:
@saschanaz  Done!

nileshgulia1:
@saschanaz  I don't know why travis is failing, while my local `npm run build` commands are executing perfectly.

saschanaz:
It seems `npm run parcel` is what failing now.

nileshgulia1:
Thanks a lot @saschanaz ;)

marcoscaceres:
Seems this needs a rebase... there are also a bunch of cache files that probably should not be there?

saschanaz:
Also, probably the target shouldn't be `builder.js` as it's just a builder script, not an actual library. And the result should be in `builds/`, not in `dist/`.

nileshgulia1:
>Also, probably the target shouldn't be builder.js as it's just a builder script

@saschanaz   Parcel does not support multiple entry points like webpack. I guess there should be one main entry point and parcel should bundle these files to the specified directory.

saschanaz:
Currently the entry point is `/js/profile-w3c-common.js`, as you can see in `examples/basic.html`:

https://github.com/w3c/respec/blob/cab4974fcab6cb564533a4ccce46ce79debb2970/examples/basic.html#L6-L7

Issue #651, opened 2016-03-14
"Open an Issue" button


halindrome:
If there is an issue tracking link available to respec, it would be nice to expose a standard "open an issue" connection to make it easier for document reviewers to quickly file comments.  Could be in the pill...  could be somewhere else.  

This might require making the otherLinks object a bit more sophisticated.

marcoscaceres:
A link to open an issue is now provided via `github`. 

It might be nice to add the ability to select text and file an issue about it, like they have in the WHATWG specs.

marcoscaceres:
In the URL, one can do: 
 * title=
 * body=

himanish-star:
@marcoscaceres, I went through this issue and also tested out how WHATWG works. So may I give this a try?

himanish-star:
So do you want to create a new button or you want to add this functionality to the existing `File it!` button shown in the pic below. But in this case the user would have to open `ReSpec > About ReSpec` then click on `File it!`

![image](https://user-images.githubusercontent.com/30361728/35763476-f93ddaec-08d2-11e8-8a79-9dc12e0f8976.png)

himanish-star:
I have implemented a solution. Please see to the GIF below for viewing the changes. Can I file a PR now?
![ezgif com-video-to-gif](https://user-images.githubusercontent.com/30361728/35778846-1973cb60-09ea-11e8-8f41-853ccae65ab1.gif)

marcoscaceres:
Sure, send it up. However, I have some suggestions of how we can make this better from a UI and UX perspective.

xfq:
FYI - this is what WebDriver uses: https://github.com/w3c/webdriver/blob/master/issue.js

marcoscaceres:
Closing this feature request. The idea of filing an issue about selected text from the ReSpec pill is interesting, but after 10 years with no implementation or demand, it's not something we're planning to pursue. Modern GitHub issue templates with URL parameters achieve a similar workflow. Thanks for the idea!

PR #1473, opened 2018-02-05
File issue related to the selected text

himanish-star:
fixes: #651 
Are the changes made in the right file and location?
EDIT: ` ✖ removes left padding before processing markdown content`. This is the error the CI test is showing. How do I fix this?

### DIFF
--- a/js/ui/ui.css
+++ b/js/ui/ui.css
@@ -79,6 +79,20 @@
   transition: color 0.2s;
 }
 
+#respec-file-issue-btn {
+  height: 20px;
+  width: 15px;
+  background-color: rgb(255, 7, 5);
+  color: rgb(251, 249, 255);
+  border-radius: 5px;
+  border: none;
+  font-weight: bold;
+}
+
+.respec-file-issue-hidden {
+  display: none;
+}
+
 #respec-menu {
   position: absolute;
   margin: 0;
--- a/src/core/github.js
+++ b/src/core/github.js
@@ -5,9 +5,106 @@
  */
 
 import l10n from "core/l10n";
+import css from "deps/text!deps/balloon.css";
 import { pub } from "core/pubsubhub";
 
 export const name = "core/github";
+
+function getSelectionCoordinates() {
+  let doc = window.document;
+  let selection = doc.selection, range, rects, rect;
+  let x = 0, y = 0;
+  if (selection) {
+    if (selection.type !== "Control") {
+      range = selection.createRange();
+      range.collapse(false);
+      x = range.boundingLeft;
+      y = range.boundingTop;
+    }
+  } else if (window.getSelection) {
+    selection = window.getSelection();
+    if (selection.rangeCount) {
+      range = selection.getRangeAt(0).cloneRange();
+      if (range.getClientRects) {
+        range.collapse(false);
+        rects = range.getClientRects();
+        if (rects.length > 0) {
+          rect = rects[0];
+        }
+        x = rect.left;
+        y = rect.top;
+      }
+      // Fall back to inserting a temporary element
+      if (x === 0 && y === 0) {
+        const span = doc.createElement("span");
+        if (span.getClientRects) {
+          // Ensure span has dimensions and position by
+          // adding a zero-width space character
+          span.appendChild( doc.createTextNode("\u200b") );
+          range.insertNode(span);
+          rect = span.getClientRects()[0];
+          x = rect.left;
+          y = rect.top;
+          const spanParent = span.parentNode;
+          spanParent.removeChild(span);
+
+          // Glue any broken text nodes back together
+          spanParent.normalize();
+        }
+      }
+    }
+  }
+  return { x: x, y: y };
+}
+
+
+function enableFileIssuesOnSelect(issueBase){
+  const { body, head } = document;
+
+  const codeStyle = document.createElement("style");
+  codeStyle.textContent = css;
+  codeStyle.classList.add('removeOnSave');
+  head.appendChild(codeStyle);
+
+  var fileIssueLink = document.createElement('button');
+  fileIssueLink.setAttribute('data-balloon', 'file an issue related to the text selected');
+  fileIssueLink.setAttribute('data-balloon-pos', 'up');
+  fileIssueLink.setAttribute('id', 'respec-file-issue-btn');
+  fileIssueLink.classList.add('removeOnSave', 'respec-file-issue-hidden');
+  fileIssueLink.textContent = 'i';
+  fileIssueLink.addEventListener("click", () => {
+    window.location.href = fileIssueLink.dataset.href;
+    fileIssueLink.classList.add('respec-file-issue-hidden');
+  });
+  body.appendChild(fileIssueLink);
+
+  body.addEventListener("mouseup", () => {
+    const textSelected = window.getSelection().toString();
+    const newIssueURL = new URL('./new',issueBase);
+    const { searchParams } = newIssueURL;
+    let titleText;
+
+    if (textSelected) {
+      if (textSelected.length > 40) {
+        const pos = textSelected.indexOf(" ", 40);
+        titleText = textSelected.slice(0, pos).concat("...");
+      } else {
+        titleText = textSelected;
+      }
+      searchParams.set('title', `'${titleText}'`);
+      searchParams.set('body', `> ${textSelected}`);
+      fileIssueLink.classList.remove('respec-file-issue-hidden');
+      fileIssueLink.dataset.href = newIssueURL.href;
+      const { x, y } = getSelectionCoordinates();
+      fileIssueLink.style.position = "fixed";
+      fileIssueLink.style.left = `${x}px`;
+      fileIssueLink.style.top = `${y}px`;
+    } else {
+      fileIssueLink.classList.add('respec-file-issue-hidden');
+    }
+  });
+
+}
 
 function findNext(header) {
   // Finds the next URL of paginated resources which
@@ -97,6 +194,10 @@ export async function run(conf) {
     githubAPI: `https://api.github.com/repos/${org}/${repo}`,
     issueBase: `${ghURL.href}${ghURL.pathname.endsWith("/") ? "" : "/"}issues/`,
   };
+  if (conf.enableIssueFeature && newProps.issueBase &&
+    '/https://github.com/'.search(newProps.issueBase)) {
+    enableFileIssuesOnSelect(newProps.issueBase);
+  }
   const commitsHref = `${ghURL.href}${
     ghURL.pathname.endsWith("/") ? "" : "/"
   }commits/${branch}`;
--- a/package.json
+++ b/package.json
@@ -75,6 +75,7 @@
     "prepublish": "npm run snyk-protect"
   },
   "dependencies": {
+    "balloon-css": "^0.5.0",
     "colors": "^1.1.2",
     "command-line-args": "^5.0.0",
     "command-line-usage": "^4.1.0",
--- a/tools/copydeps.js
+++ b/tools/copydeps.js
@@ -12,6 +12,7 @@ const unlink = promisify(fs.unlink);
 const srcDesMap = new Map([
   ["./node_modules/clipboard/dist/clipboard.js", "./js/deps/"],
   ["./node_modules/domReady/domReady.js", "./js/deps/"],
+  ["./node_modules/balloon-css/balloon.css", "./js/deps/"],
   [
     "./node_modules/handlebars/dist/handlebars.runtime.js",
     "./js/deps/handlebars.js",


js/ui/ui.css
```diff
@@ -79,6 +79,15 @@
   transition: color 0.2s;
 }
 
+#fileIssueLink {
```
marcoscaceres: nit: prefix with `respec-` and separate words with "-".

src/core/ui.js
```diff
@@ -24,6 +24,14 @@ styleElement.classList.add("removeOnSave");
 
 document.head.appendChild(styleElement);
 
+let body = $('body');
```
marcoscaceres: Nit, avoid using jQuery... we are trying to remove it from the project.

src/core/ui.js
```diff
@@ -24,6 +24,14 @@ styleElement.classList.add("removeOnSave");
...
+let fileIssueLink = $('<a href="https://github.com/w3c/respec/issues/new?title=&body=" id="fileIssueLink">file an issue related to the text selected</a>');
```
marcoscaceres: Nit: add css class "removeOnSave" to the element.

js/ui/ui.css
marcoscaceres: Ideally, we only want this to show when the use selects some text. So, it would be nice to fade it in. Also, putting it down the right-hand corner is kinda ugly... might be cool to experiment with:

https://kazzkiq.github.io/balloon.css/

src/core/ui.js
marcoscaceres: As this is github related, the best place for this code would be "src/github.js".

src/core/ui.js
```diff
@@ -24,6 +24,14 @@ styleElement.classList.add("removeOnSave");
...
+body.append(fileIssueLink);
+body.mouseup(function () {
+  let textSelected = window.getSelection().toString();
+  fileIssueLink.attr('href',`https://github.com/w3c/respec/issues/new?title='${textSelected}'&body=${textSelected}`)
```
marcoscaceres: Naturally, you want the spec's "issueBase" instead of ReSpec issue's URL 😉

marcoscaceres: Please construct the URL using "new URL()" and then set the `title` and `body` via the URL's `searchParams` property. Then assign the resulting string via `a.href = fileIssueLink.href;`

src/core/ui.js
marcoscaceres: this should have been "src/core/github.js"

marcoscaceres:
Just noting we then need a way of testing this, but we will get to that.

js/ui/ui.css
himanish-star: > prefix with respec- and separate words with "-".

Ya I'll do this 
_______
> Ideally, we only want this to show when the use selects some text. So, it would be nice to fade it in. Also, putting it down the right-hand corner is kinda ugly... might be cool to experiment with:

Great idea. Will do this too!

src/core/ui.js
himanish-star: Oops!  Sorry did not know about this. Will revert back to  normal  javascript

himanish-star: Got it

src/core/ui.js
himanish-star: Sure

src/core/ui.js
himanish-star: OK will do this too.

himanish-star: How do I get the `issueBase` of the `spec` ?

himanish-star:
@marcoscaceres, I have made the changes. Please see the GIF below. Altso I have some doubts
  - I have for now appended the `link` for using the `Balloon.css`. This is the reason why the class `removeOnSave` fails to do its job. Please tell me how to fix this.


ezgif com-video-to-gif

src/core/github.js
```diff
@@ -9,6 +9,44 @@ import { pub } from "core/pubsubhub";
 
 export const name = "core/github";
 
+var body = document.body;
```
marcoscaceres: nit: 
```
const { body, head } = document;  
```

src/core/github.js
```diff
@@ -9,6 +9,44 @@ import { pub } from "core/pubsubhub";
...
+var head = document.head;
+var issueBase;
```
marcoscaceres: you need to get `issueBase` from the `conf` object that is passed to the `.run()` function. Then, all this should be wrapped in a function.

```JS
function enableFileIssuesOnSelect(issueBase){
   // ... your code here...
}

// This is just an example:
function run(conf, doc, cb){
   
   // we need a way to turn off the feature, and it only works if there is an issue base.
  // and `issueBase` is pointing to github!  
   if(conf.useSelectToFileIssues && conf.issueBase && /https://github.com/.test(conf.issueBase) ){
      enableFileIssuesOnSelect(conf.issueBase)
   }
}
```

src/core/github.js
```diff
@@ -9,6 +9,44 @@ import { pub } from "core/pubsubhub";
...
+fileIssueLink.classList.add('removeOnSave');
+fileIssueLink.id = 'respec-file-issue-btn';
+fileIssueLink.innerHTML = 'File Issue';
+fileIssueLink.onclick = function () {
+  window.location.href = fileIssueLink.getAttribute('data-href');
+  fileIssueLink.style.display = 'none';
+};
+
+var link = document.createElement('link');
+link.rel = 'stylesheet';
```
marcoscaceres: Note: this link also needs `removeOnSave`, as we don't want ballon.css in the final spec.

src/core/github.js
```diff
@@ -9,6 +9,44 @@ import { pub } from "core/pubsubhub";
...
+link.href = "https://cdnjs.cloudflare.com/ajax/libs/balloon-css/0.5.0/balloon.min.css";
```
marcoscaceres: Generally, we want control our own dependencies. So, we want to:

1. `npm install balloon-css` 
1.  add the patch to [tools/copydeps.js](https://github.com/w3c/respec/blob/develop/tools/copydeps.js#L13)
1. import the css as text, and add it to this `link` (see how we do that in "w3c/styles").

src/core/github.js
```diff
@@ -9,6 +9,44 @@ import { pub } from "core/pubsubhub";
...
+
+head.appendChild(link);
+body.appendChild(fileIssueLink);
+
+body.onmouseup = function () {
```
marcoscaceres: Use `.addEventListener` here, otherwise, you risk trashing any existing event listener for mouseup.

src/core/github.js
```diff
@@ -9,6 +9,44 @@ import { pub } from "core/pubsubhub";
...
+  var textSelected = window.getSelection().toString();
```
marcoscaceres: What happens if the use selects the whole document? We should limit the length of the selection to something reasonable.

src/core/github.js
```diff
@@ -9,6 +9,44 @@ import { pub } from "core/pubsubhub";
...
+  var newIssueURL = new URL('new?title=&body=',issueBase);
```
marcoscaceres: Nit: just do, `new URL('./new', issueBase);`.

src/core/github.js
```diff
@@ -9,6 +9,44 @@ import { pub } from "core/pubsubhub";
...
+  var searchParams = newIssueURL.searchParams;
```
marcoscaceres: nit: `const { searchParams } = newIssueURL;`

src/core/github.js
```diff
@@ -9,6 +9,44 @@ import { pub } from "core/pubsubhub";
...
+
+  if (textSelected) {
+    searchParams.set('title',`'${textSelected}'`);
+    searchParams.set('body',textSelected);
+
+    fileIssueLink.style.display = 'inherit';
```
marcoscaceres: would prefer here, and below, we just set a CSS class instead (i.e., don't set `style` directly). I'm still a bit confused tho... why don't we just let balloon.css handle everything? Is that possible... then we don't need to add any CSS and we don't need "#respec-file-issue-btn" at all.

src/core/github.js
```diff
@@ -97,6 +135,7 @@ export async function run(conf) {
     githubAPI: `https://api.github.com/repos/${org}/${repo}`,
     issueBase: `${ghURL.href}${ghURL.pathname.endsWith("/") ? "" : "/"}issues/`,
   };
+  issueBase = newProps.issueBase;
```
marcoscaceres: you shouldn't need this :)

src/core/github.js
marcoscaceres: Note that my proposal for `conf.useSelectToFileIssues` is a terrible name... can you think of something better/shorter?

src/core/github.js
himanish-star: :+1:

src/core/github.js
himanish-star: Ok lemme think

src/core/github.js
himanish-star: Cool, will do this too

src/core/github.js
himanish-star: Ok

src/core/github.js
himanish-star: Ok

src/core/github.js
himanish-star: Hmm, any suggestions to what the length should be?

src/core/github.js
marcoscaceres: fixed typo above... should have been "path" not "patch".

marcoscaceres:
Couple of other things to consider... how will this work on a mobile phone? And, if the ReSpec pill menu at the top of the document is open, what should happen to this functionality?

src/core/github.js
himanish-star: ok

src/core/github.js
himanish-star: :+1:

src/core/github.js
himanish-star: :+1:

src/core/github.js
himanish-star: Fine I'll remove it

src/core/github.js
himanish-star: yes, this is better. I'll use the inbuilt css of `Balloon.css`

himanish-star:
> how will this work on a mobile phone?

Is there any `event` to record text being selected? I'll try searching on this. Do you have any idea?
______________
>  And, if the ReSpec pill menu at the top of the document is open, what should happen to this functionality?

When we select text in the `spec` then the `ReSpec pill menu` automatically closes. So I guess this won't cause any problem.

src/core/github.js
marcoscaceres: For the title, I would say, maybe 40 characters and then from there find the last whitespace character (so we don't chop words in half)... if the length was exceed, we should add "..." on the end. 

For the body, it can be much larger... but we need to test that out, to make sure we format it properly in the bug (e.g., it should appear as quoted text on github... we can get fancier from there, but this would be a great start).

himanish-star: Cool. I will incorporate these changes too. :+1:

src/core/github.js
himanish-star: Well `conf.issueBase` is `undefined`. Does user pass `issueBase` as a parameter in `respecConfig` or I'm thinking of using it from here https://github.com/w3c/respec/blob/34822ee20dcbbf186e81d9bfbccd66dc3d51913e/src/core/github.js#L98

himanish-star: Can I use it this way ?
```javascript
if (conf.enableIssueFeature && newProps.issueBase &&
    '/https://github.com/'.search(newProps.issueBase) ) {
    enableFileIssuesOnSelect(newProps.issueBase)
  }
```

himanish-star:
@marcoscaceres, I have made all the changes. Please see
GIF is as follows
ezgif com-video-to-gif

himanish-star:
Codacy is failing because of `balloon.css`:
image

marcoscaceres:
@himanish-star, could you instead have balloon.css  copied to the js/deps/ directory. We should also add it to the .gitignore file at the root of the project (i.e., we don't want it in our tree, as we don't control it... then Codacy won't complain ;)).

marcoscaceres:
@himanish-star, when you hover over the red button, the balloon is shown next to it, right? 

So, essentially, we want *that* ballon to appear underneath the selected text. You need to somehow get the (DOM) range of that selection and make sure the ballon gets positioned in the right place.

himanish-star:
On it marcos :+1:

himanish-star:
>  @himanish-star, when you hover over the red button, the balloon is shown next to it, right?
>
>So, essentially, we want that ballon to appear underneath the selected text. You need to somehow get the (DOM) range of that selection and make sure the ballon gets positioned in the right place.

Wonderful. Never thought of this though :sweat_smile:. I'll incorporate this into it.

marcoscaceres:
Sounds great. Not sure how challenging this will be (I’ve never done anything like that myself) - but I’m hoping it’s doable. If you get stuck and want me to ask around how we might do it, let me know.

marcoscaceres:
Found this: https://stackoverflow.com/questions/12603397/calculate-width-height-of-the-selected-text-javascript

himanish-star:
The problem is that it's difficult to position the `popup`. I have the position of selected text but I don't know how to position it.

src/core/github.js
```diff
@@ -5,10 +5,58 @@
...
+  fileIssueLink.innerText = 'File Issue';
+  fileIssueLink.onclick = function () {
+    window.location.href = fileIssueLink.getAttribute('data-href');
+    fileIssueLink.classList.add('respec-file-issue-hidden');
+  };
+  body.appendChild(fileIssueLink);
+
+  body.addEventListener("mouseup", function () {
+    const selectionObj = window.getSelection();
+    const { top, right, left } = selectionObj.getRangeAt(0).getBoundingClientRect();
```
himanish-star: So basically I now have `top, right and left`. So I can position the label as 
 - `y`: top
 - `x`: (right + left) / 2

But I can't find a way to change it to the css as it involves change in the `::before and ::after` pseudo css elements, fir which I have no idea of how to implement. @marcoscaceres, really need your help here :worried:

himanish-star:
ezgif com-video-to-gif
_______________________
@marcoscaceres, will this work?

marcoscaceres:
@saschanaz or @sidvishnoi, any chance either of you could have a look at this and maybe provide some suggestions for @himanish-star?

src/core/github.js
```diff
@@ -5,10 +5,58 @@
...
+  codeStyle.textContent = css;
+  codeStyle.classList.add('removeOnSave');
+  head.appendChild(codeStyle);
+
+  var fileIssueLink = document.createElement('button');
+  fileIssueLink.setAttribute('data-balloon', 'file an issue related to the text selected');
+  fileIssueLink.setAttribute('data-balloon-pos', 'up');
+  fileIssueLink.setAttribute('id', 'respec-file-issue-btn');
+  fileIssueLink.classList.add('removeOnSave', 'respec-file-issue-hidden');
+  fileIssueLink.innerText = 'File Issue';
```
saschanaz: nit: textContent instead of innerText

src/core/github.js
```diff
@@ -5,10 +5,58 @@
...
+  fileIssueLink.onclick = function () {
```
saschanaz: nit: `() => {`, and probably `addEventListener`.

src/core/github.js
```diff
@@ -5,10 +5,58 @@
...
+    window.location.href = fileIssueLink.getAttribute('data-href');
```
saschanaz: nit: `dataset.href` rather than `getAttribute`.

src/core/github.js
```diff
@@ -5,10 +5,58 @@
...
+    fileIssueLink.classList.add('respec-file-issue-hidden');
+  };
+  body.appendChild(fileIssueLink);
+
+  body.addEventListener("mouseup", function () {
```
saschanaz: nit: `() => {`

src/core/github.js
```diff
@@ -5,10 +5,58 @@
...
+    const selectionObj = window.getSelection();
+    const { top, right, left } = selectionObj.getRangeAt(0).getBoundingClientRect();
+    const textSelected = selectionObj.toString();
+    const newIssueURL = new URL('./new',issueBase);
+    const { searchParams } = newIssueURL;
+    var titleText, bodyText;
+
+    if (textSelected) {
+      if (textSelected.length > 40 ) {
+        var pos = textSelected.indexOf(" ", 40);
```
saschanaz: nit: `let`

src/core/github.js
```diff
@@ -97,6 +145,10 @@ export async function run(conf) {
     githubAPI: `https://api.github.com/repos/${org}/${repo}`,
     issueBase: `${ghURL.href}${ghURL.pathname.endsWith("/") ? "" : "/"}issues/`,
   };
+  if (conf.enableIssueFeature && newProps.issueBase &&
+    '/https://github.com/'.search(newProps.issueBase) ) {
```
saschanaz: nit: remove whitespace between `)`

src/core/github.js
```diff
@@ -97,6 +145,10 @@ export async function run(conf) {
...
+    enableFileIssuesOnSelect(newProps.issueBase)
```
saschanaz: nit: missing semicolon

src/core/github.js
```diff
@@ -5,10 +5,58 @@
...
+  body.addEventListener("mouseup", function () {
+    const selectionObj = window.getSelection();
+    const { top, right, left } = selectionObj.getRangeAt(0).getBoundingClientRect();
+    const textSelected = selectionObj.toString();
+    const newIssueURL = new URL('./new',issueBase);
+    const { searchParams } = newIssueURL;
+    var titleText, bodyText;
+
+    if (textSelected) {
+      if (textSelected.length > 40 ) {
```
saschanaz: nit: remove whitespace before `)`

src/core/github.js
```diff
@@ -5,10 +5,58 @@
...
+        var pos = textSelected.indexOf(" ", 40);
+        titleText = textSelected.slice(0, pos).concat("...");
+      } else {
+        titleText = textSelected;
+      }
+      bodyText = `> ${textSelected}`;
+      searchParams.set('title', `'${titleText}'`);
+      searchParams.set('body', bodyText);
```
saschanaz: We may inline `bodyText` instead of declaring a variable, as I don't see any other line that uses it.

src/core/github.js
```diff
@@ -5,10 +5,58 @@
...
+  };
+  body.appendChild(fileIssueLink);
+
+  body.addEventListener("mouseup", function () {
+    const selectionObj = window.getSelection();
+    const { top, right, left } = selectionObj.getRangeAt(0).getBoundingClientRect();
+    const textSelected = selectionObj.toString();
+    const newIssueURL = new URL('./new',issueBase);
+    const { searchParams } = newIssueURL;
+    var titleText, bodyText;
```
saschanaz: It seems `titleText` is used only within the following `if` block. We can use `let` to declare a variable in block scope.

src/core/github.js
saschanaz: I'm not sure why pseudo CSS elements should be manipulated, would you like to explain more for me?

saschanaz:
Personally I would prefer bottom-right position for this kind of button as I would select text from top-left to bottom-right, but I don't have any statistic data for general preference.

marcoscaceres:
Bottom right makes sense... it's closest to the mouse when you mouse up on the selection.

saschanaz:
https://stackoverflow.com/questions/9180405/detect-direction-of-user-selection-with-javascript?noredirect=1&lq=1

There also is a way to detect selection direction so that we can determine the button position.

himanish-star:
Thanks for the suggestions. I will see to them in the night :night_with_stars:

himanish-star:
I have made some changes. However, the code looks very unclean(I will clean the code once I have the functionalities working)

Here is a GIF which shows the error
ezgif com-video-to-gif

marcoscaceres:
having a look now...

marcoscaceres:
Ok, went down 4 hour rabbit hole 🐰... seems we are using the wrong event (we probably need `selectionchange` in combination with `mouseup`), and there are a bunch of other things that we should probably do different here. Need to still educate myself more on the Selection API. 

Best I think we can do now is maybe clean up some of the code. However, I'm not confident about being able to land this yet. Will require quite a bit of work still.

himanish-star:
@marcoscaceres, I actually took help from the code pasted [here](https://stackoverflow.com/questions/6846230/coordinates-of-selected-text-in-browser-page). Yep, even I don't think the UI looks great   :disappointed:

himanish-star:
Accidentally clicked the wrong button!

marcoscaceres:
Ok, spoke to @himanish-star. I think we are going to shelve this for a bit while we get a bit more up to speed to with the API. We have a fairly good idea from this code of what we need to do, but looks like it would take a bit more effort to get this to work than we originally thought.

marcoscaceres:
Closing for now till we can pick it up again.

himanish-star:
@marcoscaceres, yeah sure. Once I am done with my exams I'll be back on my work!!

PR #1474, opened 2018-02-05
chore(package): use uglify-es 3.3.7

saschanaz:
~~The last malfunctioning version was 3.3.9 so try 3.3.8.~~

The last known good version is 3.3.7, so use it rather than older 3.2.2.


marcoscaceres:
Published 19.1.0.

Issue #1427, opened 2017-11-15
Link to duplicate definition


marcoscaceres:
When there is a duplicate definition, we should just link to it from the warning. 

![screen shot 2017-11-15 at 11 23 21 am](https://user-images.githubusercontent.com/870154/32812149-ad1ac662-c9f7-11e7-9b12-273b7d915782.png)

himanish-star:
@marcoscaceres, Can I also take over this issue?

himanish-star:
I have attempted a solution to this. You can see the GIF below for the changes. Can I file a PR for this ?

![ezgif com-video-to-gif](https://user-images.githubusercontent.com/30361728/35780632-6acde08a-0a04-11e8-9faf-333f4eaa039e.gif)

himanish-star:
@marcoscaceres, please could I get to know what is the IRC of `W3C/respec`?

marcoscaceres:
@himanish-star, we use the w3c IRC. https://www.w3.org/wiki/IRC

You can find me as "marcosc" in the #specs channel. However, I might be slow to respond today as I'm catching up on a bunch of stuff from last week.

himanish-star:
Thanks a lot! :smiley:

himanish-star:
Actually I do have the things ready, but my laptop's battery was running dry so I have ordered a new battery and I'll surely file a PR after a few hours once I have my laptop back. Sorry for the inconvenience 😞

marcoscaceres:
heh, no problem @himanish-star! Take your time and hope your laptop gets well soon 🔋🚑

PR #1477, opened 2018-02-05
Link to duplicate dfns

himanish-star:
fixes: #1427 
I had already showed a GIF in https://github.com/w3c/respec/issues/1427#issuecomment-362926979 where I had added three `dfns` of `country` to cause the error
Please review marcosc

### DIFF
--- a/src/core/link-to-dfn.js
+++ b/src/core/link-to-dfn.js
@@ -3,12 +3,21 @@
 // to the matching definitions.
 import { linkInlineCitations } from "core/data-cite";
 import { pub } from "core/pubsubhub";
+import { lang as defaultLang } from "./l10n";
 export const name = "core/link-to-dfn";
+const l10n = {
+  en: {
+    "duplicate": "This is defined more than once in the document."
+  }
+};
+const lang = defaultLang in l10n ? defaultLang : "en";
+
 export function run(conf, doc, cb) {
   doc.normalize();
   var titles = {};
   Object.keys(conf.definitionMap).forEach(function(title) {
     titles[title] = {};
+    var listOfDuplicateDfns = [];
     conf.definitionMap[title].forEach(function(dfn) {
       if (dfn.attr("data-idl") === undefined) {
         // Non-IDL definitions aren't "for" an interface.
@@ -24,13 +33,14 @@ export function run(conf, doc, cb) {
         if (oldIsDfn && newIsDfn) {
           // Only complain if the user provides 2 <dfn>s
           // for the same term.
-          pub(
-            "error",
-            "Duplicate definition of '" +
-              (dfn_for ? dfn_for + "/" : "") +
-              title +
-              "'"
-          );
+          dfn.addClass("respec-offending-element");
+          if (dfn.attr("title") === undefined) {
+            dfn.attr("title", l10n[lang].duplicate);
+          }
+          if (dfn.attr("id") === undefined) {
+            dfn.makeID(null, title);
+          }
+          listOfDuplicateDfns.push(dfn[0]);
         }
         if (oldIsDfn) {
           // Don't overwrite <dfn> definitions.
@@ -46,6 +56,15 @@ export function run(conf, doc, cb) {
         }
       }
     });
+    if (listOfDuplicateDfns.length > 0) {
+      const dfnsList = listOfDuplicateDfns.map((elem, i) => {
+        return `[${i + 1}](#${elem.id})`;
+      }).join(", ");
+      pub(
+        "error",
+        `Duplicate definitions of '${title}' at: ${dfnsList}.`
+      );
+    }
   });
   $("a:not([href]):not([data-cite]):not(.logo)").each(function() {
     var $ant = $(this);
--- a/tests/spec/core/link-to-dfn-spec.js
+++ b/tests/spec/core/link-to-dfn-spec.js
@@ -44,4 +44,36 @@ describe("Core — Link to definitions", function() {
     expect(noCodeWrap.querySelector("code")).toBeFalsy();
     expect(noCodeWrap.textContent).toEqual("the request interface");
   });
+
+  it("checks for duplicate definitions", async () => {
+    const bodyText = `
+      <section>
+        <h2>Test Section</h2>
+        <dfn>Test1</dfn>
+        <dfn id="duplicate-definition">Test1</dfn>
+        <dfn>Test1</dfn>
+        <dfn title="test1">Test1</dfn>
+      </section>`;
+    const ops = {
+      config: makeBasicConfig(),
+      body: makeDefaultBody() + bodyText,
+    };
+    const doc = await makeRSDoc(ops);
+    const dfnList = doc.body.querySelectorAll("dfn");
+
+    const dfn1 = dfnList[1];
+    expect(dfn1).toBeTruthy();
+    expect(dfn1.classList).toContain("respec-offending-element");
+    expect(dfn1.id).toBe("duplicate-definition");
+
+    const dfn2 = dfnList[2];
+    expect(dfn2).toBeTruthy();
+    expect(dfn2.classList).toContain("respec-offending-element");
+    expect(dfn2.id).toBeDefined();
+
+    const dfn3 = dfnList[3];
+    expect(dfn3).toBeTruthy();
+    expect(dfn3.classList).toContain("respec-offending-element");
+    expect(dfn3.title).toBe("test1");
+  });
 });


src/core/link-to-dfn.js
```diff
@@ -24,12 +26,13 @@ export function run(conf, doc, cb) {
         if (oldIsDfn && newIsDfn) {
           // Only complain if the user provides 2 <dfn>s
           // for the same term.
+          dfn.attr('id',`${title + count}`);
```
marcoscaceres: You should check if there is already an ID before overriding.

src/core/link-to-dfn.js
```diff
@@ -24,12 +26,13 @@ export function run(conf, doc, cb) {
...
           pub(
             "error",
-            "Duplicate definition of '" +
+            `<a href="#${title + count}">Duplicate definition of '` +
               (dfn_for ? dfn_for + "/" : "") +
               title +
-              "'"
+              `'</a>`
```
marcoscaceres: Do you think there might be some way for us to group these? Like: "
```
  Duplicate definitions of term "X": 1, 2.
```  

Where 1 and 2 would be linked. 

Note also that the message, and the links, should be in markdown... so, you can just write the following to create the link:

```Markdown
[title](#id)
```

src/core/link-to-dfn.js
marcoscaceres: You might end up with strange IDs here... like `"foo bar1"`. You can use the `dfn.makeID(null, title)` here to solve that.

src/core/link-to-dfn.js
```diff
@@ -9,7 +9,9 @@ export function run(conf, doc, cb) {
   var titles = {};
   Object.keys(conf.definitionMap).forEach(function(title) {
     titles[title] = {};
+    let count = -1;
     conf.definitionMap[title].forEach(function(dfn) {
+      count++;
```
marcoscaceres: You might not need this with "makeID()" below.

src/core/link-to-dfn.js
himanish-star: I was thinking of the same thing while writing code. Sure I'll do this as well as all the changes you have requested for once I come back from college. Now got to rush :running_man:

marcoscaceres: Awesome! Have fun and speak again soon.

himanish-star: Back from college and I have started my work. Will update you soon

himanish-star: GIF of changes are as follows:
ezgif com-video-to-gif

src/core/link-to-dfn.js
```diff
@@ -7,6 +7,9 @@ export const name = "core/link-to-dfn";
 export function run(conf, doc, cb) {
   doc.normalize();
   var titles = {};
+  var prevTitle;
+  var titleChanged = false;
+  var listOfDuplicateDfns = [];
```
himanish-star: `prevTitle` is used to store the previous title so that when the title changes I can attach all the duplicate definitions related to that title all together
`titleChanged` keeps track of whether title changed or not
`listOfDuplicateDfns` used to store the list of duplicate definitions for a particular title

src/core/link-to-dfn.js
```diff
@@ -24,13 +27,27 @@ export function run(conf, doc, cb) {
...
           pub(
             "error",
-            "Duplicate definition of '" +
-              (dfn_for ? dfn_for + "/" : "") +
-              title +
-              "'"
+            "Duplicate definitions of '" +
+            (dfn_for ? dfn_for + "/" : "") +
+            prevTitle + "' at: " +
+            dfnsList
```
himanish-star: Will indent this part in my later commit

src/core/link-to-dfn.js
```diff
@@ -24,13 +27,27 @@ export function run(conf, doc, cb) {
...
+            dfn.makeID(null, title);
+          }
+          listOfDuplicateDfns.push(dfn);
+          prevTitle = title;
+          titleChanged = false
+        } else if (listOfDuplicateDfns && !titleChanged) {
+          var dfnsList = "";
+          for (var i = 0; i < listOfDuplicateDfns.length; i++) {
+            dfnsList += "["+ parseInt(i+1) + "](#" +
+              listOfDuplicateDfns[i].attr("id") + ")" + " ";
```
himanish-star: `dfnsList` being created in markdown format

src/core/link-to-dfn.js
```diff
@@ -24,13 +27,27 @@ export function run(conf, doc, cb) {
         if (oldIsDfn && newIsDfn) {
           // Only complain if the user provides 2 <dfn>s
           // for the same term.
+          if (dfn.attr("id") === undefined) {
+            dfn.makeID(null, title);
+          }
+          listOfDuplicateDfns.push(dfn);
+          prevTitle = title;
+          titleChanged = false
+        } else if (listOfDuplicateDfns && !titleChanged) {
```
himanish-star: I think this `if` condition can improve

himanish-star: This code is wrong. Don't review this yet!!!!

himanish-star:
I have made all changes. Please have a look

image

src/core/link-to-dfn.js
```diff
@@ -24,13 +25,10 @@ export function run(conf, doc, cb) {
...
-            "error",
-            "Duplicate definition of '" +
-              (dfn_for ? dfn_for + "/" : "") +
-              title +
-              "'"
-          );
+          if (dfn.attr("id") === undefined) {
+            dfn.makeID(null, title);
+          }
+          listOfDuplicateDfns.push(dfn);
```
marcoscaceres: Here, unJQuery it... as a general note, we are trying to remove jQuery. 

```JS
listOfDuplicateDfns.push(dfn[0]);
```

src/core/link-to-dfn.js
```diff
@@ -46,6 +44,20 @@ export function run(conf, doc, cb) {
         }
       }
     });
+    if (listOfDuplicateDfns.length > 0) {
```
marcoscaceres: Ok, we want to refactor this to use functional style... so, something like:

```JS
const dfnsList = listOfDuplicateDfns.map((elem, i) => {
  return `[${i + 1}](#${elem.id})`;
}).join(", ");
```

src/core/link-to-dfn.js
```diff
@@ -46,6 +44,20 @@ export function run(conf, doc, cb) {
...
+      var dfnsList = "";
+      for (var i = 0; i < listOfDuplicateDfns.length; i++) {
+        dfnsList += "["+ parseInt(i+1) + "](#" +
+          listOfDuplicateDfns[i].attr("id") + ")" + " ";
+      }
+      pub(
+        "error",
+        "Duplicate definitions of '" +
```
marcoscaceres: Nit, use template string instead: 
```
`Duplicate definitions of '${title}' at: ${dfnsList}.`

File: src/core/link-to-dfn.js [hunk 11]
```diff
@@ -7,6 +7,7 @@ export const name = "core/link-to-dfn";
 export function run(conf, doc, cb) {
   doc.normalize();
   var titles = {};
+  var listOfDuplicateDfns = [];
```
marcoscaceres: move this down to line 25, then just let the garbage collector remove it after each run of "forEach"

src/core/link-to-dfn.js
```diff
@@ -46,6 +44,20 @@ export function run(conf, doc, cb) {
...
+        dfnsList += "["+ parseInt(i+1) + "](#" +
+          listOfDuplicateDfns[i].attr("id") + ")" + " ";
+      }
+      pub(
+        "error",
+        "Duplicate definitions of '" +
+          title + "' at: " +
+          dfnsList
+      );
+      listOfDuplicateDfns = [];
```
marcoscaceres: No need for this, let the garbage collector deal with it.

src/core/link-to-dfn.js
marcoscaceres: additional nit, make this a "const"

src/core/link-to-dfn.js
marcoscaceres: Also just realized that we want to add the CSS class "respec-offending-element" to any duplicates, so that duplicate definitions are easier to spot. We might also want to set the `title` attribute of the element to "This is defined more than once in the document".

src/core/link-to-dfn.js
himanish-star: On it

src/core/link-to-dfn.js
himanish-star: On it

src/core/link-to-dfn.js
himanish-star: on it

src/core/link-to-dfn.js
himanish-star: Yep, I understand this but as `dfn` was in `jquery` I proceeded further on this.

marcoscaceres: Thats very true. But eventually we will remove jQuery (🤞), so we want any new code to not rely on it. Make sense?

himanish-star: Yes makes sense :smiley:

src/core/link-to-dfn.js
himanish-star: Line number 25 is this `if (oldIsDfn && newIsDfn) {`.


 Is this the appropriate place as I think in every run of `for each` for `dfn` the array would be refreshed. Rather this should happen in every run of `for each` for `title`

himanish-star: Moreover making it a `const` doesn't work. As it doesn't allow any modifications to the array for pushing other `duplicate definitions` of other `tittles`

src/core/link-to-dfn.js
himanish-star: Same for here also. As this will fail when I have two or more titles with duplicate deinitions for which the dfnsList is reconstructed every time for each and every `title`.

src/core/link-to-dfn.js
himanish-star: What I have done is that I am creating a `duplicate definitions` template string for each and every title

src/core/link-to-dfn.js
himanish-star: > Also just realized that we want to add the CSS class "respec-offending-element" to any duplicates, so that duplicate definitions are easier to spot. We might also want to set the title attribute of the element to "This is defined more than once in the document".

On it

src/core/link-to-dfn.js
marcoscaceres: That's odd. ok, leave it as `var` for now and I'll have a look.

src/core/link-to-dfn.js
himanish-star: Yes in a minute

himanish-star: I have made all the changes except some changes related to setting the variable as `const`. I may be wrong also, in this case I'm sincerely sorry

`I have created big spaces to show where I have made the code change`
```javascript
  Object.keys(conf.definitionMap).forEach(function(title) {
    titles[title] = {};


    var listOfDuplicateDfns = []; /*
     Now as you can see this array has to get updated every time the title changes. As this stores 
     the list of duplicate dfns. So how can I set this as const as this will be modified a number of 
     times for the same title if there are two or more duplicate dfns. **This is the only place where I haven't followed your review. Other wise everywhere below you can find the changes as per your review** */


    conf.definitionMap[title].forEach(function(dfn) {
      if (dfn.attr("data-idl") === undefined) {
        // Non-IDL definitions aren't "for" an interface.
        dfn.removeAttr("data-dfn-for");
      }
      var dfn_for = dfn.attr("data-dfn-for") || "";
      if (dfn_for in titles[title]) {
        // We want <dfn> definitions to take precedence over
        // definitions from WebIDL. WebIDL definitions wind
        // up as <span>s instead of <dfn>.
        var oldIsDfn = titles[title][dfn_for].filter("dfn").length !== 0;
        var newIsDfn = dfn.filter("dfn").length !== 0;
        if (oldIsDfn && newIsDfn) {
          // Only complain if the user provides 2 <dfn>s
          // for the same term.


          dfn.addClass("respec-offending-element");
          dfn.attr("title", "This is defined more than once in the document");
          /* I have made the changes here as you had requested*/


         if (dfn.attr("id") === undefined) {
            dfn.makeID(null, title);
          }


          listOfDuplicateDfns.push(dfn[0]);
          /*Made this change too*/
 
       }
        if (oldIsDfn) {
          // Don't overwrite <dfn> definitions.
          return;
        }
      }
      titles[title][dfn_for] = dfn;
      if (dfn.attr("id") === undefined) {
        if (dfn.attr("data-idl")) {
          dfn.makeID("dom", (dfn_for ? dfn_for + "-" : "") + title);
        } else {
          dfn.makeID("dfn", title);
        }
      }
    });
    if (listOfDuplicateDfns.length > 0) {
      
     
      const dfnsList = listOfDuplicateDfns.map((elem, i) => {
        return `[${i + 1}](#${elem.id})`;
      }).join(", ");
     /* incorporated the changes here */

      pub(
        "error",


        `Duplicate definitions of '${title}' at: ${dfnsList}.`
         /* made the changes here too as per your request.*/

      );
    }
  });
```

marcoscaceres: Ah, thanks! got it 👍 You are right.

marcoscaceres: Looks like this module a bigger refactor... but that's for another day :)

himanish-star: So shall I update this PR? Here is a GIF:

ezgif com-video-to-gif
___
> Looks like this module a bigger refactor... but that's for another day :)

I would love to do this!

marcoscaceres: Awwweesome! yeah, send up the new code!

himanish-star: Sending in a minute

himanish-star:
Done :+1:

himanish-star:
Any changes to be made or any updates on this?

marcoscaceres:
Apologies... I thought I posted a note about changes... all that's missing is at test or two. We should be testing if at least that "respect-offending-element" is set. Take a look at:

*  tests/spec/core/link-to-dfn-spec.js

As we currently don't have a good way of testing if the warnings are show, so the above will have to do. I guess you should also test that valid definitions are not affected. 

I'll be back in two hours if you want to discuss a bit more.

himanish-star:
Sure, I'll write the tests

himanish-star:
@marcoscaceres, This is the first time I have written a test. Please check if it is correct as from my side it seems to be correct

src/core/link-to-dfn.js
```diff
@@ -24,13 +25,12 @@ export function run(conf, doc, cb) {
...
           // for the same term.
-          pub(
-            "error",
-            "Duplicate definition of '" +
-              (dfn_for ? dfn_for + "/" : "") +
-              title +
-              "'"
-          );
+          dfn.addClass("respec-offending-element");
+          dfn.attr("title", "This is defined more than once in the document");
```
marcoscaceres: Just realized we should localize this string. At the top of the file:

```JS
import { lang as defaultLang } from "../l10n";
const l10n = {
  en: {
     "duplicate": "This is defined more than once in the document."
  }
};

// Fall back to English, if language is missing
```

And then here:

```JS
dfn.attr("title", l10n[defaultLang].duplicate);
```

Hope that makes sense!

tests/spec/core/link-to-dfn-spec.js
```diff
@@ -44,4 +44,21 @@ describe("Core — Link to definitions", function() {
     expect(noCodeWrap.querySelector("code")).toBeFalsy();
     expect(noCodeWrap.textContent).toEqual("the request interface");
   });
+
+  it("checks for duplicate definitions", async () => {
```
marcoscaceres: Ok, this initial test looks great.

Couple of other things we also want to check (looking at the code). 

 1. Does it automatically generate an id? 
 1. If there is an `id` attribute already present, make sure it doesn't trash it. 
 1. If there is an `title` attribute already present, make sure it doesn't trash it. The current code actually has bug, because it does trash it.

marcoscaceres: If you can think of other things, that would be great.

himanish-star: On it!

src/core/link-to-dfn.js
himanish-star: Yes it makes sense

himanish-star:
@marcoscaceres , I have made the changes and I have also written the required tests :smiley:

himanish-star:
@marcoscaceres, I'll be firing up two more PRs whose work has been completed about 80%. I'm writing the tests and will complete them in a few hours, so that it becomes easier for you to review

marcoscaceres:
@himanish-star, ok, but no stress. Remember your primary focus now should be the proposal for #1502!

marcoscaceres:
Now live, published as respec@19.4.0 🤘👨‍🎤🤘

PR #1480, opened 2018-02-06
feat(w3c): remove processVersion option

marcoscaceres:
* [ ] Update Wiki to reflect this change once this is merged.

### DIFF
--- a/src/w3c/headers.js
+++ b/src/w3c/headers.js
@@ -660,6 +660,7 @@ export function run(conf, doc, cb) {
       "[`processVersion`](https://github.com/w3c/respec/wiki/processVersion) configuration option " +
       `to one of: ${W3CProcessVersion.known.join(", ")}.`;
     pub("error", msg);
+    conf.processVersion = W3CProcessVersion.current;
   } else if (conf.processVersion < W3CProcessVersion.current) {
     const msg =
       `The W3C's ${
--- a/src/w3c/headers.js
+++ b/src/w3c/headers.js
@@ -652,11 +652,11 @@ export function run(conf, doc, cb) {
       return urlParts[pos] || "";
     },
   });
-  if (W3CProcessVersion.known.includes(conf.processVersion) === false) {
+  if (!W3CProcessVersion.known.includes(conf.processVersion)) {
     const msg =
       `Unknown W3C Process Document version: "${
         conf.processVersion
-      }". Please update your ` +
+      }". ReSpec has set your document to use the ${W3CProcessVersion.current} process. Please update your ` +
       "[`processVersion`](https://github.com/w3c/respec/wiki/processVersion) configuration option " +
       `to one of: ${W3CProcessVersion.known.join(", ")}.`;
     pub("error", msg);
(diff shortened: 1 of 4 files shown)


saschanaz:
Not sure why it should be string, will it be semver later?

marcoscaceres:
There are some equality checks (though they rely on coercion via "=="):
https://github.com/w3c/respec/blob/develop/src/w3c/headers.js#L654

I want to make these strict checks, and always just use a single data type.

marcoscaceres:
Maybe I'm being overly picky 🙄

saschanaz:
Hmm. It seems processVersion is always `20xx`, can we make it a number and fix tests that thinks it as a string?

marcoscaceres:
Yeah, that's a good suggestion. Will switch this around.

src/w3c/headers.js
```diff
@@ -651,14 +653,27 @@ export function run(conf, doc, cb) {
       return urlParts[pos] || "";
     },
   });
-  if (conf.processVersion == "2014" || conf.processVersion == "2015" || conf.processVersion == "2017") {
-    pub(
-      "warn",
-      "Process " + conf.processVersion + " has been superceded by Process 2018."
-    );
-    conf.processVersion = "2018";
+  if (W3CProcessVersion.known.includes(conf.processVersion) === false) {
```
saschanaz: nit: `!x` rather than `x === false`, as `includes` always returns Boolean.

src/w3c/headers.js
```diff
@@ -651,14 +653,27 @@ export function run(conf, doc, cb) {
...
+    const msg =
+      `Unknown W3C Process Document version: "${
+        conf.processVersion
+      }". Please update your ` +
```
saschanaz: Can this be in a single line: `"${conf.processVersion}"`?

src/w3c/headers.js
```diff
@@ -651,14 +653,27 @@ export function run(conf, doc, cb) {
...
+      `to one of: ${W3CProcessVersion.known.join(", ")}.`;
+    pub("error", msg);
+    conf.processVersion = W3CProcessVersion.current;
+  } else if (conf.processVersion < W3CProcessVersion.current) {
+    const msg =
+      `The W3C's ${
+        conf.processVersion
+      } Process Document has been superceded by the ${
+        W3CProcessVersion.current
+      } Process Document. Please update, or remove, your ` +
```
saschanaz: (Same here)

marcoscaceres:
@plehegar or @dontcallmedom, is there any reason for any document to still be using the 2005 (!) process document version? If, no, then I can just force all documents to the 2018 process and get rid of all this code.

src/w3c/headers.js
marcoscaceres: Prettier.js keeps formatting it like that... "computer says no" 😂

saschanaz: Line 662 is okay but line 660 is not okay? That's weird 🤣

marcoscaceres: I'm still hopeful we can delete all things (best kind of code). I have good feeling @plehegar will say "yes".

marcoscaceres:
In anticipation of an ok from @plehegar, I've preemptively updated this so it removes the configurable `processVersion` member entirely. 

If @plehegar decides for whatever reason we want to keep it,  we can just revert my last commit.

plehegar:
@marcoscaceres no reason to keep process 2005 around imho, so +1 from me.

marcoscaceres:
@saschanaz, given ok we got from @plehegar, could you review again?

PR #1483, opened 2018-02-07
refactor: use puppeteer instead of nightmare

saschanaz:
[`puppeteer`](https://developers.google.com/web/tools/puppeteer/) is a first-party wrapper library of Chromium made by Google. It directly uses Chromium instead of Electron (which again uses Chromium), and has less dependencies (80 total dependencies for `nightmare` whereas 16 for `puppeteer`.)

This still requires a real browser (Chromium) but in a lighter way.

Ultimately, but not in this PR, we can use puppeteer-installed Chromium to be used in karma tests so that GPU-less CLI environments can use it easily.

### DIFF
--- a/tools/respecDocWriter.js
+++ b/tools/respecDocWriter.js
@@ -7,7 +7,7 @@
 /*jshint node: true, browser: false*/
 "use strict";
 const os = require("os");
-const puppeteer = require('puppeteer');
+const puppeteer = require("puppeteer");
 const colors = require("colors");
 const { promisify } = require("util");
 const fs = require("fs");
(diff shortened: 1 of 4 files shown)


tools/respecDocWriter.js
```diff
@@ -7,7 +7,7 @@
 /*jshint node: true, browser: false*/
 "use strict";
 const os = require("os");
-const Nightmare = require("nightmare");
+const puppeteer = require('puppeteer');
```
marcoscaceres: nit: double quote... I've been using prettier.js for the project.

PR #1487, opened 2018-02-08
prettify linked tests : Fixes #1374

### LINKED ISSUE #1374: 💅-up linked tests
Given that tests names themselves contain details about what kinds of tests they are, we can use that information to make the tests a bit nicer. 

 * [ ] requires secure connection 🔒
 * [ ] manual tests 💪
 * [ ] automated test 🤖

so, it would be like:  manual test (💪),  requires secure connection (🔒) test-name. 

And we can drop the .html part.

shikhar-scs:
As suggested by @marcoscaceres , the required changes mentioned at #1374 have been made. Here's a preview

heheheheehehe

Please preview

### DIFF
--- a/examples/basic.html
+++ b/examples/basic.html
@@ -3,43 +3,17 @@
   <head>
     <meta charset='utf-8'>
     <title>Replace me with a real title</title>
-    <script src='../js/deps/require.js' data-main='../js/profile-w3c-common' 
+    <script src='../js/deps/require.js' data-main='../js/profile-w3c-common'
       async class='remove'></script>
-    <script class='remove'>
-      var respecConfig = {
-        specStatus: "ED",
-        editors: [{
-          name: "Your Name",
-          url: "https://your-site.com",
-        }],
-        processVersion: 2018,
-        edDraftURI: "https://some.github.repo",
-        shortName: "dahut"
+    <script>
+      const respecConfig = {
+        testSuiteURI: "https://wpt.fyi/payment-request/",
       };
     </script>
   </head>
   <body>
-    <section id='abstract'>
-      <p>
-        This is required.
-      </p>
-    </section>
-    <section id='sotd'>
-      <p>
-        This is required.
-      </p>
-    </section>
-    <section data-dfn-for="Foo">
-      <h2><dfn>Foo</dfn> interface</h2>
-      <pre class="idl">
-      interface Foo {
-        attribute Bar bar;
-        void doTheFoo();
-      };
-      </pre>
-     <p>The <a>Foo</a> interface is nice. Lets you do stuff.</p>
-     <p>The <dfn>bar</dfn> attribute, returns 🍺.</p>
-     <p>The <dfn>doTheFoo()</dfn> method, returns nothing.</p>
-    </section>
+  <p data-tests="test-1.html, test-2.html, hello-manual.https.html, hey.https.html">
+    The user agent MUST do this stuff...
+  </p>
   </body>
 </html>
--- a/examples/basic.html
+++ b/examples/basic.html
@@ -3,17 +3,43 @@
   <head>
     <meta charset='utf-8'>
     <title>Replace me with a real title</title>
-    <script src='../js/deps/require.js' data-main='../js/profile-w3c-common'
+    <script src='../js/deps/require.js' data-main='../js/profile-w3c-common' 
       async class='remove'></script>
-    <script>
-      const respecConfig = {
-        testSuiteURI: "https://wpt.fyi/payment-request/",
+    <script class='remove'>
+      var respecConfig = {
+        specStatus: "ED",
+        editors: [{
+          name: "Your Name",
+          url: "https://your-site.com",
+        }],
+        processVersion: 2018,
+        edDraftURI: "https://some.github.repo",
+        shortName: "dahut"
       };
     </script>
   </head>
   <body>
-  <p data-tests="test-1.html, test-2.html, hello-manual.https.html, hey.https.html">
-    The user agent MUST do this stuff...
-  </p>
+    <section id='abstract'>
+      <p>
+        This is required.
+      </p>
+    </section>
+    <section id='sotd'>
+      <p>
+        This is required.
+      </p>
+    </section>
+    <section data-dfn-for="Foo">
+      <h2><dfn>Foo</dfn> interface</h2>
+      <pre class="idl">
+      interface Foo {
+        attribute Bar bar;
+        void doTheFoo();
+      };
+      </pre>
+     <p>The <a>Foo</a> interface is nice. Lets you do stuff.</p>
+     <p>The <dfn>bar</dfn> attribute, returns 🍺.</p>
+     <p>The <dfn>doTheFoo()</dfn> method, returns nothing.</p>
+    </section>
   </body>
 </html>
--- a/src/core/data-tests.js
+++ b/src/core/data-tests.js
@@ -35,8 +35,9 @@ function toListItem(href) {
   let emojiList = document.createElement("span");
 
   if (isSecureTest) {
-    const requiresConnectionEmoji = hyperHTML.bind(document.createElement("span"))` 🔒 `;
-    requiresConnectionEmoji.setAttribute("aria-label","requires a secure connection");
+    const requiresConnectionEmoji = document.createElement("span")
+      requiresConnectionEmoji.innerHTML = `🔒`;
+      requiresConnectionEmoji.setAttribute("aria-label","requires a secure connection");
     
     testFileName.replace(".https","");
     emojiList.append(requiresConnectionEmoji)
@@ -44,9 +45,10 @@ function toListItem(href) {
   
   if(isManualTest) {
   
-    const manualPerformEmoji = hyperHTML.bind(document.createElement("span"))` 💪 `;
-    manualPerformEmoji.setAttribute("aria-label","the test must be run manually");
-  
+    const manualPerformEmoji = document.createElement("span")
+      manualPerformEmoji.innerHTML = `💪`;
+      manualPerformEmoji.setAttribute("aria-label","the test must be run manually");
+  console.log(testFileName)
     testFileName.replace("-manual","");
     emojiList.append(manualPerformEmoji)
   }
@@ -54,9 +56,8 @@ function toListItem(href) {
   const testList = hyperHTML.bind(document.createElement("li"))`
     <a href="${href}">
       ${testFileName}
-    </a>
+    </a> ${emojiList}
   `;
-  testList.append(emojiList)
   return testList
 }
 
--- a/src/core/data-tests.js
+++ b/src/core/data-tests.js
@@ -23,36 +23,41 @@ export const name = "core/data-tests";
 const lang = defaultLang in l10n ? defaultLang : "en";
 
 function toListItem(href) {
-  
+  const emojiList = document.createElement("span");
   const [testFile] = new URL(href).pathname.split("/").reverse();
-  
   const testParts = testFile.split(".");
-  const [testFileName] = testParts;
-  
-  const isSecureTest = testParts.find(part => part === "https");
-  const isManualTest = testFileName.split(".").join("-").split("-").find(part => part === "manual");
-  
-  let emojiList = document.createElement("span");
+  let [testFileName] = testParts;
 
+  const isSecureTest = testParts.find(part => part === "https");
   if (isSecureTest) {
     const requiresConnectionEmoji = document.createElement("span")
-      requiresConnectionEmoji.innerHTML = `🔒`;
-      requiresConnectionEmoji.setAttribute("aria-label","requires a secure connection");
-    
-    testFileName.replace(".https","");
-    emojiList.append(requiresConnectionEmoji)
+    requiresConnectionEmoji.innerHTML = `🔒`;
+    requiresConnectionEmoji.setAttribute(
+      "aria-label",
+      "requires a secure connection"
+    );
+    requiresConnectionEmoji.setAttribute("title", "Test requires HTTPS");
+    testFileName = testFileName.replace(".https", "");
+    emojiList.append(requiresConnectionEmoji);
   }
-  
-  if(isManualTest) {
-  
+
+  const isManualTest = testFileName
+    .split(".")
+    .join("-")
+    .split("-")
+    .find(part => part === "manual");
+  if (isManualTest) {
     const manualPerformEmoji = document.createElement("span")
       manualPerformEmoji.innerHTML = `💪`;
-      manualPerformEmoji.setAttribute("aria-label","the test must be run manually");
-  console.log(testFileName)
-    testFileName.replace("-manual","");
-    emojiList.append(manualPerformEmoji)
+      manualPerformEmoji.setAttribute(
+      "aria-label",
+      "the test must be run manually"
+    );
+    manualPerformEmoji.setAttribute("title", "Manual test");
+    testFileName = testFileName.replace("-manual", "");
+    emojiList.append(manualPerformEmoji);
   }
-  
+
   const testList = hyperHTML.bind(document.createElement("li"))`
     <a href="${href}">
       ${testFileName}
--- a/src/core/data-tests.js
+++ b/src/core/data-tests.js
@@ -23,7 +23,7 @@ export const name = "core/data-tests";
 const lang = defaultLang in l10n ? defaultLang : "en";
 
 function toListItem(href) {
-  const emojiList = document.createElement("span");
+  const emojiList = [];
   const [testFile] = new URL(href).pathname.split("/").reverse();
   const testParts = testFile.split(".");
   let [testFileName] = testParts;
--- a/src/core/data-tests.js
+++ b/src/core/data-tests.js
@@ -12,7 +12,8 @@ import { pub } from "core/pubsubhub";
 import { lang as defaultLang } from "core/l10n";
 const l10n = {
   en: {
-    missing_test_suite_uri: "Found tests in your spec, but missing '" +
+    missing_test_suite_uri:
+      "Found tests in your spec, but missing '" +
       "[`testSuiteURI`](https://github.com/w3c/respec/wiki/testSuiteURI)' in your ReSpec config.",
     tests: "tests",
     test: "test",
@@ -31,7 +32,7 @@ function toListItem(href) {
 
   const isSecureTest = testParts.find(part => part === "https");
   if (isSecureTest) {
-    const requiresConnectionEmoji = document.createElement("span")
+    const requiresConnectionEmoji = document.createElement("span");
     requiresConnectionEmoji.innerHTML = `🔒`;
     requiresConnectionEmoji.setAttribute(
       "aria-label",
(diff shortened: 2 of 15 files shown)


shikhar-scs:
@marcoscaceres I have pushed in some new changes. Please review.

src/core/data-tests.js
```diff
@@ -23,11 +23,34 @@ export const name = "core/data-tests";
 const lang = defaultLang in l10n ? defaultLang : "en";
 
 function toListItem(href) {
-  return hyperHTML.bind(document.createElement("li"))`
+  
+  let modifiedHref = href.split("/").pop().split(".html")[0];
```
marcoscaceres: nit:
```js
const [modifiedHref] = href.split("/").pop().split(".html");
```

Also, might be worth checking if the tests always end with ".html"... Some might end with just ".htm", but double check in the web platform test repository.

src/core/data-tests.js
```diff
@@ -23,11 +23,34 @@ export const name = "core/data-tests";
...
+  
+  let emojiList = document.createElement('i');
+  
+  let requiresConnectionEmoji = hyperHTML.bind(document.createElement("i"))` 🔒 `;
```
marcoscaceres: Any reason you chose `i` element in particular? I would figure maybe a `span` might be better here?

src/core/data-tests.js
```diff
@@ -23,11 +23,34 @@ export const name = "core/data-tests";
...
+    requiresConnectionEmoji.setAttribute('aria-label','require-secure-connection');
```
marcoscaceres: Note that aria-label is read out by a screen reader, so we want to make this as friendly as possible. I would suggest instead: "requires a secure connection".

We have quite a few users with visual impairments, so we do our best to support their needs.

src/core/data-tests.js
```diff
@@ -23,11 +23,34 @@ export const name = "core/data-tests";
...
+  
+  let manualPerformEmoji = hyperHTML.bind(document.createElement("i"))` 💪 `;
```
marcoscaceres: As above, and aria label "the test must be run manually"

src/core/data-tests.js
marcoscaceres: These elements should only be created inside the if statements below. Otherwise, they are getting created unnecessarily if not needed (and then thrown away).

src/core/data-tests.js
```diff
@@ -23,11 +23,34 @@ export const name = "core/data-tests";
 const lang = defaultLang in l10n ? defaultLang : "en";
 
 function toListItem(href) {
-  return hyperHTML.bind(document.createElement("li"))`
+  
+  let modifiedHref = href.split("/").pop().split(".html")[0];
+  
+  let emojiList = document.createElement('i');
```
marcoscaceres: Nit: all the `let` should be `const`, as we are not reassigning them.

marcoscaceres:
Just a general note that in the project we are using double quotes 😉 

To pretty up the code, I've been using (prettier.js)[https://github.com/prettier/prettier] via Visual Studio Code... but also works in the command line.

marcoscaceres:
@shikhar-scs, if you are unsure about any JS syntax stuff (e.g., when should I use `const` vs `let` vs `var`), please let me know. Happy to give you pointers to explain the difference.

shikhar-scs:
Thank you for the review . I'll post the corrected changes soon👍

src/core/data-tests.js
shikhar-scs: I really appreciate that :clap:

marcoscaceres:
Thanks for continuing to hack in this.  I’m just on my way home, so will take a look first thing tomorrow.

shikhar-scs:
> To pretty up the code, I've been using (prettier.js)[https://github.com/prettier/prettier] via Visual Studio 

Thankyou for your suggestions and though I do l know the differences between `var` `let` and `const`, I'll ask you in case I have any doubts.

I've pushed some new changes. And as of now, please ignore the changes made to `basic.html`

examples/basic.html
```diff
@@ -3,8 +3,8 @@
   <head>
     <meta charset='utf-8'>
     <title>Replace me with a real title</title>
-    <script src='../js/deps/require.js' data-main='../js/profile-w3c-common' 
-      async class='remove'></script>
+    <script src='../js/deps/require.js' data-main='../js/profile-w3c-common'
```
marcoscaceres: Please revert changes to this file.

src/core/data-tests.js
```diff
@@ -23,11 +23,39 @@ export const name = "core/data-tests";
 const lang = defaultLang in l10n ? defaultLang : "en";
 
 function toListItem(href) {
-  return hyperHTML.bind(document.createElement("li"))`
+  
+  // const modifiedHref = href.split("/").pop().split(".")[0];
+  
+  if ( href.split("/").pop().split(".").includes("html") ||  href.split("/").pop().split(".").includes("htm") ) {
```
marcoscaceres: Nit: make this two steps, and combine the search (otherwise, you are doing things twice):
```
// the last item will be the filename
const [testFile] = new URL(href).pathname.split("/").reverse();

// next we want to work out the type
const testParts = testFile.split(".");
const [testFile] = testParts; // the first item 
const isSecureTest = testParts.find(part => part === "https");
// this is probably incorrect, but find "manual" like this:  
const isManualTest = testFile.split("-").find(part => part === "manual");

// then 
if(isSecure){
  // add secure emoji...
}

if(isManualTest) {
  // add manual emoji
} 

```

src/core/data-tests.js
```diff
@@ -24,34 +24,37 @@ const lang = defaultLang in l10n ? defaultLang : "en";
 
 function toListItem(href) {
   
-  // const modifiedHref = href.split("/").pop().split(".")[0];
+  var [testFile] = new URL(href).pathname.split("/").reverse();
```
marcoscaceres: nit: rename this to something else, don't reassign it below (otherwise it's confusing to know what this actually is)

src/core/data-tests.js
```diff
@@ -23,11 +23,42 @@ export const name = "core/data-tests";
 const lang = defaultLang in l10n ? defaultLang : "en";
 
 function toListItem(href) {
-  return hyperHTML.bind(document.createElement("li"))`
+  
+  var [testFile] = new URL(href).pathname.split("/").reverse();
```
marcoscaceres: ok, call this something else and don't reassign it... otherwise, its confusing to know what this actually is.

src/core/data-tests.js
```diff
@@ -23,11 +23,42 @@ export const name = "core/data-tests";
...
+  const isSecureTest = testParts.find(part => part === "https");
+  const isManualTest = testFile.split(".").join("-").split("-").find(part => part === "manual");
+  
+  let emojiList = document.createElement("span");
+
+  if (isSecureTest) {
+    const requiresConnectionEmoji = hyperHTML.bind(document.createElement("span"))` 🔒 `;
+    requiresConnectionEmoji.setAttribute("aria-label","requires a secure connection");
+    
+    console.log(testFile)
```
marcoscaceres: nit: remove console.log

src/core/data-tests.js
```diff
@@ -23,11 +23,42 @@ export const name = "core/data-tests";
...
+    testFile.split(".https").join("");
```
marcoscaceres: this line doesn't do anything :)

shikhar-scs: It does. It removes the .https parts from the end

shikhar-scs:
Please review now @marcoscaceres . I've also squashed all my commits

src/core/data-tests.js
```diff
@@ -23,11 +23,41 @@ export const name = "core/data-tests";
...
+  
+  const [testFile] = new URL(href).pathname.split("/").reverse();
+  
+  const testParts = testFile.split(".");
+  const [testFileName] = testParts;
+  
+  const isSecureTest = testParts.find(part => part === "https");
+  const isManualTest = testFileName.split(".").join("-").split("-").find(part => part === "manual");
+  
+  let emojiList = document.createElement("span");
```
marcoscaceres: nit: this should be a `const`, as it's not being reassigned, right?

src/core/data-tests.js
```diff
@@ -23,11 +23,50 @@ export const name = "core/data-tests";
...
 function toListItem(href) {
-  return hyperHTML.bind(document.createElement("li"))`
+  const emojiList = document.createElement("span");
+  const [testFile] = new URL(href).pathname.split("/").reverse();
+  const testParts = testFile.split(".");
+  let [testFileName] = testParts;
+
+  const isSecureTest = testParts.find(part => part === "https");
+  if (isSecureTest) {
+    const requiresConnectionEmoji = hyperHTML.bind(
```
marcoscaceres: It seems unnecessary to use hyperHTML here, and same in the `isManualTest` block below. Just create spans here, and don't bind them.

src/core/data-tests.js
```diff
@@ -23,11 +23,50 @@ export const name = "core/data-tests";
...
+    emojiList.append(requiresConnectionEmoji);
+  }
+
+  const isManualTest = testFileName
+    .split(".")
+    .join("-")
+    .split("-")
+    .find(part => part === "manual");
+  if (isManualTest) {
+    const manualPerformEmoji = hyperHTML.bind(
```
marcoscaceres: As above.

src/core/data-tests.js
```diff
@@ -23,11 +23,50 @@ export const name = "core/data-tests";
...
+    emojiList.append(manualPerformEmoji);
+  }
+
+  const testList = hyperHTML.bind(document.createElement("li"))`
     <a href="${href}">
-      ${href.split("/").pop()}
+      ${testFileName}
     </a>
   `;
+  testList.append(emojiList);
```
marcoscaceres: Here, I think testList should "render" like this:

```
 <a href="${href}">
    ${testFileName}
 </a> ${emojiList}
```

src/core/data-tests.js
```diff
@@ -23,11 +23,50 @@ export const name = "core/data-tests";
...
+  if (isSecureTest) {
+    const requiresConnectionEmoji = hyperHTML.bind(
+      document.createElement("span")
+    )` 🔒 `;
+    requiresConnectionEmoji.setAttribute(
+      "aria-label",
+      "requires a secure connection"
+    );
+    requiresConnectionEmoji.setAttribute("title", "Test requires HTTPS");
+    testFileName = testFileName.replace(".https", "");
```
marcoscaceres: Not that previously, you were doing the `replace()` but not assigning it to anything.

src/core/data-tests.js
```diff
@@ -23,11 +23,50 @@ export const name = "core/data-tests";
 const lang = defaultLang in l10n ? defaultLang : "en";
 
 function toListItem(href) {
-  return hyperHTML.bind(document.createElement("li"))`
+  const emojiList = document.createElement("span");
```
marcoscaceres: Note, this should be an array really... otherwise, this could end up with either and empty span or with a nested set of spans:

```html
<a href="...">name-of-test</a> 
<!-- nested span are not great -->
<span>
   <span>💪</span>
</span>
```

We want to end up with markup like this:
```HTML
<li>
  <a href="...">name-of-test</a> <span>💪</span> <span>🔒</span>
</li>
```

src/core/data-tests.js
```diff
@@ -23,11 +23,50 @@ export const name = "core/data-tests";
...
+
+  const isManualTest = testFileName
+    .split(".")
+    .join("-")
+    .split("-")
+    .find(part => part === "manual");
+  if (isManualTest) {
+    const manualPerformEmoji = hyperHTML.bind(
+      document.createElement("span")
+    )` 💪 `;
```
marcoscaceres: nit: we probably don't need the whitespace. We should add that between the spans. If we then want them to have more padding, we should add that using CSS.

src/core/data-tests.js
shikhar-scs: ok thanx for that :+1:

src/core/data-tests.js
shikhar-scs: sure :+1:

src/core/data-tests.js
shikhar-scs: ok I'll keep all of that in mind :+1:

src/core/data-tests.js
```diff
@@ -48,10 +47,9 @@ function toListItem(href) {
     .split("-")
     .find(part => part === "manual");
   if (isManualTest) {
-    const manualPerformEmoji = hyperHTML.bind(
-      document.createElement("span")
-    )` 💪 `;
-    manualPerformEmoji.setAttribute(
+    const manualPerformEmoji = document.createElement("span")
+      manualPerformEmoji.innerHTML = `💪`;
```
marcoscaceres: Nit: note that the indentation will get reset by [prettier.js](https://github.com/prettier/prettier)

shikhar-scs: oh completely forgot about that. Apart from that any more changes you would like to see ?

marcoscaceres: Lastly, we would need to add tests to tests/spec/core/data-tests-spec.js, that check all the new changes we've made. 

So, pretty close to being done here! Thanks for hanging in there!

shikhar-scs:
@marcos I've written the tests related to my code and pushed them here. Please review 😅

marcoscaceres:
@shikhar-scs, just stepping out, but promise to review first thing tomorrow! Thanks for putting in the work.

shikhar-scs:
@marcoscaceres Not an issue at all. I'll wait for tomorrow :+1:

src/core/data-tests.js
marcoscaceres: This is still remaining to be fixed, right? (i.e., `const emojiList = [];`, instead of using a span)

tests/spec/core/data-tests-spec.js
```diff
@@ -17,6 +17,11 @@ describe("Core — data-tests attribute", () => {
             <a id="t1" data-cite="#test">a</a>
             <dfn id="t2" data-cite="#test">a</dfn>
           </p>
+          <p id="metadata-tests"
+            data-tests="this-is-secure.https.html,
+             this-is-manual.html, this-is-secure-and-manual.https.html">
```
marcoscaceres: Nit: put "this-is-secure-and-manual.https.html" on a new line, so it's visually distinct from the other two. Basically, we want:
```HTML
data-tests="
   this-is-secure.https.html,
   this-is-manual.html,
   this-is-secure-and-manual.https.html"
```

tests/spec/core/data-tests-spec.js
```diff
@@ -63,5 +68,33 @@ describe("Core — data-tests attribute", () => {
         expect(pathname.endsWith(`test${i}.html`)).toBe(true);
       });
     });
+    it(`adds emojis for manual and https related tests`, () => {
+      
+      let test = doc.querySelector("#metadata-tests ul").querySelectorAll("li")
```
marcoscaceres: As you have the tests here,  we should split the tests into three separate `it()` statements:

```JS
it("adds emojis for secure tests", () => {
  const li = doc.querySelector("#metadata-tests ul>li:nth-child(1)");
  // other stuff
});
it("adds emojis for manual tests", () => {
  const li = doc.querySelector("#metadata-tests ul>li:nth-child(2)");
  // other stuff
});
it("adds emojis for secure manual tests", () => {
  const li = doc.querySelector("#metadata-tests ul>li:nth-child(3)");
  // other stuff
});
```

tests/spec/core/data-tests-spec.js
```diff
@@ -63,5 +68,33 @@ describe("Core — data-tests attribute", () => {
...
+
+      // for test 1
+      expect(test[0].children[0].innerText.trim()).toEqual("this-is-secure");   // filename
```
marcoscaceres: might be better here:

```JS
const filename = li.innerText.trim();
expect(filename).toEqual("this-is-secure");
```

Then it's much more clear exactly what you are testing, same applies to the ones below. Basically, if you find that you have to leave a comment, e.g., `// filename`, that's a strong sign you should be using a variable.

marcoscaceres: note also here, the fact that you are calling `.trim()` seems like it might be a bug in the implementation. You shouldn't need to call trim, as the implementation should have trimmed it.

tests/spec/core/data-tests-spec.js
```diff
@@ -63,5 +71,78 @@ describe("Core — data-tests attribute", () => {
         expect(pathname.endsWith(`test${i}.html`)).toBe(true);
       });
     });
+    it(`adds emojis for secure tests`, () => {
+      const li = doc.querySelector("#metadata-tests ul>li:nth-child(1)");
+
+      const fileName = li.children[0].innerText.trim();
```
marcoscaceres: So, having to do "li.children[0]" and "li.children[1]" is ideal here, because it's not clear what element you are working with. Can you please destructure the array into named variables, like (you can probably come up with better names):

```JS
const [fileElement, secureIconElement] = li.children;
```

Similarly for the ones below...

tests/spec/core/data-tests-spec.js
```diff
@@ -63,5 +71,78 @@ describe("Core — data-tests attribute", () => {
...
+
+      const spanEmoji2 = li.children[1].children[1].innerText;
+      expect(spanEmoji2).toEqual("💪");
+
+      const spanAriaLabel2 = li.children[1].children[1].getAttribute(
+        "aria-label"
+      );
+      expect(spanAriaLabel2).toEqual("the test must be run manually");
+
+      const spanTitle2 = li.children[1].children[1].getAttribute("title");
```
marcoscaceres: Seems odd that they are nested? I think this might be a bug. The tree structure should be:

```JS
<span>file-name</span> <span>emoji</span> <span>emoji</span>
```

Or am I reading this wrong?

tests/spec/core/data-tests-spec.js
marcoscaceres: typo above, I mean "is not ideal"

shikhar-scs: Yeah Sure 👍

tests/spec/core/data-tests-spec.js
shikhar-scs: Its something like 

`<li><a>fileName</a><span>emoji</span></li>`4

So, yeah roughly the same

shikhar-scs:
@marcoscaceres I've refactored the code, removing all `children[i]` parts of code by destructuring the array. Also I realised that because of introducing the `emojilist = []` instead of initialising it as a `querySelector` , the structure of the `li` element had to be changed a bit for selecting the `spans` . 
Thanks for reminding.

Please review the code now.

shikhar-scs:
Ready for merge ??

marcoscaceres:
yep, was just waiting for green light from TravisCI.

marcoscaceres:
respec@19.6.0 going live now. Great work @shikhar-scs!

marcoscaceres:
Looking hella nice in a real spec (make sure you shift-refresh) https://w3c.github.io/payment-request/ 😻

shikhar-scs:
Thanks a lot @marcoscaceres . My first contribution to respec 🎉🎉🎉

Issue #1373, opened 2017-08-28
Github links should include link to pull requests


marcoscaceres:
I keep having to click on the issues link to get to the pull requests... it's getting a little boring.

OrionStar25:
Hey, I'd like to work on this issue! Can you please elaborate a bit more though?

marcoscaceres:
@OrionStar25 that's great, and should be pretty easy. 

See the screenshot below of what is actually needed:

![screenshot_2018-02-08_13_19_00](https://user-images.githubusercontent.com/870154/35952160-d0ddb488-0cd2-11e8-8aaf-17a42f14cb80.png)

You basically just need to add one of these objects here:
https://github.com/w3c/respec/blob/develop/src/core/github.js#L117

And add the localization string in English here:
https://github.com/w3c/respec/blob/develop/src/core/l10n.js#L19

And then add a test to check that the value showed up in "tests/spec/core/github-spec.js".

marcoscaceres:
@OrionStar25, if you haven't already, please also see the get started guide for devs https://github.com/w3c/respec/wiki/Developers-Guide

OrionStar25:
Sent a pr 👍

PR #1491, opened 2018-02-09
Github links should include link to pull requests #1373

OrionStar25:


OrionStar25:
@marcoscaceres can you please review this? thanks

OrionStar25:
@marcoscaceres Should we be updating the docs related to this? 
[Like this one](https://github.com/w3c/respec/wiki/github)
Or is there another workflow for that?

marcoscaceres:
@OrionStar25, good question... `pullBase` is mostly internal, so we probably don't want people using it directly, right? 

However, in the "github:" wiki page, have a look to see if the description that is there makes sense.

OrionStar25:
> It also adds an object to otherLinks for under "Participate", with the appropriate links to your github repository.
Since pullBase does add an object to otherLinks, and is automatically generated after giving a repoURL, I thought it should be included as well. Please tell me if I'm interpreting it wrongly though!

marcoscaceres:
So, consider the following ReSpec configuration:

```
      var respecConfig = {
        specStatus: "ED",
        editors: [{
          name: "Your Name",
          url: "https://your-site.com",
        }],
        github: "https://github.com/w3c/some-API/",
        pullBase: "foo"
      };
```

Adding `pullBase` above has no effect on the final pull request URL, right? So, for us to include it in the documentation, it would need to allow the spec author to override the pull request URL. 

Make sense?

OrionStar25:
Makes absolute sense! Thanks for clarifying that 💃

PR #1492, opened 2018-02-09
fix: add all logos in the first p element 

### LINKED ISSUE #1392: No longer possible to override W3C logo in an official document?
From time to time, W3C working groups develop specs in collaboration with external organizations. In such cases, we need to add another logo next to the W3C one. Specs developed by the Spatial Data on the Web (SDW) Working Group, a joint W3C/OGC effort, provide a concrete example, e.g. the [SSN Ontology](https://w3c.github.io/sdw/ssn/)

To add the OGC logo, the config uses the `logos` setting to [override the standard W3C logo](https://github.com/w3c/respec/wiki/logos), and specifies both the W3C and the OGC logo. This used to work fine, at least until a few months ago, I don't remember what update made behavior change.

However, the [code in headers.html](https://github.com/w3c/respec/blob/ad4a19af0b8b64e9e8af8582a45bc0d806a85f16/js/w3c/templates/headers.html) now adds the W3C logo unless the document is flagged as "unofficial". The SDW WG specs are not "unofficial".

The W3C logo is not overridden and appears duplicated as a result. Duplication is easy to fix: dropping the W3C logo from ReSpec config would do the trick. But then, ideally, I'd like the two logos to appear side by side within the same `<p>`.

saschanaz:
Fixes #1392 

Current:

image

Fix:

image


package.json
```diff
@@ -16,7 +16,8 @@
   },
   "contributors": [
     "Marcos Cáceres <marcos@marcosc.com> (https://marcosc.com/)",
-    "Robin Berjon"
+    "Robin Berjon",
+    "Kagami Sascha Rosylight <saschaplas@outlook.com>"
```
marcoscaceres: 🙏🎉

marcoscaceres:
Should we have a slightly meaner test with a few more icons?

saschanaz:
>Should we have a slightly meaner test with a few more icons?

I'm sorry but I'm not sure I understood well, what do you mean by *a few more icons*?

marcoscaceres:
Argh, sorry, I meant logos. Shouldn’t  review code at 3am 😓

src/w3c/headers.js
```diff
@@ -95,6 +95,7 @@ import { concatDate, joinAnd, ISODate } from "core/utils";
 import hb from "handlebars.runtime";
 import { pub } from "core/pubsubhub";
 import tmpls from "templates";
+import { w3cDefaults } from "w3c/defaults";
```
marcoscaceres: Just realized that this is imported but never used, right?

tests/spec/w3c/headers-spec.js
```diff
@@ -1013,4 +1013,29 @@ describe("W3C — Headers", function() {
...
+    it("adds logos defined by configuration", async () => {
+      const ops = makeStandardOps();
+      const logo = {
+        src: "http://myimage/",
+        alt: "this is a logo",
+        height: 765,
+        width: 346,
+        url: "http://hyperlink/"
+      };
+      const newProps = {
```
marcoscaceres: To clarify my comment from yesterday, we should probably test with at least 2 logos. If possible, we should add an local image here that won't result in a 404, as it makes logs a little less messy.

src/w3c/headers.js
saschanaz: Oops, I used it once and removed later, but forgot about this. Good catch!

saschanaz:
I used some small data URIs because I just didn't want to add files 😁

marcoscaceres:
Smart! I should do the same for other images.

PR #1493, opened 2018-02-09
Bundled some files

nileshgulia1:
@saschanaz  @marcoscaceres  .I have closed old PR for some rebasing and conflicting issues.This patch bundles required `js` files which are dependencies in basic.html including w3c common.js to builds directory.


saschanaz:
(The result should be respec-w3c-common.js rather than basic.js)

package.json
```diff
@@ -71,7 +71,7 @@
     "test:karma": "npm run karma",
     "prepare": "npm run snyk-protect",
     "prepublish": "npm run snyk-protect",
-    "parcel": "parcel ./examples/basic.html -d ./builds"
+    "parcel": "parcel  ./builds/respec-w3c-common.js -d ./builds/output"
```
nileshgulia1: @saschanaz  Done!

nileshgulia1:
@marcoscaceres  Can we have it merged now?

nileshgulia1:
@marcoscaceres  Parcel  bundles ``w3.respec.common.js`` and its linked dependencies to build folder. I thought that parcel-bundler is very good ,but it doesn't support multiple entry points which is needed here, I have used it because of its easy setup, no config file etc. It also enables `uglify.js` within itself to minify the code so we can refrain from using some packages here.

marcoscaceres:
@saschanaz, what’s your opinion? Worth using?

saschanaz:
And can you push your package-lock.json changes, as it helps seeing dependency changes + boosts installation speed?

marcoscaceres:
@nileshgulia1, you probably want to give this a rebase before adding your package-lock.

tools/builder.js
```diff
@@ -77,16 +76,6 @@ See original source for licenses: https://github.com/w3c/respec */
 window.respecVersion = "${version}";
```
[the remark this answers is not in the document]
saschanaz: (Note that this is also broken in master. (Comments disappear!)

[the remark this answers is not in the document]
nileshgulia1: >npm run parcel currently skips adding boilerplate, we should add it.

Which boilerplate?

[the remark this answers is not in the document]
saschanaz: This part:

https://github.com/w3c/respec/blob/1aa589d19fa99e05768cb95bebc156854727c221/tools/builder.js#L71-L76

saschanaz:
yarn is good, but I'm not sure we should maintain both lock files...

.travis.yml
```diff
@@ -6,7 +6,6 @@ node_js:
 install:
   - npm install
   - npm run build
-  - npm run build:respec-w3c-common
```
saschanaz: We should add `npm run parcel` here, but `npm run parcel` never ends on my local environment. Not sure what's happening.

package.json
```diff
@@ -71,7 +70,8 @@
     "test:headless": "node ./tests/headless.js",
     "test:karma": "npm run karma",
     "prepare": "npm run snyk-protect",
-    "prepublish": "npm run snyk-protect"
+    "prepublish": "npm run snyk-protect",
+    "parcel": "parcel  ./builds/respec-w3c-common.js -d ./builds/output"
```
saschanaz: `build` is omitted: `"parcel build ./builds/respec-w3c-common.js -d ./builds/output"`

saschanaz:
I'm getting error logs when building:

```
> npm run parcel

> respec@19.2.0 parcel C:\Users\sasch\Documents\GitHub\respec
> parcel build ./builds/respec-w3c-common.js -d ./builds/output

∞  Building respec-w3c-common.js...
Condition always false [0:2608,20]
Condition left of && always false [0:2617,87]
Dropping unreachable code [0:2867,384]
Declarations in unreachable code! [0:2867,384]
Dropping unreachable code [0:3099,151]
Declarations in unreachable code! [0:3099,151]
Function.protoype.arguments not supported [0:4182,628]
Function.protoype.arguments not supported [0:4188,164]
Dropping duplicated definition of variable ADa [0:4435,14]
Function.protoype.arguments not supported [0:4616,49]
Function.protoype.arguments not supported [0:4616,67]
Function.protoype.arguments not supported [0:4630,120]
Function.protoype.arguments not supported [0:4707,361]
Function.protoype.arguments not supported [0:4798,41]
Function.protoype.arguments not supported [0:4798,73]
Function.protoype.arguments not supported [0:4830,24]
Dropping duplicated definition of variable dLa [0:4840,25]
Function.protoype.arguments not supported [0:4930,22]
Non-strict equality against boolean: == false [0:5302,6]
Non-strict equality against boolean: == false [0:5302,63]
Dropping unreachable code [0:5652,184]
√  Built in 16.58s.
```

Too hard to understand as it never says what file is in problem.

Also, we [should be able to get a non-minified version of file](https://github.com/w3c/respec/pull/1461#issuecomment-363009075):

>Ok, but it would be a hard requirement that we can generate a concatenated, but not minified, version for development... otherwise, debugging is going to be extremely hard. Sourcemaps are very very broken across browsers, so definitely don't want to be reliant on them (been there, done that, never again 🤪).

package.json
nileshgulia1: I Have tried adding build before but is repeatedly throws this error:
```
parcel build ./builds/respec-w3c-common.js -d ./builds/output

🚨  /home/nilesh/respec/builds/respec-w3c-common.js: sym.definition is not a function
    at may_modify (eval at <anonymous> (/home/nilesh/respec/node_modules/uglify-es/tools/node.js:21:1), <anonymous>:8931:31)
    at TreeTransformer.abort [as before] (eval at <anonymous> (/home/nilesh/respec/node_modules/uglify-es/tools/node.js:21:1), <anonymous>:8541:74)
    at AST_VarDef.eval [as transform] (eval at <anonymous> (/home/nilesh/respec/node_modules/uglify-es/tools/node.js:21:1), <anonymous>:4570:35)
    at eval (eval at <anonymous> (/home/nilesh/respec/node_modules/uglify-es/tools/node.js:21:1), <anonymous>:4586:25)
    at doit (eval at <anonymous> (/home/nilesh/respec/node_modules/uglify-es/tools/node.js:21:1), <anonymous>:130:23)
    at MAP (eval at <anonymous> (/home/nilesh/respec/node_modules/uglify-es/tools/node.js:21:1), <anonymous>:156:52)
    at do_list (eval at <anonymous> (/home/nilesh/respec/node_modules/uglify-es/tools/node.js:21:1), <anonymous>:4585:16)
    at eval (eval at <anonymous> (/home/nilesh/respec/node_modules/uglify-es/tools/node.js:21:1), <anonymous>:4664:28)
    at AST_Const.eval [as transform] (eval at <anonymous> (/home/nilesh/respec/node_modules/uglify-es/tools/node.js:21:1), <anonymous>:4573:17)
    at collapse (eval at <anonymous> (/home/nilesh/respec/node_modules/uglify-es/tools/node.js:21:1), <anonymous>:8617:39)
```

.travis.yml
```diff
@@ -6,6 +6,7 @@ node_js:
 install:
   - npm install
   - npm run build
+  - npm run parcel
```
nileshgulia1: Adding `npm run parcel` is causing travis to fail.

package.json
saschanaz: Ah yes, this is the uglify-es problem we met in #1489 that will be fixed in its next release. See https://github.com/mishoo/UglifyJS2/issues/2896

nileshgulia1: Let's do not use it until new release ,Using `--no-minify` removes this error for me. https://github.com/parcel-bundler/parcel/issues/8. What do you think @saschanaz  @marcoscaceres ?

marcoscaceres: We ultimately want to minify this, otherwise the bundle size will be huge... so we are kinda stuck until the next Uglify release.

.travis.yml
nileshgulia1: @saschanaz  Should I remove this ?

saschanaz: No, this is the required step. I think `npm run build` should be completely replaced by `npm run parcel`, as that's the whole purpose of using parcel.

package.json
saschanaz: `--no-minify` may still be good for debugging purpose, where source mapping is frequently not perfect.

saschanaz:
Using parcel may remove the need of `npm run copydeps`, which will break `examples/basic.html` which uses copied dependencies.

saschanaz:
Currently we are generating transpiled non-concatenated files into `js/` and `examples/basic.html` uses this. parcel removes the need of them, and then we probably should remove `basic.html` as we already have `basic.built.html`.

marcoscaceres:
> Currently we are generating transpiled non-concatenated files into js/ and examples/basic.html uses them. parcel removes the need of them, and then we probably should remove basic.html as we already have basic.built.html.

I'm still super concerned about trying to work with sourcemaps and a fully bundled app (debugging experience is still really terrible), so I'd prefer to keep basic.html... at least, till I'm confident that source mapping support is sane.

marcoscaceres:
(I'm willing to try tho - super interested to see what comes out of all this)

saschanaz:
>I'm still super concerned about trying to work with sourcemaps and a fully bundled app (debugging experience is still really terrible)

Deeply agreed, but that means we should't use parcel at all, we are just adding a new dependency then.

marcoscaceres:
> Deeply agreed, but that means we should't use parcel at all, we are just adding a new dependency then.

Yeah, I'm kinda getting the same feeling - and can definitely live with the current setup. I guess eventually (in around 2 years) we will just be able to use ES Modules and native `require()` to import things we need. That seems fine to me.

saschanaz:
For a reference: this 2015 webpack issue is still open 😞

marcoscaceres:
Lol, it was filed by an old friend on mine! Hi @Stuk!

Stuk:
Haha, hi @marcoscaceres 👋

PR #1498, opened 2018-02-12
chore: update dependencies

marcoscaceres:


saschanaz:
```
FAILED TESTS:
  W3C — Bibliographic References
    ✖ displays the publisher when present
      HeadlessChrome 64.0.3282 (Linux 0.0.0)
    Error: SpecRef seems to be down. Can't proceed with this spec.
        at UserContext.it (tests/spec/core/biblio-spec.js:89:13)
        at <Jasmine>
```

This happens fairly frequently, Hmm.

marcoscaceres:
> This happens fairly frequently, Hmm.

Yeah, Tobie is looking into it. We haven't been able to work out why specref keeps failing on us.

saschanaz:
It seems the spec tests themselves has a problem with the new jasmine ordering behavior, as `"pings biblio service to see if it's running"` test never fails.

marcoscaceres:
We might need to do some kind of check in `beforeAll()` https://github.com/w3c/respec/blob/develop/tests/spec/core/biblio-spec.js#L66

If I recall correctly, you said there was some way to force tests to be run in order, right? If so, we might need to enable that here. Alternatively, we could do the ping on every test.

marcoscaceres:
.... Let's see how bad this gets tho.

PR #1501, opened 2018-02-12
fix(tools/release): use build:components instead

marcoscaceres:
The hb:build command is no longer available.


marcoscaceres:
Ok, will try to do a release with the stuff from the last couple of days... will be 19.2.0, I think.

PR #1504, opened 2018-02-13
feat: support --disable-sandbox on respec2html

saschanaz:
This introduces sandbox-disabling option for some limited environments. See https://github.com/w3c/spec-generator/pull/64

cc: @marcoscaceres @deniak


marcoscaceres:
It’s ok, version numbers are best effort and I missed it too in review. Sometimes things slip through. Sometimes a fix is ambiguous.

Issue #857, opened 2016-07-05
Support different bibliographical styles


riannella:
The publisher is listed after the Authors, which is not a common bibliographic style.
Can the publisher item be moved to the end (before the URL ref)?

marcoscaceres:
I tend to agree that the current bibliographical style is pretty awful. 

Can I suggest we change it to:

Linked-Document-Title (refNote). Publisher/Consortium. Date. Author/Editor. URL.

So: 
A. Phillips; M. Davis. IETF. Tags for Identifying Languages. September 2009. IETF Best Current Practice. URL: https://tools.ietf.org/html/bcp47

Becomes:
Tags for Identifying Languages (IETF Best Current Practice). IETF. September 2009. A. Phillips; M. Davis. URL: https://tools.ietf.org/html/bcp47

URL is also redundant, as it's already in the link... it should only be added when document is printed.

marcoscaceres:
So to avoid debates, we should have a `biblioStyle` option... so we can just have some of the standard ones: MLA, APA, or whatever and people can just pick.

riannella:
Agree - the biblioStyle option is the best solution (with a default setting).

marcoscaceres:
@riannella, sorry, only now coming back to this... if I point you to the code for the current bibliographical style, could you make a derivative version that produces a "standard" bilbiographical style? It would be trivial to enable.

riannella:
What biblio properties are currently supported? I can then indicate which ones are used (and the order) for APA and MLA citation styles.

marcoscaceres:
@riannella we use this data/API: https://github.com/tobie/specref/blob/master/README.md#api

So, basically, we take a reference and covert it here: 
https://github.com/w3c/respec/blob/develop/src/core/biblio.js#L49

What we would do instead is add an optional "style" argument: 
```
function stringifyReference(ref, style) {
  switch(style):
    case "APA":
      output =  // do conversion here....
       break;
    case "MLA":
       output = // do conversion here....
       break;
   default:
     output = // what ReSpec currently does... 
  }
  return output;
}
```

marcoscaceres:
Expanded example above... but that's basically what needs to happen.

riannella:
APA and MLA have different rules depending if it is a Book, Journal, Website etc...but I think this is a good compromise (given the set of biblio properties we have):

MLA: authors, title, status, publisher, date, href

APA: authors, date, title, status, publisher, href

marcoscaceres:
[...]

PR #1505, opened 2018-02-13
 Support different bibliographical styles #857

OrionStar25:
I didn't quite understand how wireReference would use style as an argument. I tried to modify stringifyReference to accomadate different citation styles. Can you please tell me how to procede further!


src/core/biblio.js
```diff
@@ -63,7 +63,7 @@ const endNormalizer = function(endStr) {
 
 const endWithDot = endNormalizer(".");
 
-export function wireReference(rawRef, target = "_blank") {
+export function wireReference(rawRef, target = "_blank",style) {
```
marcoscaceres: Sorry! I was wrong here. `stringifyReference` below was correct.

src/core/biblio.js
```diff
@@ -93,24 +93,61 @@ export function wireReference(rawRef, target = "_blank") {
...
 
-export function stringifyReference(ref) {
+export function stringifyReference(ref,style) {
   if (typeof ref === "string") return ref;
-  let output = `<cite>${ref.title}</cite>`;
-  if (ref.href) {
-    output = `<a href="${ref.href}">${output}</a>. `;
+  let output = "";
+  switch (style) {
+  	case "APA":
```
marcoscaceres: It would be great to move each case into its own function, like:

```JS
case "APA":
  output = entryToAPA(entry);
  break;
case "MLA": 
  output = entryToMLA(entry);
  break;
default:
   output = entryToW3C(entry);
return output; 
```

And then:

```JS
function entryToAPA(entry) {
   // logic here...
}
```

Makes it easier to test and maintain.

src/core/biblio.js
OrionStar25: Oh alright! So I should remove that argument from there then. I also realized I have to add " " for title in MLA. Should I be doing anything else differently?

src/core/biblio.js
```diff
@@ -93,24 +93,61 @@ export function wireReference(rawRef, target = "_blank") {
...
-export function stringifyReference(ref) {
+export function stringifyReference(ref,style) {
   if (typeof ref === "string") return ref;
-  let output = `<cite>${ref.title}</cite>`;
-  if (ref.href) {
-    output = `<a href="${ref.href}">${output}</a>. `;
+  let output = "";
+  switch (style) {
+  	case "APA": 
+  		// Author, A. (Year, Month Date of Publication). Article title. Retrieved from URL
```
marcoscaceres: The logic is good, but instead of building out the output incrementally, would it be possible to gather all the parts, then have a template string at the end where all the parts are put together?

```JS
  const authors = ...;
  const date = ...;
  return `${authors} ${date} <cite>${title}</cite> ${URL}`;
```

OrionStar25:
@marcoscaceres I did the requested changes!

src/core/biblio.js
```diff
@@ -93,24 +93,52 @@ export function wireReference(rawRef, target = "_blank") {
...
-  }
-  if (ref.authors && ref.authors.length) {
-    output += ref.authors.join("; ");
-    if (ref.etAl) output += " et al";
-    output += ".";
-  }
-  if (ref.publisher) {
-    const publisher = ref.publisher + (/\.$/.test(ref.publisher) ? "" : ".");
-    output = `${output} ${publisher} `;
+  let output = "";
```
marcoscaceres: Ok, I think what we want to do here is "normalize" each reference. That way, we don't need to check for missing properties, types, etc. 
 
```JS
// this would  go outside stringifyReference()
const canonicalRef = {
   authors: [],
   date: "",
   status: "",
   href: "",
   publisher: "",
};

// Then in stringifyReference(), we plug all the holes!
const nomarlizedRef = {...canonicalRef, ref};
```

src/core/biblio.js
```diff
@@ -93,24 +93,52 @@ export function wireReference(rawRef, target = "_blank") {
   `;
 }
 
-export function stringifyReference(ref) {
+// Author, A. (Year, Month Date of Publication). Article title. Retrieved from URL
+function entryToAPA(ref) {
+	const authors = ((ref.authors && ref.authors.length) ? (ref.authors.join("; ") + (ref.etAl ? " et al." : ".")) : "");
+	const date = ref.date ?  "(" + ref.date + "). " : "";
```
marcoscaceres: nit: `(${ref.date}).` ... bit less noisy.

src/core/biblio.js
```diff
@@ -93,24 +93,52 @@ export function wireReference(rawRef, target = "_blank") {
...
+    return `${authors} ${date} ${title} ${URL}`;
+}
+
+//Author’s Last name, First name. “Title of the Article or Individual Page.” Title of the website, Name of the publisher, Date of publication, URL.
+function entryToMLA(ref) {
+	const authors = ((ref.authors && ref.authors.length) ? (ref.authors.join("; ") + (ref.etAl ? " et al." : ".")) : "");
+	const date = ref.date ?  ref.date + ", " : "";
+	const title = ref.href ? `<a href="${ref.href}"> <cite>${ref.title}</cite> </a>. ` : "";
+	const URL = ref.href ? `<a href="${ref.href}">${ref.href}</a>` : "";
+	const publisher = ref.publisher ? ref.publisher + (/\.$/.test(ref.publisher) ? "" : ",") : "";
```
marcoscaceres: for readability, maybe use:

```JS
ref.publisher + (ref.publisher.endsWith(".") ? "" : ",") : ""
```

Might be worth a splitting into two lines also, as nested ternary statements can be confusing.

src/core/biblio.js
```diff
@@ -93,24 +93,52 @@ export function wireReference(rawRef, target = "_blank") {
...
+      	 
+  	return 	`${authors} "${title}" ${publisher} ${date} ${URL}`;
+}
+
+// what ReSpec currently does...
+function entryToW3C(ref) {
+	const title = ref.href ? `<a href="${ref.href}"> <cite>${ref.title}</cite> </a>. ` : "";
+	const authors = ((ref.authors && ref.authors.length) ? (ref.authors.join("; ") + (ref.etAl ? " et al." : ".")) : "");
```
marcoscaceres: Same comments as above.

src/core/biblio.js
marcoscaceres: fixed typo above

OrionStar25:
@marcoscaceres please review this!

marcoscaceres:
@OrionStar25, cleaned up the style of the pull request a bit. To get those:

```bash
git pull 
```

Generally, the code looks good, but the changes are causing test failures that we need to investigate. Do:

```
npm start
```

and navigate to the URL given... you want to run the tests at:  "https://localhost:8080/tests/SpecRunner.html". 

For the ones failing, open up the "tests/spec/core/data-cite-spec.js" and add some "[debugger](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/debugger)" statements for the tests that are failing. It might just be that we changed the formatting a little bit, which is causing things to fail now.

OrionStar25:
Sorry for the late response! Navigating to SpecRunner.html shows me a blank screen.

marcoscaceres:
@OrionStar25, hmmm... how odd. Can you try the following:

```Bash
git pull
npm install
npm run build
npm start
```

And then navigate to the SpecRunner.html again. 

Check also in the browser's developer tools that you have cache disabled, and in the network panel that it's actually loading stuff. 

Hopefully that should work! If not, let know and I'll see what I can do.

OrionStar25:
Had to update my node and disabling caching worked! Ill update things in about an hour.

OrionStar25:
This is a preview of SpecRunner.html
screenshot from 2018-03-03 04-14-41

I went on to the individual files (eg : tests/spec/core/biblio-spec,js ,etc) to understand what could be breaking, and went to "tests/spec/core/data-cite-spec.js and added debugger statements, but in both cases I'm clueless as to what is happening :-/ 
Can you please help me a little?

marcoscaceres:
Unchecking running the tests in random order should help. That’s something new that was added by Jasmine, but we currently don’t support. Hopefully then it should all be ok. Just in case, also check you are on latest version of Firefox (anything greater than 57).

OrionStar25:
Screenshot without randomizing tests
screenshot from 2018-03-03 04-47-08
I'm using the lastest Firefox, yes. Also, I'm not understanding how debugger is working here because when I inserted them in data-cite-spec.js , no break points were being created.

OrionStar25:
BTW, here is something I found : 

`// Make sure publisher is shown even when there is no author `
  ` ref = doc.querySelector("#bib-TestRef3 + dd"); `
   ` expect(ref).toBeTruthy(); `
    ` expect(ref.textContent).toMatch(/Publisher Here\.\s/);`

In APA citations, publishers aren't needed, hence this doesn't hold anymore!

marcoscaceres:
> Also, I'm not understanding how debugger is working here because when I inserted them in data-cite-spec.js , no break points were being created. 

Must be crashing before. Any JavaScript error being reported the developer console?

marcoscaceres:
The bug was an attempt to reassign a `const` variable. See https://github.com/w3c/respec/pull/1505/commits/29c79f3472f22236ba0abe0ee52d2b845a01b43d

src/core/biblio.js
```diff
@@ -93,24 +92,78 @@ export function wireReference(rawRef, target = "_blank") {
...
+    authors = ref.authors.join("; ") + (ref.etAl ? " et al." : ".");
+  }
+  const date = ref.date ? `${ref.date}, ` : "";
+  const title = ref.href
+    ? `<a href="${ref.href}"> "<cite>${ref.title}</cite>" </a>. `
+    : "";
+  const URL = ref.href ? `<a href="${ref.href}">${ref.href}</a>` : "";
+  const publisher = "";
+  if (ref.publisher) {
+    publisher = ref.publisher + ref.publisher.endsWith(".");
```
saschanaz: This concatenates a string and a Boolean, what's the purpose here?

src/core/biblio.js
```diff
@@ -93,24 +92,78 @@ export function wireReference(rawRef, target = "_blank") {
...
-    output += ref.authors.join("; ");
-    if (ref.etAl) output += " et al";
-    output += ".";
+    authors = ref.authors.join("; ") + (ref.etAl ? " et al." : ".");
   }
+  const publisher = "";
   if (ref.publisher) {
-    const publisher = ref.publisher + (/\.$/.test(ref.publisher) ? "" : ".");
-    output = `${output} ${publisher} `;
+    publisher = ref.publisher + ref.publisher.endsWith(".");
```
saschanaz: Same concatenation.

src/core/biblio.js
marcoscaceres: This will also fail, as it’s reassigning a const.

src/core/biblio.js
marcoscaceres: Adding some tests should pick up any other bugs like this.

src/core/biblio.js
```diff
@@ -93,24 +92,78 @@ export function wireReference(rawRef, target = "_blank") {
...
   }
+
+  return `${authors} ${title} ${publisher} ${date} ${URL}`;
+}
+
+// what ReSpec currently does...
+function entryToW3C(ref) {
+  const title = ref.href
+    ? `<a href="${ref.href}"> <cite>${ref.title}</cite> </a>. `
+    : "";
```
saschanaz: Do we want to skip title if no `ref.href`?

marcoscaceres: ah, well spotted. Definitely, no.

src/core/biblio.js
```diff
@@ -93,24 +92,78 @@ export function wireReference(rawRef, target = "_blank") {
...
   if (ref.publisher) {
-    const publisher = ref.publisher + (/\.$/.test(ref.publisher) ? "" : ".");
-    output = `${output} ${publisher} `;
+    publisher = ref.publisher + ref.publisher.endsWith(".");
+  }
+  const date = ref.date ? `${ref.date}. ` : "";
+  const status = ref.status
+    ? (REF_STATUSES.get(ref.status) || ref.status) + ". "
+    : "";
+  const URL = ref.href ? `URL: <a href="${ref.href}">${ref.href}</a>` : "";
```
saschanaz: I would prefer `url` over `URL` as a variable name.

saschanaz:
It seems the original `wireReference` and `stringifyReference` roughly do the same thing, is the latter deprecated?

src/core/biblio.js
marcoscaceres: Agree... we generally only use uppercase variable names for constants.

marcoscaceres:
> It seems the original `wireReference` and `stringifyReference` roughly do the same thing, is the latter deprecated?

Need to check which one is used where... and we should maybe just have one. My feeling is that `wireReference` should probably go away, and the caller of `stringifyReference()` can wire the stringified reference.

saschanaz:
https://github.com/w3c/respec/blob/4602ac0020741391f2c5d46460f6c518880e67e6/src/ui/search-specref.js#L44-L51

saschanaz:
https://github.com/w3c/respec/blob/4602ac0020741391f2c5d46460f6c518880e67e6/src/ui/search-specref.js#L44-L51

marcoscaceres:
Yeah, that `wireReference` seems a bit unnecessary.

saschanaz:
I think we should ultimately use `wireReference`-like hyperHTML pattern rather than current string concatenation.

marcoscaceres:
I guess you are right. I was thinking it's easier to pass strings around and transform them into HTML as needed... but if the output is always real HTML DOM elements, then working with DOM objects makes sense.

OrionStar25:
@saschanaz @marcoscaceres I have tried to make the requested changes. As I understand from the discussion above, we need to modify `wireReference` and remove `stringifyReference` completely. 
So what should be my next move then?

OrionStar25:
The only test case failing now is that of the publisher not getting printed when it is present (APA citation), I'll change the test for that.

OrionStar25:
Also, I need some help here. When I finished running the tests, some files themselves got added (as can be seen in this commit). Idk what these are and they are making codacy fail too :cry:

marcoscaceres:
@OrionStar25, not sure what's happened... but the easiest thing to do is to copy the "src/core/biblio.js" somewhere else on your computer and do:

```
git checkout develop 
git pull
git checkout -b biblio2
``` 

And override "src/core/biblio.js" with the changes you have made. 

With regards to `wireReference()` and `stringilyReference()`, don't worry about that for now :) We can deal with that later.

marcoscaceres:
Closing because of unrecoverable issues. @OrionStar25, don't worry too much, just follow the steps I gave above, and send a new pull request... git mistake happen all the time (I've been using it for like 5+ years every single day, and still get into tricky situation like this - it happens 😊).

OrionStar25:
Oh okay! Sure I'll send a fresh PR asap :)

saschanaz:
A simple rebasing will help, as the added files are from the recent changes on the main branch.

marcoscaceres:
@OrionStar25, try @saschanaz suggestion first (as that's the right way to do it):

```bash
git rebase develop
git push origin biblio -f
```

But if it gets lots of errors, please fall back to my suggestion.

OrionStar25:
Should I make a new PR after rebasing? Also, the new files aren't going away.

saschanaz:
You can reopen this PR or just open a new one. Either way is okay.

For those new files, delete them, push that commit, and then you won't see them in git anymore.

Issue #670, opened 2016-03-24
Issues that are closed shouldn't show up in the spec


marcoscaceres:
With the GH module, when issues are closed, they still show up in the spec but as empty... they should be removed from the document and from the summary:

![screenshot 2016-03-24 13 45 18](https://cloud.githubusercontent.com/assets/870154/14007004/c41a4000-f1c6-11e5-9286-71733d7bed92.png)

A warning should be generated instead.

himanish-star:
I can fix this maybe.

marcoscaceres:
+ respec@20.0.5

PR #1506, opened 2018-02-14
Closed issues are now removed from the spec

himanish-star:
fixes: #670 
I tried writing tests but when I was logging inside the test console I found out the changes made in `issue-notes.js` were not reflecting somehow inside the test written for the same file inside `issue-notes-spec.js`. Please tell me how to move about with the test

### DIFF
--- a/src/core/issues-notes.js
+++ b/src/core/issues-notes.js
@@ -15,116 +15,128 @@ import css from "deps/text!core/css/issues-notes.css";
 export const name = "core/issues-notes";
 
 function handleIssues($ins, ghIssues, conf) {
-  const { issueBase } = conf;
+  const { issueBase, githubAPI } = conf;
   var hasDataNum = $(".issue[data-number]").length > 0,
     issueNum = 0,
     $issueSummary = $(
       "<div><h2>" + conf.l10n.issue_summary + "</h2><ul></ul></div>"
     ),
     $issueList = $issueSummary.find("ul");
-  $ins.each(function (i, inno) {
-    var $inno = $(inno),
-      isIssue = $inno.hasClass("issue"),
-      isWarning = $inno.hasClass("warning"),
-      isEdNote = $inno.hasClass("ednote"),
-      isFeatureAtRisk = $inno.hasClass("atrisk"),
-      isInline = $inno[0].localName === "span",
-      dataNum = $inno.attr("data-number"),
-      report = {
-        inline: isInline,
-        content: $inno.html(),
-      };
-    report.type = isIssue
-      ? "issue"
-      : isWarning ? "warning" : isEdNote ? "ednote" : "note";
-    if (isIssue && !isInline && !hasDataNum) {
-      issueNum++;
-      report.number = issueNum;
-    } else if (dataNum) {
-      report.number = dataNum;
-    }
-    // wrap
-    if (!isInline) {
-      var $div = $(
-        "<div class='" +
-        report.type +
-        (isFeatureAtRisk ? " atrisk" : "") +
-        "'></div>"
-      ),
-        $tit = $(
+  if (githubAPI) {
+    Array.from($ins)
+      .filter(({ dataset : { number : value } }) => ghIssues.get(Number(value)).state === "closed")
+      .forEach(issue => {
+        const { dataset: { number } } = issue;
+        const msg = `Github issue ${number} was closed on GitHub, so removing from spec`;
+        pub("warn", msg);
+        issue.remove();
+      });
+  }
+  $ins
+    .filter((i, issue) => issue.parentNode)
+    .each(function (i, inno) {
+      var $inno = $(inno),
+        isIssue = $inno.hasClass("issue"),
+        isWarning = $inno.hasClass("warning"),
+        isEdNote = $inno.hasClass("ednote"),
+        isFeatureAtRisk = $inno.hasClass("atrisk"),
+        isInline = $inno[0].localName === "span",
+        dataNum = $inno.attr("data-number"),
+        report = {
+          inline: isInline,
+          content: $inno.html(),
+        };
+      report.type = isIssue
+        ? "issue"
+        : isWarning ? "warning" : isEdNote ? "ednote" : "note";
+      if (isIssue && !isInline && !hasDataNum) {
+        issueNum++;
+        report.number = issueNum;
+      } else if (dataNum) {
+        report.number = dataNum;
+      }
+      // wrap
+      if (!isInline) {
+        var $div = $(
+          "<div class='" +
+          report.type +
+          (isFeatureAtRisk ? " atrisk" : "") +
+          "'></div>"
+          ),
+          $tit = $(
           "<div role='heading' class='" +
           report.type +
           "-title'><span></span></div>"
-        ),
-        text = isIssue
-          ? isFeatureAtRisk ? "Feature at Risk" : conf.l10n.issue
-          : isWarning
-            ? conf.l10n.warning
-            : isEdNote ? conf.l10n.editors_note : conf.l10n.note,
-        ghIssue;
-      $tit.makeID("h", report.type);
-      report.title = $inno.attr("title");
-      if (isIssue) {
-        if (hasDataNum) {
-          if (dataNum) {
-            text += " " + dataNum;
-            // Set issueBase to cause issue to be linked to the external issue tracker
-            if (!isFeatureAtRisk && issueBase) {
-              $tit
-                .find("span")
-                .wrap($("<a href='" + issueBase + dataNum + "'/>"));
-            } else if (isFeatureAtRisk && conf.atRiskBase) {
-              $tit
-                .find("span")
-                .wrap($("<a href='" + conf.atRiskBase + dataNum + "'/>"));
+          ),
+          text = isIssue
+            ? isFeatureAtRisk ? "Feature at Risk" : conf.l10n.issue
+            : isWarning
+              ? conf.l10n.warning
+              : isEdNote ? conf.l10n.editors_note : conf.l10n.note,
+          ghIssue;
+        $tit.makeID("h", report.type);
+        report.title = $inno.attr("title");
+        if (isIssue) {
+          if (hasDataNum) {
+            if (dataNum) {
+              text += " " + dataNum;
+              // Set issueBase to cause issue to be linked to the external issue tracker
+              if (!isFeatureAtRisk && issueBase) {
+                $tit
+                  .find("span")
+                  .wrap($("<a href='" + issueBase + dataNum + "'/>"));
+              } else if (isFeatureAtRisk && conf.atRiskBase) {
+                $tit
+                  .find("span")
+                  .wrap($("<a href='" + conf.atRiskBase + dataNum + "'/>"));
+              }
+              ghIssue = ghIssues.get(Number(dataNum));
+              if (ghIssue && !report.title) {
+                report.title = ghIssue.title;
+              }
             }
-            ghIssue = ghIssues.get(Number(dataNum));
-            if (ghIssue && !report.title) {
-              report.title = ghIssue.title;
+          } else {
+            text += " " + issueNum;
+          }
+          if (report.number !== undefined) {
+            // Add entry to #issue-summary.
+            var id = "issue-" + report.number,
+              $li = $("<li><a></a></li>"),
+              $a = $li.find("a");
+            $div.attr("id", id);
+            $a.attr("href", "#" + id).text(conf.l10n.issue + " " + report.number);
+            if (report.title) {
+              $li.append(
+                $(
+                  "<span style='text-transform: none'>: " +
+                  report.title +
+                  "</span>"
+                )
+              );
             }
+            $issueList.append($li);
           }
-        } else {
-          text += " " + issueNum;
         }
-        if (report.number !== undefined) {
-          // Add entry to #issue-summary.
-          var id = "issue-" + report.number,
-            $li = $("<li><a></a></li>"),
-            $a = $li.find("a");
-          $div.attr("id", id);
-          $a.attr("href", "#" + id).text(conf.l10n.issue + " " + report.number);
-          if (report.title) {
-            $li.append(
-              $(
-                "<span style='text-transform: none'>: " +
-                report.title +
-                "</span>"
-              )
-            );
-          }
-          $issueList.append($li);
+        $tit.find("span").text(text);
+        if (report.title) {
+          $tit.append(
+            $("<span style='text-transform: none'>: " + report.title + "</span>")
+          );
+          $inno.removeAttr("title");
         }
+        $tit.addClass("marker");
+        $div.append($tit);
+        $inno.replaceWith($div);
+        var body = $inno.removeClass(report.type).removeAttr("data-number");
+        if (ghIssue && !body.text().trim()) {
+          body = ghIssue.body_html;
+        }
+        $div.append(body);
+        const level = $tit.parents("section").length + 2;
+        $tit.attr("aria-level", level);
       }
-      $tit.find("span").text(text);
-      if (report.title) {
-        $tit.append(
-          $("<span style='text-transform: none'>: " + report.title + "</span>")
-        );
-        $inno.removeAttr("title");
-      }
-      $tit.addClass("marker");
-      $div.append($tit);
-      $inno.replaceWith($div);
-      var body = $inno.removeClass(report.type).removeAttr("data-number");
-      if (ghIssue && !body.text().trim()) {
-        body = ghIssue.body_html;
-      }
-      $div.append(body);
-      const level = $tit.parents("section").length + 2;
-      $tit.attr("aria-level", level);
-    }
-    pub(report.type, report);
-  });
+      pub(report.type, report);
+    });
   if ($(".issue").length) {
     if ($("#issue-summary"))
       $("#issue-summary").append($issueSummary.contents());
--- a/tests/spec/core/issues-notes-spec.js
+++ b/tests/spec/core/issues-notes-spec.js
@@ -160,6 +160,33 @@ describe("Core — Issues and Notes", function() {
     );
   });
 
+  it("removes closed issues from spec", async () => {
+    const githubConfig = {
+      github: "https://github.com/mock-company/mock-repository",
+      githubAPI: `${window.location.origin}/tests/data`
+    };
+    const ops = {
+      config: githubConfig,
+      body:
+        makeDefaultBody() +
+        "<div class='issue' id='this-should-not-exist' data-number='1548'></div>" +
+        "<div class='issue' data-number='1540'></div>" +
+        "<div class='issue' id='i-should-be-here-too'></div>" +
+        "<section id='issue-summary'></section>",
+    };
+    const doc = await makeRSDoc(ops);
+    const issueDiv1 = doc.getElementById("this-should-not-exist");
+    expect(issueDiv1).toBeFalsy();
+    const issueDiv2 = doc.getElementById("issue-1540");
+    expect(issueDiv2).toBeTruthy();
+    const issueDiv3 = doc.getElementById("i-should-be-here-too");
+    expect(issueDiv3).toBeTruthy();
+    const summarySection = doc.getElementById("issue-summary");
+    expect(summarySection).toBeTruthy();
+    const  { innerText } = summarySection.querySelector("[href='#issue-1540']");
+    expect(innerText).toBe("Issue 1540");
+  });
+
   it("should link to external issue tracker for features at risk", function(
     done
   ) {
--- a/tests/data/issues/1540.json
+++ b/tests/data/issues/1540.json
@@ -1,0 +1,6 @@
+[
+  {
+    "number": 1540,
+    "title": "A mock open issue for testing"
+  }
+]


saschanaz:
>but when I was logging inside the test console I found out the changes made in issue-notes.js were not reflecting somehow inside the test written for the same file inside issue-notes-spec.js.

Probably because of browser cache? ^F5 may help.

himanish-star:
Oh, you might be right let me check. If it doesn't work, I'll explain a bit more on this about my test. Do you think this change is what you need?

saschanaz:
I think this PR is in the right direction, but we are removing jQuery codes, any new code should not use jQuery. You may use `item` instead of `$(item)` in this case.

src/core/issues-notes.js
```diff
@@ -23,6 +23,14 @@ export function run(conf, doc, cb) {
         "<div><h2>" + conf.l10n.issue_summary + "</h2><ul></ul></div>"
       ),
       $issueList = $issueSummary.find("ul");
+    $ins = $ins.filter(function (index, item, array) {
+      var $item = $(item),
+        dataNum = $item.attr("data-number");
+      if (ghIssues[dataNum] === undefined) {
+        $item.hide();
```
saschanaz: Perhaps `.remove()` rather than just hiding?

src/core/issues-notes.js
```diff
@@ -23,6 +23,14 @@ export function run(conf, doc, cb) {
         "<div><h2>" + conf.l10n.issue_summary + "</h2><ul></ul></div>"
       ),
       $issueList = $issueSummary.find("ul");
+    $ins = $ins.filter(function (index, item, array) {
```
saschanaz: `array` is not used so it can be removed.

himanish-star: yup I'll make these changes

saschanaz:
(Don't forget that we need warnings here. `core/pubsubhub` module handles it.)

himanish-star:
Sure I'll implement that too! I'll update this PR tomorrow maybe. Today I got some college assignments which I got to complete 😅

himanish-star:
@saschanaz, I have made the changes please see

src/core/issues-notes.js
```diff
@@ -23,6 +23,16 @@ export function run(conf, doc, cb) {
...
       ),
       $issueList = $issueSummary.find("ul");
+    if (!(Object.keys(ghIssues).length === 0 && ghIssues.constructor === Object)) {
+      $ins = $ins.filter(function (index, item) {
+        var dataNum = item.getAttribute("data-number");
+        if (ghIssues[dataNum] === undefined) {
+          item.remove();
+        }
+        pub("warn", `issue #${dataNum} is a closed issue`);
+        return ghIssues[dataNum] !== undefined;
```
saschanaz: How about:

```
if issue is defined {
  return true;
}
warn about the undefined thing;
```

This way we don't need to do `[=|!]== undefined` twice.

src/core/issues-notes.js
```diff
@@ -23,6 +23,16 @@ export function run(conf, doc, cb) {
         "<div><h2>" + conf.l10n.issue_summary + "</h2><ul></ul></div>"
       ),
       $issueList = $issueSummary.find("ul");
+    if (!(Object.keys(ghIssues).length === 0 && ghIssues.constructor === Object)) {
```
saschanaz: Line 140 (150 in your PR) always creates `ghIssues` as an object, so probably no need to check constructor. Probably worth making `ghIssues` as a map rather than a dictionary object, so that we can just check `ghIssues.size` instead of doing `Object.keys`.

saschanaz: Wait, what happens if `Object.keys(ghIssues).length === 0`? No filtering at all?

src/core/issues-notes.js
himanish-star: Ok

src/core/issues-notes.js
himanish-star: If I remove this check there are a number of tests which fail :disappointed:

saschanaz: Hmm, we should fix those failures then. Maybe the failing tests include some closed issues 😄

himanish-star: Maybe. Fine then I'll update this in 10-15 mins

src/core/issues-notes.js
```diff
@@ -23,6 +23,16 @@ export function run(conf, doc, cb) {
...
+      $ins = $ins.filter(function (index, item) {
+        var dataNum = item.getAttribute("data-number");
+        if (ghIssues[dataNum] === undefined) {
+          item.remove();
```
marcoscaceres: Filtering should have no side effects (i.e., don't `remove()` things)... return the things you are going to remove, then call `.forEach()` to remove them.

himanish-star: ok

himanish-star: so shall I keep this or remove? 
```javascript
Object.keys(ghIssues).length === 0
```

marcoscaceres: Probably not

marcoscaceres: Check is probably not needed, that is

himanish-star: Ok marcos I'll change this once I find some time. Currently focusing on my proposal :sweat_smile:

himanish-star: image
What do I do in such a case?

himanish-star:
@saschanaz, I have made some changes but some tests are still failing :disappointed:

saschanaz:
The errors are related with removals. Maybe just adding ~~strikethrough~~ rather than removing is better for testing and for users (they can explicitly see an issue existed but now is fixed). What do you think? @marcoscaceres

himanish-star:
Sure then once you decide let me know. I'm working on the other PR of mine for now

himanish-star:
@marcoscaceres, any suggestions on this?

marcoscaceres:
Sorry, didn't have time to get to this today @himanish-star :( Will try to get to it tomorrow.

himanish-star:
Sure no problem

himanish-star:
@marcoscaceres, any updates on this?

src/core/issues-notes.js
```diff
@@ -22,7 +22,18 @@ export function run(conf, doc, cb) {
       $issueSummary = $(
         "<div><h2>" + conf.l10n.issue_summary + "</h2><ul></ul></div>"
       ),
-      $issueList = $issueSummary.find("ul");
+      $issueList = $issueSummary.find("ul"),
+      closedIssues = [];
+    closedIssues = $ins.filter(function (index, item) {
```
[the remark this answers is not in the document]
marcoscaceres: I think here we actually want something like this:

```JS
    const closedIssues = Array.from($ins)
       // only issues with a number that are not in ghIssues
      .filter(({ dataset: { number } }) => number && !ghIssues[dataNum])
      .forEach(elem => {
        const msg = `Github issue ${dataNum} was closed on GitHub, so removing from spec`;
        pub("warn", msg);
        elem.remove(); // remove from DOM
      });
```

As elements above get removed from the DOM tree, you then would need to also filter them out of `$ins` on line 37 (otherwise, we are doing needless work on things that are not in the DOM tree). You can filter DOM nodes that are not longer attached using `document.contains(element)`.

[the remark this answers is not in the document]
marcoscaceres: Fixed typo above... didn't need `const closedIssues = ` at all.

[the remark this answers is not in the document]
himanish-star: Thanks marcos :smiley:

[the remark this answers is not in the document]
saschanaz: >You can filter DOM nodes that are not longer attached using `document.contains(element)`.

(Checking `element.parentNode` will be better for performance.)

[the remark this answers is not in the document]
himanish-star: I'll look into this too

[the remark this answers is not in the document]
marcoscaceres: @saschanaz is right. Though generally we prefer readability/maintainability over performance (unless it really needs super high performance code). I'm happy with either, so long as it's clear what the code is doing and why.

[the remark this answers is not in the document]
himanish-star: Ok, I will be updating my work here very soon :smile:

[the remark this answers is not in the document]
himanish-star: The reason you have used `Array.from` is to avoid another declaration of an array right ? So basically you wanted a use and throw sort of a method. Cool got it :+1:

src/core/issues-notes.js
```diff
@@ -23,7 +23,18 @@ export function run(conf, doc, cb) {
...
+    Array.from($ins)
+      .filter(({ dataset: { number } }) => number && !ghIssues[number] && conf.githubAPI)
+      .forEach(elem => {
+        const { dataset: { number } } = elem;
+        const msg = `Github issue ${number} was closed on GitHub, so removing from spec`;
+        pub("warn", msg);
+        elem.remove();
+      });
     $ins.each(function(i, inno) {
+      if (!document.contains(inno.parentNode)) {
```
saschanaz: Oops, what I meant was `if (!inn.parentNode)` as any removed elements does not have a parent node.

himanish-star: Oh sorry, even I intended on writing the same but maybe I might have accidentally written that

src/core/issues-notes.js
```diff
@@ -23,7 +23,18 @@ export function run(conf, doc, cb) {
         "<div><h2>" + conf.l10n.issue_summary + "</h2><ul></ul></div>"
       ),
       $issueList = $issueSummary.find("ul");
+    Array.from($ins)
+      .filter(({ dataset: { number } }) => number && !ghIssues[number] && conf.githubAPI)
```
himanish-star: I have added an extra test for `conf.githubAPI` as without this two-three tests fail. The reason being that there are cases where user sets up an issue base that is not related to any github repo so in that case `ghIssues[number]` by default goes to `undefined` and thus leading to removal of that issue from the spec. I found this out when I was analysing the tests located inside https://github.com/w3c/respec/blob/develop/tests/spec/core/issues-notes-spec.js and an example can be given for https://github.com/w3c/respec/blob/3e505449ad98fbc3e6977de1c976fbf05df5dc38/tests/spec/core/issues-notes-spec.js#L181

himanish-star: @saschanaz you can now have a look. Shall I begin writing tests for this ?

marcoscaceres: It does that... remember, you can test everything locally (see dev guide). Don't rely on Travis during development as it can be very slow.

himanish-star: Yep that is how I test using localhost:8080/tests/specRunner.html

himanish-star: So shall I write tests now. If everything is ok

src/core/issues-notes.js
```diff
@@ -23,7 +23,18 @@ export function run(conf, doc, cb) {
...
+      .forEach(elem => {
+        const { dataset: { number } } = elem;
+        const msg = `Github issue ${number} was closed on GitHub, so removing from spec`;
+        pub("warn", msg);
+        elem.remove();
+      });
     $ins.each(function(i, inno) {
+      if (!inno.parentNode) {
```
marcoscaceres: Nit: as this `if` is "filtering", can you do something like this here (I don't know how to do it in jQeury, but...):
```
$ins
  .filter((i, elem) => elem.parentNode)
  .each(function(i, inno) { 
```

We should also rename "inno"... I don't know who named it that, but it's not a great name.

src/core/issues-notes.js
marcoscaceres: Tests would be great!

saschanaz: >I have added an extra test for `conf.githubAPI` as without this two-three tests fail.

So it means the filter requires `conf.githubAPI`. In that case maybe we should add if-statement to skip the filtering itself if not needed..

himanish-star:
@saschanaz, you're right. Slipped out of my head :sweat_smile:

src/core/issues-notes.js
himanish-star: `$ins` is actually in normal JavaScript that is why we do `$inno = $(inno)`. So elem.parentNode works, tested this just now things do work.

himanish-star: Made changes except that I forgot to change the name of `inno`

himanish-star:
@marcoscaceres, the test written by me is an adhoc one, where I am checking whether the issue is present or not. One problem is that the test will fail if the issue is reopened. I'm not sure if this is possible but I'm thinking of using the `githubAPI` to fetch all the issues and make some relevant checks after this. Is this a feasible plan ?

src/core/issues-notes.js
himanish-star: I have renamed inno to elem. Is this better ?

src/core/issues-notes.js
```diff
@@ -113,12 +125,12 @@ export function run(conf, doc, cb) {
               "<span style='text-transform: none'>: " + report.title + "</span>"
             )
           );
-          $inno.removeAttr("title");
+          $elem.removeAttr("title");
         }
         $tit.addClass("marker");
```
saschanaz: Lol that name 🤣

src/core/issues-notes.js
```diff
@@ -23,17 +23,29 @@ export function run(conf, doc, cb) {
...
+        .forEach(elem => {
+          const { dataset: { number } } = elem;
+          const msg = `Github issue ${number} was closed on GitHub, so removing from spec`;
+          pub("warn", msg);
+          elem.remove();
+        });
+    }
+    $ins
+      .filter((i, elem) => elem.parentNode)
+      .each(function(i, elem) {
```
saschanaz: `elem` looks too general, how about `issueNumber`, `issueNum`, or just `issue`?

tests/spec/core/issues-notes-spec.js
```diff
@@ -160,6 +160,31 @@ describe("Core — Issues and Notes", function() {
...
+      }
+    };
+    const ops = {
+      config: githubConfig,
+      body:
+        makeDefaultBody() +
+        "<div class='issue' data-number='1530'></div>" +
+        "<section id='issue-summary'></section>",
+    };
+    makeRSDoc(
```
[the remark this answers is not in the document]
himanish-star: Yep I have seen this new pattern inside the test for link-to-dfn-spec.js, but as I saw that all tests were written in the old way, I thought of not breaking the pattern. No problem I'll change it. 

Are you satisfies with the test (in terms of the content)?

src/core/issues-notes.js
```diff
@@ -23,17 +23,29 @@ export function run(conf, doc, cb) {
...
+      var $elem = $(elem),
```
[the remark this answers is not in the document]
himanish-star: :sweat_smile:

tests/spec/core/issues-notes-spec.js
```diff
@@ -160,6 +160,31 @@ describe("Core — Issues and Notes", function() {
...
+      ops,
+      function(doc) {
+        const issueDiv = doc.querySelector("#issue-1530");
+        expect(issueDiv).toBeFalsy();
+        const summarySection = doc.querySelector("#issue-summary");
```
saschanaz: I would prefer `doc.getElementById("issue-summary")` as it skips parsing selector syntax, but not a big problem.

tests/spec/core/issues-notes-spec.js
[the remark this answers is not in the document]
saschanaz: 👍

tests/spec/core/issues-notes-spec.js
himanish-star: done :+1:

himanish-star: @marcoscaceres, any suggestions? A lot of changes have occurred due to indentation of the code.

src/core/issues-notes.js
```diff
@@ -23,111 +23,123 @@ export function run(conf, doc, cb) {
...
+        .forEach(issue => {
+          const { dataset: { number } } = issue;
+          const msg = `Github issue ${number} was closed on GitHub, so removing from spec`;
+          pub("warn", msg);
+          issue.remove();
+        });
+    }
+    $ins
+      .filter((i, issue) => issue.parentNode)
+      .each(function(i, issue) {
```
marcoscaceres: Seems odd that it thinks you changed all these lines. Can you please ensure the formatting of all this doesn't change?

tests/spec/core/issues-notes-spec.js
```diff
@@ -160,6 +160,26 @@ describe("Core — Issues and Notes", function() {
     );
   });
 
+  it("should remove closed issues from spec", async () => {
```
marcoscaceres: nit: `it(removes closed issues from spec...`. "should" is unnecessary.

tests/spec/core/issues-notes-spec.js
marcoscaceres: Don't worry about the indentation for now... worst case, I can run prettier on "develop" and you just rebase on-top. Then things should match.

src/core/issues-notes.js
marcoscaceres: Ignore this... I just saw your question below

marcoscaceres:
Rebased and (hopefully) fixed the merge conflicts - @himanish-star, can you please double check that it's still fine?

himanish-star:
Sure

src/core/issues-notes.js
himanish-star: ok

marcoscaceres:
To see the diff and ignore whitespace, try: 
https://github.com/w3c/respec/pull/1506/files?w=1

tests/spec/core/issues-notes-spec.js
himanish-star: Everything now seems fine. But I have a doubt what if issue 1530 get reopened in the future again, in that case the test would fail

marcoscaceres: here ya go: https://github.com/w3c/respec/issues/1548 😸

marcoscaceres:
Talked with @himanish-star, we are going to mock the github response using a JSON file instead. We can do that by overriding the github API config option, by posting it to window.location's origin instead (so, like `${window.location.origin}/tests/data/mock-github-api-v3-response.json`).

saschanaz:
Another option would be using dummy closed issue by GitHub. https://github.com/octocat/Hello-World/issues/7

That Hello-World repo is being used for API documentation. https://developer.github.com/v4/guides/migrating-from-rest/

himanish-star:
Resuming my work on this :smile:

marcoscaceres:
Great to hear! Please keep an eye on  https://github.com/w3c/respec/pull/1560 as we will need to align them.

himanish-star:
@marcoscaceres, You can have a look now

himanish-star:
This is strange as none of my tests fail locally but one test fails in CI
Below is the test log
image

```Javascript
    const issueDiv2 = doc.getElementById("issue-1540");
    expect(issueDiv2).toBeTruthy();
```
This is the region of the failing test
This shouldn't fail as issueDiv2 is truthy

tests/spec/core/issues-notes-spec.js
```diff
@@ -160,6 +160,32 @@ describe("Core — Issues and Notes", function() {
     );
   });
 
+  it("removes closed issues from spec", async () => {
+    const githubConfig = {
+      github: "https://github.com/mock-company/mock-repository",
```
marcoscaceres: this might conflict with the next property. Make sure you check that it's correctly overridden.

tests/spec/core/issues-notes-spec.js
```diff
@@ -160,6 +160,32 @@ describe("Core — Issues and Notes", function() {
...
+      githubAPI: `${window.location.origin}/tests/data/mock-github-api-v3-response.json`
```
marcoscaceres: this may end up being incompatible with:
https://github.com/w3c/respec/pull/1560/files#diff-0cfbbabe5bc2a8772222d1152309dd3eR144

Here we probably want: `${window.location.origin}/tests/data/`

And we should resolve to: `${window.location.origin}/tests/data/issues/1540`

"1540" should just be a the json file. Alternatively, you might be able to just add "?testing=", which should also work: 

```js
${window.location.origin}/tests/data/mock-github-api-v3-issue-response.json?testing=
```

tests/spec/core/issues-notes-spec.js
```diff
@@ -160,6 +160,32 @@ describe("Core — Issues and Notes", function() {
...
+    const doc = await makeRSDoc(ops);
+    const issueDiv1 = doc.getElementById("this-should-not-exist");
+    expect(issueDiv1).toBeFalsy();
+    const issueDiv2 = doc.getElementById("issue-1540");
+    expect(issueDiv2).toBeTruthy();
+    const issueDiv3 = doc.getElementById("i-should-be-here-too");
+    expect(issueDiv3).toBeTruthy();
+    const summarySection = doc.getElementById("issue-summary");
+    expect(summarySection).toBeTruthy();
+    expect(summarySection.querySelector("[href='#issue-1540']").innerText).toBe("Issue 1540");
```
marcoscaceres: nit: 
````js
const  { innerText } = summarySection.querySelector("[href='#issue-1540']");
expect(innerText).toBe("Issue 1540");
````

File: tests/data/mock-github-api-v3-response.json [hunk 20]
```diff
@@ -0,0 +1,5 @@
+{
+  "name": "mock-repository",
+  "issues_url": "http://localhost:8080/tests/data/mock-github-api-v3-issue-response.json",
```
marcoscaceres: If this URL is going to be used, it might be a problem. There is no assurance that "ttp://localhost:8080" will be available on Travis or elsewhere.

tests/spec/core/issues-notes-spec.js
himanish-star: Ok, got this

tests/spec/core/issues-notes-spec.js
himanish-star: Yeah sure, this was just a sort of demo

tests/data/mock-github-api-v3-response.json
himanish-star: Oh, that's the reason why Travis is failing. But I couldn't find a way to change it to something like window.location as this is a JSON file

marcoscaceres: yeah :) I've change how issues-and-notes.js works now, so it doesn't download the index from github anymore. So now it should work... but you will need to rebase this.... sorry about the merge conflict 😭

marcoscaceres: by "it should work", I mean, it should be easier to test.

himanish-star: So all in all what should I do, only a rebase?

himanish-star: @marcoscaceres, I like the conflict :laughing:

marcoscaceres: @himanish-star let's get this back to a good state, just showing your changes. Then we can work through it.

himanish-star: Actually @marcoscaceres. I'm outside my home and will get back to work in the night after 2hrs

himanish-star: @marcoscaceres, stuck now for a while
image

himanish-star: Also @marcoscaceres, I'll manage to update the code as you want, as now I can clearly see the new changes and how to use them. In the meanwhile it is my sincere request to you, to please find some time from your busy schedule to review my proposal today :pray:

himanish-star: @marcoscaceres, I have pushed the changes and there's only one problem now, that is resolving to route `$(window.location.origin)/tests/data/issues/1540` won't fetch the JSON file we'll have to resolve to `$(window.location.origin)/tests/data/issues/1540.json`

himanish-star:
I have  made the changes

tests/spec/core/issues-notes-spec.js
himanish-star: @marcoscaceres, I think I do have to set `github` to something like `https://github.com/mock-company/mock-repository` otherwise it fires an error saying that github links should start with `https://github.com`. So I guess the conflict would persist :confused:

PR #1507, opened 2018-02-14
Refactor showPeople helper function

saschanaz:
This PR uses hyperHTML for showPeople function, replacing previous simple-but-hard-to-read string concatenations.

cc: @marcoscaceres 

Probably a good chance to do some chat and make things in right direction, before any migrations for core IDL templates.

### DIFF
--- a/src/w3c/headers.js
+++ b/src/w3c/headers.js
@@ -170,14 +170,11 @@ hb.registerHelper("showPeople", function(name, items = []) {
     }
     if (p.note) ret += ` (${p.note})`;
     if (p.extras) {
-      var self = this;
       var resultHTML = p.extras
         // Remove empty names
-        .filter(function(extra) {
-          return extra.name && extra.name.trim();
-        })
+        .filter(extra => extra.name && extra.name.trim())
         // Convert to HTML
-        .map(function(extra) {
+        .map(extra => {
           var span = document.createElement("span");
           var textContainer = span;
           if (extra.class) {
@@ -188,7 +185,7 @@ hb.registerHelper("showPeople", function(name, items = []) {
             span.appendChild(a);
             a.href = extra.href;
             textContainer = a;
-            if (self.doRDFa) {
+            if (this.doRDFa) {
               a.setAttribute("property", "rdfs:seeAlso");
             }
           }
--- a/src/w3c/headers.js
+++ b/src/w3c/headers.js
@@ -112,29 +112,6 @@ const W3CDate = new Intl.DateTimeFormat(["en-AU"], {
 hb.registerHelper("showPeople", function(name, items = []) {
   const html = (...args) => hyperHTML.wire()(...args);
   // stuff to handle RDFa
-  var re = "",
-    rp = "",
-    rm = "",
-    rn = "",
-    rwu = "",
-    rpu = "",
-    bn = "",
-    editorid = "",
-    propSeeAlso = "";
-  if (this.doRDFa) {
-    if (name === "Editor") {
-      bn = "_:editor0";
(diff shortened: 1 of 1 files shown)


saschanaz:
I still have a concern about the comment nodes hyperHTML implicitly creates (https://github.com/WebReflection/hyperHTML/issues/194), what do you think?

src/w3c/headers.js
```diff
@@ -109,137 +110,96 @@ const W3CDate = new Intl.DateTimeFormat(["en-AU"], {
...
-        bn = i < n - 1 ? "_:editor" + (i + 1) : "rdf:nil";
-        re = " resource='" + bn + "'";
-      }
-    } else {
-      ret += "<dd class='p-author h-card vcard'" + editorid + ">";
-    }
+  const attr = Object.freeze(self.doRDFa ? {
+    ...name === "Editor" ? {
+      rp1: "rdf:first",
+      rp2: "foaf:Person"
```
marcoscaceres: Based on the discussion in the related bug, it seems safe to remove this stuff.

src/w3c/headers.js
```diff
@@ -109,137 +110,96 @@ const W3CDate = new Intl.DateTimeFormat(["en-AU"], {
...
+    } : {},
+    rn: "foaf:name",
+    rm: "foaf:mbox",
+    rwu: "foaf:workplaceHomepage",
+    rpu: "foaf:homepage",
+    propSeeAlso: "rdfs:seeAlso"
+  } : {});
+  let ret = "";
+  const bns = [];
+  if (self.doRDFa && name === "Editor") {
```
marcoscaceres: As above.

src/w3c/headers.js
```diff
@@ -109,137 +110,96 @@ const W3CDate = new Intl.DateTimeFormat(["en-AU"], {
...
-          p.name +
-          "</a>";
-      } else
-        ret +=
-          "<a class='u-url url p-name fn' href='" +
-          p.url +
-          "'>" +
-          p.name +
-          "</a>";
+      if (self.doRDFa) {
```
marcoscaceres: Same here

src/w3c/headers.js
```diff
@@ -109,137 +110,96 @@ const W3CDate = new Intl.DateTimeFormat(["en-AU"], {
...
+  }
+
+  function getExtra(extra) {
+    const span = html`<span class='${extra.class || null}'></span>`
+    let textContainer = span;
+    if (extra.href) {
+      textContainer = html`<a
+        href='${extra.href || null}'
+        property='${self.doRDFa ? "rdfs:seeAlso" : null}'
+      ></a>`;
```
marcoscaceres: Here too. Let’s get rid of all the property attributes.

src/w3c/headers.js
saschanaz: Oh, was that the conclusion? I thought https://github.com/w3c/respec/issues/1503#issuecomment-365561010 is saying that Google actively uses this.

marcoscaceres: Yes and no :) Consider:

> "Whether the specific vocabulary and data patterns you're publishing map to a current public facing question is another matter."

Ideally, we should only include things that *do map* to "rich snippets". None of these seem to show up in any "rich snippets", so IMO they should go.

marcoscaceres: I made a proposal in the bug: when refactoring code, if we find stuff that doesn't show up as a rich snippet, it should be removed from ReSpec (as it has little, to no, value). However, we should keep anything that does show up.

saschanaz:
This was included in #1514 and now merged. We can discuss more in #1503 👍

PR #1510, opened 2018-02-15
WIP - rewrite of linking and WebIDL

marcoscaceres:
Incremental things that could be taken from this:

 * [off main thread IDL parsing](https://github.com/w3c/respec/pull/1510/files#diff-2d8d7c73d5131300bf7e13d4f42d9bb5)
 * [automatic pluralization of defined words](https://github.com/w3c/respec/pull/1510/files#diff-47300e7ecd8989b6246221de88fc9a3cR8). 
* [Symbols list](https://github.com/w3c/respec/pull/1510/files#diff-c47894de9d07998f30a863548f575207) 
* [Initial structure for validator](https://github.com/w3c/respec/pull/1510/files#diff-40ef9727ceebcaf7fb2d29e7606b4b4d)
* [Well-known cross references](https://github.com/w3c/respec/pull/1510/files#diff-bf690d7137b4c9600290f74c1538f83e) 
* Rewrite of WebIDL and linking code.


marcoscaceres:
@saschanaz ... this is basically what I've been hacking on... it's not fully function yet, but wanted to share it with you so we can discuss where I want to go with the WebIDL support. This has lots of parts, which can be split into separate tasks.

marcoscaceres:
Added description and links to various things above.

marcoscaceres:
Sorry, it's a ton of code, I know... so hopefully don't need to throw it away, as I spend quite a bit of time on it. I'm hopeful that by us picking bits, they can be incrementally included.

saschanaz:
Thank you for sharing your thoughts! It may take some time 😵 but I definitely will read and discuss about it 😁

worker/respec-worker.js
```diff
@@ -10,20 +10,45 @@ try {
   console.error("Network error loading/configuring highlighter", err);
 }
 
+try {
+  importScripts("https://www.w3.org/Tools/respec/webidl2.js");
```
saschanaz: I think this may cause test failure when upgrading webidl2.js. Not sure what would be the best way though...

src/w3c/abstract.js
```diff
@@ -4,10 +4,10 @@ import { pub } from "core/pubsubhub";
 import { l10n, lang } from "core/l10n";
 export const name = "w3c/abstract";
 
-export async function run(conf) {
+export async function run() {
   const abs = document.getElementById("abstract");
   if (!abs) {
-    pub("error", `Document must have one element with \`id="abstract"`);
+    pub("error", "Document must have one element with `id=\"abstract\".");
```
saschanaz: It looks like there is a nit in the master branch. `` `id="abstract" `` vs `` `id="abstract"` ``

worker/respec-worker.js
marcoscaceres: ah, true. We might need to signal debug mode somehow.

saschanaz: webpack has a solution specifically for workers, maybe we can do a similar thing. https://github.com/webpack-contrib/worker-loader

src/core/utils.js
```diff
@@ -18,6 +19,158 @@ const dashes = /\-/g;
...
+    id = "generated-id";
+  }
+  if (id.endsWith(".") || !/^[a-z]/i.test(id)) {
+    id = "x" + id;
+  }
+  if (elem.ownerDocument.getElementById(id)) {
+    let i = 0;
+    let nextId = id + "-" + i;
+    while (elem.ownerDocument.getElementById(nextId)) {
+      nextId = id + "-" + i++;
```
saschanaz: `++i`?

marcoscaceres:
@saschanaz, sorry, I didn't mean for you to review this in any detail. I just wanted to put up the stuff I'd been hacking on in the background, so we could pick and choose things to potentially include (totally ok with throwing away stuff too). 

I'll leave this branch, but split off the things I listed above into different project. Then I can send the parts for each and we can decide if they are worth keeping or not. 

About the pluralization - automatic pluralization sounds great.... as I imagine it doesn't work for some cases, we just need a means for an author to explicitly disable it (globally or locally).

saschanaz:
>as I imagine it doesn't work for some cases

`data-no-pluralize` may make sense. BTW, what will happen when a document already defines `data-lt` for pluralization purpose?

marcoscaceres:
> BTW, what will happen when a document already defines data-lt for pluralization purpose?

We should just merge them into `data-lt`.

saschanaz:
I like auto-pluralization and partial removal of jquery-enhanced part. For IDL writer part, IMO it is mainly for replacing handlebars templates?

marcoscaceres:
> For IDL writer part, IMO it is mainly for replacing handlebars templates?

Yes, but also to get rid of the C++ style indentation, add constructor support, etc.

marcoscaceres:
Might be able to salvage some of this at some point.

PR #1511, opened 2018-02-15
refactor: small syntax changes and hyperHTML

saschanaz:
A test fails on Firefox because of jQuery issue: https://github.com/jquery/jquery/issues/3980

```
FAILED TESTS:
  W3C — Headers
    logos
      × adds logos defined by configuration
        Firefox 58.0.0 (Windows 10.0.0)
      Expected 'data:image/svg+xml,<svg%20xmlns="http://www.w3.org/2000/svg"></svg%20xmlns="http:>' to equal 'data:image/svg+xml,<svg%20xmlns="http://www.w3.org/2000/svg"/>'.
      <Jasmine>
      @tests/spec/w3c/headers-spec.js:1046:9
```

I can replace the SVG data URI example (added on #1492)  but instead removed jQuery 😁


src/w3c/headers.js
```diff
@@ -663,17 +663,15 @@ export function run(conf, doc, cb) {
...
-  // insert into document and mark with microformat
-  var bp;
-  if (conf.isCGBG) bp = cgbgHeadersTmpl(conf);
-  else bp = headersTmpl(conf);
-  $("body", doc)
-    .prepend($(bp))
-    .addClass("h-entry");
+  // insert into document
+  const bp = hyperHTML.bind(document.createDocumentFragment())`${
+    [(conf.isCGBG ? cgbgHeadersTmpl : headersTmpl)(conf)]}`;
```
marcoscaceres: I think you may want a `.wire(obj)` here. On my phone reviewing, so kindly please check the hyperHTML docs.

saschanaz: `.wire(obj)` here returns an hyperHTML-specific object that contains a `_hyper` comment. I really don't like that behavior, but anyway this is a workaround to get a single node.

Otherwise we can do this:

```js
const bp = hyperHTML`${...}`; // hyperHTML can be used as an alias for hyperHTML.wire()
document.body.insertBefore(bp.childNodes[0], document.body.firstChild);
```

This may break if we choose to insert multiple top elements to the header template by whatever reason, but still works.

marcoscaceres: Can you help me understand your concerns about the comment nodes? Are you worried they will end up in the final spec?

saschanaz: >Are you worried they will end up in the final spec?

Partially yes, as my experiment shows a lot of resulting comments (one for every if-branch) when porting the full header templates.

But in this specific PR, I just don't like the result of ``hyperHTML`${[...]}` `` is not an HTMLElement just because of that additional comment node.

saschanaz:
#1514 made `headersTmpl(conf)` return HTMLDivElement so we don't need `bind()` nor `wire()` anymore in this PR 👍

src/core/jquery-enhanced.js
```diff
@@ -63,11 +63,9 @@ window.$.fn.getDfnTitles = function(args) {
     // prefer @data-lt for the list of title aliases
     titleString = this.attr(theAttr).toLowerCase();
     if (normalizedText !== "") {
-      //Regex: starts with the "normalizedText|"
-      var startsWith = new RegExp("^" + normalizedText + "\\|");
       // Use the definition itself, so to avoid
       // having to declare the definition twice.
-      if (!startsWith.test(titleString)) {
+      if (!titleString.startsWith(`${normalizedText}|`)) {
```
marcoscaceres: much nicer

src/w3c/headers.js
```diff
@@ -501,17 +501,16 @@ export function run(conf, doc, cb) {
...
-  // insert into document and mark with microformat
-  var bp;
-  if (conf.isCGBG) bp = cgbgHeadersTmpl(conf);
-  else bp = headersTmpl(conf);
-  $("body", doc)
-    .prepend($(bp))
-    .addClass("h-entry");
+  // insert into document
+  document.body.insertBefore(
+    (conf.isCGBG ? cgbgHeadersTmpl : headersTmpl)(conf),
```
marcoscaceres: A lot is happening here, might be a a little easier on the eyes/easier to debug if things go bad if:

```JS
const tmpl = conf.isCGBG ? cgbgHeadersTmpl : headersTmpl;
const elem = tmpl(conf);
document.body.insertBefore(elem, document.body.firstChild) 
```

`elem` should probably be named whatever comes out of template? you said it was a div, right?

src/w3c/headers.js
```diff
@@ -501,17 +501,16 @@ export function run(conf, doc, cb) {
...
+    document.body.firstChild
+  );
+  document.body.classList.add("h-entry");
```
marcoscaceres: Are you sure this is correct? In the previous code, I looks like it was adding the "h-entry" to the "bp", instead of the body.

marcoscaceres: fixed typo, "it looks like"

saschanaz: At first I thought so, but try this:

```js
$("<div>").prepend($("<a>"))[0];
// returns <div>
```

`.prepend()` returns the container instead of the child.

PR #1514, opened 2018-02-16
Convert headers/sotd templates into hyperHTML

saschanaz:
@marcoscaceres 

This is my experiment to migrate to hyperHTML, and this is my concern:

image


marcoscaceres:
Ok, for the comments, we need to make sure those get stripped out on save. They will be quite easy to find and eliminate.

src/w3c/templates/headers.js
```diff
@@ -0,0 +1,159 @@
...
+      `}
+    ` : ""}
+    <dt>${conf.multipleEditors ? html`${conf.l10n.editors}` : html`${conf.l10n.editor}`}</dt>
+    ${showPeople(conf, "Editor", conf.editors)}
+    ${conf.authors ? html`
+      <dt>${conf.multipleAuthors ? [conf.l10n.authors] : [conf.l10n.author]}</dt>
+      ${showPeople(conf, "Author", conf.authors)}
+    ` : ""}
+    ${conf.otherLinks ? html`
+      ${conf.otherLinks.map(link => html`
```
marcoscaceres: This one seems a little hard to work with. Might be worth slitting into its own function?

marcoscaceres:
BTW, this is where we would remove those hyperHTML nodes:
https://github.com/w3c/respec/blob/develop/src/ui/save-html.js#L83

We can do that in a separate bug.

src/ui/save-html.js
```diff
@@ -112,6 +113,21 @@ function cleanup(cloneDoc) {
   pub("beforesave", documentElement);
 }
 
+function cleanupComments(node) {
+  // collect first, or walker will cease too early
+  const comments = [...walkTree(node, NodeFilter.SHOW_COMMENT)];
+  for (const comment of comments) {
+    comment.parentNode.removeChild(comment);
```
marcoscaceres: only the hyperHTML comments, no? Sometimes Editors add random comments that they don't mind having on TR.

PR #1515, opened 2018-02-18
Update RDFa support

gkellogg:
to simplify and use schema.org properties and classes, as much as possible.


marcoscaceres:
Thanks @gkellogg, will review over next few days. I'm AFK for the next few days 🛫. 

Could you remove the changes to:

* builds/respec-w3c-common.build.js.map
* builds/respec-w3c-common.build.js

We only build on release. We probably shouldn't need the additional Babel plugin either, I think. @saschanaz, could you kindly check?

gkellogg:
Trying to run `npm run build` failed without the babel-plugin-transform-es2015-block-scoping being installed for me.

gkellogg:
Build passes for me: https://travis-ci.org/gkellogg/respec, not sure why it fails CI here, but there may be a babel issue

src/w3c/templates/show-people.js
```diff
@@ -1,60 +1,63 @@
...
     if (p.mailto) {
-      contents.push(html`, <span class='ed_mailto'><a class='u-email email'
-        property='${attr.rm}' href='${`mailto:${p.mailto}`}'>${p.mailto}</a></span>`);
+      if (conf.doRDFa) {
+        contents.push(html`, <span class='ed_mailto'><a class='u-email email'
+          property='foaf:mbox' href='${`mailto:${p.mailto}`}'>${p.mailto}</a></span>`);
+      } else {
+        contents.push(html`, <span class='ed_mailto'><a class='u-email email'
+          href='${`mailto:${p.mailto}`}'>${p.mailto}</a></span>`);
+      }
```
saschanaz: Same here

src/w3c/templates/show-people.js
```diff
@@ -1,60 +1,63 @@
...
     const contents = [];
     if (p.url) {
       if (conf.doRDFa) {
-        contents.push(html`<meta property='${attr.rn}' content='${p.name}' />`);
+        contents.push(html`<meta property='schema:name' content='${p.name}' />
+          <a class='u-url url p-name fn'
+          property='schema:url' href='${p.url}'>${p.name}</a>`);
+      } else {
+        contents.push(html`<a class='u-url url p-name fn'
+          href='${p.url}'>${p.name}</a>`);
```
saschanaz: This additional branch is unneeded if we keep `attr`.

src/w3c/templates/show-people.js
```diff
@@ -1,60 +1,63 @@
...
-        contents.push(html`, <a property='${attr.rwu}'
-          class='p-org org h-org h-card' href='${p.companyURL}'>${p.company}</a>`);
+        if (conf.doRDFa) {
+          contents.push(html`, <span property='schema:worksFor' typeof='schema:Organization'>
+            <meta property='schema:name' content='${p.company}' />
+            <a property='schema:url'
+            class='p-org org h-org h-card' href='${p.companyURL}'>${p.company}</a>
+            </span>`);
+        } else {
+          contents.push(html`, <a class='p-org org h-org h-card' href='${p.companyURL}'>${p.company}</a>`);
```
saschanaz: We can perhaps use `<a property='${attr.url}' ... />` outside of if-branch to reduce lines.

src/w3c/templates/show-people.js
```diff
@@ -1,60 +1,63 @@
...
-    } : name === "Author" ? {
-      rp1: "dc:contributor",
-      rp2: "foaf:Person"
-    } : {},
-    rn: "foaf:name",
-    rm: "foaf:mbox",
-    rwu: "foaf:workplaceHomepage",
-    rpu: "foaf:homepage",
-    propSeeAlso: "rdfs:seeAlso"
-  } : {});
```
saschanaz: I would keep `attr` here. It makes `attr.rp`, for example, to be `undefined` when `conf.deRDFa` is falsey, and that helps hyperHTML by explicitly asking attribute removal:

```js
html`<span property=${undefined} />`
```

This is exactly same as:

```js
html`<span />` // .hasAttribute("property") will return false for both examples
```

This enables us to keep if-statements minimal.

saschanaz:
This PR completely removes `builds/` directory, instead it should make no diff for that one.

src/w3c/templates/show-people.js
```diff
@@ -1,60 +1,63 @@
...
-      contents.push(html`<span property='${attr.rn}' class='p-name fn'>${p.name}</span>`);
+      contents.push(html`<span class='p-name fn'>${p.name}</span>`);
     }
     if (p.company) {
       if (p.companyURL) {
-        contents.push(html`, <a property='${attr.rwu}'
-          class='p-org org h-org h-card' href='${p.companyURL}'>${p.company}</a>`);
+        if (conf.doRDFa) {
+          contents.push(html`, <span property='schema:worksFor' typeof='schema:Organization'>
+            <meta property='schema:name' content='${p.company}' />
```
marcoscaceres: nit: we no longer need `" />"` on elements. The XML serializer auto-closes this when exporting.

src/w3c/templates/show-people.js
marcoscaceres: We can probably get rid of the `span` too, right? just move the `ed_mailto` into the `<a>` - same above.

src/w3c/templates/show-people.js
gkellogg: Good to know, I'll go back to the old way.

src/w3c/templates/show-people.js
```diff
@@ -2,55 +2,54 @@ export default (conf, name, items = []) => {
...
-        contents.push(html`, <a property='${attr.rwu}'
-          class='p-org org h-org h-card' href='${p.companyURL}'>${p.company}</a>`);
+        if (conf.doRDFa) {
+          contents.push(html`, <span property='schema:worksFor' typeof='schema:Organization'>
+            <meta property='${attr.rn}' content='${p.company}'>
+            <a property='${attr.ru}'
+            class='p-org org h-org h-card' href='${p.companyURL}'>${p.company}</a>
+            </span>`);
+        } else {
+          contents.push(html`, <a class='p-org org h-org h-card' href='${p.companyURL}'>${p.company}</a>`);
```
saschanaz: We can remove an else branch here as we use `attr`.

src/w3c/headers.js
```diff
@@ -492,13 +492,13 @@ export function run(conf, doc, cb) {
...
+      $("html").attr("typeof", "schema:TechArticle " + conf.rdfStatus);
+    else $("html").attr("typeof", "schema:TechArticle ");
     var prefixes =
-      "bibo: http://purl.org/ontology/bibo/ w3p: http://www.w3.org/2001/02pd/rec54#";
+      "w3p: http://www.w3.org/2001/02pd/rec54#";
     $("html").attr("prefix", prefixes);
     $("html>head").prepend(
-      $("<meta lang='' property='dc:language' content='en'>")
+      $("<meta lang='' property='schema:inLanguage' content='en'>")
     );
```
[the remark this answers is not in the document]
gkellogg: The JSON-LD implementation uses `doc.documentElement.lang || "en"` The root element should do the same. However, setting the `schema:inLanguage` property is metadata.

src/w3c/permalinks.js
```diff
@@ -40,24 +40,22 @@ export function run(conf, doc, cb) {
         }
       }
       // if we still have resourceID
-      if (resourceID !== null) {
+      if (resourceID !== undefined && resourceID !== null) {
```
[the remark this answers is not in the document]
gkellogg: Sure, it was a quick fix.

gkellogg:
If we're going to simply remove RDFa, we could probably not pull, but I think it would be good to have in the commit chain, and then follow up with another PR that removes RDFa entirely.

I'll do some cleanup here, and prepare another PR for removing RDFa and thereby cleanup the code base. Note that there are some uses of `property` and `resource` that are not controlled using `doRDFa` which should go too.

marcoscaceres:
@gkellogg agree. I'd be ok to commit this also, and then work on removing later. We can run both this and the JSON-LD side-by-side for a while too.

src/w3c/headers.js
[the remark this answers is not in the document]
marcoscaceres: ah, go it.

src/w3c/templates/show-people.js
gkellogg: I don't think this can be cleaned up any further, without leaving an empty span surrounding the anchor.

marcoscaceres: @saschanaz, if you can see a more optimal solution, maybe you can do the changes on-top of this PR? I can live with what is currently here, as there is a high probability we are going to remove all this.

saschanaz:
Great, I’m okay with it then 👍

Sent from my Windows 10 phone

Issue #1084, opened 2017-02-16
Make it possible to test errors and warning


marcoscaceres:
There is currently no way to hook into the errors and warnings system. Need a means to do that that for testing the linter properly, for instance. A bunch of other modules emit warnings, and it would be good to test those too.

devanshbatra04:
Been thinking over this. 
We could have some debug functions/test functions. They would be very specific to the critical code paths that are most likely to yield an error/warning. 

And then when the user clicks the button to list the warning/error, add a button "Test With Custom Input". 
On clicking it, we could have a modal pop up with a description specific to each such function. 

Should I come up with a mockup?

marcoscaceres:
Can probably avoid the clicking part... I was thinking we could just have callbacks on the configuration objects in the tests... so:

```JS
const ops = {}

const errorPromise = new Promise(r => {
  ops.onError(error){
    expect(error).toBe("whatever");
    r();
  },
});
const warningPromise = new Promise(r => {
  ops.onWarning(error){
    expect(error).toBe("whatever");
    r();
  },
});
const doc = await makeRSDoc(ops);
await Promise.all([errorPromise, warningPromise])

```

marcoscaceres:
Fixed some typos above...

devanshbatra04:
Oh okay. Thanks. 
I guess I misunderstood. Will work on this today.

marcoscaceres:
No problem - you were definitely on the right track :) Eventually we might want to do UI testing.

devanshbatra04:
So I was working on this and figured this was probably not the best approach.
1. This does not give us all the errors and warning to work with. That means we wouldn't be able to test them in conjunction. (One warning/error may be related with another but we have no way of interacting with the two. As both are dealt with by different promises)
2. Also there would be several unnecessary errors. Most of the test documents made by makeRSdoc would not have the SoTD for instance. Resulting in an error.

Here's a solution I came up with. 
create conf flags. collectErrors and collectWarnings. 
This gets automatically applied during tests. 
Add a conditional at ```sub("error")``` and ```sub("warn")``` to collect errors and warnings in globally available Sets. 

During Tests the tester can then 
```javascript
expect(collectedErrors.has("some-error")).toBeTruthy();
expect(collectedErrors.has("some-other-error")).toBeFalsy();
// (and possibly, although I would avoid)
expect(collectedErrors.size).toEqual(3);
```

This way he doesn't have to care about all the other errors popping up and handling them.
[...]

PR #1516, opened 2018-02-18
fix(w3c/templates/headers): use <h1> exists for title (fixes #460)

### LINKED ISSUE #460: problems generating h1 from title
http://w3c.github.io/clreq/ (and its TR equivalen) needs to add markup to the `h1` to apply language information, (a) so that chinese fonts are used for the ideographs, and (b) to enable the language in the title to be filtered by language.

since the `title` element takes only plain text, generating the `h1` from it, as respec does, is problematic.

i can think of other scenarios where you might possibly want to style the `h1` text, for just one example. see the `h1` for http://www.w3.org/International/questions/qa-navigation-select

can someone please de-couple this arrangement if there is an `h1` defined in the document already.

shikhar-scs:
> Note that the title of the document is reused as the title of the specification in the resulting document's h1. It's a small win, but that's always something less to repeat.

mentioned here, https://github.com/w3c/respec/wiki/ReSpec-Editor's-Guide#the-very-basics .The same was to be rectified.

Suggested by @marcoscaceres, I've made some changes and pushed them here. Please review.
@saschanaz 

P.S. I'm clueless about the build errors :cry:


src/w3c/templates/headers.js
```diff
@@ -7,7 +7,12 @@ export default conf => {
   const html = hyperHTML;
   return html`<div class='head'>
   ${conf.logos.map(showLogo)}
-  <h1 class='title p-name' id='title' property='${conf.doRDFa ? "dcterms:title" : null}'>${conf.title}</h1>
+  <script>
+    if (document.querySelectorAll("h1") ?
+      document.write("") :
```
marcoscaceres: So, here we want to get just the first, so `const h1 = querySelector(“body>h1”)`.

marcoscaceres: We want to avoid document.write (we are actually trying to remove it from the web entirely). It’s not allowed in many security contexts and has a huge performance penalty. 

Instead, use either the found h1, or just use the current code.

marcoscaceres: Will provide you with some code suggestions once I’m back in front of a computer.

shikhar-scs: > Will provide you with some code suggestions once I’m back in front of a computer

Fine I'll wait. 👍 Also, please check  Slack whenever you are free

shikhar-scs:
@marcoscaceres any updates ?

marcoscaceres:
No, sorry... super busy still and about to jump on another plane. But should be as simple as:

```js
const h1 = querySelector(“h1#title, body>h1”);
if(h1){
   conf.specTile = h1.textContent.trim();
   // set attributes if required to do so...  
   h1.classList.add('title') 
   h1.id = "title";
} else {
  // whatever is currently in ReSpec
}
```

marcoscaceres:
Note that we removed the `property` attribute, etc. so this will need a rebase.

shikhar-scs:
Thanks @marcoscaceres for finding time in this busy schedule. I have made the required changes so you can review them whenever you are free :sweat_smile:

src/w3c/templates/headers.js
```diff
@@ -7,7 +7,20 @@ export default conf => {
   const html = hyperHTML;
   return html`<div class='head'>
   ${conf.logos.map(showLogo)}
-  <h1 class='title p-name' id='title'>${conf.title}</h1>
+  <script>
```
marcoscaceres: Not quite right. You need to look at what the `html` tagged template string is doing 😉 

We definitely don’t want a `script` element here.

shikhar-scs:
@marcoscaceres Is this what you were expecting ?

BTW, I also tried 
```
${document.querySelector("h1#title, body>h1") ?
        const h1 = document.querySelector("h1#title, body>h1"),
        conf.specTile = h1.textContent.trim(),
        h1.classList.add("title"),
        h1.id = "title"
          :html`<h1 class="title p-name">${conf.title}</h1>`}

```

However, `const` didn't work with a ternary operator so I had to go with this. :sweat_smile: . Please suggest any more changes, if the need be. :smile:

marcoscaceres:
@shikhar-scs, sorry, I wasn't very clear in the code sample I sent the other day... I was trying to illustrate the logic, but not exactly the solution. Where the code goes in relation to the template string doesn't matter too much (like, it could be a function call that returns the `h1` element).  

Having a look now at your latest changes.

src/w3c/templates/headers.js
```diff
@@ -7,7 +7,13 @@ export default conf => {
   const html = hyperHTML;
   return html`<div class='head'>
   ${conf.logos.map(showLogo)}
-  <h1 class='title p-name' id='title'>${conf.title}</h1>
+  
+  ${document.querySelector("h1#title, body>h1") ? (
```
marcoscaceres: Above the `return html` line, you could have a check for `const specTitleElem = document.querySelector("h1#title, body>h1")` or just create a HTML element if it's missing... then:

```JS
const specTitleElem = document.querySelector("h1#title, body>h1") || document.createElement("h1") ;
if (specTitleElem.parentElement) { // remove it from the DOM tree, because we are going to relocate it
  specTitleElem.remove();
}
docorateSpecTitle(specTitleElem);
return html`<div class='head'>
    ${conf.logos.map(showLogo)}		    ${conf.logos.map(showLogo)}
    ${specTitleElem}
```

And `function docorateSpecTitle()` would check or add the `title` css class, `id` attribute, and whatever else is needed. 

The goal is never to repeat ourselves ([DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)) and our goal is always to make code as legible and efficient as possible... note your original code has "document.querySelector" 3x times for the same thing, which is not ideal - both because it's hard to read, and because it's very inefficient to ask the browser to perform the query.   

Makes sense? Again, I don't know if the above code I proposed works - it's probably buggy (thus we need test for it), but should be hopefully enough to go from.

shikhar-scs: This is an awesome review! I'll make the changes accordingly. And yeah, [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself).

shikhar-scs:
@marcoscaceres I've pushed in some changes. This ensures that if an `<h1>` tag already exists, then it is the one to be kept. Otherwise, a new one is generated and used.

src/w3c/templates/headers.js
```diff
@@ -43,12 +55,12 @@ export default conf => {
...
     ${conf.showPreviousVersion ? html`
       <dt>Previous version:</dt>
-      <dd><a href='${conf.prevVersion}'>${conf.prevVersion}</a></dd>
+      <dd><a rel='${conf.doRDFa ? "dcterms:replaces" : null}' href='${conf.prevVersion}'>${conf.prevVersion}</a></dd>
     ` : ""}
     ${conf.prevRecURI ? html`
       ${conf.isRec ? html`
           <dt>Previous Recommendation:</dt>
-          <dd><a href='${conf.prevRecURI}'>${conf.prevRecURI}</a></dd>
+          <dd><a rel='${conf.doRDFa ? "dcterms:replaces" : null}' href='${conf.prevRecURI}'>${conf.prevRecURI}</a></dd>
```
saschanaz: You may want to remove these `rel` and `property` things, as the main branch removed them. (See #1503)

src/w3c/templates/headers.js
```diff
@@ -5,13 +5,25 @@ import showLink from "./show-link";
...
+    specTitleElem.classList.add("title", "p-name");
+    specTitleElem.id = "title";
+  }
   return html`<div class='head'>
   ${conf.logos.map(showLogo)}
-  <h1 class='title p-name' id='title'>${conf.title}</h1>
+  ${specTitleElem}
   ${conf.subtitle ? html`
-    <h2 id='subtitle'>${conf.subtitle}</h2>
+    <h2 property='${conf.doRDFa ? "bibo:subtitle" : null}' id='subtitle'>${conf.subtitle}</h2>
```
marcoscaceres: nit: seems your rebase has re-introduced the `doRDFa` stuff... we recently removed that.

src/w3c/templates/headers.js
```diff
@@ -5,13 +5,25 @@ import showLink from "./show-link";
...
   ` : ""}
-  <h2>${conf.prependW3C ? `W3C ` : ""}${conf.textStatus} <time class='dt-published' datetime='${conf.dashDate}'>${conf.publishHumanDate}</time></h2>
+  <h2>${conf.prependW3C ? `W3C ` : ""}${conf.textStatus} <time property='${conf.doRDFa ? "dcterms:issued" : null}' class='dt-published' datetime='${conf.dashDate}'>${conf.publishHumanDate}</time></h2>
```
marcoscaceres: as above, and same below

src/w3c/templates/headers.js
```diff
@@ -5,13 +5,25 @@ import showLink from "./show-link";
...
   const html = hyperHTML;
+  const specTitleElem = document.querySelector("h1#title, body>h1") || document.createElement("h1") ;
+  if (specTitleElem.parentElement) { // remove it from the DOM tree, because we are going to relocate it
+    specTitleElem.remove();
+    specTitleElem.innerText.trim();
+  } else {
+    specTitleElem.innerText = conf.title;
+  }
+  decorateSpecTitle(specTitleElem);
+  function decorateSpecTitle(specTitleElem) {
```
marcoscaceres: nit: move this function outside this export block (above would be great).

src/w3c/templates/headers.js
```diff
@@ -5,13 +5,25 @@ import showLink from "./show-link";
 
 export default conf => {
   const html = hyperHTML;
+  const specTitleElem = document.querySelector("h1#title, body>h1") || document.createElement("h1") ;
+  if (specTitleElem.parentElement) { // remove it from the DOM tree, because we are going to relocate it
+    specTitleElem.remove();
+    specTitleElem.innerText.trim();
```
saschanaz: It should be `specTitleElem.innerText = specTitleElem.innerText.trim();` to be effective, but why should we do this? Also, I would prefer `textContent` which is a whole lot faster than `innerText`, [at least on some browsers](https://kellegous.com/j/2013/02/27/innertext-vs-textcontent/):

image

marcoscaceres: I think here we want:

```JS
conf.specTitle = specTitleElem.innerText.trim(); 
```

src/w3c/templates/headers.js
```diff
@@ -5,13 +5,25 @@ import showLink from "./show-link";
 
 export default conf => {
   const html = hyperHTML;
+  const specTitleElem = document.querySelector("h1#title, body>h1") || document.createElement("h1") ;
+  if (specTitleElem.parentElement) { // remove it from the DOM tree, because we are going to relocate it
```
marcoscaceres: We have a potential conflict here... if `conf.specTitle` is set, then `conf.specTitle` should win... and we should emit a an error about duplicate titles.

src/w3c/templates/headers.js
marcoscaceres: We definitely don't want to rewrite the content BTW. See the original bug, as the whole point of this is to retain the custom markup in the `h1`.

marcoscaceres: @saschanaz good suggestion about accessing `.textContent` btw.

saschanaz: >We definitely don't want to rewrite the content BTW.

Right, I just had no idea about what the line does.

marcoscaceres: I think @shikhar-scs had overlooked that :) 

@shikhar-scs, be mindful that calling `.trim()` doesn't actually trim the `.textContent` - it returns a trimmed *copy* of whatever is in `.textContent`.

marcoscaceres: Err.. I mean "innerText"... you know what I'm saying :)

shikhar-scs: Yes yes I get your point. Thankyou @marcoscaceres and @saschanaz for this detailed review. Ill get started with the changes soon :sweat_smile:

shikhar-scs:
@marcoscaceres & @saschanaz I've introduced new changes and it does what is intended. Here's a preview 

Look at the Line#20 in the editor. This is when the h1 exists
.
h1exists1

This is when it doesn't

h1exists2

src/w3c/templates/headers.js
```diff
@@ -3,11 +3,25 @@ import showLogo from "./show-logo";
...
 import showLink from "./show-link";
 
+function decorateSpecTitle(specTitleElem) {
+  specTitleElem.classList.add("title", "p-name");
+  specTitleElem.id = "title";
+}
+
 export default conf => {
   const html = hyperHTML;
+  const specTitleElem = document.querySelector("h1#title, body>h1") || document.createElement("h1");
```
marcoscaceres: We should actually just require `"h1#title"`.  That is, to use this, Editors should explicitly write: `<h1 id="title">Foo</h1> `. Other `h1` should be ignored.

src/w3c/templates/headers.js
```diff
@@ -3,11 +3,25 @@ import showLogo from "./show-logo";
 import showPeople from "./show-people";
 import showLink from "./show-link";
 
+function decorateSpecTitle(specTitleElem) {
+  specTitleElem.classList.add("title", "p-name");
+  specTitleElem.id = "title";
```
marcoscaceres: given suggestion below, this line can go away, and we can drop `decorateSpecTitle()`, but keep `specTitleElem.classList.add("title", "p-name");` as it's the only thing we need to do to the `h1`.

src/w3c/templates/headers.js
```diff
@@ -3,11 +3,25 @@ import showLogo from "./show-logo";
...
+}
+
 export default conf => {
   const html = hyperHTML;
+  const specTitleElem = document.querySelector("h1#title, body>h1") || document.createElement("h1");
+  if (specTitleElem.parentElement) {
+    // remove it from the DOM tree, because we are going to relocate it
+    specTitleElem.remove();
+    conf.specTitle = specTitleElem.innerText.trim();
```
marcoscaceres: this should be `conf.title = `

src/w3c/templates/headers.js
```diff
@@ -3,11 +3,25 @@ import showLogo from "./show-logo";
...
+  } else {
+    specTitleElem.innerText = conf.title;
+  }
+  decorateSpecTitle(specTitleElem);
   return html`<div class='head'>
   ${conf.logos.map(showLogo)}
-  <h1 class='title p-name' id='title'>${conf.title}</h1>
+  ${specTitleElem}
   ${conf.subtitle ? html`
     <h2 id='subtitle'>${conf.subtitle}</h2>
```
marcoscaceres: We should support the same behaviour for `conf.subTitle` too - we just match on "h2#subtitle".

marcoscaceres:
This is starting to take shape! @shikhar-scs, start thinking about tests too.

src/w3c/templates/headers.js
```diff
@@ -3,11 +3,25 @@ import showLogo from "./show-logo";
...
+}
+
 export default conf => {
   const html = hyperHTML;
+  const specTitleElem = document.querySelector("h1#title, body>h1") || document.createElement("h1");
+  if (specTitleElem.parentElement) {
+    // remove it from the DOM tree, because we are going to relocate it
+    specTitleElem.remove();
+    conf.specTitle = specTitleElem.innerText.trim();
+  } else {
```
marcoscaceres: We need one more check here:
```js
if(document.title !== conf.title) {
   pub("warn", "The document's title and the `<title>` element differ.") 
}
```

shikhar-scs: Done! What all should I check in testing. That if `h1` exists what all should be in place and vice versa ?

shikhar-scs: Also how do I perform a test on the `title` ? Aren't the tests in `tests/spec/w3c/headers-spec.js` mean't for the specific code block of 
```html
<section id='sotd'><p>foo</p></section><section id='toc'></section>
```
Shall I make a different one ?

marcoscaceres: > Done! What all should I check in testing. That if h1 exists what all should be in place and vice versa ?

 Yes, exactly. Also, check that `conf.title` is correctly set.

marcoscaceres: > Also how do I perform a test on the `title` ? 

Unfortunately, we won't be able to check that one :( We have a bug for that #1084.

shikhar-scs:
@marcoscaceres I've made changes for this and also introduced the changes for `subtitle`. Please have a look. 
I'll start writing tests in a while. I have a rough idea on how to do the same :+1: . Suggest changes please.

marcoscaceres:
No problem, will fix the formatting before we merge.

marcoscaceres:
Seems the fails being reported by Travis are legit. This is causing regressions in other modules - so we need to look into those.

shikhar-scs:
I'll get back to this in a week's time. Exams are holding me up :sweat_smile:

marcoscaceres:
No problem! Good luck with your exams @shikhar-scs!!! 🤓📖

src/w3c/templates/headers.js
```diff
@@ -2,16 +2,40 @@ import "deps/hyperhtml";
...
+    if (document.title !== conf.title) {
+      pub("warn", "The document's title and the `<title>` element differ.");
+    }
+    specTitleElem.textContent = conf.title;
+    specTitleElem.id = "title";
+  }
+  specTitleElem.classList.add("title", "p-name");
+
+  const specSubTitleElem =
+    document.querySelector("h2#subtitle") || document.createElement("h2");
```
sidvishnoi: You might want to add create a `h2` for subtitle only if `conf.subtitle` is given. 

Saying this if we need the same test conditions to hold for `"handles missing subtitle"` in https://github.com/w3c/respec/blob/9ae571a97d9162316c011c4946b1b16a205d23ef/tests/spec/w3c/headers-spec.js#L224-L233

shikhar-scs: Nice suggestion. Though I guess now we can do away with the keeping the subtitle optional as we are anyways adding the subtitle element by my code. whatsay @marcoscaceres  ?

sidvishnoi: If `conf.subtitle` is not specified and neither `h2#subtitle` is created by user, it creates an empty `<h2 id="subtitle"></h2>`. Might need to look into that too (although an empty `h2#subtitle` doesn't hurt)

marcoscaceres: Prefer we don’t have empty elements, as they risk getting styled.

shikhar-scs: So I should only include a `subTitleElement` if it exists in `conf.subTitle` ?

marcoscaceres: Yep, exactly.

shikhar-scs: ohk fine :+1:

shikhar-scs:
@marcoscaceres I guess this is ready to merge.

P.S. if you could check slack :sweat_smile:

marcoscaceres:
> @marcoscaceres I guess this is ready to merge. All tests pass

The code is still not right tho. I'll review in there.

src/w3c/templates/headers.js
```diff
@@ -2,16 +2,43 @@ import "deps/hyperhtml";
...
+  }  
   return html`<div class='head'>
   ${conf.logos.map(showLogo)}
-  <h1 class='title p-name' id='title'>${conf.title}</h1>
-  ${conf.subtitle ? html`
-    <h2 id='subtitle'>${conf.subtitle}</h2>
-  ` : ""}
-  <h2>${conf.prependW3C ? `W3C ` : ""}${conf.textStatus} <time class='dt-published' datetime='${conf.dashDate}'>${conf.publishHumanDate}</time></h2>
+  ${specTitleElem}
+  ${conf.subtitle ? specSubTitleElem : ""}
```
marcoscaceres: Here, don't we just want:

```JS
${specSubTitleElem}
```

If it's null, won't it just not get rendered?

tests/spec/w3c/headers-spec.js
```diff
@@ -221,6 +221,25 @@ describe("W3C — Headers", function() {
     });
   });
 
+  describe("title", () => {
+    it("uses <h1> if already present", async () => {
+      const ops = makeStandardOps();
+      ops.body = "<h1 id='title'>Hello <code>World</code></h1>" + makeDefaultBody();
+      const doc = await makeRSDoc(ops);
+
+      const titleInHead = doc.querySelectorAll(".head .title.p-name");
```
marcoscaceres: why not query `"head h1"`? And then:

```
expect(titleInHead.classList.has("p-name")).toBe(true);
expect(titleInHead.id).toEqual("title");
```

tests/spec/w3c/headers-spec.js
```diff
@@ -221,6 +221,25 @@ describe("W3C — Headers", function() {
...
+      expect(titleInHead.length).toEqual(1);
+
+      const title = doc.querySelectorAll(".title.p-name");
```
marcoscaceres: Same here... might be good to clarify exactly what's being tested here. .

tests/spec/w3c/headers-spec.js
```diff
@@ -221,6 +221,25 @@ describe("W3C — Headers", function() {
...
+      expect(title.length).toEqual(1);
+
+      const unescapedHTML = doc.querySelectorAll("#title code");
```
marcoscaceres: What we probably want here is to test that the first child is code, and the textContent is "World"... though we should change it to "pass"

tests/spec/w3c/headers-spec.js
```diff
@@ -221,6 +221,25 @@ describe("W3C — Headers", function() {
...
+      expect(unescapedHTML.length).toEqual(1);
+
+      expect(doc.getElementById("title").innerText).toEqual("Hello World");
```
marcoscaceres: This is doing the same work we did above.

tests/spec/w3c/headers-spec.js
```diff
@@ -231,6 +250,19 @@ describe("W3C — Headers", function() {
       const doc = await makeRSDoc(ops);
       expect($("#subtitle", doc).length).toEqual(0);
     });
+    
+    it("uses existing h2#subtitle as subtitle", async () => {
+      const ops = makeStandardOps();
+      ops.body = "<h2 id='subtitle'>This is a pre existing subtitle</h2>" + makeDefaultBody();
```
marcoscaceres: Make sure you add another element (e.g., `<code>pass</code>`) inside the `h2`, as that's the purpose of the change.

shikhar-scs: Sure. The other changes belong to @sidvishnoi I guess. So I hope he would correct them :+1:

sidvishnoi: @shikhar-scs on it.

tests/spec/w3c/headers-spec.js
sidvishnoi:
``` js
const title = doc.querySelectorAll(".title.p-name");
expect(title.length).toEqual(1); // original h1#title is removed
```
not sure how can I make this clearer. any help?

sidvishnoi: would this one do? https://gist.github.com/sidvishnoi/e60fa8b991c163339e38f6293458a310

shikhar-scs:
Made the changes.

tests/spec/w3c/headers-spec.js
marcoscaceres: Yep, that one is good.

shikhar-scs:
@marcoscaceres if you could see for final :sweat_smile:

tests/spec/w3c/headers-spec.js
```diff
@@ -231,6 +250,22 @@ describe("W3C — Headers", function() {
...
+      const ops = makeStandardOps();
+      ops.body = "<h2 id='subtitle'>This is a pre existing subtitle <code>pass</code></h2>" + makeDefaultBody();
+      
+      const doc = await makeRSDoc(ops);
+
+      const subTitleElement = doc.querySelectorAll("h2#subtitle");
+      expect(subTitleElement.length).toEqual(1);
+      
+      const [forInnerText] = subTitleElement;
+      expect(forInnerText.textContent).toEqual("This is a pre existing subtitle");
```
marcoscaceres: Travis is falling this because "pass" is missing:

> Expected 'This is a pre existing subtitle pass' to equal 'This is a pre existing subtitle'.

tests/spec/w3c/headers-spec.js
```diff
@@ -221,6 +221,25 @@ describe("W3C — Headers", function() {
     });
   });
 
+  describe("title", () => {
+    it("uses <h1> if already present", async () => {
+      const ops = makeStandardOps();
```
marcoscaceres: I think we are missing a test checking the effect on `conf.title`. 

```JS
const { title } = doc.defaultView.respecConfig; 
expect(title).toEqual(...);
```

tests/spec/w3c/headers-spec.js
```diff
@@ -231,6 +250,22 @@ describe("W3C — Headers", function() {
       const doc = await makeRSDoc(ops);
       expect($("#subtitle", doc).length).toEqual(0);
     });
+    
+    it("uses existing h2#subtitle as subtitle", async () => {
+      const ops = makeStandardOps();
```
marcoscaceres: As above, we are missing a test to check `conf.subtitle`.

tests/spec/w3c/headers-spec.js
marcoscaceres: note that here we want to be sure we hit both code paths for the if statement. So, we might actually need two independent tests: 

1. one with the `respecConfig.title` and `h1`
1. one without the `respecConfig.title` but with `h1`

Same applies to `subtitle`.

shikhar-scs: Even I was thinkning about the same. But in the case of `subtitle` we do have a test for a missing `subtitle` and moreover the subtitle element is not mandatorily present on the doc. So, do we still need it in the case of `subtitile` ?

marcoscaceres: > ...not mandatorily present on the doc. So, do we still need it in the case of subtitile ?

So, mostly we are interested in making sure we walk all code paths, and think of weird edge cases. Although we'd like to assume things about what will be common usage, things break when users do things we don't expect.

marcoscaceres: To be clear, I'm particularly interested in this being tested:
https://github.com/w3c/respec/pull/1516/files#diff-ebb1a497c097951bb868645bd73ade4fR29

sidvishnoi: > we want to be sure we hit both code paths for the if statement

@marcoscaceres you mean one with `h1` and one without `h1`? and not (w/ and w/o `respecConfig.title`)
so that our tests will be "*uses `respecConfig.title` if `<h1>` not present*" and "*uses `<h1>` for title if already present*"?

src/w3c/templates/headers.js
```diff
@@ -2,16 +2,42 @@ import "deps/hyperhtml";
...
+    if (document.title !== conf.title) {
+      pub("warn", "The document's title and the `<title>` element differ.");
+    }
+    specTitleElem.textContent = conf.title;
+    specTitleElem.id = "title";
+  }
+  specTitleElem.classList.add("title", "p-name");
+
+  let specSubTitleElem = null;
+  if (conf.subtitle) {
```
sidvishnoi: @shikhar-scs As you are using condition `if (conf.subtitle)`, if `conf.subtitle` is not set, then even if a `<h2 id="subtitle">SUB</h2>` is present, it won't be used (i.e. the subtitle won't appear in `.head`).
Please check. (https://i.imgur.com/AO1cmSG.png)

shikhar-scs: Ohk. I'll see this in a while :+1:

tests/spec/w3c/headers-spec.js
shikhar-scs: > To be clear, I'm particularly interested in this being tested:
https://github.com/w3c/respec/pull/1516/files#diff-ebb1a497c097951bb868645bd73ade4fR29

I guess we can test for this here : 

https://github.com/shikhar-scs/respec/blob/e3ddc6f13af4d4a965f4b5f22944dda35e895caf/tests/spec/w3c/headers-spec.js#L270-L281

src/w3c/templates/headers.js
marcoscaceres: @sidvishnoi nice find. Definitely need add that as a test!

tests/spec/w3c/headers-spec.js
marcoscaceres: It's best to just leave the old tests and just add new ones (even if they overlap). Changing existing tests risks introducing new regressions.

shikhar-scs: Ok I'll implement that. Also @marcoscaceres is there any way I could run only a single test. I mean I have to go to `http://localhost:8080/tests/SpecRunner.html` after every little change I make (and then it starts all over again ). So, is there any way I could run just a single test or test file ?

marcoscaceres: Ah yeah, just put "f" in front of either `it(` or `describe(`. So:

```
fit();// "focus" this one
fdescribe(); // focus all these tests. 
```

You an also click spec in the results page, to only run the files associated with that spec.

marcoscaceres: More info https://jasmine.github.io/2.1/focused_specs.html

shikhar-scs: This is a life saviour. Earlier I used to sit for minutes just to see small changes. Thanks Marcos

marcoscaceres: Feel your pain. Just remember to remove the `fdescribe` and `fit` when done. I've accidentally checked in a few of those ☺️.

src/w3c/templates/headers.js
shikhar-scs: In such cases should I then introduce `conf.subtitle` in the configs, based on the condition `if(!conf.subtitle && document.querySelector("h2#subtitle")` ? 
Or just keep a test to ensure that all of this is in place ?

tests/spec/w3c/headers-spec.js
shikhar-scs: Ok so this has been covered :+1:

src/w3c/templates/headers.js
marcoscaceres: Both, I would say. But basically, the tests should guide the code. Just add sufficient tests to cover all possible situations. If nothing breaks, you are good :)

shikhar-scs: Ok :+1:

shikhar-scs:
@marcoscaceres Please have a look now :sweat_smile:

tests/spec/w3c/headers-spec.js
```diff
@@ -250,19 +250,89 @@ describe("W3C — Headers", function() {
...
       ops.body = "<h2 id='subtitle'>This is a pre existing subtitle <code>pass</code></h2>" + makeDefaultBody();
+    
+      const doc = await makeRSDoc(ops);
       
+      const subTitleElement = doc.querySelectorAll("h2#subtitle");
+      expect(subTitleElement.length).toEqual(1);
+    
+      const confSubTitle = doc.defaultView.respecConfig.subtitle;
+    
+      if (confSubTitle) {
```
marcoscaceres: This can't be an `if` here. It must equal exactly one thing.

tests/spec/w3c/headers-spec.js
```diff
@@ -250,19 +250,89 @@ describe("W3C — Headers", function() {
...
+        expect(confSubTitle).toEqual("This is a pre existing subtitle pass");
+      } else {
+        expect(confSubTitle).toEqual("");
+      }
+    
+      const [forInnerText] = subTitleElement;
```
marcoscaceres: I'm still not sure what "forInnerText" is? It's clearly an element, but this seems a strange/non-standard way to get at it? Can you help me understand this bit?

tests/spec/w3c/headers-spec.js
shikhar-scs: I get it. This `if` would vary for each of the 3 tests that I have introduced. Correct ?

tests/spec/w3c/headers-spec.js
```diff
@@ -250,19 +250,89 @@ describe("W3C — Headers", function() {
...
+      const initialSubtitle = doc.defaultView.respecConfig.subtitle;
+      expect(initialSubtitle).toEqual("This is a pre existing subtitle pass")
+    
       const subTitleElement = doc.querySelectorAll("h2#subtitle");
       expect(subTitleElement.length).toEqual(1);
-      
+    
+      const confSubTitle = doc.defaultView.respecConfig.subtitle;
+    
+      if (confSubTitle) {
```
marcoscaceres: As above - tests should never have a conditional :)

tests/spec/w3c/headers-spec.js
```diff
@@ -250,19 +250,89 @@ describe("W3C — Headers", function() {
...
+      const newProps = {
+        subtitle: "sub",
+      };
+    
+      Object.assign(ops.config, newProps);
+    
       const doc = await makeRSDoc(ops);
-
+    
+      const initialSubtitle = doc.defaultView.respecConfig.subtitle;
```
marcoscaceres: nit: calling this "initialSubtitle" makes it sound like it could change. Maybe just:

```JS
 const { subtitle } = doc.defaultView.respecConfig;
```

tests/spec/w3c/headers-spec.js
```diff
@@ -250,19 +250,89 @@ describe("W3C — Headers", function() {
...
+      expect(initialSubtitle).toEqual("This is a pre existing subtitle pass")
+    
       const subTitleElement = doc.querySelectorAll("h2#subtitle");
       expect(subTitleElement.length).toEqual(1);
-      
+    
+      const confSubTitle = doc.defaultView.respecConfig.subtitle;
```
marcoscaceres: this will be the same as initialSubtitle.

tests/spec/w3c/headers-spec.js
marcoscaceres: If you want the actual "initial subtitle", that's "ops.config.subtitle", as that's the one you passed. 

So, somewhere:

```JS
expect(doc.defaultView.respecConfig.subtitle).toEqual(ops.config.subtitle)
```

tests/spec/w3c/headers-spec.js
```diff
@@ -250,19 +250,89 @@ describe("W3C — Headers", function() {
...
+    });
+    
+    it("introduces conf.subtitle if it doesn't exist but h2#subtitle exists", async () => {
+      const ops = makeStandardOps();
+    
+      ops.body = "<h2 id='subtitle'>This is a pre existing subtitle <code>pass</code></h2>" + makeDefaultBody();
+    
+      const doc = await makeRSDoc(ops);
+    
+      const initialSubtitle = doc.defaultView.respecConfig.subtitle;
```
marcoscaceres: As above about initial subtitle

shikhar-scs:
Thankyou for the immediate feedback. I'll get back to this ASAP

tests/spec/w3c/headers-spec.js
marcoscaceres: Correct. If you need to split into multiple tests, that's totally ok.

tests/spec/w3c/headers-spec.js
shikhar-scs: This `forInnerText` is for checking the entire `innertext` of the subtitle element (normal text + text inside <code\> tags )at once. Now, I think it is becoming a bit repetitive to check this again and again. Shall I remove it ?

Also, I was in college at that time so the late reply :sweat_smile:

tests/spec/w3c/headers-spec.js
shikhar-scs: Sure :+1:

tests/spec/w3c/headers-spec.js
shikhar-scs: I guess we don't need it now. But thanks for the information :sweat_smile:

shikhar-scs:
@marcoscaceres Is this it ?

marcoscaceres:
Did a couple of fixes directly on github... so no idea if it works. But I think that's closer to what we want.

src/w3c/templates/headers.js
```diff
@@ -2,16 +2,42 @@ import "deps/hyperhtml";
...
+  } else {
+    if (document.title !== conf.title) {
+      pub("warn", "The document's title and the `<title>` element differ.");
+    }
+    specTitleElem.textContent = conf.title;
+    specTitleElem.id = "title";
+  }
+  specTitleElem.classList.add("title", "p-name");
+
+  let specSubTitleElem = document.querySelector("h2#subtitle");
```
marcoscaceres: This gives us `null` for free, which is nice.

src/w3c/templates/headers.js
```diff
@@ -2,16 +2,42 @@ import "deps/hyperhtml";
...
+  if (specSubTitleElem && specSubTitleElem.parentElement) {
+    specSubTitleElem.remove();
+    conf.subtitle = specSubTitleElem.textContent.trim();
+  } else if (conf.subtitle) {
+    specSubTitleElem = document.createElement("h2");
+    specSubTitleElem.textContent = conf.subtitle;
+    specSubTitleElem.id = "subtitle";
+  }
+  if (specSubTitleElem) {
```
marcoscaceres: in either element case, but ignoring null...

shikhar-scs:
> so no idea if it works

I'll pull for once and see.

shikhar-scs:
> so no idea if it works

I'll pull for once and see.

shikhar-scs:
Really thankful to you for all this painstaking work :clap: :sweat_smile: . Let's wait for saschanz then :+1:

src/w3c/templates/headers.js
```diff
@@ -2,16 +2,42 @@ import "deps/hyperhtml";
 import showLogo from "./show-logo";
 import showPeople from "./show-people";
 import showLink from "./show-link";
+import { pub } from "core/pubsubhub";
 
 export default conf => {
   const html = hyperHTML;
+  const specTitleElem =
+    document.querySelector("h1#title") || document.createElement("h1");
```
saschanaz: Generally okay, but maybe we can split these into `getSpecTitleElement()` and `getSpecSubtitleElement()` functions for readability.

shikhar-scs: If I could get a confirmation on this it, I'll incorporate :sweat_smile:

src/w3c/templates/headers.js
```diff
@@ -2,16 +2,42 @@ import "deps/hyperhtml";
...
   return html`<div class='head'>
   ${conf.logos.map(showLogo)}
-  <h1 class='title p-name' id='title'>${conf.title}</h1>
-  ${conf.subtitle ? html`
-    <h2 id='subtitle'>${conf.subtitle}</h2>
-  ` : ""}
-  <h2>${conf.prependW3C ? `W3C ` : ""}${conf.textStatus} <time class='dt-published' datetime='${conf.dashDate}'>${conf.publishHumanDate}</time></h2>
+  ${specTitleElem}
+  ${specSubTitleElem}
+  <h2>${conf.prependW3C ? "W3C " : ""}${ conf.textStatus } <time class='dt-published' datetime='${conf.dashDate}'>${conf.publishHumanDate}</time></h2>
```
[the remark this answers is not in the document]
shikhar-scs: Done :+1:

src/w3c/templates/headers.js
marcoscaceres: Good idea about splitting this out into functions.

shikhar-scs:
@marcoscaceres I've changed it to functions now. Please have a look. 

P.S. Also, if you could check slack for a while (I know you don't do it on saturdays but it would be great if you could see it for once )

src/w3c/templates/headers.js
```diff
@@ -2,16 +2,50 @@ import "deps/hyperhtml";
 import showLogo from "./show-logo";
 import showPeople from "./show-people";
 import showLink from "./show-link";
+import { pub } from "core/pubsubhub";
 
 export default conf => {
   const html = hyperHTML;
+  
+  function getSpecTitleElem () {
```
marcoscaceres: These should be moved outside this function.

src/w3c/templates/headers.js
```diff
@@ -2,16 +2,50 @@ import "deps/hyperhtml";
...
+        pub("warn", "The document's title and the `<title>` element differ.");
+      }
+      specTitleElem.textContent = conf.title;
+      specTitleElem.id = "title";
+    }
+    specTitleElem.classList.add("title", "p-name");
+    return specTitleElem;
+  }
+  
+  function getSpecSubTitleElem () {
```
marcoscaceres: As above

src/w3c/templates/headers.js
```diff
@@ -2,16 +2,50 @@ import "deps/hyperhtml";
...
+    let specSubTitleElem = document.querySelector("h2#subtitle");
+    
+    if (specSubTitleElem && specSubTitleElem.parentElement) {
+      specSubTitleElem.remove ();
```
marcoscaceres: Nit: spaces before (). Also elsewhere.

src/w3c/templates/headers.js
```diff
@@ -2,16 +2,50 @@ import "deps/hyperhtml";
...
   ${conf.logos.map(showLogo)}
-  <h1 class='title p-name' id='title'>${conf.title}</h1>
-  ${conf.subtitle ? html`
-    <h2 id='subtitle'>${conf.subtitle}</h2>
-  ` : ""}
-  <h2>${conf.prependW3C ? `W3C ` : ""}${conf.textStatus} <time class='dt-published' datetime='${conf.dashDate}'>${conf.publishHumanDate}</time></h2>
+  ${getSpecTitleElem()}
+  ${getSpecSubTitleElem()}
+  <h2>${conf.prependW3C ? "W3C":""}${conf.textStatus}<time class='dt-published' datetime='${conf.dashDate}'>${conf.publishHumanDate}</time></h2>
   <dl>
```
marcoscaceres: Please add a space to “W3C “. See TravisCI error.

marcoscaceres:
I’m not on my computer today so can’t access slack (reviewing on my phone). Happy to answer questions on email tho.

src/w3c/templates/headers.js
shikhar-scs: I had initially kept these 2 functions outside the `export` statement only. However, `conf` would then be undefined and thus, 5 out of 6 tests fail :disappointed: . I guess we'll have to keep them inside `export` only

marcoscaceres: Ah, you can just pass conf in:

```js
function getSpecTitleElem(conf) {

}
```

shikhar-scs: how can i miss that. So sorry

marcoscaceres: Heh, it happens to all of us :) why reviews matter.

marcoscaceres:
Nice work @shikhar-scs!

shikhar-scs:
yayy !!! Thanks a lot @marcoscaceres for merging this :tada:

PR #1517, opened 2018-02-18
JSON-LD SEO support

gkellogg:
creates script element with basic document information in JSON-LD. Depends on `config.toJsonld` being true.

Provides an effective and simple alternative to RDFa support. See w3c/respec#1503.


src/core/jsonld.js
```diff
@@ -0,0 +1,101 @@
...
+      "citation": {"@type": "@id"},
+      "contributor": {"@type": "@id"},
+      "datePublished": {"@type": "xsd:date"},
+      "dependencies": {"@type": "@id"},
+      "editor": {"@type": "@id"},
+      "hasPart": {"@type": "@id"},
+      "inLanguage": {"@language": null},
+      "isBasedOn": {"@type": "@id"},
+      "license": {"@type": "@id"},
+      "license": {"@type": "@id"},
```
saschanaz: A duplication!

src/core/jsonld.js
```diff
@@ -0,0 +1,101 @@
...
+      "worksFor": {"@type": "@id"}
+    }],
+    id: (conf.canonicalURI || conf.thisVersion),
```
saschanaz: probably no need to parenthesize?

src/core/jsonld.js
```diff
@@ -0,0 +1,101 @@
...
+    type: types,
+    name: conf.title,
+    inLanguage: ($('html', doc).attr('lang') || 'en'),
```
saschanaz: We are trying to remove jQuery dependency, so I would use `doc.documentElement.getAttribute("lang")`. (Double quotes are preferred)

src/core/jsonld.js
```diff
@@ -0,0 +1,101 @@
...
+      url: "https://www.w3.org/"
+    },
+    discussionUrl: conf.issueBase
+  };
+
+  if (conf.subtitle) jsonld.alternativeHeadline = conf.subtitle;
+  if (conf.prevVersion) jsonld.isBasedOn = conf.prevVersion;
+
+  // description from abstract
+  const $abs = $("#abstract", doc);
```
saschanaz: Same jQuery thing: `doc.getElementById("abstract");`

src/core/jsonld.js
```diff
@@ -0,0 +1,101 @@
...
+
+  // normative and informative references
+  jsonld.dependencies = Array.from(conf.normativeReferences).map(ref => addRef(conf, ref));
+  jsonld.citation = Array.from(conf.informativeReferences).map(ref => addRef(conf, ref));
+
+  var $jsonld = $(
+    "<script type='application/ld+json'>" +
+      JSON.stringify(jsonld) +
+      "</script>"
+  ).appendTo($("head"));
```
saschanaz: hyperHTML is preferred here:

```js
hyperHTML`<script type="application/ld+json">${JSON.stringify(jsonld)}</script>`
```

This returns HTMLElement, so we can use `.appendChild`.

tests/spec/core/jsonld-spec.js
```diff
@@ -0,0 +1,125 @@
...
+        url: 'http://COMPANY'
+      }
+    });
+  });
+
+  it("should describe contributors", async () => {
+    const ops = {config, body};
+    const doc = await makeRSDoc(ops);
+
+    const $script = $("head>script[type='application/ld+json']", doc);
```
saschanaz: Same jQuery thing.

tests/spec/core/jsonld-spec.js
```diff
@@ -0,0 +1,125 @@
...
+      type: 'Person',
+      name: 'Shane McCarron'
+    });
+  });
+
+  it("should describe dependencies and citations", async () => {
+    const ops = {config, body};
+    const doc = await makeRSDoc(ops);
+
+    const $script = $("head>script[type='application/ld+json']", doc);
```
saschanaz: Same jQuery thing.

src/core/jsonld.js
```diff
@@ -62,14 +52,13 @@ export function run(conf, doc, cb) {
...
+    .concat(Array.from(conf.informativeReferences));
+  jsonld.citation = refs.map(ref => addRef(conf, ref));
 
-  var $jsonld = $(
-    "<script type='application/ld+json'>" +
-      JSON.stringify(jsonld) +
-      "</script>"
-  ).appendTo($("head"));
+  var $jsonld = hyperHTML`<script type="application/ld+json"></script>`;
+  $jsonld.appendChild(doc.createTextNode(JSON.stringify(jsonld)));
```
saschanaz: `${JSON.stringify(jsonld)}` within script tag should be okay, as that's how hyperHTML works. (It will automatically create a text node)

gkellogg: No, it created an error. I found a reference to it here: https://www.danielcrabtree.com/blog/25/gotchas-with-dynamically-adding-script-tags-to-html. I tried your suggestion before resorting to this.

Perhaps, could put the `doc.createTextNode` within the hyperHTML string.

saschanaz: Oh, I tried and it breaks only for `<script>`. I didn't know that, thanks!

src/core/jsonld.js
```diff
@@ -0,0 +1,92 @@
...
+    },
+    discussionUrl: conf.issueBase
+  };
+
+  if (conf.subtitle) jsonld.alternativeHeadline = conf.subtitle;
+  if (conf.prevVersion) jsonld.isBasedOn = conf.prevVersion;
+
+  // description from abstract
+  const $abs = doc.getElementById("abstract");
+  if ($abs.textContent.length > 0) {
```
saschanaz: The `.length` check on RDFa part was to check the existence of `abstract`, probably we should do the same to prevent a catastrophic error here.

saschanaz:
`js/core/jsonld.js` shouldn't be pushed, I opened #1518 to prevent this.

gkellogg:
@saschanaz if all concerns have been addressed, can you revise your review comments?

gkellogg:
Ultimately having both JSON-LD and RDFa support are not in conflict, but based on the perceived value of the JSON-LD SEO, the RDFa support is redundant and could be deprecated/removed. This implementation is much more modular and does not send tentacles out into other bits of code, so should be less sensitive to semantic drift.

Eventually, perhaps the `doJsonLd` flag becomes default `true`.

saschanaz:
This still have js/core/jsonld.js, you may want to merge from master as #1518 is now merged 👍

src/core/jsonld.js
```diff
@@ -0,0 +1,92 @@
...
+
+  // description from abstract
+  const $abs = doc.getElementById("abstract");
+  if ($abs && $abs.textContent.length > 0) {
+    jsonld.description = $abs.textContent;
+  }
+
+  // Editors
+  if (conf.editors) {
+    jsonld.editor = conf.editors.map(person => addPerson(person));
```
saschanaz: nit: a shorter equivalent is `.map(addPerson)` 😎

src/core/jsonld.js
```diff
@@ -0,0 +1,92 @@
...
+  }
+  if (conf.authors) {
+    jsonld.contributor = conf.authors.map(person => addPerson(person));
```
saschanaz: (same nit)

src/core/jsonld.js
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  // normative and informative references
+  const refs = Array.from(conf.normativeReferences)
+    .concat(Array.from(conf.informativeReferences));
```
saschanaz: nit: we may do `[...conf.foo, ...conf.bar]` instead of copying and concatenating by `Array.from`.

saschanaz:
I don't have much knowledge about JSON-LD so I looked more at code styles, @marcoscaceres may want to comment more.

gkellogg:
Rebased and addressed code nits. I had attempted to remove js/core/jsonld.js previously, but it didn't take; should be gone now.

src/core/jsonld.js
```diff
@@ -0,0 +1,91 @@
...
+  if (conf.editors) {
+    jsonld.editor = conf.editors.map(addPerson);
+  }
+  if (conf.authors) {
+    jsonld.contributor = conf.authors.map(addPerson);
+  }
+
+  // normative and informative references
+  const refs = [...conf.normativeReferences, ...conf.informativeReferences];
+  jsonld.citation = refs.map(ref => addRef(conf, ref));
```
marcoscaceres: This should probably change to the following, as some refs can be strings (we haven't finished converting them all in SpecRef yet). Note also that now `addRef` can just take a single argument `ref` - no need to pass the `conf` in:

```JS
jsonld.citation = refs
  .map(ref => conf.biblio[ref])
  .filter(ref => typeof ref === "object")
  .map(addRef);
```

src/core/jsonld.js
```diff
@@ -0,0 +1,91 @@
...
+    ed.worksFor = {
+      name: person.company,
+      url: person.companyURL
+    };
+  }
+  return ed;
+}
+
+// Create a reference URL from the ref
+function addRef(conf, ref) {
```
marcoscaceres: As noted above, this one could just be:

```JS
function addRef(ref) {
  const { href: id, title: name, href: url } = ref;
  return {
    id,
    type: "TechArticle",
    name,
    url,
  };
}
```

src/core/jsonld.js
```diff
@@ -0,0 +1,91 @@
...
+
+  var $jsonld = hyperHTML`<script type="application/ld+json"></script>`;
+  $jsonld.appendChild(doc.createTextNode(JSON.stringify(jsonld)));
+  doc.head.appendChild($jsonld);
+
+  cb();
+}
+
+// Turn editors and authors into a list of JSON-LD relationships
+function addPerson(person) {
```
marcoscaceres: This one can just become:
```JS
function addPerson({ name, url, mailto, company, companyURL }) {
  const ed = {
    type: "Person",
    name,
    url,
    "foaf:mbox": mailto,
    worksFor: {
      name: company,
      url: companyURL,
    },
  };
  return ed;
}
```

When `company` and `companyURL` are missing, `worksFor` will be an empty object when JSON.stringify() is called, which should be fine.

src/core/jsonld.js
```diff
@@ -0,0 +1,91 @@
...
+  }
+  if (conf.authors) {
+    jsonld.contributor = conf.authors.map(addPerson);
+  }
+
+  // normative and informative references
+  const refs = [...conf.normativeReferences, ...conf.informativeReferences];
+  jsonld.citation = refs.map(ref => addRef(conf, ref));
+
+  var $jsonld = hyperHTML`<script type="application/ld+json"></script>`;
```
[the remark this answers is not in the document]
marcoscaceres: Simplified the above a little bit, to just set `script.textContent`.

src/core/jsonld.js
```diff
@@ -0,0 +1,91 @@
...
+      "isBasedOn": {"@type": "@id"},
+      "license": {"@type": "@id"}
+    }],
+    id: conf.canonicalURI || conf.thisVersion,
+    type: types,
+    name: conf.title,
+    inLanguage: doc.documentElement.getAttribute("lang") || "en",
+    license: conf.licenseInfo.url,
+    datePublished: conf.dashDate,
+    copyrightHolder: {
```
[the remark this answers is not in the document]
gkellogg: I thought about that, as well as `overrideCopyright`, but we don't get quite the information out of this. But, presuming that it's a simple text string, could create something like the following:

```json
  copyrightHolder: [{
      name: "World Wide Web Consortium",
      url: "https://www.w3.org/"
    }, {
      name: ${conf.additionalCopyrightHolders}
    }
```

although, need to consider that it may be an array.

Similarly, the w3c bit could be left out if there is an `overrideCopyright`. There's also the non-w3c case, but I wasn't clear on what could be gleaned from the configuration in that case.

src/core/jsonld.js
```diff
@@ -0,0 +1,91 @@
...
+    },
+    discussionUrl: conf.issueBase
+  };
+
+  if (conf.subtitle) jsonld.alternativeHeadline = conf.subtitle;
+  if (conf.prevVersion) jsonld.isBasedOn = conf.prevVersion;
+
+  // description from abstract
+  const $abs = doc.getElementById("abstract");
+  if ($abs && $abs.textContent.length > 0) {
```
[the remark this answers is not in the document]
gkellogg: I wasn't clear if a zero-length string was `falsey` in JavaScript, good to know.

src/core/jsonld.js
```diff
@@ -0,0 +1,91 @@
+// Module core/jsonld
+// Create script element with JSON-LD for SEO document description
+export const name = "core/jsonld";
+
+export function run(conf, doc, cb) {
+  if (!conf.doJsonLd) {
```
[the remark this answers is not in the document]
gkellogg: It's certainly useful for non-W3C use, and for Unofficial documents, but we may need some config hints to go off of, mostly for the copyright bits, if you have a suggestion..

src/core/jsonld.js
gkellogg: `worksFor` is an empty object, but that's not reasonable JSON-LD, IMHO. It implies that the person works for something, which is not what's asserted.

src/core/jsonld.js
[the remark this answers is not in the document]
gkellogg: The problem with `overrideCopyright` is that it's described as being markup, which isn't too helpful here.

src/core/jsonld.js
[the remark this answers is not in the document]
marcoscaceres: What we could do for now is just take this and say that it's initially for use with W3C specs.

marcoscaceres:
@gkellogg give this a `git pull`, I fixed some additional nits.

gkellogg:
If/when we remove RDFa support, we should add the following to the default configuration:

```
if (conf.doJsonLd === undefined) conf.doJsonLd = true;
```

marcoscaceres:
W3C defaults go here:
https://github.com/w3c/respec/blob/develop/src/w3c/defaults.js#L50

gkellogg:
I was referring to [this code](https://github.com/w3c/respec/blob/develop/src/w3c/headers.js#L234), which sounds like it's there in error.

marcoscaceres:
> I was referring to this code, which sounds like it's there in error.

Indeed! Thanks!

marcoscaceres:
Send a PR to address the duplicate defaults: #1525

marcoscaceres:
*sent

marcoscaceres:
🎉!

@gkellogg, before we ship, could you kindly help us document `toJsonld` in the wiki? You can use, for example, [crEnd](https://github.com/w3c/respec/wiki/crEnd) as the basis to copy/paste from. 

If you want to post some suggested text here, that works too... then I can merge add to the wiki. Whatever is easiest for you. Happy to review.

gkellogg:
Wiki Entry added: https://github.com/w3c/respec/wiki.

Feel free to improve.

Issue #1419, opened 2017-10-26
add test for spec writer


marcoscaceres:
The "save-html.js" module is attached to the UI, which makes it  hard to test. We should split it out to its own module.

shubhshrma:
@marcoscaceres Can I take this forward? Meaning can this be completed keeping upcoming submissions in mind.

marcoscaceres:
@shubhshrma, it could definitely be. However, we'd have to think about how to actually test this. It's not super trivial.

shubhshrma:
So @marcoscaceres, what I need to do is move ```save-html.js``` in ```src/core``` and add tests for it. Please correct me if I am wrong.

marcoscaceres:
That is correct, @shubhshrma.

marcoscaceres:
Closing this 9-year-old testing task. The save-html module has evolved significantly and is tested through integration. The specific refactoring concern is outdated.

PR #1521, opened 2018-02-21
make save-html a core module

shubhshrma:
For #1419 
@marcoscaceres @saschanaz  I just moved ```save-html``` to ```core``` and changed wherever it is referenced. Tests are to be written. It would be great if you could help taking forward this PR. What tests are to be written? How to capture downloadable files inside tests?


marcoscaceres:
Closing this one, as it's not quite correct.

Issue #1522, opened 2018-02-21
Respec can't check local source files anymore


vivienlacourba:
respec was able to check a local source file now it seems it is not able to check local files anymore whereas it works fine if the file is retrieved from the web.

With local source file ... **Fails**:
```bash
~/W3C/WebRTC/webrtc-pc$ node support/respec/tools/respec2html.js -e --timeout 30 --src file://`pwd`/webrtc.html --out build/output.html
Error: 📡 HTTP Error 0: file:///home/vivien/W3C/WebRTC/webrtc-pc/webrtc.html
   at fetchAndWrite (/home/vivien/W3C/WebRTC/webrtc-pc/support/respec/tools/respecDocWriter.js:78:13)
   at <anonymous>
   at process._tickCallback (internal/process/next_tick.js:188:7)
```

With same source file fetched from the web ... **Works**:
```bash
~/W3C/WebRTC/webrtc-pc$ node support/respec/tools/respec2html.js -e --timeout 30 --src https://w3c.github.io/webrtc-pc/ --out build/output.html
```

If possible please add a new test for using local files as the test you have (number 8 and 9) are using a local webserver, eg.

```bash
node ./tools/respec2html.js -e --disable-sandbox --timeout 30 --src http://localhost:3000/examples/basic.built.html --out /dev/null (test 8/9)
```

Issue faced by @suhasHere @alvestrand @dontcallmedom on various webrtc drafts

vivienlacourba:
Hi @marcoscaceres I hope you can help us here.

Side note respec tests are also broken https://travis-ci.org/w3c/respec/jobs/344196977

marcoscaceres:
@saschanaz, this is related to us switching to Puppeteer. File URLs might not work until this get addressed: https://github.com/GoogleChrome/puppeteer/issues/1506

@vivienlacourba, the work around in just to spin up a local server for now.

```Bash
python -m SimpleHTTPServer 8000
```

(or you can npm install `http-server` and use that)

marcoscaceres:
@vivienlacourba, 
> Side note respec tests are also broken https://travis-ci.org/w3c/respec/jobs/344196977

Kindly note the above is a random feature branch, which happens to be failing - but it's not part of the build. 

To check the build status of ReSpec (the "develop" branch), go directly to https://travis-ci.org/w3c/respec

vivienlacourba:
[...]

PR #1524, opened 2018-02-21
fix(tools/respec2html): resolve src to file URL

marcoscaceres:
* relates to #1522

### DIFF
--- a/tools/respec2html.js
+++ b/tools/respec2html.js
@@ -2,6 +2,7 @@
 
 /*jshint node: true, browser: false*/
 "use strict";
+const { URL } = require('url');
 const colors = require("colors");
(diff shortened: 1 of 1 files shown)


marcoscaceres:
This doesn't fix the issue, but at least we get always get a URL based off file... it it's absolute URL, it just resolves as normal.

saschanaz:
We already have `express` (and `http-server` for dev) dependency. Should we automatically open a local server until the Chrome issue gets fixed?

marcoscaceres:
@saschanaz, we could... could you give me a rough sketch of how it might work?

marcoscaceres:
Going to use this for now, as it technically correct. We can discuss a longer term solution separately.

PR #1527, opened 2018-02-22
jsonld uses undefined "xsd" namespace

gkellogg:
Set JSON-LD `@langauge` from HTML `@lang`.
Fix `datePublished` `@type` to not use undefined `xsd` namespace.


gkellogg:
Note that the SDTT doesn't care, as it doesn't look at contexts, but any other conforming JSON-LD processor will complain at the use of an undefined prefix.

Issue #1502, opened 2018-02-13
META: Google Summer of Code 2018


marcoscaceres:
🎉 ReSpec has been selected to be part of Google Summer of Code 2018 as a Mozilla project 🎉

This means that there is a really good chance of getting funding for one of you - but it's not yet guaranteed! 

## Your proposals 
Applicants, please read Google's [student guidelines for project proposals](https://google.github.io/gsocguides/student/writing-a-proposal) carefully: 

Important to note from above:
 
> This is a competitive program, each year Google turns down many more students than it funds. While pre-proposal activities are key to improving your chances of success, a poorly-written proposal is an easy way to fail. 

If the proposals are not strong, all of us risk not getting to work together - so it's on me to make sure our project proposals are highly competitive: we are competing for funds with other projects at Mozilla, so make sure that you choose wisely. 

Students interested in working on this (there is about 10 of you!), please:
 
* create and share a **private [Google Doc](http://docs.google.com/)** with your project proposal with me. I will review this when you tell me to, and help you make sure it's competitive.  
* use the [Elements of a Quality Proposal](https://google.github.io/gsocguides/student/writing-a-proposal#elements-of-a-quality-proposal) to structure your proposal. 
* Explore this project: please research how it is used in the wild, explore the user interface, explore the code, and see the [issues people are currently having](https://github.com/w3c/respec/issues), and start thinking about how you would make the project better.  

## Proposal due date
I expect the **first round of proposals by the 26th of February** for me to review. 

You have two options:

 1. Marcos gives you a list of tasks to work on. I'll label the issues on github as "GSoC". 
 1. You propose a list of things that are of interest to you.

In either case, I expect applicants to do some degree of deep diving into the project and understand what it's being used for, by who, and why. Consider this part of the application process, as it's demonstrating your ability to be self directed.    

Please also subscribe to this issue for updates. 

If you want to contact me privately with questions, please use "mcaceres@mozilla.com".  

I'll update this with more details and large things that need fixing.  

## Tasks (big ticket items)
More stars, the more important. 

## WebIDL module rewrite ⭐️⭐️⭐️ ⭐️⭐️
WebIDL is absolutely critical to web specifications - it's how we define APIs for the web.

Unfortunately, the current code base for WebIDL has maintainability problems, bugs, and needs a rewrite. 

[...]

PR #1532, opened 2018-02-25
converting promise to async/await

himanish-star:
fixes; #1502 
@marcoscaceres, I have really filed this PR in a hurry so please don't mind those function names. 
I hope my solution is the right fix to this :smiley:


marcoscaceres:
See feedback in https://github.com/w3c/respec/pull/1533, as these are really quite similar.

PR #1533, opened 2018-02-25
Converting .then to async/await 

shikhar-scs:
As suggested by @marcoscaceres at https://github.com/w3c/respec/issues/1502#issuecomment-368250097 the `.then()` segments have been changed to `async` functions. @himanish-star had already filed a PR so his can be the one to be merged. However, I did this just for practice.

Please review.

PR #1534, opened 2018-02-25
refactor(core/issues-notes): replace then with async- await

shubhshrma:
@marcoscaceres, as requested, my solution to #1502. Please review.

### DIFF
--- a/src/core/issues-notes.js
+++ b/src/core/issues-notes.js
@@ -141,24 +141,27 @@ export function run(conf, doc, cb) {
     issueBase = conf.issueBase;
   if ($ins.length) {
     if (conf.githubAPI) {
-      ghFetch(conf.githubAPI)
-        .then(function (json) {
-          issueBase = issueBase || json.html_url + "/issues/";
-          return fetchIndex(json.issues_url, {
-            // Get back HTML content instead of markdown
-            // See: https://developer.github.com/v3/media/
-            Accept: "application/vnd.github.v3.html+json",
-          });
-        })
-        .then(function (issues) {
-          issues.forEach(function (issue) {
-            ghIssues[issue.number] = issue;
-          });
-          handleIssues($ins, ghIssues, issueBase);
-          cb();
-        }).catch(err => {
-          pub("error", err.message);
-          handleIssues($ins, ghIssues, issueBase);
+      let issuesUrl;
+      const getIssues = (json) => {
+        issuesUrl = issueBase || json.html_url + "/issues/";
+        return fetchIndex(json.issues_url, {
+          // Get back HTML content instead of markdown
+          // See: https://developer.github.com/v3/media/
+          Accept: "application/vnd.github.v3.html+json",    
+        });
+      };
+      const asyncFetch = async () => {
+        const json = await ghFetch(conf.githubAPI);
+        const issues = await getIssues(json);
+        issues.reduce((collector, issue) => {
+          collector[issue.number] = issue;
+          return collector;
+        }, ghIssues);
+      };
+      asyncFetch()
+        .catch(err => pub("error", err.message))
+        .finally(() => {
+          handleIssues($ins, ghIssues, issuesUrl);
           cb();
         });
     } else {


marcoscaceres:
@shikhar-scs, @himanish-star, can you review @shubhshrma's code? I want to get a sense of your collaboration and review skills. 

@shikhar-scs, @himanish-star, @shubhshrma - the same applies for #1532 and #1533. Please review each other's code and work together to improve each other's solutions - or ultimately, decide on the best solution. 

Please verify that there is a test or two to proves that the solution works 😉 There might already be tests for this, but please confirm.

shikhar-scs:
Yeah sure!. I'll do that in a while . As of now I guess the try catch blocks are missing.

himanish-star:
Well currently I'm in class but I can see that the try catch block is missing

shubhshrma:
> shubhshrma requested review from marcoscaceres and removed request for marcoscaceres

Sorry, that was some inability to handle new collaborator privileges. :laughing: 

@shikhar-scs I intentionally avoided try-catch blocks because we have catch chaining. I guess this will work and I am looking into it now.

src/core/issues-notes.js
```diff
@@ -141,22 +141,25 @@ export function run(conf, doc, cb) {
...
+      const asyncFetch = async () => {
+        const json = await ghFetch(conf.githubAPI);
+        const issues = await getIssues(json);
+        issues.forEach( (issue) => {
+          ghIssues[issue.number] = issue;
+        });
+        handleIssues($ins, ghIssues, issueBase);
+        cb();
+      }
+      const getIssues = (json) => {
```
shikhar-scs: A function expression, that is:

[const | let | var] = function () {} (or () =>
Is the creation of an anonymous function (function () {}) and the creation of a variable, and then the assignment of that anonymous function to that variable.

So the usual rules around variable hoisting within a scope -- block-scoped variables (let and const) do not hoist as undefined to the top of their block scope.

This means:

if (true) {
    doSomething() // will fail
    const doSomething = function () {}
}
Will fail since doSomething is not defined. (It will throw a  ReferenceError)

If you switch to using var you get your hoisting of the variable, but it will be initialized to undefined so that block of code above will still not work. (This will throw a TypeError since doSomething is not a function at the time you call it)

As far as standard practices go, you should always use the proper tool for the job.



This is precisely the reason why `getIssues` at Line#146 `Cannot find declaration to go to` and at Line#153, `getIssues` shows as `no Usages found`. (These are the log messages in my editor)

src/core/issues-notes.js
```diff
@@ -141,22 +141,25 @@ export function run(conf, doc, cb) {
...
+        issueBase = issueBase || json.html_url + "/issues/";
+        return fetchIndex(json.issues_url, {
+          // Get back HTML content instead of markdown
+          // See: https://developer.github.com/v3/media/
+          Accept: "application/vnd.github.v3.html+json",    
+        });
+      }
+      asyncFetch()
+        .catch(err => {
```
shikhar-scs: Yes, I tried testing this, it works as of now, i.e., the promise returned was chained with the catch. Sorry for pointing that out

src/core/issues-notes.js
```diff
@@ -141,22 +141,25 @@ export function run(conf, doc, cb) {
...
-          issues.forEach(function(issue) {
-            ghIssues[issue.number] = issue;
-          });
-          handleIssues($ins, ghIssues, issueBase);
-          cb();
-        }).catch(err => {
+      const asyncFetch = async () => {
+        const json = await ghFetch(conf.githubAPI);
+        const issues = await getIssues(json);
+        issues.forEach( (issue) => {
```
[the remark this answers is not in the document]
marcoscaceres: Just note for all code/idiomaticness, the project uses: https://github.com/prettier/prettier 

... we should document that somewhere... probably in CONTRIBUTING.md :D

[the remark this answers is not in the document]
shikhar-scs: Yeah. Even I was stuck with that initially . 😅

src/core/issues-notes.js
shubhshrma: Yes @shikhar-scs I got your point. Here a solution could be just moving the function expression ```const getIssues = ...``` above ```const asyncFetch =...```.

himanish-star: > Here a solution could be just moving the function expression const getIssues = ... above const asyncFetch =....

Yep @shubhshrma you should change it to `function`

shikhar-scs:
@marcoscaceres 
> Please verify that there is a test or two to proves that the solution works wink There might already be tests for this, but please confirm.

Yeah tests do exist at https://github.com/w3c/respec/blob/develop/tests/spec/core/issues-notes-spec.js at Line#4 , Line#99 & Line#129.

Console.log of various jquery variables inside each gives us insight as to what happens in each test (which can also be gathered by the name of the tests partially.)

src/core/issues-notes.js
shubhshrma: > I would still suggest to follow the conventional way as after you a lot of people would be re-reading this code

This is arguable but let's go with what respec has used everywhere, function declarations.

himanish-star: Yes please

shubhshrma: > This is precisely the reason why getIssues at Line#146 Cannot find declaration to go to and at Line#153, getIssues shows as no Usages found. (These are the log messages in my editor)

Just to confirm these are log messages in your editor or browser? Because the PR is passing all tests and locally the issue referencing respec feature is working normal.

shikhar-scs: No. No. No. I use Webstorm IDE. It always auto-suggests changes and any minor errors in tooltips on hover. Similar to these. 
> This is precisely the reason why getIssues at Line#146 Cannot find declaration to go to and at Line#153, getIssues shows as no Usages found. (These are the log messages in my editor)

shubhshrma: > I use Webstorm IDE

Got it.

src/core/issues-notes.js
shubhshrma: NP :)

src/core/issues-notes.js
marcoscaceres: Nit: use `reduce`, see https://github.com/w3c/respec/pull/1533/files#r172387219

src/core/issues-notes.js
```diff
@@ -141,22 +141,25 @@ export function run(conf, doc, cb) {
...
+          ghIssues[issue.number] = issue;
+        });
+        handleIssues($ins, ghIssues, issueBase);
+        cb();
+      }
+      const getIssues = (json) => {
+        issueBase = issueBase || json.html_url + "/issues/";
```
marcoscaceres: Make a new variable here, don't trash issueBase over and over.

src/core/issues-notes.js
```diff
@@ -141,22 +141,25 @@ export function run(conf, doc, cb) {
...
-          handleIssues($ins, ghIssues, issueBase);
-          cb();
-        }).catch(err => {
+      const asyncFetch = async () => {
+        const json = await ghFetch(conf.githubAPI);
+        const issues = await getIssues(json);
+        issues.forEach( (issue) => {
+          ghIssues[issue.number] = issue;
+        });
+        handleIssues($ins, ghIssues, issueBase);
```
marcoscaceres: I wonder, if we are doing `handleIssues()` and `cb()` both here had in `.catch()`, can we make use of `.finally()`:

```Js
asyncFetch()
  .catch(err => pub("error", err.message))
  .finally(() => { 
     handleIssues($ins, ghIssues, issueBase)
     cb();
  })
```

marcoscaceres:
Quite like your approach, @shubhshrma. Couple of suggestions that you could try out to make it even more concise.

shubhshrma:
> Quite like your approach

Thanks :smile: , updating.

src/core/issues-notes.js
shubhshrma: Yes you are right. ```handleIssues()``` and ```cb()``` are present in both ```then()``` and ```catch()```. We can avoid duplicacy here. Updating.

src/core/issues-notes.js
```diff
@@ -141,24 +141,27 @@ export function run(conf, doc, cb) {
...
+          // Get back HTML content instead of markdown
+          // See: https://developer.github.com/v3/media/
+          Accept: "application/vnd.github.v3.html+json",    
+        });
+      };
+      const asyncFetch = async () => {
+        const json = await ghFetch(conf.githubAPI);
+        const issues = await getIssues(json);
+        issues.reduce((collector, issue) => {
+          ghIssues[issue.number] = issue;
```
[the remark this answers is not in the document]
marcoscaceres: Probably just  typo :) @shubhshrma, you want:

```JS
collector[issue.number] = issue;
```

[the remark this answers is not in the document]
marcoscaceres: Just noting that a lot of people are new to things like `reduce()`. If anyone would like to learn more, there is a great [MDN page](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce) about it.

goelrohan6:
Hey @marcoscaceres @shubhshrma @shikhar-scs @himanish-star what do you guys think about #1560 .
We can have a discussion there

src/core/issues-notes.js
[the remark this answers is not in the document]
shubhshrma: Yes this is a typo made in hurry.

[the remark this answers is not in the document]
shubhshrma: Updated.

marcoscaceres:
Closing, as this was just an exercise - @shubhshrma, I really liked your initial approach and where you took this. Was really concise and easy to read, which was what I was looking for.

PR #1538, opened 2018-02-28
refactor(core/contrib): use ES2015+

saschanaz:


src/core/contrib.js
```diff
@@ -10,107 +10,90 @@ import { pub } from "core/pubsubhub";
...
-  };
+  return o => o[prop];
 }
 
 function findUsers(...thingsToFind) {
-  var users = {};
-  thingsToFind.forEach(function(things) {
-    things.forEach(function(thing) {
+  const users = new Set();
+  for (const things of thingsToFind) {
```
marcoscaceres: Although this work better than existing code, I wonder if we can .map() the URLs out of things.users and .reduce() into users set? Declaritve loops are easier to rationalize about and work with (for me, at least 😊)

src/core/contrib.js
```diff
@@ -10,107 +10,86 @@ import { pub } from "core/pubsubhub";
...
 function join(things) {
   if (!things.length) {
     return "";
   }
-  things = things.slice(0);
-  var last = things.pop();
-  var length = things.length;
-  if (length === 0) {
+  const length = things.length;
+  const last = things[length - 1];
```
marcoscaceres: Nit: const {length} = things;

src/core/contrib.js
```diff
@@ -10,107 +10,86 @@ import { pub } from "core/pubsubhub";
...
-        });
-      names.sort(function(a, b) {
-        return a.toLowerCase().localeCompare(b.toLowerCase());
-      });
-      $(element).html(join(names)).attr("id", null);
-    });
+async function toHTML(urls, editors, element) {
+  const args = await Promise.all(...urls.map(ghFetch));
+  const names = args
+    .map(user => user[0].name || user[0].login)
```
marcoscaceres: Nit: ([usedDetails]) => userDetails.name || userDetails.login

src/core/contrib.js
```diff
@@ -10,107 +10,86 @@ import { pub } from "core/pubsubhub";
...
-        .filter(function(name) {
-          return editors.indexOf(name) < 0;
-        });
-      names.sort(function(a, b) {
-        return a.toLowerCase().localeCompare(b.toLowerCase());
-      });
-      $(element).html(join(names)).attr("id", null);
-    });
+async function toHTML(urls, editors, element) {
+  const args = await Promise.all(...urls.map(ghFetch));
```
marcoscaceres: Might be a bug here: Promise.all is not variadic. You might want `Promise.all([...` instead.

src/core/contrib.js
```diff
@@ -10,107 +10,86 @@ import { pub } from "core/pubsubhub";
...
-    .then(cb, function(error) {
-      pub(
-        "error",
-        "Error loading contributors and/or commenters from  Error: " + error
-      );
-      cb();
-    });
+  const json = await ghFetch(conf.githubAPI);
+  const [issues, comments, contributors] = await Promise.all(
+    fetchIndex(json.issues_url),
```
marcoscaceres: As above, regarding Promise.all. This should be in an array, no?

src/core/contrib.js
```diff
@@ -10,107 +10,86 @@ import { pub } from "core/pubsubhub";
...
   if (!things.length) {
     return "";
   }
-  things = things.slice(0);
-  var last = things.pop();
-  var length = things.length;
-  if (length === 0) {
+  const length = things.length;
+  const last = things[length - 1];
+  if (length === 1) {
```
marcoscaceres: This is really asking to be a switch statement, wdyt?

src/core/contrib.js
marcoscaceres: Apologies if I’m totally wrong about this. Maybe I’m misremembering?

src/core/contrib.js
```diff
@@ -10,107 +10,86 @@ import { pub } from "core/pubsubhub";
...
+    fetchIndex(json.issues_url),
+    fetchIndex(json.issue_comment_url),
+    fetchIndex(json.contributors_url)
+  );
+  const editors = respecConfig.editors.map(prop("name"));
+  const commenterUrls = findUsers(issues, comments);
+  const contributorUrls = contributors.map(prop("url"));
+  try {
+    return Promise.all(
+      toHTML(commenterUrls, editors, ghCommenters),
```
marcoscaceres: I’m not sure we want to return the result of Promise.all here. Shouldn’t this await first the return void instead? Alternatively, we could return a promise that resolves with undefined once Promise.all() resolves.

src/core/contrib.js
saschanaz: I'm not very familiar with `.reduce` but we probably can do:

```js
const urls = thingsToFind
    .map(things => things.filter(thing => thing.user).map(thing => thing.user.url))
    .reduce((x, y) => x.concat(y));
return [...new Set(urls)];
```

Would this be declarative enough?

marcoscaceres: The nesting is still a bit hard to read (in which case the for loop is more clear)... I’ll think about it and propose something tomorrow (😴 2am here) but otherwise, I’m ok for this to land

src/core/contrib.js
```diff
@@ -10,107 +10,87 @@ import { pub } from "core/pubsubhub";
...
-      });
-      $(element).html(join(names)).attr("id", null);
-    });
+async function toHTML(urls, editors, element) {
+  const args = await Promise.all(urls.map(ghFetch));
+  const names = args
+    .map(([user]) => user.name || user.login)
+    .filter(name => !editors.includes(name))
+    .sort((a, b) => a.toLowerCase().localeCompare(b.toLowerCase()));
+  element.textContent = join(names);
```
marcoscaceres: Note to self: .reduceRight() for `join()` here maybe.

PR #1541, opened 2018-03-01
refactor(core/best-practices): replace jQuery

saschanaz:
Because hyperHTML is great 😆

### DIFF
--- a/src/core/aria.js
+++ b/src/core/aria.js
@@ -10,7 +10,6 @@ sub(
   { once: true }
 );
 
-export function run(conf, doc, cb) {
-  doc.body.setAttribute("aria-busy", "true");
-  cb();
+export function run(conf) {
+  document.body.setAttribute("aria-busy", "true");
 }
--- a/src/core/best-practices.js
+++ b/src/core/best-practices.js
@@ -4,34 +4,32 @@
 // Best practices are marked up with span.practicelab.
 import css from "deps/text!core/css/bp.css";
 import { pub } from "core/pubsubhub";
+import "deps/hyperhtml";
 
 export const name = "core/best-practices";
 
-export function run(conf, doc, cb) {
-  var num = 0;
-  var $bps = $("span.practicelab", doc);
-  var $content = $("<div><h2>Best Practices Summary</h2><ul></ul></div>");
-  var $ul = $content.find("ul");
-  $bps.each(function() {
-    var $bp = $(this);
-    var id = $bp.makeID("bp");
-    var $li = $("<li><a></a></li>");
-    var $a = $li.find("a");
-    num++;
-    $a.attr("href", "#" + id).text("Best Practice " + num);
-    $li.append(doc.createTextNode(": " + $bp.text()));
-    $ul.append($li);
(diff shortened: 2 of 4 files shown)


tools/release.js
```diff
@@ -395,7 +395,7 @@ const run = async () => {
       colors.info(" Making sure the generated version is ok... 🕵🏻")
     );
     await node(
-      `./tools/respec2html.js -e --timeout 30 --src file://${process.cwd()}/examples/basic.built.html --out /dev/null`,
+      `./tools/respec2html.js -e --timeout 30 --src file:///${__dirname}/../basic.built.html --out /dev/null`,
```
saschanaz: I accidentally missed `../` part, but still think `__dirname` is better because it's cwd-safe. What do you think? @marcoscaceres

saschanaz:
Codacy doesn't like the unused `conf` but it's currently needed for base-runner to detect modern modules...

tools/release.js
marcoscaceres: I'm ok with __dirname, but it was breaking. I usually run the release script from:

```Bash
./tools/release.js
``` 

Thus need to make sure that it always resolves to the right place.

src/core/aria.js
```diff
@@ -10,7 +10,6 @@ sub(
   { once: true }
 );
 
-export function run(conf, doc, cb) {
-  doc.body.setAttribute("aria-busy", "true");
-  cb();
+export function run(conf) {
+  document.body.setAttribute("aria-busy", "true");
```
marcoscaceres: Oh, I thought I'd removed this module? This module is no good. Could you delete it?

src/core/best-practices.js
```diff
@@ -4,34 +4,33 @@
...
-    $bp.prepend(doc.createTextNode("Best Practice " + num + ": "));
-  });
-  if ($bps.length) {
-    $(doc).find("head link").first().before($("<style/>").text(css));
-    if ($("#bp-summary")) $("#bp-summary").append($content.contents());
-  } else if ($("#bp-summary").length) {
+    const id = window.$.fn.makeID.call(bp, "bp");
+    const li = hyperHTML`<li><a href="${`#${id}`}">Best Practice ${num}</a>: ${bp.textContent}</li>`;
+    ul.appendChild(li);
+    bp.insertBefore(document.createTextNode(`Best Practice ${num}: `), bp.firstChild);
```
marcoscaceres: still not a `wire()` fan, huh? 😉 That's ok.

saschanaz:
```js
const num = 1;
hyperHTML`Best Practice ${num}:`
```

This returns 4 nodes. Inserting multiple nodes with `insertBefore` is kind of pain :/

marcoscaceres: Yeah, that makes sense... good point.

saschanaz:
No idea why the test fails, it passes on my machine. I should investigate more...

saschanaz:
I had wrong configuration 😱 Now fixed. Unfortunately the Codacy warnings should be ignored for now...

PR #1546, opened 2018-03-02
Refactor .then() to async/await

vinicius0197:
I'm a little bit late for this challenge, but wanted to do it anyway for practice! I was unable to run tests locally to test these changes, so I would appreciate all feedback.

### DIFF
--- a/src/core/issues-notes.js
+++ b/src/core/issues-notes.js
@@ -163,11 +163,9 @@ export function run(conf, doc, cb) {
       issues.forEach(function(issue) {
         ghIssues[issue.number] = issue;
       }); 
-      handleIssues($ins, ghIssues, issueBase);
-      cb();
-      
     } catch(err) {
-        pub("error", err.message);
+        pub("error", err.message); 
+    } finally {
         handleIssues($ins, ghIssues, issueBase);
         cb();
     }
--- a/src/core/issues-notes.js
+++ b/src/core/issues-notes.js
@@ -141,26 +141,27 @@ export function run(conf, doc, cb) {
     issueBase = conf.issueBase;
   if ($ins.length) {
     if (conf.githubAPI) {
-      ghFetch(conf.githubAPI)
-        .then(function(json) {
+      (async function(githubAPI) {
+        try {
+          let json = await ghFetch(githubAPI);
           issueBase = issueBase || json.html_url + "/issues/";
-          return fetchIndex(json.issues_url, {
+          let issues = fetchIndex(json.issues_url, {
             // Get back HTML content instead of markdown
             // See: https://developer.github.com/v3/media/
-            Accept: "application/vnd.github.v3.html+json",    
+            Accept: "application/vnd.github.v3.html+json",
           });
-        })
-        .then(function(issues) {
+          
           issues.forEach(function(issue) {
             ghIssues[issue.number] = issue;
           });
           handleIssues($ins, ghIssues, issueBase);
           cb();
-        }).catch(err => {
+        } catch(err) {
           pub("error", err.message);
           handleIssues($ins, ghIssues, issueBase);
           cb();
-        });
+        }
+      })(conf.githubAPI)
     } else {
       handleIssues($ins, ghIssues, issueBase);
       cb();
--- a/src/core/issues-notes.js
+++ b/src/core/issues-notes.js
@@ -141,27 +141,7 @@ export function run(conf, doc, cb) {
     issueBase = conf.issueBase;
   if ($ins.length) {
     if (conf.githubAPI) {
-      (async function(githubAPI) {
-        try {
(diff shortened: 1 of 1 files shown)


marcoscaceres:
Thanks @vinicius0197! Not too late at all. Will have a look.

vinicius0197:
#1502

src/core/issues-notes.js
```diff
@@ -141,31 +141,32 @@ export function run(conf, doc, cb) {
...
-          return fetchIndex(json.issues_url, {
+          let issues = fetchIndex(json.issues_url, {
             // Get back HTML content instead of markdown
             // See: https://developer.github.com/v3/media/
-            Accept: "application/vnd.github.v3.html+json",    
+            Accept: "application/vnd.github.v3.html+json",
           });
-        })
-        .then(function(issues) {
+
```
OrionStar25: I see that you've broken the whole .then chain and converted it to behave like a synchronous function. LGTM. I just feel that we could modularize the code better so that maintaining it would become easier.

src/core/issues-notes.js
```diff
@@ -141,31 +141,32 @@ export function run(conf, doc, cb) {
...
   if ($ins.length) {
     if (conf.githubAPI) {
-      ghFetch(conf.githubAPI)
-        .then(function(json) {
+      (async function(githubAPI) {
+        try {
+          let json = await ghFetch(githubAPI);
           issueBase = issueBase || json.html_url + "/issues/";
-          return fetchIndex(json.issues_url, {
+          let issues = fetchIndex(json.issues_url, {
```
OrionStar25: Adding await before fetchIndex would ensure returning us a promise, I guess.

src/core/issues-notes.js
```diff
@@ -141,31 +141,32 @@ export function run(conf, doc, cb) {
...
             // Get back HTML content instead of markdown
             // See: https://developer.github.com/v3/media/
-            Accept: "application/vnd.github.v3.html+json",    
+            Accept: "application/vnd.github.v3.html+json",
           });
-        })
-        .then(function(issues) {
+          
           issues.forEach(function(issue) {
```
shubhshrma: This can execute before ```issues``` has actually got its value ```fetchIndex```. You need to add another waiting step here.

vinicius0197: hm, you are saying the waiting step should go right before `fetchIndex`, like @OrionStar25 mentioned?

I will make the changes according to your feedback and push again soon

shubhshrma: Yes, precisely.

src/core/issues-notes.js
marcoscaceres: For what it's worth, @OrionStar25, you are correct... this whole module is terribly written. Part of this simple exercise was to see how people deal with jumping into fairly bad code to make it better.

marcoscaceres:
@vinicius0197, this was a nice solution. Thanks for putting it together.

src/core/issues-notes.js
```diff
@@ -141,32 +141,36 @@ export function run(conf, doc, cb) {
...
+    issues.forEach(function(issue) {
+      ghIssues[issue.number] = issue;
+    }); 
+    handleIssues($ins, ghIssues, issueBase);
+    cb();
+  
+  } catch(err) {
+      pub("error", err.message);
+      handleIssues($ins, ghIssues, issueBase);
+      cb();
```
marcoscaceres: Not cb() is no longer in scope

marcoscaceres: Err, “note”

vinicius0197: Thanks for noticing that! I didn't realize I'd put the function outside run().

src/core/issues-notes.js
```diff
@@ -141,31 +141,35 @@ export function run(conf, doc, cb) {
...
-            ghIssues[issue.number] = issue;
-          });
-          handleIssues($ins, ghIssues, issueBase);
-          cb();
-        }).catch(err => {
-          pub("error", err.message);
-          handleIssues($ins, ghIssues, issueBase);
-          cb();
-        });
+        fetchFunction(conf.githubAPI);
```
marcoscaceres: See comments I left in https://github.com/w3c/respec/pull/1534#discussion_r172388979 about try/catch/finally... would probably apply here too.

marcoscaceres:
Thanks for taking the time to put this together @vinicius0197.

src/core/issues-notes.js
vinicius0197: Makes sense! I will update that and push again tomorrow, since I'm in a hurry right now. Thanks for the hint

marcoscaceres:
Closing as this was just an exercise. @vinicius0197, your solution was also quite elegant and showed you have good command over async code. Really nice work! Thanks again.

Issue #1545, opened 2018-03-01
Back to Top button doesn't work


OrionStar25:
eg :- https://www.w3.org/TR/payment-request/
The arrow button used for navigating to the top hasn't been assigned any top link in any of the specs.

marcoscaceres:
@OrionStar25, could you clarify a little bit more? I'm having trouble finding the "Back to Top" button you mentioned. If possible, please include a screenshot.

himanish-star:
![image](https://user-images.githubusercontent.com/30361728/36880692-0d109514-1df0-11e8-96f0-464becfaaee7.png)

@marcoscaceres this is the `back to top` button which is not functioning. @OrionStar25 Am I right?

marcoscaceres:
Ah, got it! ok, it actually _is_ functioning... but it's kinda useless because it links back to the "#toc", which is being displayed in the side panel. If you "collapse side bar", and click it, you can see it working. 

But technically, @OrionStar25 is correct that this is confusing. I'll just note that this is something that's being added by the W3C's "inject.js" script, so it's technically not a ReSpec issue... but, we can definitely bring this issue to the W3C's attention so they can fix it on their side. 

The project for that is: http://github.com/w3c/tr-design

himanish-star:
> it actually is functioning

:+1:

marcoscaceres:
Actually, I lied, it's a ReSpec bug: 
https://github.com/w3c/respec/blob/develop/src/core/structure.js#L147

@OrionStar25, you want to fix this?

OrionStar25:
Yeah! I'll work on this! :D

OrionStar25:
@marcoscaceres So we want the TOC when the side panel exists and top of the spec when its been collapsed. Or we could just bring it to the top of the spec in both cases as TOC mostly is concise enough (?)

marcoscaceres:
No, let’s make it simpler. Just take the reader back to the very top of the document. Check if you can either hook into the #head or change the scroll position.

OrionStar25:
PR sent!

PR #1550, opened 2018-03-02
Link back to top to title of spec

OrionStar25:
Fixes : #1545 
Since title is always present.

### DIFF
--- a/src/core/structure.js
+++ b/src/core/structure.js
@@ -145,7 +145,7 @@ export function run(conf, doc, cb) {
     }
 
     var $link = $(
-      "<p role='navigation' id='back-to-top'><a href='#toc'><abbr title='Back to Top'>&uarr;</abbr></a></p>"
+      "<p role='navigation' id='back-to-top'><a href='#title'><abbr title='Back to Top'>&uarr;</abbr></a></p>"
     );
     $("body").append($link);
   }
--- a/tests/spec/core/structure-spec.js
+++ b/tests/spec/core/structure-spec.js
@@ -17,7 +17,7 @@ describe("Core - Structure", () => {
   it("should build a ToC with default values", async () => {
     const ops = {
       config: makeBasicConfig(),
-      body: body,
+      body: body
     };
     const doc = await makeRSDoc(ops);
     // test default values
@@ -42,7 +42,7 @@ describe("Core - Structure", () => {
     // test with noTOC
     var ops = {
       config: makeBasicConfig(),
-      body: "<section class='sotd'><p>.</p></section>",
+      body: "<section class='sotd'><p>.</p></section>"
     };
     ops.config.noTOC = true;
     makeRSDoc(ops, function(doc) {
@@ -50,12 +50,10 @@ describe("Core - Structure", () => {
     }).then(done);
   });
 
-  it("should include introductory sections in ToC with tocIntroductory", function(
-    done
-  ) {
+  it("should include introductory sections in ToC with tocIntroductory", function(done) {
     var ops = {
       config: makeBasicConfig(),
-      body: body,
+      body: body
     };
     ops.config.tocIntroductory = true;
     makeRSDoc(ops, function(doc) {
@@ -63,7 +61,12 @@ describe("Core - Structure", () => {
       expect($toc.find("h2").text()).toEqual("Table of Contents");
       expect($toc.find("> ol > li").length).toEqual(6);
       expect($toc.find("li").length).toEqual(18);
-      expect($toc.find("> ol > li a").first().text()).toEqual("Abstract");
+      expect(
+        $toc
+          .find("> ol > li a")
+          .first()
+          .text()
+      ).toEqual("Abstract");
       expect($toc.find("> ol > li a[href='#intro']").length).toEqual(1);
     }).then(done);
   });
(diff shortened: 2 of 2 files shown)


src/core/structure.js
```diff
@@ -145,7 +145,7 @@ export function run(conf, doc, cb) {
     }
 
     var $link = $(
-      "<p role='navigation' id='back-to-top'><a href='#toc'><abbr title='Back to Top'>&uarr;</abbr></a></p>"
+      "<p role='navigation' id='back-to-top'><a href='#title'><abbr title='Back to Top'>&uarr;</abbr></a></p>"
```
marcoscaceres: Ah, good choice 👌

marcoscaceres:
Trying to decide if we should add a test for this... I’m thinking, check if this element was added, it links to #title, and that the spec has an element #title. That should catch any potential regressions. @OrionStar25, what do you think?

OrionStar25:
According to the user guide
https://github.com/w3c/respec/wiki/ReSpec-Editor's-Guide

It says " the title of the document is reused as the title of the specification in the resulting document's h1. That way, they are always in sync and you need not worry about specifying it twice. Including a title is required."
I thought hence its always mandatory to specify a title and hence we wouldn't have a case when the #title element isn't present. Please tell me if I'm wrong!

marcoscaceres:
You are definitely not wrong, but things/code changes. For example, we are reworking the title code in a separate PR:
 https://github.com/w3c/respec/pull/1516

So the test would assure that what you stated above remains demonstratively true. It’s a bit of insurance against our future selves:)

OrionStar25:
Then all the more reason to go ahead with a test! I have a generic doubt though, when you mention "tests", would if-else statements count as test statements? or they always have to be something like toBeTruthy() ?

marcoscaceres:
Yeah, they always have to be an assertion of some sort - if/else wouldn’t count.

OrionStar25:
Alright then, could you please tell me where I should be adding tests for this @marcoscaceres ?

marcoscaceres:
That would be the right place for tests. Note that we have a new async way of writing test, for example:

https://github.com/w3c/respec/blob/develop/tests/spec/core/best-practices-spec.js#L5

Makes them a little easier to work with.

OrionStar25:
@marcoscaceres please review this.

tests/spec/core/structure-spec.js
```diff
@@ -42,49 +42,72 @@ describe("Core - Structure", () => {
...
   });
+
+  it("should link to the title of the document", async () => {
+    const ops = {
+      config: makeBasicConfig(),
+      body
+    };
+    const doc = await makeRSDoc(ops);
+    const title = doc.getElementById("title");
+    expect(title).toEqual(!null);
```
marcoscaceres: Right now, this is checking: element === true, which is not correct (hence it’s failing). We actually want `.toBeTruthy()` or, `.not.toEqual(null)`.

OrionStar25: Oh! I'm sorry, I shouldve checked the differences between these assertions before sending this. Somehoe my tests weren't showing any errors. I'll make the change.

OrionStar25:
@marcoscaceres I changed it to .toBeTruthy()

tests/spec/core/structure-spec.js
```diff
@@ -42,49 +42,72 @@ describe("Core - Structure", () => {
...
   });
+
+  it("should link to the title of the document", async () => {
+    const ops = {
+      config: makeBasicConfig(),
+      body
+    };
+    const doc = await makeRSDoc(ops);
+    const title = doc.getElementById("title");
+    expect(title).toBeTruthy();
```
marcoscaceres: Ok, now we need to check that the actual element was added and that it links to the title.

OrionStar25: Alright, just to make sure Im understanding this right,
this test confirmed that a title element is present. The next assertion to be written would be that that arrow should link to our title. 
Is that right?

marcoscaceres: Exactly right. Remember, the goal of the tests is to make sure it exercises every line of code you wrote - so always double check each new line is being run.

OrionStar25:
I wasn't quite sure what to do after selecting the link so I just checked if it exists. :sweat_smile:

OrionStar25:
Oh, alright! I should remember to rebate my branch often, sorry about that. Is there anything left to do here?

marcoscaceres:
Nope all done! Nice work, @OrionStar25. Just waiting for TravisCI to run the tests and let us know it's ok... Travis is being very slllloooooooowwww today.

OrionStar25:
And Travis build failed 😭

marcoscaceres:
Thankfully, not related to your code (see logs). I’ll restart it and should be fine 👌

marcoscaceres:
Restarted the build.

OrionStar25:
I'm out of the city for 3 days. And logs don't show on mobile. I'll work on it, if it needs modification,  once I get back for sure 🤘

marcoscaceres:
Ok, going to take this in for now.

OrionStar25:
@marcoscaceres should anything be done here about the build error? I can't seem to understand this. Also, its still working the same way (linking back to #toc).

marcoscaceres:
It's working perfectly and was deployed a few days ago 😃 

You might have a cached version of the spec. Check:
https://w3c.github.io/payment-request/#informative-references

You can see it's all in order: 
screenshot 2018-03-14 11 21 13

PR #1551, opened 2018-03-02
Add support for former editors

### LINKED ISSUE #1097: Add support for former editors
A bit superfluous, but in some cases the editor section can be a bit of a mess. For example, here is the APG’s editor listing:

List of 7 APG editors including 3 previous editors. It shows editor’s name, company, and e-mail

ARIA’s editor listing is a bit better, but that’s mainly because their company names are shorter.

List of 4 ARIA editors. It shows editor’s name, company, and e-mail.

To compare that with the cleaner look of the HTML spec editor listing:

Separate lists for 5 current HTML editors and 3 former HTML editors. It lists their name (which is linked to either their email or website) and their company name between brackets.

Thoughts about moving towards the HTML look?

sidvishnoi:
Make editor section cleaner. Fixes #1097

I've made a new `conf` attribute called `formerEditors`. It is same as `editors`, but shown in different section.
Alternative: add an optional attribute (like `isFormer: bool` to `People`) and filter based on it (but I think adding a separate variable for former editors is better, keeps things at one place)

TODO: i10n


marcoscaceres:
I think we should go with `formerEditors`. I’ll try to review soon.

saschanaz:
Side note: You may want to add "Fixes #1097" on the PR description instead of title, as it enables automatic linking and issue closing.

src/w3c/headers.js
```diff
@@ -377,10 +378,14 @@ export function run(conf, doc, cb) {
...
     conf.editors.forEach(peopCheck);
   }
+  if (conf.formerEditors) {
+    conf.formerEditors.forEach(peopCheck);
+  }
   if (conf.authors) {
     conf.authors.forEach(peopCheck);
   }
   conf.multipleEditors = conf.editors && conf.editors.length > 1;
+  conf.multipleFormerEditors = conf.formerEditors && conf.formerEditors.length > 1;
```
marcoscaceres: Maybe: 
```JS
conf.multipleFormerEditors = Array.isArray(conf.formerEditors) && conf.formerEditors.length;
```

src/w3c/templates/cgbg-headers.js
```diff
@@ -49,6 +49,10 @@ export default conf => {
     ` : ""}
     <dt>${conf.multipleEditors ? html`${conf.l10n.editors}` : html`${conf.l10n.editor}`}</dt>
     ${showPeople(conf, "Editor", conf.editors)}
+    ${conf.formerEditors ? html`
+      <dt>${conf.multipleFormerEditors ? html`${conf.l10n.former_editors}` : html`${conf.l10n.former_editor}`}</dt>
```
marcoscaceres: nit:
```HTML
<dt>${conf.multipleFormerEditors ? conf.l10n.former_editors : conf.l10n.former_editor}</dt>
```
These are always a string, not HTML.

src/w3c/templates/headers.js
```diff
@@ -56,6 +56,10 @@ export default conf => {
     ` : ""}
     <dt>${conf.multipleEditors ? html`${conf.l10n.editors}` : html`${conf.l10n.editor}`}</dt>
     ${showPeople(conf, "Editor", conf.editors)}
+    ${conf.formerEditors ? html`
+      <dt>${conf.multipleFormerEditors ? html`${conf.l10n.former_editors}` : html`${conf.l10n.former_editor}`}</dt>
```
marcoscaceres: As above.

src/w3c/templates/show-people.js
```diff
@@ -12,23 +12,22 @@ export default (conf, name, items = []) => {
...
     } else {
       contents.push(html`<span class='p-name fn'>${p.name}</span>`);
     }
     if (p.company) {
       if (p.companyURL) {
-        contents.push(html`, <a class='p-org org h-org h-card' href='${p.companyURL}'>${p.company}</a>`);
+        contents.push(html` (<a class='p-org org h-org h-card' href='${p.companyURL}'>${p.company}</a>)`);
       } else {
-        contents.push(document.createTextNode(`, ${p.company}`));
+        contents.push(document.createTextNode(` (${p.company})`));
```
marcoscaceres: Would "html` (${p.company})`" not do the same thing (i.e. create the text node)?

src/w3c/templates/cgbg-headers.js
sidvishnoi: did that based on code above (like for `multipleEditors`). will fix them also.

sidvishnoi:
I've made the fixes as you've mentioned

The tests I've added are primarily based on tests of "editors"

tests/spec/w3c/headers-spec.js
```diff
@@ -179,6 +179,88 @@ describe("W3C — Headers", function() {
...
+          companyURL: "http://COMPANY",
+          mailto: "EMAIL",
+          w3cid: "1234",
+        }],
+      };
+      Object.assign(ops.config, newProps);
+      const doc = await makeRSDoc(ops);
+
+      expect($("dt:contains('Former Editors:')", doc).length).toEqual(0);
+
```
marcoscaceres: Sorry 😐 we are trying not to use jQuery in the project any more. Could you change these to just use querySelectorAll().

sidvishnoi: Oh. I thought we could use jQuery in tests.
No problem. I'll update this.

marcoscaceres: It actually turns out to be bad in tests too, as it adds a layer of indirection that makes things harder to debug when things go wrong.

sidvishnoi: I'm thinking of adding these helper functions to `specHelper.js`: https://gist.github.com/sidvishnoi/2614f24af7d11051451a6a244bcecefb

Please do comment on that.

marcoscaceres: Sorry, I’m not clear which helper functions you mean. Could you clarify?

sidvishnoi: As javascript doesn't have one-liners for selectors like `dt:contains(someText)` and `el.next("dd")`, I'm writing the helper functions

For example, `selectorContains(selector, text, context = document)` equivalent to jQuery's `:contains` selector

I've added them in gist above.

Unless we plan to use some other selector library like Sizzle or create all the helper functions as we need them.

marcoscaceres: I could be convinced, but I’ve personally not found a need for them.

sidvishnoi: Can you suggest alternative to the following? 
``` js
expect($("dt:contains('Former Editors:')", doc).length).toEqual(0);
```

marcoscaceres: So, a proposal elsewhere was to bring in Sizzle as a starting point. I’d be open to that - but would only want us to use it on the rare case the web platform can’t do something (e.g. “contains”)

sidvishnoi: In the test files, I've seen a huge usage of `:contains` and `el.next(tagname)`. Most of selectors are supported with `querySelector[All]`. 

So what do you suggest for the test cases for this module? Should I continue to use jQuery like selectors until we finalise on this? Or create helpers?

sidvishnoi: I will create a list of helper functions we may need that can help in making a decision whether to use some library like Sizzle or create our own (if a few).

marcoscaceres: Sounds great!

marcoscaceres: About suggestion from above:

const eds= [...doc.querySelectorAll(“dt”)].find({textContent}) => textContent.trim() === “Former Editors:”);

Or some variation of the above.

sidvishnoi: gets a little verbose and repetitive though

marcoscaceres: Yes, sometimes it definitely does. For the tests at least, the most important thing is limiting indirection (even at the cost of some repetition). But we make exceptions, like the makeRSDoc() function.

marcoscaceres:
@sidvishnoi, I had misunderstood what you mean by repetition with the lookup. I think it's totally fine to make helper functions in the tests, so long as they are super light weight. I've added a generic content lookup there for now. I also cleaned up a few things. Please take a look!

sidvishnoi:
This looks a lot better! Thank you!

Just one question. Instead of doing `[...dtElems]` multiple times, wouldn't it be better to do `const dtElems = [... doc.querySelectorAll("dt")];` once?

marcoscaceres:
Absolutely, do it! 👌

sidvishnoi:
> document this in the wiki

on it :smile:

PR #1553, opened 2018-03-03
refactor(w3c/informative): remove jQuery

saschanaz:
This one was easy.

### DIFF
--- a/src/core/base-runner.js
+++ b/src/core/base-runner.js
@@ -28,7 +28,7 @@ function toRunnable(plug) {
       }
       try {
         // Modern plugins are async or normal functions, take one argument (conf)
-        if (plug.run.length === 1) {
+        if (plug.run.length <= 1) {
           await plug.run(config);
           resolve();
         } else {
--- a/src/core/base-runner.js
+++ b/src/core/base-runner.js
@@ -27,7 +27,7 @@ function toRunnable(plug) {
         performance.mark(name + "-start");
       }
       try {
-        // Modern plugins are async or normal functions, take one argument (conf)
+        // Modern plugins are async or normal functions, take zero or one argument (conf)
         if (plug.run.length <= 1) {
           await plug.run(config);
           resolve();
--- a/src/w3c/informative.js
+++ b/src/w3c/informative.js
@@ -3,7 +3,7 @@
 import "deps/hyperhtml";
 export const name = "w3c/informative";
 
-export function run(conf) {
+export function run() {
   Array.from(document.querySelectorAll("section.informative"))
     .map(informative => Array.from(informative.children).find(child => child instanceof HTMLHeadingElement))
     .filter(heading => heading)
--- a/src/w3c/informative.js
+++ b/src/w3c/informative.js
@@ -5,7 +5,7 @@ export const name = "w3c/informative";
 
 export function run() {
   Array.from(document.querySelectorAll("section.informative"))
-    .map(informative => Array.from(informative.children).find(child => child instanceof HTMLHeadingElement))
+    .map(informative => informative.querySelector("h2, h3, h4, h5, h6"))
     .filter(heading => heading)
     .forEach(heading => {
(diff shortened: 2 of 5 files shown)


src/w3c/informative.js
```diff
@@ -1,10 +1,13 @@
...
 export const name = "w3c/informative";
 
-export function run(conf, doc, cb) {
-  $("section.informative")
-    .find("> h2:first, > h3:first, > h4:first, > h5:first, > h6:first")
-    .after("<p><em>This section is non-normative.</em></p>");
-  cb();
+export function run() {
+  Array.from(document.querySelectorAll("section.informative"))
+    .map(informative => Array.from(informative.children).find(child => child instanceof HTMLHeadingElement))
```
marcoscaceres: Nice!

src/w3c/informative.js
```diff
@@ -1,10 +1,13 @@
...
+    .filter(heading => heading)
+    .forEach(heading => {
+      heading.parentNode.insertBefore(hyperHTML`<p><em>This section is non-normative.</em></p>`, heading.nextSibling);
```
marcoscaceres: Nit: this should be a reduce.

saschanaz: I'm not very familiar with `.reduce`, how should the change look like?

src/core/base-runner.js
```diff
@@ -28,7 +28,7 @@ function toRunnable(plug) {
       }
       try {
         // Modern plugins are async or normal functions, take one argument (conf)
-        if (plug.run.length === 1) {
+        if (plug.run.length <= 1) {
```
marcoscaceres: Nit, should update comment to say zero argument.

src/w3c/informative.js
marcoscaceres: Actually, you are correct here. I misread the code. No need for reduce.

src/w3c/informative.js
marcoscaceres: I’m now kinda unsure if we shouldn’t let a selector find the headings. In addition to the array creation, we are walking potentially all the children per section, which seems a little excessive. Wdyt?

saschanaz: Hmm yes, that may be a problem, although I think an informative section should only contain ~10 direct descendants.

saschanaz: My drive was that the original jQuery selector doesn't have a widely available standard equivalent.

```
> h2:first, > h3:first, > h4:first, > h5:first, > h6:first
```

We can just ignore `:first` part (which forces the selector to stop at the first match) but not `>`, which limits the selector for the direct descendants and has a non-widely available equivalent [`:scope`](https://developer.mozilla.org/en-US/docs/Web/CSS/:scope).

But I'm starting to think that we probably also can ignore `>` as anyway we shouldn't see an indirect child headings. Wdyt?

marcoscaceres: All sections are supposed to have a heading (and ReSpec complains if they don't)... so, yeah, that could work. Let's try it.

src/core/base-runner.js
```diff
@@ -27,8 +27,8 @@ function toRunnable(plug) {
         performance.mark(name + "-start");
       }
       try {
-        // Modern plugins are async or normal functions, take one argument (conf)
-        if (plug.run.length === 1) {
+        // Modern plugins are async or normal functions, take zero or one argument (conf)
+        if (plug.run.length <= 1) {
```
marcoscaceres: nit: `<==` ... just as matter of style, we use strict equality throughout the project

marcoscaceres:
Tiny nit aside, feel free to merge.

src/core/base-runner.js
saschanaz: I think `<==` is not even a JS operator...

marcoscaceres: Argh... I quit.

saschanaz: I haven't heard there is a strict comparison operator in JavaScript, we probably should just ignore the type strictness here.

marcoscaceres: You are too kind @saschanaz. I just need more coffee.

Issue #1529, opened 2018-02-22
Highlight variable matching in algorithm


marcoscaceres:
Given a spec algorithm like:
https://www.w3.org/TR/payment-request/#dfn-update-a-paymentrequest-s-details-algorithm

If a `<var>` is set, it should be possible to click on it and see all the places where the variable is used. 

If a `<var>` is set but never used, it that's a bug.

sidvishnoi:
Hello sir,
Is this what you're looking for? https://streamable.com/i7kns (something like this, needs update on second part (set, never used))

If yes, I can create a PR for this. Hopefully you can help where this needs to be implemented.
If no, sorry I misunderstood the problem. Can you describe more?

marcoscaceres:
You nailed it @sidvishnoi. That's exactly what I was looking for! 

Let's continue the discussions over in the pull request.

PR #1554, opened 2018-03-04
Highlight variable matching in algorithm #1529

sidvishnoi:


src/core/highlight-vars.js
```diff
@@ -0,0 +1,22 @@
...
+  cb();
+}
+
+function highlightVars({ target }) {
+  const value = target.innerText;
+  const parent = target.closest(".algorithm");
+  const toHightlight = [...parent.querySelectorAll("var")]
+    .filter(el => el.innerText === value);
+  const classToAdd = toHightlight.length === 1 ? "bug" : "active";
+  toHightlight.map(el => el.classList.toggle(classToAdd));
```
marcoscaceres: This isn't really a `map`... maybe just:

```JS
[...parent.querySelectorAll("var")]
  .filter(el => el.innerText === value);
  .forEach((el, _, {length}) => {
     const classToAdd = length === 1 ? "bug" : "active";
     el.classList.toggle(classToAdd);
   });
```

However, we have a problem when exporting (as we can't export JS). We either need to see if there is a way to do this with just CSS, or we need to make sure the CSS for this is not exported out.

src/core/highlight-vars.js
```diff
@@ -0,0 +1,22 @@
+// Module core/highlight-vars
+// Highlights occurrences of a <var> in an algorithm on click
+// if a <var> is set only once (unused/undeclared), it's highlighted as "var.bug"
+
+export const name = "core/highlight-vars";
+
+export function run(conf, doc, cb) {
+  const vars = [...document.querySelectorAll(".algorithm var")];
```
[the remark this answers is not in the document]
sidvishnoi: I realized this results in better performance (0.2ms to 0.1s now)
Thank you!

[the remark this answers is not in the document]
sidvishnoi: are you saying that I should remove the `export` also?

src/core/highlight-vars.js
sidvishnoi: 1. About doing in CSS: we can use `[data-var="value"]` as selector. That will need adding a style for each variable value (during export) on page. That will make page stylesheet really long. Should I do it?
2. We can remove the `var` CSS in export altogether by adding `.removeOnSave` on stylesheet for `var`, which is really simple thing to do but removes the highlight functionality in exports.

I'll try to find out more ways for CSS only part.

sidvishnoi: I tried the first approach, and it won't work, as we can't highlight (tried on `:hover`) all the vars with same `data-var` as CSS has no parent selector.
Why don't we export JS though?

src/core/highlight-vars.js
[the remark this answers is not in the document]
marcoscaceres: No, sorry, when I said “export” below I meant exporting the generated document as just HTML without any JS.

src/core/highlight-vars.js
marcoscaceres: The W3C doesn’t allow us to. The only JS allowed in specs is “fixup.js”. However, if we prove this would be really useful, we can add the JS to fixup.js.

src/core/highlight-vars.js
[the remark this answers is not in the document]
sidvishnoi: I asked because you highlighted lines 5:8 here.

src/core/highlight-vars.js
sidvishnoi: For now, I'll just remove the highlight functionality from exported content, if that's fine.

src/core/highlight-vars.js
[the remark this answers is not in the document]
marcoscaceres: Ah, no, that’s just GitHub giving context. The line I was commenting on was 8.

src/core/highlight-vars.js
marcoscaceres: Let’s not bloat the css. Let’s just start with this JS version to show it’s useful.

js/core/css/var.css
```diff
@@ -0,0 +1,18 @@
+.algorithm var:hover {
```
marcoscaceres: `hover` won't work, as some algorithms are very long. For example: 
https://w3c.github.io/payment-request/#update-a-paymentrequest-s-details-algorithm

Ideally, want to be able to click, then scroll to see the instances.

marcoscaceres: Argh, sorry... ignore this comment.

marcoscaceres:
Ok, had a small play with it. Working well. Few things: 

- [ ]  please add "core/highlight-vars" to the plugins in the js/profile-w3c-common.js profile. 
- [ ]  once selected, I think clicking again on the document should toggle the highlighted ones off.

marcoscaceres:
Also, I'm wondering if we should only allow selecting one at a time - or if we allow selecting multiple, we should maybe change the color to differentiate a bit. Not sure.

src/core/highlight-vars.js
```diff
@@ -0,0 +1,29 @@
...
+
+import hlVars from "deps/text!core/css/var.css";
+const styleElement = document.createElement("style");
+styleElement.textContent = hlVars;
+styleElement.classList.add("removeOnSave");
+document.head.appendChild(styleElement);
+
+export const name = "core/highlight-vars";
+
+export function run(conf, doc, cb) {
```
marcoscaceres: Nit: new plugins don't require `(conf, doc, cb)`, you can just do: `run(){ ... }`. Also gets rid of the `cb()`.

sidvishnoi:
> change the color to differentiate

I think this might be a good idea. We need to pick colors that are accessible though. I'll look into it soon. (another exam on 8th)

> clicking again on the document should toggle the highlighted ones off.

Yes. Will do.

I'll also fix rest of the things you've said.

Should I remove the `.active` and `.bug` classes from `<var>` in `cleanUp` before export? Will result in cleaner HTML. Currently I just removed the styles so they are not highlighted.

src/core/highlight-vars.js
sidvishnoi: I tried using this new plugin syntax. but `document.respecIsReady` didn't resolve. Am I missing something?

marcoscaceres: could be, send up your code (or put it in gist) and I can have a look.

sidvishnoi: https://gist.github.com/sidvishnoi/dc57028a6703948ecd14a62a20317f43

marcoscaceres: yeah, that's odd... it looks fine. Just stick to the old style for now, I think. Otherwise, throw in a couple of `console.log()`s.

src/core/highlight-vars.js
```diff
@@ -0,0 +1,29 @@
...
+  for (const elem of document.querySelectorAll(".algorithm var")) {
```
marcoscaceres: so, taking a look at, for instance:
https://w3c.github.io/payment-request/#constructor

The some of the `<var>`s come from the `<p>` that defines the algorithm. For example:

```HTML
<p>
  The <code><dfn>PaymentRequest(<var>methodData</var>,
  <var>details</var>, <var>options</var>)</dfn></code> constructor <em class="rfc2119" title="MUST">MUST</em> act as follows:
</p>
```

So maybe we should match on each `section`?

sidvishnoi: Yes sir. I think we should highlight matches in the entire `<section>`. I took the "algorithm"  stated in issue too literally. Haha.

``` js
function highlightVars({ target }) {
  const value = target.innerText;
  const parent = target.closest(".algorithm"); // <<< need a match here also. to closest section
```

Will fix.
Currently working on different highlight colors.

sidvishnoi:
Here is what I've made so far: https://streamable.com/so82n

I'll need advice on what colors/styles should be used for highlighting matched variables.

marcoscaceres:
That’s looking really awesome Sid 👏 I think a lot of people are going to find this really useful.

sidvishnoi:
1. Should I write tests for this feature? I'm not sure what we'll test though.
2. Should I add some stuff to `cleanUp.js` so the exports are cleaner? Like remove empty `<var class="">` or such things are managed by some pre-deploy scripts (like cleanHTML/cleanCSS)? 


PS: For some reason, the Travis CI didn't run.

marcoscaceres:
> 1. Should I write tests for this feature? I'm not sure what we'll test though.

So, we should be able to test this with synthetic clicks: i.e., `element.click()`, then we can attach an click  handler to check the CSS was correctly set, and then `click()` the document to check that the css was removed. 

We can then check at multiple levels to check that the right colors are applied at the right level - and that clicks are not leaking out or trashing each other.

Basically, this could be the input: 
```
<section id="section1">
  <p><var id="section1-foo">foo</var></p>
  <ol>
     <li><var id="section1-bar">bar</var>
     <li><var> foo</foo>
  </ol>
</section>

<section id="section2">
  <p><var id="section2-foo">foo</var> </p>
  <ol>
     <li><var> foo</foo>
     <li><var id="section2-bar">bar</var>
  </ol>
</section>
```

> 2. Should I add some stuff to cleanUp.js so the exports are cleaner? Like remove empty <var class=""> or such things are managed by some pre-deploy scripts (like cleanHTML/cleanCSS)?

Yeah, would be good to remove the CSS classes... just to be safe, we should then rename the CSS classes to be `respec-`, so we don't accidentally trash any Editor set classes. 

In this module, you can add:
```
sub("beforesave", outputDoc => {
    // do cleanup here.
});
```

marcoscaceres:
@sidvishnoi, for a bit more context, the "beforesave" signal is emitted here: 
https://github.com/w3c/respec/blob/develop/src/ui/save-html.js#L113

And in your module, you will need:
```JS
import { sub } from "core/pubsubhub"
```

marcoscaceres:
> PS: For some reason, the Travis CI didn't run.

Travis was a bit sick yesterday... seems to be running ok today.

sidvishnoi:
I tried to write tests, but realised that the doc returned by `makeRSDoc` includes the `fixUp.js` which is doesn't include `highlight-vars.js`, yet.

Is their some alternative way that won't require me writing a totally new RSDoc for testing this module?

js/core/css/var.css
```diff
@@ -0,0 +1,21 @@
...
+  border-bottom: 1px dotted;
+  cursor: pointer;
+}
+
+var.respec-active {
+  background: var(--respec-background-color);
+  color: var(--respec-color);
+}
+
+var.respec-bug {
```
marcoscaceres: I'm thinking maybe we shouldn't do this one. There might be legit cases for having a single `var`. 
At least, we should be really sure before we show anything red... or people are going to get upset :)

src/core/highlight-vars.js
```diff
@@ -0,0 +1,83 @@
...
+import hlVars from "deps/text!core/css/var.css";
+const styleElement = document.createElement("style");
+styleElement.textContent = hlVars;
+styleElement.classList.add("removeOnSave");
+document.head.appendChild(styleElement);
+
+export const name = "core/highlight-vars";
+
+export function run(conf, doc, cb) {
+  document.addEventListener("click", initHighlight);
```
marcoscaceres: We should have a config option to toggled this on. 

```JS
if(!conf.highlightVars){
  return;
}
```

src/core/highlight-vars.js
```diff
@@ -0,0 +1,83 @@
...
+  // return current colors if applicable
+  const bg = target.style.getPropertyValue("--respec-background-color");
+  const col = target.style.getPropertyValue("--respec-color");
+  if (bg) return `${bg}${col ? `,${col}` : ""}`;
+
+  // first color preference
+  if (HL_COLORS["yellow"] === true) return "yellow";
+
+  // otherwise get some other available color
+  return Object.keys(HL_COLORS).filter(c => HL_COLORS[c] === true).pop()
```
marcoscaceres: use `find()` here.

src/core/highlight-vars.js
```diff
@@ -0,0 +1,83 @@
...
+    || "yellow";
+}
+
+function removeHighlight(el) {
+  // done so that only the respec classes are removed
+  el.classList.remove("respec-active");
+  el.classList.remove("respec-bug");
+  el.setAttribute("style", "");
```
marcoscaceres: won't need this anymore, right?

src/core/highlight-vars.js
```diff
@@ -0,0 +1,83 @@
...
+  el.removeAttribute("style");
+  if (!el.classList.value.trim()) el.removeAttribute("class");
```
marcoscaceres: this seems overly destructive, no?... another plugin in the future might do other things with `var`. We should only remove things we control in this module.

src/core/highlight-vars.js
```diff
@@ -0,0 +1,83 @@
...
+}
+
+function highlightVars(target) {
+  const innerText = target.innerText;
```
marcoscaceres: nit:
```
const { innerText } = target;
```

src/core/highlight-vars.js
```diff
@@ -0,0 +1,83 @@
...
+  cb();
+}
+
+// availability of highlight colors.
+const HL_COLORS = {};
+[ // background[,color]
+  "yellow",
+  "#726012,#fff",
+  "#c78e00",
+  "#3949ab,#fff",
```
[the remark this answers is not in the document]
sidvishnoi: the first one is `background`, the second one is for `color`.
If the second one is not given, it'll be default color (black). this is done so that we have good color contrasts

js/core/css/var.css
sidvishnoi: will remove that then.

src/core/highlight-vars.js
sidvishnoi: which one are you talking about? the style one or the class one or both?
(github context not being helpful)

sidvishnoi:
Don't know how, but the test cases passed. :rofl: (I need to check my local setup. Failed locally)

src/core/highlight-vars.js
```diff
@@ -0,0 +1,95 @@
...
+  }
+  const styleElement = document.createElement("style");
+  styleElement.textContent = hlVars;
+  styleElement.classList.add("removeOnSave");
+  document.head.appendChild(styleElement);
+
+  document.addEventListener("click", initHighlight);
+
+  // remove highlights, cleanup empty class/style attributes
+  sub("beforesave", (outputDoc) => {
```
marcoscaceres: Nit: remove parans around outputDoc

tests/spec/core/highlight-vars-spec.js
```diff
@@ -0,0 +1,68 @@
...
+      </ol>
+    </section>
+
+    <section id="section2">
+      <p><var id="section2-foo">foo</var></p>
+      <ol>
+        <li><var>foo</foo>
+        <li><var id="section2-bar">bar</var>
+      </ol>
+    </section>`
```
marcoscaceres: Nit: missing semicolon

tests/spec/core/highlight-vars-spec.js
```diff
@@ -0,0 +1,68 @@
...
+
+  const getBgColor = (el) => {
```
marcoscaceres: Nit: remove parans around el

marcoscaceres:
Closing in favor of https://github.com/w3c/respec/pull/1588

Issue #1238, opened 2017-05-24
caniuse integration 


marcoscaceres:
Would like to dynamically add caniuse integration. Basically, given a "key" or keys would like to read from:

https://github.com/Fyrd/caniuse/blob/master/features-json/

And then display the support. This could be either at the top of a spec, or per section. So:

```JS
var config = {
   caniuse: "payment-request",
}
```

And per section: 

```HTML
<section data-caniuse="payment-request">
   <h2>The foo interface</h2>
</section>
```

saschanaz:
Recently Mozilla launched [its own data repository](https://github.com/mdn/browser-compat-data) for [MDN](https://developer.mozilla.org/en-US/docs/MDN/Contribute/Structures/Compatibility_tables), it may be a possible alternative.

marcoscaceres:
Oh, nice to see! We could definitely add support for that. Looks like they are still sorting out the data format.

marcoscaceres:
Actually, this is nice checking interface compat against WebIDL in a spec. Definitely want to add support for this at some point.

marcoscaceres:
...so, the caniuse.com integration is nice for the head of the document... and the MDN compat tables are good for when WebIDL is declared.

marcoscaceres:
Closed via #1238

PR #1555, opened 2018-03-04
feat(core/caniuse): Add caniuse integration #1238

sidvishnoi:
Here is what I made for #1238. Please review.

Adds a caniuse table for a "key" which can be given in following two ways:
1. If `conf.caniuse` is set, the table is added below `.head`
2. If `data-caniuse` is set to a key, the table is added after first element where `data-caniuse` is used (on an onclick handler). `data-caniuse` is defined on a `<section>`, so table will be added after section's first element (usually section heading)

Optional:
- Use `conf.caniuseBrowsers` to set which browsers are to be shown in the table. VALUES: `"ALL"` or and array of caniuse supported browser names  (otherwise `DEFAULT_BROWSERS` will be used).

TODO:
1.(HELP WANTED): L141 (`json.notes`) does not render as html even after using `marked()` (html is shown as it is)
2. Write tests (hopefully someone else can do that)


marcoscaceres:
This is really cool, @sidvishnoi - but we definitely wouldn't want to replicate caniuse inside the spec. At most, we would just want (roughly, this can be made prettier): 

screenshot 2018-03-05 11 30 01

And if a reader wants more info, they should be able to see something like: 

screenshot 2018-03-05 11 30 07

And "more info" would link too the caniuse page, where people can get full details.

sidvishnoi:
Thanks for feedback sir.
I will definitely work on it (although the previous design took quite a time :rofl: )
So how many older browser versions should be show in "more info"?
And the main pills tell about status of latest browser? (asking to be sure)

sidvishnoi:
Updated look: https://streamable.com/e0zjm

(I hope these videos are helpful in feature testing)

marcoscaceres:
Looking much better... some immediate thoughts: 

1. the space at the top of the spec is fairly precious, so having the two rows is going to eat up too much space. Why not just fold the latest version number into the first row, like "Edge 14" "Firefox 52". 
1. knowing more than 4 browser version back is probably overkill (that's basically 6 months to a year on most browsers). That also gets need to have scroll auto. 
1. the "more info" is too disconnected visually from what it's providing more information about. I'd suggest just placing it next to the colored boxes.

marcoscaceres:
Oh, the "fetching data" should probably say, "fetching data from caniuse.com..." and have it in a `dd` element. 

Also, should we consider caching that data more aggressively? Could you check the response cache directive headers from caniuse to see if they are suitable? (at a minimum, should hopefully cache for 1 day)

sidvishnoi:
I think this ones looks even better and cleaner. 


I'm keeping 4 old + 1 current browser version in the list. Should I make this configurable or hard-code? 

> Could you check the response cache directive headers

I checked the cache directives. Sadly, GitHub (raw.githubusercontent.com) gives only `max-age=300`.
Should we look into services like gitcdn/rawgit?
As we are keeping the response in browser's `localStorage`, why do we need the cache-control from Github?

sidvishnoi:
Should I clear the stats from exported html as we want dynamic results?

marcoscaceres:
I think it’s ok to keep them. The snapshot can just contain the one row of stats. 

No problem about the caching. We can look at storing the results later.

sidvishnoi:
Should I auto-load the caniuse data bound on the `<section data-caniuse="key">` without clicking the button? It is currently loaded on click (via JS) so it won't be available in exports.
I did that because I expected many sections to have the `caniuse` tables. So the onclick was added  to reduce unnecessary requests

> The snapshot can just contain the one row of stats

The tables which are loaded before export will be interactive in exported HTML (not just first row) as it is done by CSS. I'm not sure about EPUB and XML ones.

src/core/caniuse.js
```diff
@@ -0,0 +1,172 @@
...
+ * @param  {Array:string} [browsers=DEFAULT_BROWSERS] list of browsers to show
+ */
+function canIUse(key, el, browsers = DEFAULT_BROWSERS) {
+  const url = `https://raw.githubusercontent.com/Fyrd/caniuse/master/features-json/${key}.json`;
+
+  // use data from localStorage data if valid and render
+  const cached = localStorage.getItem(`caniuse-${key}`);
+  if (cached) {
+    const stats = JSON.parse(cached);
+    const CACHE_DURATION = 24 * 60 * 60 * 1000; // in ms
```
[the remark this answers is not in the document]
marcoscaceres: We should maybe make this configurable - so we can force a refresh.

src/core/caniuse.js
```diff
@@ -0,0 +1,172 @@
...
+  and_ff: "Firefox (Android)",
+};
+
+/**
+ * main canIUse function
+ * @param  {String} key                         which api to look for
+ * @param  {Node} el                          add table after el
+ * @param  {Array:string} [browsers=DEFAULT_BROWSERS] list of browsers to show
+ */
+function canIUse(key, el, browsers = DEFAULT_BROWSERS) {
```
marcoscaceres: let's make this `async`

src/core/caniuse.js
```diff
@@ -0,0 +1,172 @@
...
+    const CACHE_DURATION = 24 * 60 * 60 * 1000; // in ms
+    if (new Date() - new Date(stats.$cacheTime) < CACHE_DURATION) {
+      return showData(key, stats, el, browsers);
+    }
+  }
+  // otherwise fetch new data and render
+  const placeholder = createPlaceholder(key, el);
+  const handleResponse = ({ stats }) => showData(key, stats, el, browsers);
+  const handleError = err => showError(err, placeholder);
+  ghFetch(url)
```
marcoscaceres: await/try/catch.

src/core/caniuse.js
```diff
@@ -0,0 +1,172 @@
...
+      <dt class="caniuse-title">
+        Can I Use this API? (${key})
+      </dt>
+      <dd>fetching data from caniuse.com... </dl>
+    </div>`;
+  return parent.parentNode.insertBefore(placeholder, parent.nextSibling);
+}
+
+function showError(err, placeholder) {
+  placeholder.style.height = "40px";
```
marcoscaceres: Maybe we should set a CSS class for this instead of inlining the style.

src/core/caniuse.js
```diff
@@ -0,0 +1,172 @@
...
+  placeholder.querySelector("dd").innerText = `Error: ${err.message}`;
```
marcoscaceres: We need to consider what the error might be... often, the reader can't do much about these errors - so maybe the best recovery would be to just add a link to the canIuse.com page. We can then route the actual error to `console.error()`.

src/core/caniuse.js
```diff
@@ -0,0 +1,172 @@
...
+
+/**
+ * render the canIUse support table
+ * @param  {String} key      API name
+ * @param  {Object} stats     CanIUse API results
+ * @param  {Node} parent   where to render table
+ * @param  {Array:string} browsers lost of browsers to show support for
+ */
+function showData(key, stats, parent, browsers) {
+  if (typeof browsers === "string") {
```
marcoscaceres: nit, un-nest this - and always help the user fix problems. 
```
if (typeof browsers !== "string") {
   return;
}
switch (browsers){
   case "ALL": 
      browsers = Object.keys(stats);
   default:  {
      const msg = "[`caniuseBrowsers`](https://github.com/w3c/respec/wiki/caniuseBrowsers) was set incorrectly. It should contain any of the follow: ${DEFAULT_BROWSERS.map(item => `"${item}"`).join(", ")}. "; 
      pub("warn", msg);
    }
}
```

marcoscaceres: Looking at this, I think we should have an object. 

```js
conf.caniuse: {
   feature: "foo", 
   browsers: ["firefox", "chrome"], // maybe we have a special "all", including everything.  
   maxAge: 10000, // user override for cache 
}
```

src/core/caniuse.js
```diff
@@ -0,0 +1,172 @@
...
+    }
+  }
+
+  // utils
+  const canIUseId = `caniuse-${key}`;
+  const permalink = `http://caniuse.com/#feat=${key}`;
+
+  // cache the response
+  stats.$cacheTime = new Date();
+  localStorage.setItem(canIUseId, JSON.stringify(stats));
```
marcoscaceres: We usually prefix "$" for jQuery things, so this might be confusing. Any reasons you chose $ here?

src/core/caniuse.js
```diff
@@ -0,0 +1,172 @@
...
+
+  const validBrowsers = browsers.filter(b => b in stats);
+  if (validBrowsers.length !== browsers.length) {
+    pub("warn", "Unsupported value(s) in `conf.caniuseBrowsers`");
```
marcoscaceres: Say which ones, and that they are ignored.

src/core/caniuse.js
```diff
@@ -0,0 +1,172 @@
...
+
+  // utils
+  const canIUseId = `caniuse-${key}`;
+  const permalink = `http://caniuse.com/#feat=${key}`;
+
+  // cache the response
+  stats.$cacheTime = new Date();
+  localStorage.setItem(canIUseId, JSON.stringify(stats));
+
+  const validBrowsers = browsers.filter(b => b in stats);
```
marcoscaceres: I think we should normalize user input. I can see people are going to do:

```JS
browsers: ["Chrome", "FireFox"] 
```

So we should map them to lower case.

src/core/caniuse.js
```diff
@@ -0,0 +1,172 @@
...
+  if (validBrowsers.length !== browsers.length) {
+    pub("warn", "Unsupported value(s) in `conf.caniuseBrowsers`");
+  }
+
+  // render the support table
+  const caniuse = hyperHTML`
+    <div class="caniuse" id="${canIUseId}">
+      <dt>Can I Use this API? (${key})</dt>
+      <div class="caniuse-stats">${validBrowsers.map(addBrowser)} <a href="${permalink}" title="Get details on caniuse.com">More info</a></div>
```
marcoscaceres: nit: line length

src/core/caniuse.js
```diff
@@ -0,0 +1,172 @@
...
+    </div>`;
+
+  const old = document.getElementById(canIUseId);
+  if (old) old.parentNode.removeChild(old);
```
marcoscaceres: nit: `if (old) old.remove();`

src/core/caniuse.js
```diff
@@ -0,0 +1,172 @@
...
+
+    const getSupport = version =>
+      browserData[version].split("#", 1)[0].trim();
+
+    const addBrowserVersion = version =>
+      `<div class="ciu-cell ${getSupport(version)}">${version}</div>`;
+
+    const browserVersions = Object.keys(browserData)
+      .sort(semverCompare)
+      .slice(-5) // 4 browser versions back + 1 current
```
marcoscaceres: Let's make this also user configurable, defaulting to 4

src/core/caniuse.js
```diff
@@ -0,0 +1,172 @@
...
+      .reverse();
+    const currentVersion = `ciu-cell ${getSupport(browserVersions[0])}`;
```
marcoscaceres: nit, just inline this below.

src/core/caniuse.js
sidvishnoi: Chose $ as a personal habit to denote system generated variables (sometimes I use _).
Will fix as while moving from localStorage to IDB

sidvishnoi:
As a little preview: https://imgur.com/a/S4lcv

marcoscaceres:
> Can I know the reasons why we don't use localStorage (I can guess a big one being synchronous)?

yeah, synchronous. People accessing the spec want to be able to start scrolling and interacting with specs immediately, so offloading asynchronously to IDB can potentially make that better. 

It's also a crappy API, in that you can't store objects... you have to JSON.stringify() etc.

src/core/caniuse.js
```diff
@@ -0,0 +1,289 @@
...
+      <div class="caniuse-col">
+        ${browserVersions.slice(1).map(addBrowserVersion)}
+      </div>
+    </div>`;
+  }
+}
+
+// promise based interface to IDB
+// supported methods: get, set, remove, clear, keys
+async function IDBCache(name, stores = ["caniuse"], version = 1) {
```
marcoscaceres: so, let's move this to another module. And yeah, this would be super useful on its own! 

Also, what you have here is basically a class... so let's convert it to one:

```JS
// store databases by name here: 
const databases = new Map();

class IDBCache {
    constructor(name, stores = [], version = 1) {
        //  if databases has "name", use it. Otherwise, 
        // open new one and store it in databases.   
    }
    async ready(){ 
        //basically your getDatabase() function
    }
    async get(){
       const db = await this.ready();
    }
    async set(){} 
    async remove(){}
    async clear(){} 
    async keys(){}
}
```

src/core/caniuse.js
```diff
@@ -0,0 +1,289 @@
...
+
+// Opportunistically insert the style into the head to reduce FOUC.
+const codeStyle = document.createElement("style");
+codeStyle.textContent = caniuseCss;
+document.head.appendChild(codeStyle);
+
+export const name = "core/caniuse";
+
+export function run(conf) {
+  normalizeConf(conf);
```
marcoscaceres: We should bail out early here. Adding the defaults when they won't do anything doesn't seem great. 

```
if(!conf.caniuse){
   return; // nothing to do
}
normalizeConf(conf);
```

sidvishnoi: I did this because we have an option to add caniuse by `<section data-caniuse>` which doesn't require `conf.caniuse` to declared.

src/core/caniuse.js
```diff
@@ -0,0 +1,289 @@
...
+}
+
+/**
+ * main canIUse function
+ * @param  {String} key              which api to look for
+ * @param  {Node} parent             adds table after parent
+ * @param {Object} conf              normalized respecConfig.caniuse
+ */
+async function canIUse(key, parent, conf) {
+  const url = `https://raw.githubusercontent.com/Fyrd/caniuse/master/features-json/${key}.json`;
```
marcoscaceres: This url should be user configurable too. Then we can test it locally, etc.

src/core/caniuse.js
```diff
@@ -0,0 +1,289 @@
...
+    console.error("[core/caniuse]", err);
+  }
+
+  // otherwise fetch new data, cache and render
+  const placeholder = createPlaceholder(key, parent);
+  try {
+    const json = await getJson(url);
+    showData(key, json.stats, conf, parent);
+    cache.set(key, { stats: json.stats, cacheTime: new Date() })
+      .catch(err => console.error("[core/caniuse] (Could not cache)", err));
```
marcoscaceres: no need for "core/caniuse", the error will contain the `.stack`

src/core/caniuse.js
```diff
@@ -0,0 +1,289 @@
...
+ * @param  {Node} parent    where to create placeholder
+ * @return {Node}           inserted placeholder
+ */
+function createPlaceholder(key, parent) {
+  const placeholder = hyperHTML`
+    <div class="caniuse" id="${`caniuse-${key}`}">
+      <dt class="caniuse-title">
+        Can I Use this API? (${key})
+      </dt>
+      <dd>fetching data from caniuse.com... </dl>
```
marcoscaceres: hyperHTML supports placeholders... I wonder if we can make better use of that functionality? I can't recall exactly what that looks like... but it will show the place holder until promises resolve.

src/core/caniuse.js
```diff
@@ -0,0 +1,289 @@
...
+  if (validBrowsers.length !== browsers.length) {
+    pub("warn", `Unsupported value(s) in \`conf.caniuse.browsers\`.
+    The following were ignored: \`[${browsers.filter(b => !(b in stats))}]\``);
+  }
+
+  // render the support table
+  const caniuse = hyperHTML`
+    <div class="caniuse" id="${canIUseId}">
+      <dt>Can I Use this API? (${key})</dt>
+      <div class="caniuse-stats">
```
marcoscaceres: this should be a `dd`

src/core/caniuse.js
```diff
@@ -0,0 +1,289 @@
...
+
+  const validBrowsers = browsers.filter(b => b in stats);
+  if (validBrowsers.length !== browsers.length) {
+    pub("warn", `Unsupported value(s) in \`conf.caniuse.browsers\`.
+    The following were ignored: \`[${browsers.filter(b => !(b in stats))}]\``);
+  }
+
+  // render the support table
+  const caniuse = hyperHTML`
+    <div class="caniuse" id="${canIUseId}">
```
marcoscaceres: can we do it without this outer div?

src/core/caniuse.js
```diff
@@ -0,0 +1,289 @@
...
+    const addBrowserVersion = version =>
+      `<div class="caniuse-cell ${getSupport(version)}">${version}</div>`;
+
+    const browserVersions = Object.keys(browserData)
+      .sort(semverCompare)
+      .slice(-(numVersions + 1)) // plus 1 current
+      .reverse();
+
+    return hyperHTML`
+    <div class="caniuse-browser">
```
marcoscaceres: I feel like this should be a `ul` and list items?

src/core/caniuse.js
sidvishnoi: `<dd class="caniuse-stats">` ?

src/core/caniuse.js
sidvishnoi: i'll look into it :)

src/core/caniuse.js
sidvishnoi: I've finally understood how to read the GitHub contexts! LOL. (need to read on files page instead on conversation one)

js/profile-w3c-common.js
```diff
@@ -70,6 +70,7 @@ define(
     "core/data-tests",
     "core/list-sorter",
     "w3c/jsonld",
+    "w3c/jsonld",
```
marcoscaceres: Oops, looks like jsonld is loaded twice

sidvishnoi: tried to solve a merge conflict, went wrong :P
will fix on next review

sidvishnoi:
I don't understand what is this build error. Local build is working fine. Need help

src/core/idb-cache.js
```diff
@@ -0,0 +1,109 @@
...
+   * @private returns a promise to get an IDBObjectStore
+   * @param {string} storeName      store name
+   * @param {string} mode           IDBTransaction.mode
+   */
+  async ready(storeName, mode) {
+    const db = await databases.get(this.name);
+    return new Promise((resolve, reject) => {
+      const transaction = db.transaction(storeName, mode);
+      transaction.onerror = () => reject(transaction.error);
+      transaction.oncomplete = () => resolve();
```
marcoscaceres: this line and the next compete for `resolve()`, which will be confusing.

src/core/idb-cache.js
```diff
@@ -0,0 +1,109 @@
...
+      return resolve(transaction.objectStore(storeName));
+    });
+  }
+
+  /**
+   * @private returns a promise to get result of IDBRequest
+   * @param {IDBRequest} request    IDBRequest interface
+   */
+  getResponse(request) {
```
marcoscaceres: if this is private, then this should not be on the class definition (as that's all public). Make it a function:

```JS
function getResponse(idbCache, request) { 

}

```

marcoscaceres:
> I don't understand what is this build error. Local build is working fine. Need help

We will need to see what Esprima is spitting out. However, don't worry about this for now :)

src/core/idb-cache.js
```diff
@@ -0,0 +1,109 @@
...
+export default class IDBCache {
+  /**
+   * @constructor
+   * @param {string} name             name of database
+   * @param {array} stores            name of stores
+   * @param {object} Options
+   *    @param {integer} version      database version
+   *    @param {string} defaultStore  name of default store
+   */
+  constructor(name, stores = [], { version = 1, defaultStore }) {
```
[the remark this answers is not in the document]
sidvishnoi: i think it should not be optional. it would be pretty useless if user didn't provide a store (didn't think about this earlier). So, I'll make this required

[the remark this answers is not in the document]
marcoscaceres: Good point! Should we then make it mandatory and require 1?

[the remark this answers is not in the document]
sidvishnoi: yes. atleast one must be required.
(unless we want to create another method like `addStore` - which adds unnecessary complexity)

src/core/idb-cache.js
sidvishnoi: didn't get what you mean here.

src/core/idb-cache.js
```diff
@@ -0,0 +1,109 @@
...
+      databases.set(this.name, dbPromise);
+    }
+  }
+
+  /**
+   * @private returns a promise to get an IDBObjectStore
+   * @param {string} storeName      store name
+   * @param {string} mode           IDBTransaction.mode
+   */
+  async ready(storeName, mode) {
```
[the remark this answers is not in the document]
sidvishnoi: should i put this promise into a separate function (`getStore`) as you said ready is doing too much work? if we put this into each method (get, set), it'll become redundant

src/core/idb-cache.js
marcoscaceres: Sorry, will try to clarify. You have .oncomplete trying to use resolve(), but the following line calls resolve(). You can only ever call resolve once, so .oncomplete, even if it fires, won’t do anything. Make sense?

src/core/idb-cache.js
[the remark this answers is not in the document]
marcoscaceres: I don’t have a strong opinion. I guess we don’t want the caller getting access to the raw store, so maybe just inline or make getStore() a private function to this module.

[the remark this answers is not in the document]
sidvishnoi: Shouldn't it be 
``` js
async ready(){
  const db = await databases.get(this.name);
  databases.set(this.name, db); // the resolved IDBDatabase, so we can access it in getStore
  return; 
}
```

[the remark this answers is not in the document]
sidvishnoi: also, shouldn't ready be also private?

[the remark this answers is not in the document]
marcoscaceres: First question, I respectfully think no because you still need to “tick” on await. So there is no advantage to changing the type held by the map. 

With second question, maybe. It depends if we want a generic means to check if the cache is ready to be used (without doing any actual operation on it). So, just:

```js
const cache = new IDBChace();
const readyPromise = cache.ready();
// do other stuff
readyPromise.then(addThings);
```

All the other operations require some interaction with the database.

sidvishnoi:
Made some changes. I hope this is what you were expecting :)

src/core/caniuse.js
```diff
@@ -0,0 +1,218 @@
+/*
+Module: "core/caniuse"
+Adds a caniuse support table for a "feature" #1238
+
+`conf.caniuse = { feature: key }` =>
+  .. the table is added before Copyright
+`<section data-caniuse=key>` =>
```
marcoscaceres: I don't think we should do this yet. There are a number of issues with the current approach that will make the HTML invalid. We should work out what this will look like first, then add this in.

sidvishnoi: sure

src/core/caniuse.js
```diff
@@ -0,0 +1,212 @@
...
+
+import { semverCompare } from "core/utils";
+import IDBCache from "core/idb-cache";
+import { pub } from "core/pubsubhub";
+import "deps/hyperhtml";
+import caniuseCss from "deps/text!core/css/caniuse.css";
+
+export const name = "core/caniuse";
+
+export async function run(conf) {
```
marcoscaceres: So, can I use is not critical spec content, so maybe we could do something like:

```JS
export async function run(conf, doc, cb) {
  cb(); // this is low priority, so let's keep going
  if (!conf.caniuse) {
    return; // nothing to do.
  }
  document.head.appendChild(hyperHTML`<style>${caniuseCss}</style>`);
  const cache = new IDBCache("respec-caniuse", ["caniuse"]);
  normalizeConf(conf);
  const { caniuse } = conf;

  if (!caniuse.feature) {
    return; // no feature to show
  }
  // Insert placeholder dt/dd into head
  // kick off fetch here too...
  await Promise.all([doc.respecIsReady, cache.ready(), /* fetch */]);

  await canIUse(
    caniuse.feature,
    document.querySelector(".head dl").lastChild,
    cache,
    caniuse
  );
}
```

src/core/caniuse.js
```diff
@@ -0,0 +1,212 @@
...
+  if (conf.caniuse.maxAge === undefined) conf.caniuse.maxAge = DEFAULTS.maxAge;
+  if (!conf.caniuse.versions) conf.caniuse.versions = DEFAULTS.versions;
+}
+
+/**
+ * main canIUse function
+ * @param  {String} key              which api to look for
+ * @param  {Node} refNode            adds table after refNode
+ * @param {IDBCache} cache           IDB cache
+ * @param {Object} conf              normalized respecConfig.caniuse
```
marcoscaceres: The `conf` contains the `key`, so we could pass `conf` as the first argument?

src/core/caniuse.js
```diff
@@ -0,0 +1,212 @@
...
+      resolve(getTableHtml(key, json.stats, conf));
+    } catch (err) {
+      const permalink = `http://caniuse.com/#feat=${key}`;
+      resolve(hyperHTML`
+        Error [core/caniuse]: ${err.message}.
+        <br>Please check directly on <a href="${permalink}">${permalink}</a>.`);
+    }
+  });
+
+  const content = hyperHTML`
```
marcoscaceres: The `title` should just be inside here:
```JS
  const content = hyperHTML`
    <dt class="caniuse-title" id="${`caniuse-${key}`}">
      Can I Use this API? (${key})
    </dt>
    <dd class="caniuse-stats">${{
      any: contentPromise,
      placeholder: "Fetching data from caniuse.com...",
    }}</dd>`;

```

src/core/caniuse.js
```diff
@@ -0,0 +1,212 @@
...
+      caniuse
+    );
+  }
+}
+
+/**
+ * MUTATES `conf.caniuse` object to hold normalized configurarion
+ * @param {Object} conf   configuration settings
+ */
+function normalizeConf(conf) {
```
marcoscaceres: We can optimize this a little bit as follows. Also, we should do the browser filtering/checks here, and just emit warnings:

```JS
function normalizeConf(conf) {
  const DEFAULTS = {
    maxAge: 24 * 60 * 60 * 1000, // 24 hours (in ms)
    browsers: ["chrome", "firefox", "safari", "edge"],
    versions: 4,
  };
  if (typeof conf.caniuse === "string") {
    conf.caniuse = { feature: conf.caniuse, ...DEFAULTS };
    return;
  } else {
    conf.caniuse = { ...DEFAULTS, ...conf.caniuse };
  }
  if (Array.isArray(conf.caniuse.browsers)) {
    conf.caniuse.browsers = conf.caniuse.browsers
      .map(b => b.toLowerCase())
      .filter(b => { /*
          warn here if invalid value
      */});
  } else if (conf.caniuse.browsers !== "ALL") {
    conf.caniuse.browsers = [...DEFAULTS.browsers];
  }
}
```

src/core/caniuse.js
```diff
@@ -0,0 +1,212 @@
...
+      throw new Error("Could not fetch GitHub resource (HTTP 404)");
+    }
+    const errorMsg = `GitHub Response not OK. Probably exceeded request limit. (HTTP ${response.status})`;
+    console.error(`${errorMsg}. Resource = ${url}`);
+    throw new Error(errorMsg);
+  }
+  return await response.json();
+}
+
+const BROWSERS = { // browser name dictionary
```
marcoscaceres: This should be a `new Map()`:

```JS
const BROWSERS = new Map([
  ["chrome", "Chrome"],
  ["firefox", "Firefox"],
  ["ie", "IE"],
  ["edge", "Edge"],
  ["android", "Android"],
  ["safari", "Safari"],
  ["opera", "Opera"],
  ["bb", "Blackberry"],
  ["and_uc", "UC (Android)"],
  ["and_ff", "Firefox (Android)"],
]);
```

Because:

```JS
conf.browsers: ["hasOwnProperty"]; // or anything in the prototype chain
```

Would return true when you do the "in" check.

marcoscaceres: Nit: move to the top of the file

src/core/idb-cache.js
```diff
@@ -0,0 +1,130 @@
...
+    stores,
+    { version, defaultStore } = { version: 1, defaultStore: null }
+  ) {
+    this.name = name;
+    if (!Array.isArray(stores) || stores.length < 1) {
+      throw new TypeError("`stores` must be an array of length atleast 1");
+    }
+    this.stores = [...stores];
+    this.version = version;
+    this.defaultStore = defaultStore || this.stores[0];
```
marcoscaceres: Maybe here we should just do the following, so `.ready` can just be a property:

```KS
this.ready = new Promise(async resolve=>{
  await databases.get(name);
  resolve();
});
```

Eventually, we should do this properly with a private field and getter. We would need to add the right Babel plugin.

marcoscaceres: updated the above

src/core/caniuse.js
sidvishnoi: Yes that sounds valid now as we longer plan to support `<section data-caniuse>` (It was used earlier like this: https://github.com/w3c/respec/pull/1555/commits/bddc74fd62889bfde10d64099546b3834db8fa6e#diff-7ac308e51d1448eeee88084fe250ff44L45)

src/core/caniuse.js
sidvishnoi: Yes Bu the `content` won't be a single node then so can't do `insertBefore` on this (Will figure this one out, currently gives:  `TypeError: Failed to execute 'insertBefore' on 'Node': parameter 1 is not of type 'Node'.`)

src/core/caniuse.js
sidvishnoi: What's the purpose of doing a `fetch` here?
We should do a fetch only if the cache is invalid so as to avoid requests on github

sidvishnoi:
I've made some changes.

I'm not sure about the second commit (idb-cache) but i just did it

src/core/caniuse.js
```diff
@@ -0,0 +1,206 @@
...
+      resolve(hyperHTML`
+        Error [core/caniuse]: ${err.message}.
+        <br>Please check directly on <a href="${permalink}">${permalink}</a>.`);
+    }
+  });
+}
+
+// TODO: replace with fetch in core/github ?
+async function getJson(url) {
+  const response = await window.fetch(url);
```
marcoscaceres: So, essentially, `getJson()` should actually be `fetchAndCache(feature){}`. Essentially, the `canIUse()` function shouldn't care where the data comes from, just that it got data to render. 

So, we should do this here instead:

 1. When this module loads, create the cache instance. That can kick off initialization of IDB in the background. 
 1. When we are about to render, `fetchAndCache(feature)` (see below). 
 1.  We create the URL, which we can actually use as the unique identifier (this is good also if we want to use the browser's [Cache API](https://developer.mozilla.org/en-US/docs/Web/API/CacheStorage), once it lands in Safari in a few months).
 1. await cache.ready.
 1. Check the cache for URL. If we have it, return the object.
 1. await, go to the network. 
 1. If error, deal with error. 
 1. await json.   
 1. Cache response.
 1. return json();

I think if we do the above here, the `canIUse()` function itself can go away. Then, this all boils down to:

```JS
hyperHTML.bind(placeholder)`${{
  any: getAndCacheJson(feature).then( json => createTableHTML(feature, json)),
  placeholder: "Fetching data from caniuse.com...",
}}`;
```

src/core/caniuse.js
```diff
@@ -0,0 +1,206 @@
...
+    return;
+  }
+  normalizeConf(conf);
+  const { caniuse } = conf;
+  if (!caniuse.feature) {
+    return; // no feature to show
+  }
+  // Opportunistically insert the style into the head to reduce FOUC.
+  document.head.appendChild(hyperHTML `<style>${caniuseCss}</style>`);
+
```
marcoscaceres: Another thing we want to do here is insert a preconnect:

```JS
import { createResourceHint } from "core/utils";

const link = createResourceHint({
  hint: "preconnect",
  href: "https://raw.githubusercontent.com",
}); 
document.head.appendChild(link);
```

This will warm up the connection for us, ready to download the JSON.

src/core/caniuse.js
```diff
@@ -0,0 +1,206 @@
...
+        <br>Please check directly on <a href="${permalink}">${permalink}</a>.`);
+    }
+  });
+}
+
+// TODO: replace with fetch in core/github ?
+async function getJson(url) {
+  const response = await window.fetch(url);
+  if (!response.ok) {
+    if (response.status === 404) {
```
marcoscaceres: use switch/case here, in case we need to add more conditions later.

src/core/caniuse.js
```diff
@@ -0,0 +1,206 @@
...
+  // Opportunistically insert the style into the head to reduce FOUC.
+  document.head.appendChild(hyperHTML `<style>${caniuseCss}</style>`);
+
+  const parent = document.querySelector(".head dl");
+  parent.appendChild(hyperHTML`
+    <dt class="caniuse-title" id="${`caniuse-${caniuse.feature}`}">
+      Can I Use this API?
+    </dt>`);
+  const placeholder = parent.appendChild(hyperHTML`
+    <dd class="caniuse-stats">fetching data from caniuse.com...</dd>`);
```
marcoscaceres: Nit, you shouldn't need the "fetching data from caniuse.com...", as it's set by hyperHTML below.

src/core/caniuse.js
```diff
@@ -0,0 +1,206 @@
...
+
+  const cache = new IDBCache("respec-caniuse", ["caniuse"]);
```
marcoscaceres: With the suggestions I made below, these two lines can go away :) Then `run()` goes back to being a sync function, which is nice.

src/core/caniuse.js
sidvishnoi: I like this one! `fetchAndCache`!

Thoughts - What if we add some  more arguments to this function (or wrap it in some other function) so we can make use of a general `fetchAndCache` function that can be used in more places?
We can pass the cache and url to it as arguments.

Also, I feel like a function with "And" its name name should be refactored to two separate functions, although this might not be a case for doing that.

A little doubt - Use URL as the key for IDB?

marcoscaceres: > Thoughts - What if we add some more arguments to this function (or wrap it in some other function) so we can make use of a general fetchAndCache function that can be used in more places?

I think we will definitely get there, but we basically want to evolve into that by road testing the code first. If we over generalize too early, we might end up over-engineering something.  

> Also, I feel like a function with "And" its name name should be refactored to two separate functions, although this might not be a case for doing that.

I don't have a strong opinion, tbh.   

> A little doubt - Use URL as the key for IDB?

I'd eventually like for us to just put things into the Cache API, which stores requests (which can be URLs) and responses. So, was trying to basically model it on that.

src/core/caniuse.js
```diff
@@ -45,22 +46,26 @@ export async function run(conf, doc, cb) {
...
   parent.appendChild(hyperHTML`
     <dt class="caniuse-title" id="${`caniuse-${caniuse.feature}`}">
       Can I Use this API?
     </dt>`);
   const placeholder = parent.appendChild(hyperHTML`
-    <dd class="caniuse-stats">fetching data from caniuse.com...</dd>`);
-
-  const cache = new IDBCache("respec-caniuse", ["caniuse"]);
-  await Promise.all([doc.respecIsReady, cache.ready]);
+    <dd class="caniuse-stats"></dd>`);
```
marcoscaceres: Is there some reason that lines 59-61 can't go in here too? Does hyperHTML get upset for some reason?

src/core/caniuse.js
```diff
@@ -45,22 +46,26 @@ export async function run(conf, doc, cb) {
...
 
   hyperHTML.bind(placeholder)`${{
-    any: canIUse(caniuse, cache),
+    any: fetchAndCacheJson(caniuse)
+      .then(stats => createTableHTML(caniuse, stats))
+      .catch(err => err),
```
marcoscaceres: we should probably console.error() here, otherwise we won't know there has been an error.

src/core/caniuse.js
sidvishnoi: That'll return two nodes (dd, dt) while `appendChild` accepts a single node (that's why I had earlier wrapped it in a div)

src/core/caniuse.js
sidvishnoi: the `console.error` are in the `try-catch` blocks already. This `catch` is here just to show the error and fallback (caniuse link) to user.
Should i remove the console.error from there and add here?

src/core/caniuse.js
```diff
@@ -35,7 +36,7 @@ const BROWSERS = new Map([ // browser name dictionary
 
 export const name = "core/caniuse";
 
-export async function run(conf, doc, cb) {
+export function run(conf, doc, cb) {
```
marcoscaceres: as this is now sync again, we can just:

```
export function run(conf) {
```

And remove `cb();`

Apologies for getting you to change this so many times. I think we got it this time tho 🤞

src/core/caniuse.js
```diff
@@ -103,60 +108,69 @@ function normalizeConf(conf) {
  * promises to get content for canIUse table
  * @param {Object} conf              normalized respecConfig.caniuse
  */
-async function canIUse(conf, cache) {
-  return new Promise(async (resolve) => {
+async function fetchAndCacheJson(conf) {
+  const cache = new IDBCache("respec-caniuse", ["caniuse"]);
```
marcoscaceres: I'm still wondering if we should just add this up on line 23. We can spin up the database, even if we don't end up using it... it's cheap. WDYT? It's ok if we don't.

src/core/caniuse.js
```diff
@@ -103,60 +108,69 @@ function normalizeConf(conf) {
...
+
+  const url = conf.apiURL ?
+    conf.apiURL.replace("{FEATURE}", conf.feature)
+    : `https://raw.githubusercontent.com/Fyrd/caniuse/master/features-json/${conf.feature}.json`;
+
+  return new Promise(async (resolve, reject) => {
```
marcoscaceres: I'm not sure we need, `return new Promise(async (resolve, reject) => ` here... looking at the code, looks like we can  naturally return or throw.

src/core/caniuse.js
sidvishnoi: It won't hurt. I'll do it.
We are calling `cache.ready` when needed anyway

src/core/caniuse.js
marcoscaceres: Ah, does hyperHTML display the error?

marcoscaceres: If yes, then it probably shouldn't. If it errors, we should resolve with the html content, but we shouldn't use `.catch()` here.

src/core/caniuse.js
sidvishnoi: why did i fail to notice that! On it!

src/core/caniuse.js
marcoscaceres: > Should i remove the console.error from there and add here?

Yes, let's remove from here. I think we can simplify the error handling a little bit more over in `fetchAndCacheJson()`, but will need to think about it.

sidvishnoi: I'll need to move the content of third try-catch into second, unless I'm missing something. The `then-catch` is there just because if the fetch (or `.json()`) fails, we can't `createTableHTML` as it expects `json`. So we show the error to user instead using that `catch`

src/core/idb-cache.js
```diff
@@ -0,0 +1,137 @@
...
+        };
+        request.onsuccess = () => resolve(request.result);
+      });
+      databases.set(this.name, dbPromise);
+    }
+  }
+
+  get ready() {
+    return new Promise(async (resolve) => {
+      const db = await databases.get(this.name);
```
marcoscaceres: Generally, a getter shouldn’t return a new instance. This is why we should set it via `this.ready = ... `.

src/core/caniuse.js
```diff
@@ -0,0 +1,238 @@
...
+      .sort(semverCompare)
+      .slice(-numVersions)
+      .reverse();
+
+    const { support, title } = getSupport(latestVersion);
+    return hyperHTML`
+      <ul class="caniuse-browser">
+        <li class="${"caniuse-cell " + support}" title="${title}">
+          ${BROWSERS.get(browser) || browser} ${latestVersion}
+          <ul class="caniuse-col">
```
marcoscaceres: This throws off the design a little bit, but this is the markup we want here. Otherwise, the lists are disjoint. We can probably make this a bit better... we should really only have one list per browser.

src/core/caniuse.js
```diff
@@ -0,0 +1,238 @@
...
+    conf.caniuse.browsers = DEFAULTS.browsers;
+  }
+  Object.assign(conf.caniuse, DEFAULTS, { ...conf.caniuse });
+  function isValidBrowser(browser) {
+    if (BROWSERS.has(browser)) {
+      return true;
+    }
+    pub(
+      "warn",
+      `Ignoring invalid browser "\`${browser}\`" in [\`respecConfig.caniuse.browsers\`](https://github.com/w3c/respec/wiki/)`
```
marcoscaceres: We need to point to the right wiki page here, once we add it.

src/core/caniuse.js
```diff
@@ -0,0 +1,238 @@
+/*
+Module: "core/caniuse"
+Adds a caniuse support table for a "feature" #1238
+
+`conf.caniuse = { feature: key }` =>
```
marcoscaceres: All this needs to move to the wiki

src/core/caniuse.js
sidvishnoi: I tried to do that at first. But I couldn't figure out CSS based hover (whatever I figured out was causing layout to move  :cry: ). Will try that again after some time.

src/core/caniuse.js
```diff
@@ -0,0 +1,238 @@
...
+    const getSupport = version => {
+      const supportKeys = browserData[version]
+        .split("#", 1) // don't care about footnotes.
+        .reduce((collector, items) => collector.concat(items.split(" ")), [])
+        .filter(item => item);
+      let titles = [];
+      for (const key of supportKeys) {
+        // Keys from https://github.com/Fyrd/caniuse/blob/master/CONTRIBUTING.md
+        switch (key) {
+          case "y":
```
sidvishnoi: what about using a Map for this?

src/core/caniuse.js
```diff
@@ -0,0 +1,238 @@
...
+   * @param {String} browser      name of browser (as in CanIUse API response)
+   * @param {Number} numVersions  number of old browser versions to show
+   * @param {Object} browserData  stats data from api response
+   */
+  function addBrowser(browser, numVersions, browserData) {
+    if (!browserData) return "";
+    const getSupport = version => {
+      const supportKeys = browserData[version]
+        .split("#", 1) // don't care about footnotes.
+        .reduce((collector, items) => collector.concat(items.split(" ")), [])
```
sidvishnoi: why not this?
``` js
      const supportKeys = browserData[version]
        .split("#", 1)[0] // don't care about footnotes.
        .split(" ")
        .filter(item => item);
```

src/core/caniuse.js
```diff
@@ -178,23 +176,61 @@ function createTableHTML(conf, stats) {
...
+      };
+    };
+    const addBrowserVersion = version => {
+      const { support, title } = getSupport(version);
+      const cssClass = "caniuse-cell " + support;
+      return `<li class="${cssClass}" title="${title}">${version}</li>`;
+    };
 
-    const browserVersions = Object.keys(browserData)
+    const [latestVersion, ...olderVersions] = Object.keys(browserData)
```
sidvishnoi: :dancer: :fire:

sidvishnoi:
Thanks for the edits :heart:

src/core/caniuse.js
```diff
@@ -0,0 +1,229 @@
...
+  ["android", "Android"],
+  ["safari", "Safari"],
+  ["opera", "Opera"],
+  ["bb", "Blackberry"],
+  ["and_uc", "UC (Android)"],
+  ["and_ff", "Firefox (Android)"],
+]);
+
+// Keys from https://github.com/Fyrd/caniuse/blob/master/CONTRIBUTING.md
+const supportTitles = new Map([
```
marcoscaceres: Much cleaner 👌

src/core/caniuse.js
```diff
@@ -0,0 +1,229 @@
...
+  function addBrowser(browser, numVersions, browserData) {
+    if (!browserData) return "";
+    const getSupport = version => {
+      const supportKeys = browserData[version]
+        .split("#", 1)[0] // don't care about footnotes.
+        .split(" ")
+        .filter(item => item);
+      let titles = [];
+      for (const key of supportKeys) {
+        if (supportTitles.has(key)) {
```
marcoscaceres: `const tiles = supportKeys.filter(key => suppotTiles.has(key)).map(key => supportTiles.get(key) `

src/core/caniuse.js
```diff
@@ -0,0 +1,238 @@
...
+}
+
+/**
+ * get HTML element for the canIUse support table
+ * @param  {Object} stats     CanIUse API results
+ * @param  {Object} conf      respecConfig.caniuse
+ */
+function createTableHTML(conf, stats) {
+  let browsers = conf.browsers;
+  if (conf.browsers === "ALL") {
```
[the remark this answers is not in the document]
sidvishnoi: Did that because we don't have all the browsers in `BROWSERS` (that are in json response)
But as we are doing a `isValidBrowser` filter, this one is here is meaningless. 

Actually we should remove the `"ALL"` feature altogether as there really are too many browsers.

src/core/caniuse.js
sidvishnoi: even cleaner :fire: :smiley: 

Anyway, have you noticed we can't do `filter(supportTitles.has).map(supportTitles.get)` (it's a function, isn't it?). It bothers me.

src/core/idb-cache.js
sidvishnoi: I'm not clearly understanding this one thing that you're saying. Can you edit the code for this?
I've added a separate PR for IDBCache here: https://github.com/w3c/respec/pull/1586

marcoscaceres: Will do!

src/core/caniuse.js
```diff
@@ -0,0 +1,213 @@
...
+  // use data from cache data if valid and render
+  const cache = await caches.open("caniuse");
+  const cached = await cache.match(url);
+  if (cached && new Date(cached.headers.get("Expires")) > new Date()) {
+    const { stats } = await cached.json();
+    return stats;
+  }
+  // otherwise fetch new data and cache
+  const response = await window.fetch(url);
+  const clonedResponse = response.clone();
```
marcoscaceres: Nit: Let's move this down 149, so it's a bit more obvious what it's used for.

src/core/caniuse.js
```diff
@@ -0,0 +1,213 @@
...
+        pub("error", msg);
+      }
+    }
+    throw new Error(
+      `Response not ok from URL ${url} (HTTP Status ${response.status})`
+    );
+  }
+  const { stats } = await response.json();
+  // set it, and forget it (there is no recovery if it throws, but that's ok).
+  const customHeaders = new Headers(response.headers);
```
marcoscaceres: Ah, I'd forgotten that headers in the clone were immutable.

src/core/caniuse.js
```diff
@@ -0,0 +1,213 @@
...
+        const msg = `Couldn't find feature "${feature}" on caniuse.com? Please check the feature key on [caniuse.com](https://caniuse.com)`;
+        pub("error", msg);
+      }
+    }
+    throw new Error(
+      `Response not ok from URL ${url} (HTTP Status ${response.status})`
+    );
+  }
+  const { stats } = await response.json();
+  // set it, and forget it (there is no recovery if it throws, but that's ok).
```
marcoscaceres: Nit: this comment should now move down

src/core/caniuse.js
```diff
@@ -0,0 +1,213 @@
...
+  const customHeaders = new Headers(response.headers);
+  customHeaders.set("Expires", new Date(new Date().valueOf() + maxAge));
```
marcoscaceres: Nit, maybe:
 
```JS
const expiryDate = new Date(new Date().valueOf() + maxAge) ;
```

Looks a bit cleaner than doing Date + number math.

src/core/caniuse.js
```diff
@@ -0,0 +1,213 @@
...
+  const cacheResponse = new Response(await clonedResponse.blob(), { headers: customHeaders });
+  cache.put(url, cacheResponse).catch(console.error);
```
marcoscaceres: If possible, please confirm this works in Edge. This capability of calling `console.error` directly is really new (though it works in Safari, Chrome, and Firefox).

marcoscaceres:
Small thing we should thinking about is adding to "core/utils" based on the discussions here (rough sketch):

```JS
class NetworkError extends Error {
  constructor(message, response) {
    super(message);
    Object.assign(this, "response", { get() { return response; } })
  }
}

export async function fetchAndCache(request, maxAge = 86400000) {
  const url = new URL(request.url);
  const cache = await caches.open(url.origin);
  const cached = await cache.match(url);
  if (cached && new Date(cached.headers.get("Expires")) > new Date()) {
    return cached;
  }
  const response = await window.fetch(url);
  if (!response.ok) {
    const msg = `Response not OK from ${url} (HTTP Status: ${response.status})`;
    throw new NetworkError(msg, response);
  }
  // cache it
  const clonedResponse = response.clone();
  const expiryDate = new Date(new Date().valueOf()) + maxAge;
  clonedResponse.headers.set("Expires", expiryDate);
  // set it, and forget it (there is no recovery if it throws, but that's ok).
  cache.put(url, cacheResponse).catch(err => console.error);
  return response;
}
```

marcoscaceres:
Oops, missing the following: 

```js
const cacheResponse = new Response(await clonedResponse.blob(), { headers: customHeaders });
```
But the basic idea is there.

saschanaz:
I think if 1) we have expired cache 2) fetch has failed then we should use cache rather than just failing.

src/core/caniuse.js
marcoscaceres: Thanks for confirming!

sidvishnoi:
Please see https://github.com/sidvishnoi/respec/commit/9d7ff2775ac6288a0656054dfe9f9d707ebab76c in https://github.com/sidvishnoi/respec/tree/ISSUE_1238-handle-error. Is this error handling fine/needed?

sidvishnoi:
Any advice on writing tests for this?

sidvishnoi:
Any advice on writing tests for this?

src/core/caniuse.js
```diff
@@ -0,0 +1,192 @@
...
+    resolve(content);
+  });
+  const definitionPair = hyperHTML.bind(document.createDocumentFragment())`
+    <dt class="caniuse-title"
+      id="${`caniuse-${caniuse.feature}`}">Can I Use this API?</dt>
+    <dd class="caniuse-stats">${{
+      any: contentPromise,
+      placeholder: "Fetching data from caniuse.com...",
+    }}</dd>`;
+  headDlElem.appendChild(definitionPair);
```
marcoscaceres: As we are running this module second last, we might as well `await contentPromise` before run() ends. Then we don’t need setTimeout() in the tests, as it’s fairly unreliable.

tests/spec/core/caniuse-spec.js
```diff
@@ -0,0 +1,133 @@
...
+
+    setTimeout(() => {
+      const link = doc.querySelector(".caniuse-stats a");
+      expect(link.textContent).toBe("caniuse.com");
+      expect(link.href).toBe("https://caniuse.com/#feat=FEATURE");
+      done();
+    }, 50);
+  });
+
+  fit("shows caniuse browser support table", async (done) => {
```
marcoscaceres: Nit: fit

src/core/caniuse.js
sidvishnoi: will do that!

sidvishnoi: Not able to run tests for placeholders now though (shown before data being fetched/error). Any tip?

marcoscaceres: Yeah, don't `any: await contentPromise,`, just: 

```
any: contentPromise,
```

And then, underneath:

```
await contentPromise;
```

marcoscaceres: Hopefully should be good now :)

sidvishnoi: The placeholder test still doesn't work :frowning_face: 
Anyway we can throttle the request?

marcoscaceres: heh, crapify :) let me try a few things...

marcoscaceres: Ah, ok... yeah, we won't be able to test for the placeholder because we only get the document after all the plugins have run. That's fine tho. So, let's drop this test.

src/core/caniuse.js
```diff
@@ -0,0 +1,193 @@
...
+    </a>`;
+
+  /**
+   * add a browser and it's support to table
+   * @param {String} browser      name of browser (as in CanIUse API response)
+   * @param {Number} numVersions  number of old browser versions to show
+   * @param {Object} browserData  stats data from api response
+   */
+  function addBrowser(browser, numVersions, browserData) {
+    if (!browserData) return "";
```
marcoscaceres: Was testing, and this causes a bug because hyperHTML is expecting DOM Nodes. So, these need to be filtered out. Just `return;`  here instead.

src/core/caniuse.js
```diff
@@ -0,0 +1,193 @@
...
+ * get HTML element for the canIUse support table
+ * @param  {Object} stats     CanIUse API results
+ * @param  {Object} conf      respecConfig.caniuse
+ */
+function createTableHTML(conf, stats) {
+  // render the support table
+  return hyperHTML`
+    ${conf.browsers.map(browser =>
+      addBrowser(browser, conf.versions, stats[browser])
+    )}
```
marcoscaceres: Here, `.filter(elem => elem)` (see below)

tests/spec/core/caniuse-spec.js
```diff
@@ -0,0 +1,133 @@
...
+
+    const title = doc.getElementById("caniuse-FEATURE");
+    const placeholder = title.nextElementSibling;
+    expect(title.textContent).toBe("Can I Use this API?");
+    expect(placeholder.textContent).toBe("Fetching data from caniuse.com...");
+    expect(placeholder.localName).toBe("dd");
+  });
+
+  it("uses meaningful defaults", async () => {
+    const ops = makeStandardOps({ caniuse: "FEATURE" });
```
marcoscaceres: For most of these, we probably want to override with `apiURL`, otherwise, we just end up with a bunch of 404s. However, if we want to do a "real" run, we can grab a well-known feature here.

src/core/caniuse.js
sidvishnoi: should add this as a test also?

marcoscaceres:
Tested without the setTimeout(): 
https://gist.github.com/marcoscaceres/c6306924b077e015ddacfa7e01937e0f

Mostly Ok, but hit that bug I mentioned in the comments.

sidvishnoi:
> Tested without the setTimeout():

Oh. I just pushed same

sidvishnoi:
> Mostly Ok, but hit that bug I mentioned in the comments.

Which bug? 404s?

PR #1558, opened 2018-03-06
chore(package): build, then start server

marcoscaceres:


marcoscaceres:
I've gotten a few emails of people confused about not seeing their changes when they `npm start`. This should hopefully help. 

Eventually, we should set up a proper service that both spins up the server and runs babel watch too. Otherwise, it's annoying because you need have two console tabs open at once (one for the http server and one for `babel watch`).

sidvishnoi:
Almost related: Do I have to run `npm run build:respec-w3c-common` each time when I make a change to some plugin? `babel:watch` only builds the plugins and not the profile.
Could be helpful during development or maybe I'm missing something.

> Side note: I'm currently using this as a workaround

``` bash
while inotifywait -e close_write PATH_TO_FILE_I_AM_EDITING; 
  do npm run build:respec-w3c-common; 
done
```

marcoscaceres:
> Almost related: Do I have to run npm run build:respec-w3c-common each time when I make a change to some plugin?

No, that's only to build the actual minified version of the code. 

> babel:watch only builds the plugins and not the profile.

It's only necessary to build the profile when deploying. If things are not showing up, it usually means the browser's cache is being overly aggressive (best to disable it in your browser's developer's tools).

PR #1560, opened 2018-03-07
refactor(core/issues-notes): use async/await + simplify

goelrohan6:
w3c/respec#1502

Converting code to async await as stated in https://github.com/w3c/respec/issues/1502#issuecomment-368250097

### DIFF
--- a/src/core/issues-notes.js
+++ b/src/core/issues-notes.js
@@ -22,7 +22,7 @@ function handleIssues($ins, ghIssues, conf) {
       "<div><h2>" + conf.l10n.issue_summary + "</h2><ul></ul></div>"
     ),
     $issueList = $issueSummary.find("ul");
-  $ins.each(function(i, inno) {
+  $ins.each(function (i, inno) {
     var $inno = $(inno),
       isIssue = $inno.hasClass("issue"),
       isWarning = $inno.hasClass("warning"),
@@ -46,15 +46,15 @@ function handleIssues($ins, ghIssues, conf) {
     // wrap
     if (!isInline) {
       var $div = $(
-          "<div class='" +
-            report.type +
-            (isFeatureAtRisk ? " atrisk" : "") +
-            "'></div>"
-        ),
+        "<div class='" +
+        report.type +
+        (isFeatureAtRisk ? " atrisk" : "") +
+        "'></div>"
+      ),
         $tit = $(
           "<div role='heading' class='" +
-            report.type +
-            "-title'><span></span></div>"
+          report.type +
+          "-title'><span></span></div>"
         ),
         text = isIssue
           ? isFeatureAtRisk ? "Feature at Risk" : conf.l10n.issue
@@ -78,7 +78,7 @@ function handleIssues($ins, ghIssues, conf) {
                 .find("span")
                 .wrap($("<a href='" + conf.atRiskBase + dataNum + "'/>"));
             }
-            ghIssue = ghIssues[dataNum];
+            ghIssue = ghIssues.get(Number(dataNum));
             if (ghIssue && !report.title) {
               report.title = ghIssue.title;
             }
@@ -97,8 +97,8 @@ function handleIssues($ins, ghIssues, conf) {
             $li.append(
               $(
                 "<span style='text-transform: none'>: " +
-                  report.title +
-                  "</span>"
+                report.title +
+                "</span>"
               )
             );
           }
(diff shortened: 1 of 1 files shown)


src/core/issues-notes.js
```diff
@@ -139,28 +139,28 @@ export function run(conf, doc, cb) {
...
+  async function asyncFetch() {
+    try {
+      const json = await ghFetch(conf.githubAPI);
+      const issueUrl = issueBase || json.html_url + "/issues/";
+      const issues = await fetchIndex(json.issues_url, {
+        // Get back HTML content instead of markdown
+        // See: https://developer.github.com/v3/media/
+        Accept: "application/vnd.github.v3.html+json",
+      });
+      ghIssues = issues.reduce((issuesObj, issue) => {
```
marcoscaceres: Re-assinging `ghIssues` seems redundant, as we already initialized it to an empty object above.

src/core/issues-notes.js
```diff
@@ -139,28 +139,28 @@ export function run(conf, doc, cb) {
   var $ins = $(".issue, .note, .warning, .ednote"),
     ghIssues = {},
     issueBase = conf.issueBase;
+  async function asyncFetch() {
```
marcoscaceres: I'm not a huge fan of relying on the variables in outer scope, tbh. Can we move `async` up whatever contains this function?  The we won't need `asyncFetch` at all.

src/core/issues-notes.js
goelrohan6: you want me to over ride the global object inside reduce?

This doesn't seem nice as recuce should have it's own local variable inside it

marcoscaceres: > you want me to over ride the global object inside reduce? 

Consider, it's "reducing prop/value into", but not overriding. ghIssues remains the same reference.

src/core/issues-notes.js
goelrohan6: we are using  ``` if ( $ins.length ) ``` that is in the ``` run ``` and then ``` asyncFetch ``` is called.
so essentially we are just invoking the funcion inside if condition, how is ``` asyncfetch ``` relying on the global variables?

Am I missing something, correct me if i'm wrong.

I didn't get your solution.

marcoscaceres: > relying on the global variables?

Sorry, bad terminology on my part... "global" in the sense of the function scope. That can get confusing, specially when things change async. 

to be clear, we don't want things like this:

```JS
function run(){
   var x;
   async function y(){
        x++;
   }
   x = 1;
   y().then(somethingElse);
}
```

We want:

```JS
async function run(){
   var x;
   x = 1;
   x++;
   await y();
   await somethingElse();
}
```

marcoscaceres: couple of fixes above... hopefully that's more clear.

goelrohan6: got it, made the changes

src/core/issues-notes.js
goelrohan6: thanks, made the changes

goelrohan6:
@marcoscaceres i made the changes requested by you in 2nd commit 
1) removed brackets after finally
2) passing global ghissues in reduce
3) removed asyncfetch function and made run async instead

Can this be merged now?

marcoscaceres:
This is just a coding exercise to evaluate GSoC coding skills and participation style - we might merge one, but would need review and agreement from all participants, as we need to evaluate if it’s the best solution (and everyone worked together to come up with the final design).

goelrohan6:
const issuesUrl was declared in try which I was unable to use in finally, so changed it to var.
Should work now @marcoscaceres

src/core/issues-notes.js
```diff
@@ -143,7 +143,7 @@ export async function run(conf, doc, cb) {
     if (conf.githubAPI) {
       try {
         const json = await ghFetch(conf.githubAPI);
-        const issueUrl = issueBase || json.html_url + "/issues/";
+        var issueUrl = issueBase || json.html_url + "/issues/";
```
marcoscaceres: Can you explain why var here?

goelrohan6: As const is blocked scoped and I declared it in try, I couldn't use the same variable in finally so had  to make it var. 
Is there another turn around?

marcoscaceres: I'll have a look, but in principle, if we need a `var` then it's likely there is a bug in the design somewhere.

goelrohan6: I can make another wrapper function which would take run function as argument and return run.catch() and we can handle errors there then. 
What do you say @marcosscaceres ?

goelrohan6: we can seperate it in another functinon like before and then use .catch on it ? @marcoscaceres

marcoscaceres: No, that would not be ideal either. This module is not doing anything special that would warrant it having a different design than any other module.

goelrohan6: re-assigning the previous global variable is also not an option, right @marcoscaceres

marcoscaceres:
Ok, so, I moved `handleIssues()` and simplified run a little bit. So, now we have two paths: 

1. if we have a github URL, let's async grab the github issues. 
1. otherwise, just process the issues in the spec. 

This gets rid of the whole `finally` part, doesn't require any tricky variable handling, etc.

src/core/issues-notes.js
```diff
@@ -14,158 +14,150 @@ import { pub } from "core/pubsubhub";
...
-          cb();
-        }).catch(err => {
-          pub("error", err.message);
-          handleIssues($ins, ghIssues, issueBase);
-          cb();
-        });
-    } else {
-      handleIssues($ins, ghIssues, issueBase);
-      cb();
+  if (conf.githubAPI) {
```
marcoscaceres: I'm still not super happy with this part... I think we should ONLY got to network if:

```
if (conf.githubAPI && document.querySelector("issue[data-number]") ){

}
```

src/core/issues-notes.js
```diff
@@ -14,155 +14,151 @@ import { pub } from "core/pubsubhub";
...
+  const $ins = $(".issue, .note, .warning, .ednote");
+  const ghIssues = {};
+  const { issueBase } = conf;
+  if (!$ins.length) {
+    return; // nothing to do.
+  }
+  if (conf.githubAPI) {
+    try {
+      const json = await ghFetch(conf.githubAPI);
+      const issueUrl = issueBase || json.html_url + "/issues/";
```
goelrohan6: issueUrl is not being used anywhere, we declared it so that I can pass it to ``` handleIssues ``` function @marcoscaceres

src/core/issues-notes.js
marcoscaceres: Another significant improvement would be to _only_ fetch the issues in the spec... we are downloading all the issues for the repo... that's hugely wasteful.

marcoscaceres:
Ok, now I'm liking this a lot more. `handleIssues()` is still a complete train wreck, but at least we are only fetching just what we need. We've also make `run()` super clean. 

Can other folks take a look and let me know if they have more suggestion?

goelrohan6:
looks good to me @marcoscaceres, much more optimized and no more tricks here and there handling variables

PR #1561, opened 2018-03-07
feat(core/biblio): support different bibliographical styles

### LINKED ISSUE #857: Support different bibliographical styles

OrionStar25:
Fixes #857. 

Adds support for APA, MLA, and retains ReSpec's custom citation style as the default.

### DIFF
--- a/src/core/biblio.js
+++ b/src/core/biblio.js
@@ -235,7 +235,6 @@ function bibref(conf) {
         } else {
           key = refcontent.aliasOf;
           refcontent = conf.biblio[key];
-          { biblioStyle: refstyle }  = conf;
           circular[key] = true;
         }
       }
--- a/src/core/biblio.js
+++ b/src/core/biblio.js
@@ -93,24 +93,76 @@ export function wireReference(rawRef, target = "_blank") {
   `;
 }
 
-export function stringifyReference(ref) {
+// Author, A. (Year, Month Date of Publication). Article title. Retrieved from URL
+function entryToAPA(ref) {
+  const authors =
+    ref.authors && ref.authors.length
+      ? `${ref.authors.join("; ")} ${ref.etAl ? " et al" : ""}.`
+      : "";
+  const date = ref.date ? `(${ref.date}).` : "";
+  const title = ref.href
+    ? `<a href="${ref.href}"><cite>${ref.title}</cite></a>. `
+    : `${ref.title}`;
+  const url = ref.href
+    ? `Retrieved from URL: <a href="${ref.href}">${ref.href}</a>`
+    : "";
+
+  return `${authors} ${date} ${title} ${url}`;
+}
+
+//Author’s Last name, First name. “Title of the Article or Individual Page.” Title of the website, Name of the publisher, Date of publication, URL.
+function entryToMLA(ref) {
+  const authors =
+    ref.authors && ref.authors.length
+      ? `${ref.authors.join("; ")} ${ref.etAl ? " et al" : ""}.`
+      : "";
+  const date = ref.date ? `${ref.date}, ` : "";
+  const title = ref.href
+    ? `<a href="${ref.href}"> "<cite>${ref.title}</cite>" </a>. `
+    : `${ref.title}`;
+  const url = ref.href ? `<a href="${ref.href}">${ref.href}</a>` : "";
+  const publisher = ref.publisher
+    ? `${ref.publisher} ${ref.publisher.endsWith(".")}`
+    : "";
+
+  return `${authors} ${title} ${publisher} ${date} ${url}`;
+}
+
+// what ReSpec currently does...
+function entryToW3C(ref) {
+  const title = ref.href
+    ? `<a href="${ref.href}"> <cite>${ref.title}</cite> </a>. `
+    : `${ref.title}`;
+  const authors =
+    ref.authors && ref.authors.length
+      ? `${ref.authors.join("; ")} ${ref.etAl ? " et al" : "."}.`
+      : "";
+  const publisher = ref.publisher
+    ? `${ref.publisher} ${ref.publisher.endsWith(".")}`
+    : "";
+  const date = ref.date ? `${ref.date}. ` : "";
+  const status = ref.status
+    ? (REF_STATUSES.get(ref.status) || ref.status) + ". "
+    : "";
+  const url = ref.href ? `URL: <a href="${ref.href}">${ref.href}</a>` : "";
+
+  return `${title} ${authors} ${publisher} ${date} ${status} ${url}`;
+}
+
+export function stringifyReference(ref, style) {
   if (typeof ref === "string") return ref;
-  let output = `<cite>${ref.title}</cite>`;
-  if (ref.href) {
-    output = `<a href="${ref.href}">${output}</a>. `;
+  const nomarlizedRef = Object.assign({}, defaultsReference, ref);
+  let output = "";
+  switch (style) {
+    case "APA":
+      output = entryToAPA(nomarlizedRef);
+      break;
+    case "MLA":
+      output = entryToMLA(nomarlizedRef);
+      break;
+    default:
+      output = entryToW3C(nomarlizedRef);
   }
-  if (ref.authors && ref.authors.length) {
-    output += ref.authors.join("; ");
-    if (ref.etAl) output += " et al";
-    output += ".";
-  }
-  if (ref.publisher) {
-    const publisher = ref.publisher + (/\.$/.test(ref.publisher) ? "" : ".");
-    output = `${output} ${publisher} `;
-  }
-  if (ref.date) output += ref.date + ". ";
-  if (ref.status) output += (REF_STATUSES.get(ref.status) || ref.status) + ". ";
-  if (ref.href) output += `URL: <a href="${ref.href}">${ref.href}</a>`;
   return output;
 }
 
@@ -154,6 +206,7 @@ function bibref(conf) {
         .appendTo($dl);
       var $dd = $("<dd></dd>").appendTo($dl);
       var refcontent = conf.biblio[ref];
+      var refstyle = conf.biblioStyle[key];
       var circular = {};
       var key = ref;
       circular[ref] = true;
@@ -165,13 +218,14 @@ function bibref(conf) {
         } else {
           key = refcontent.aliasOf;
           refcontent = conf.biblio[key];
+          refstyle = conf.biblioStyle[key];
           circular[key] = true;
         }
       }
       aliases[key] = aliases[key] || [];
       if (aliases[key].indexOf(ref) < 0) aliases[key].push(ref);
       if (refcontent) {
-        $dd.html(stringifyReference(refcontent) + "\n");
+        $dd.html(stringifyReference(refcontent, refstyle) + "\n");
       } else {
         if (!badrefs[ref]) badrefs[ref] = 0;
         badrefs[ref]++;
--- a/src/core/biblio.js
+++ b/src/core/biblio.js
@@ -93,6 +93,82 @@ export function wireReference(rawRef, target = "_blank") {
   `;
 }
 
+// const html = (...args) => hyperHTML.bind(doc.createDocumentFragment())(...args);
+
+// // Author, A. (Year, Month Date of Publication). Article title. Retrieved from URL
+// function entryToAPA(ref) {
+//   const { title } = ref.title;
+//   const authors =
+//     ref.authors && ref.authors.length
+//       ? `${ref.authors.join("; ")} ${ref.etAl ? " et al" : ""}.`
+//       : "";
+//   const date = ref.date ? `(${ref.date})` : "";
+//   return html`
+//     ${authors} ${date} ${
+//     ref.href
+//       ? html`<a href="${ref.href}"><cite>${title}</cite></a>.`
+//       : `${ref.title}`
+//   }
+//     ${
+//       ref.href
+//         ? html`Retrieved from URL: ${html`<a href="${ref.href}">${
+//             ref.href
+//           }</a>`}`
+//         : ""
+//     }
+//   `;
+// }
+
+// //Author’s Last name, First name. “Title of the Article or Individual Page.” Title of the website, Name of the publisher, Date of publication, URL.
+// function entryToMLA(ref) {
+//   const authors =
+//     ref.authors && ref.authors.length
+//       ? `${ref.authors.join("; ")} ${ref.etAl ? " et al" : ""}.`
+//       : "";
+//   const date = ref.date ? `${ref.date}, ` : "";
+//   const title = ref.title;
+//   const publisher = ref.publisher
+//     ? `${ref.publisher} ${ref.publisher.endsWith(".")}`
+//     : "";
+
+//   return html`
+// 	${authors} ${
+//     ref.href
+//       ? html`<a href="${ref.href}"><cite>${title}</cite></a>.`
+//       : `${ref.title}`
+//   } ${publisher} ${date} ${
+//     ref.href ? html`<a href="${ref.href}">${ref.href}</a>` : ""
+//   }`;
+// }
+
+// //what ReSpec currently does...
+// function entryToW3C(ref) {
+//   const title = ref.title;
+//   const authors =
+//     ref.authors && ref.authors.length
+//       ? `${ref.authors.join("; ")} ${ref.etAl ? " et al" : "."}.`
+//       : "";
+//   const publisher = ref.publisher
+//     ? `${ref.publisher} ${ref.publisher.endsWith(".")}`
+//     : "";
+//   const date = ref.date ? `${ref.date}. ` : "";
+//   const status = ref.status
+//     ? (REF_STATUSES.get(ref.status) || ref.status) + ". "
+//     : "";
+
+//   return html`
+//   	${
+//       ref.href
+//         ? html`<a href="${ref.href}"><cite>${title}</cite></a>.`
+//         : `${ref.title}`
+//     } ${authors} ${publisher} ${date} ${status} ${
+//     ref.href ? html`<a href="${ref.href}">${ref.href}</a>` : ""
+//   }`;
+// }
+
+
+
+
 // Author, A. (Year, Month Date of Publication). Article title. Retrieved from URL
 function entryToAPA(ref) {
   const authors =
--- a/src/core/biblio.js
+++ b/src/core/biblio.js
@@ -206,7 +206,7 @@ function bibref(conf) {
         .appendTo($dl);
       var $dd = $("<dd></dd>").appendTo($dl);
       var refcontent = conf.biblio[ref];
-      var refstyle = conf.biblioStyle[key];
+      var refstyle = conf.biblioStyle;
       var circular = {};
       var key = ref;
       circular[ref] = true;
@@ -218,7 +218,7 @@ function bibref(conf) {
         } else {
           key = refcontent.aliasOf;
           refcontent = conf.biblio[key];
-          refstyle = conf.biblioStyle[key];
+          refstyle = conf.biblioStyle;
           circular[key] = true;
         }
       }
--- a/src/core/biblio.js
+++ b/src/core/biblio.js
@@ -93,97 +93,30 @@ export function wireReference(rawRef, target = "_blank") {
   `;
 }
 
-// const html = (...args) => hyperHTML.bind(doc.createDocumentFragment())(...args);
-
-// // Author, A. (Year, Month Date of Publication). Article title. Retrieved from URL
-// function entryToAPA(ref) {
-//   const { title } = ref.title;
-//   const authors =
-//     ref.authors && ref.authors.length
-//       ? `${ref.authors.join("; ")} ${ref.etAl ? " et al" : ""}.`
-//       : "";
-//   const date = ref.date ? `(${ref.date})` : "";
-//   return html`
-//     ${authors} ${date} ${
-//     ref.href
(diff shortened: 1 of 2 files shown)


OrionStar25:
Should I move on to writing a test @marcoscaceres?

marcoscaceres:
Yes please! Let me know what you plan to tests. Happy to provide your with additional ideas.

OrionStar25:
I have a doubt. Right now what happens is that wire Reference creates an Html frame for our stringifyReference strings to fit into.
But now we are giving an option to make that HTML frame flexible, so shouldn't wireReference be changed too?

OrionStar25:
Also, localBiblio should now have a "style" configuration present.  
Checking biblio-spec for what can be done.

OrionStar25:
I was thinking  "style" should be sent to wireReference and then according to the chosen style, an appropriate HTML frame is returned. Then stringifyReference adds our output string in the frame.
Does that make sense?

marcoscaceres:
Ah, so instead of `localBiblio` having the configuration option, we should just have a global config option `biblioStyle`.

marcoscaceres:
global meaning `conf.biblioStyle`

OrionStar25:
Oh yes that's better. Alright so 2 extra test cases shall be added in biblio-spec.js for APA and MLA. Then I'll check if the chosen style is only getting implemented. I'm.not quite sure if I should be rewriting any assertion mentioned there, could you please help me with that?

marcoscaceres:
Only write new tests cases, as all existing test cases should continue to work exactly as before. If any existing ones fail, then that’s a bug 🐛

OrionStar25:
@marcoscaceres , I used `push --force` hence had to reopen this. I didn't add a test for MLA as the structure for MLA is being tested twice already (when publisher is present). I couldn't find where we define our `global conf`, but I did pass it as a parameter.

tests/spec/core/biblio-spec.js
```diff
@@ -102,6 +107,18 @@ describe("W3C — Bibliographic References", () => {
     expect(ref.textContent).toMatch(/Publisher Here\.\s/);
   });
 
+  it("displays the reference for APA citation", async() => {
+    // Make sure the reference is added.
+    let ref = doc.querySelector("#bib-TestRef4 + dd");
+    expect(ref).toBeTruthy();
+    // This prevents Jasmine from taking down the whole test suite if SpecRef is down.
+    if (!specRefOk) {
```
marcoscaceres: We shouldn't need this here, as this is a local reference.

src/core/biblio.js
```diff
@@ -93,24 +93,76 @@ export function wireReference(rawRef, target = "_blank") {
   `;
 }
 
-export function stringifyReference(ref) {
+// Author, A. (Year, Month Date of Publication). Article title. Retrieved from URL
+function entryToAPA(ref) {
```
marcoscaceres: So, based on @saschanaz's feedback, we should just be treating these as HTML objects, and not text... so we need refactor these like so:

```JS
const html = (...args) => hyperHTML.bind(doc.createDocumentFragment())(...args)

function entryToAPA(ref) {
  const { title } = ref;
  const authors =
    ref.authors && ref.authors.length
      ? `${ref.authors.join("; ")} ${ref.etAl ? " et al" : ""}.`
      : "";
  const date = ref.date ? `(${ref.date})` : "";
  return html`
    ${authors} ${date} <a href="${ref.href}"><cite>${title}</cite></a>
    ${ref.href ? html`Retrieved from URL: ${html`<a href="${ref.href}">${ref.href}</a>`}` : ""}
  `;
}

function entryToMLA(ref) {
   // as above, but for MLA
}

function entryToW3C(ref) {
  // as above, but for W3C...
}
```

The we can just add them directly to the document as real HTML elements.

tests/spec/core/biblio-spec.js
OrionStar25: Alright, so no assertions needed?

marcoscaceres: just the `if (!specRefOk) { ... ` can go away, as it doesn't serve any useful purpose here. `specRefOk` is for checking if we can talk to the Specref server - but here, we don't talk to the server.

OrionStar25:
@marcoscaceres I'm stuck :disappointed:  . I can't seem to get past these build errors. Also, its making biblio-db.js to fail.  I haven't changed `stringifyReference` yet, because I wanted the tests to pass first.

OrionStar25:
@marcoscaceres

src/core/biblio.js
```diff
@@ -93,24 +93,152 @@ export function wireReference(rawRef, target = "_blank") {
   `;
 }
 
-export function stringifyReference(ref) {
+// const html = (...args) => hyperHTML.bind(doc.createDocumentFragment())(...args);
```
marcoscaceres: We should probably get rid of this commented out code.

tests/spec/core/biblio-spec.js
```diff
@@ -1,10 +1,11 @@
 "use strict";
 describe("W3C — Bibliographic References", () => {
-  const config = {
+  //Configuration for default of ReSpec
+  const conf1 = {
```
marcoscaceres: call this, "confDefault"

tests/spec/core/biblio-spec.js
```diff
@@ -14,37 +15,147 @@ describe("W3C — Bibliographic References", () => {
...
       },
       BARBAR: {
-        title: "The BARBAR Spec",
+        title: "The BARBAR Spec"
+      }
+    }
+  };
+
+  //Configuration for APA citation
+  const conf2 = {
```
marcoscaceres: here, you can just do:

```JS
const confAPA = {
   ...confDefault,
   biblioStyle: "APA",
};

const confMLA = {
   ...confDefault,
   biblioStyle: "MLA",
};
```

So then you don't need all the duplicate code ✨

tests/spec/core/biblio-spec.js
```diff
@@ -14,37 +15,147 @@ describe("W3C — Bibliographic References", () => {
...
+      FOOBARGLOP: {
+        aliasOf: "BARBAR"
+      },
+      BARBAR: {
+        title: "The BARBAR Spec"
+      }
+    }
+  };
+
+  const body1 = `
```
marcoscaceres: You should be able to use just 1 body, don't need 3. Each body can then have a different citation style, based on the config.

tests/spec/core/biblio-spec.js
```diff
@@ -58,13 +169,43 @@ describe("W3C — Bibliographic References", () => {
...
+
   afterAll(flushIframes);
   const bibRefsURL = new URL("https://specref.herokuapp.com/bibrefs");
 
-  let doc;
+  let doc1, doc2, doc3;
   let specRefOk;
   beforeAll(async () => {
-    doc = await makeRSDoc({ config, body });
+    doc1 = await makeRSDoc({ conf1, body1 });
```
marcoscaceres: rename: "docDefault", "docMLA", "docAPA"... so it's clear which is which.

src/core/biblio.js
marcoscaceres: Ah, sorry... this is the updated code... can you uncomment this, and remove the other stuff?

tests/spec/core/biblio-spec.js
OrionStar25: Alright, but how do I manage the change in number of tests cases? Or should I reduce all. To 3 test cases?

tests/spec/core/biblio-spec.js
OrionStar25: Spread the values, got it.

OrionStar25:
@marcoscaceres made the requested changes.

tests/spec/core/biblio-spec.js
```diff
@@ -36,6 +37,13 @@ describe("W3C — Bibliographic References", () => {
         href: "http://test.com",
         publisher: "Publisher Here",
       },
+      TestRef5: {
```
marcoscaceres: This should be remade to "RefWithOnlyHref"

tests/spec/core/biblio-spec.js
```diff
@@ -36,6 +37,13 @@ describe("W3C — Bibliographic References", () => {
...
+        href: "http://test.com",
+      },
+      TestRef6: {
```
marcoscaceres: this one should be renamed to "RefWithOnlyHrefAndAuthor"

src/core/biblio.js
```diff
@@ -154,6 +223,7 @@ function bibref(conf) {
         .appendTo($dl);
       var $dd = $("<dd></dd>").appendTo($dl);
       var refcontent = conf.biblio[ref];
+      var refstyle = conf.biblioStyle;
```
marcoscaceres: nit: 
```JS
const { biblioStyle: refstyle }  = conf;
```

tests/spec/core/biblio-spec.js
```diff
@@ -92,26 +117,104 @@ describe("W3C — Bibliographic References", () => {
...
     expect(ref.textContent).toMatch(/Publisher Here\.\s/);
   });
 
+  it("displays the reference for APA citation", () => {
+    // Make sure the reference is added.
+    let ref = docAPA.querySelector("#bib-TestRef1 + dd");
+    expect(ref).toBeTruthy();
+    expect(ref.textContent).toMatch(/William Shakespeare\.\s/);
+    expect(ref.textContent).toMatch(/Test ref title\.\s/);
+    expect(ref.textContent).toMatch("Retrieved from URL: http://test.com/");
```
marcoscaceres: Ok, but now we want to test the finished product:
```JS
const finalString = "William Shakespeare. Test ref title. Retrieved from URL: http://test.com/";
expect(ref.textContent.trim()).toEqual(finalString); 
```

The same applies to each below.

marcoscaceres: We also want to make sure that the right things are hyperlinked:

```JS
const anchors = ref.querySelectorAll("a");
expect(anchors.length).toEqual(2);
const [title, retrievedFrom] = [...anchors];
expect(title.href).toEqual("http://test.com/");
expect(title.firstElementChild.localName).toEqual("cite");
expect(title.firstElementChild.textContent).toEqual("Test ref title");
expect(retrievedFrom.href).toEqual("http://test.com/");
```

And so on... have a look at the code that created this, and make sure the HTML structure is what you expect it to be.

tests/spec/core/biblio-spec.js
```diff
@@ -92,26 +117,104 @@ describe("W3C — Bibliographic References", () => {
...
+    ref = null;
```
marcoscaceres: If you find you need to reset things, then this signals you need a new test, so here:
```JS
it(".... describe what is different...", ()=>{

}); 
```

tests/spec/core/biblio-spec.js
```diff
@@ -92,26 +117,104 @@ describe("W3C — Bibliographic References", () => {
...
+  it("displays the reference for MLA citation", () => {
+    // Make sure the reference is added.
+    let ref = docMLA.querySelector("#bib-TestRef1 + dd");
+    expect(ref).toBeTruthy();
+    expect(ref.textContent).toMatch(/William Shakespeare\.\s/);
+    expect(ref.textContent).toMatch(/"Test ref title\."\s/);
+    expect(ref.textContent).toMatch(/Publishers Inc,/);
+    expect(ref.textContent).toMatch("http://test.com");
+    ref = null;
+    // title,url
```
marcoscaceres: As above, but for MLA :)

OrionStar25:
@marcoscaceres please review this!

OrionStar25:
@marcoscaceres I hope that's what you meant!

src/core/biblio.js
```diff
@@ -93,24 +93,93 @@ export function wireReference(rawRef, target = "_blank") {
...
-  }
-  if (ref.authors && ref.authors.length) {
-    output += ref.authors.join("; ");
-    if (ref.etAl) output += " et al";
-    output += ".";
+const html = (...args) => hyperHTML.bind(doc.createDocumentFragment())(...args);
+
+// Author, A. (Year, Month Date of Publication). Article title. Retrieved from URL
+function entryToAPA(ref) {
+  const { title } = ref.title;
```
marcoscaceres: Here it should be:

const { title } = ref;

Otherwise it wont work.

src/core/biblio.js
```diff
@@ -93,24 +93,93 @@ export function wireReference(rawRef, target = "_blank") {
...
+    ref.authors && ref.authors.length
+      ? `${ref.authors.join("; ")} ${ref.etAl ? " et al" : ""}.`
+      : "";
+  const date = ref.date ? `(${ref.date})` : "";
+  return html`
+    ${authors} ${date} ${
+    ref.href
+      ? html`<a href="${ref.href}"><cite>${title}</cite></a>.`
+      : `${ref.title}`
   }
```
marcoscaceres: I think we should use a cite element here too.

src/core/biblio.js
```diff
@@ -93,24 +93,93 @@ export function wireReference(rawRef, target = "_blank") {
...
+  const publisher = ref.publisher
+    ? `${ref.publisher} ${ref.publisher.endsWith(".")}`
+    : "";
+
+  return html`
+	${authors} ${
+    ref.href
+      ? html`<a href="${ref.href}"><cite>${title}</cite></a>.`
+      : `${title}`
+  } ${publisher} ${date} ${
```
marcoscaceres: As above regarding cite

src/core/biblio.js
```diff
@@ -93,24 +93,93 @@ export function wireReference(rawRef, target = "_blank") {
...
+  const date = ref.date ? `${ref.date}. ` : "";
+  const status = ref.status
+    ? (REF_STATUSES.get(ref.status) || ref.status) + ". "
+    : "";
+
+  return html`
+  	${
+      ref.href
+        ? html`<a href="${ref.href}"><cite>${title}</cite></a>.`
+        : `${ref.title}`
```
marcoscaceres: As above

src/core/biblio.js
```diff
@@ -165,13 +235,14 @@ function bibref(conf) {
         } else {
           key = refcontent.aliasOf;
           refcontent = conf.biblio[key];
+          refstyle = conf.biblioStyle;
```
marcoscaceres: I don’t think you need this line.

OrionStar25:
@marcoscaceres Can you review the tests now please? :smiley:

tests/spec/core/biblio-spec.js
```diff
@@ -92,26 +117,148 @@ describe("W3C — Bibliographic References", () => {
     expect(ref.textContent).toMatch(/Publishers Inc\.\s/);
     ref = null;
     // Make sure the ". " is automatically added to publisher.
-    ref = doc.querySelector("#bib-TestRef2 + dd");
+    ref = docDefault.querySelector("#bib-TestRef2 + dd");
```
marcoscaceres: Ok, here looks like you are testing a new thing (ref = ...), so it's best to split this test into a new:
```
 it("makes sure the '. ' is automatically added to publisher.", () =>{
    const ref =  docDefault.querySelector("#bib-TestRef2 + dd");
    // and so on... 
 })
```

tests/spec/core/biblio-spec.js
```diff
@@ -92,26 +117,148 @@ describe("W3C — Bibliographic References", () => {
...
     expect(ref).toBeTruthy();
     expect(ref.textContent).toMatch(/Testing 123\.\s/);
     ref = null;
     // Make sure publisher is shown even when there is no author
-    ref = doc.querySelector("#bib-TestRef3 + dd");
+    ref = docDefault.querySelector("#bib-TestRef3 + dd");
```
marcoscaceres: Same here... it should be: 
```JS
it("makes sure publisher is shown even when there is no author", ()=>{
  const ref = docDefault.querySelector("#bib-TestRef3 + dd");
  // and so on...
})
```

tests/spec/core/biblio-spec.js
```diff
@@ -92,26 +117,148 @@ describe("W3C — Bibliographic References", () => {
...
+    expect(ref.textContent.trim()).toEqual(finalString); 
+    //Make sure that the right things are hyperlinked
+    let anchors = ref.querySelectorAll("a");
+    expect(anchors.length).toEqual(2);
+    let [title, retrievedFrom] = [...anchors];
+    expect(title.href).toEqual("http://test.com/");
+    expect(title.firstElementChild.localName).toEqual("cite");
+    expect(title.firstElementChild.textContent).toEqual("Test ref title");
+    expect(retrievedFrom.href).toEqual("http://test.com/");
+    ref = finalString = null;
```
marcoscaceres: As above... these can be slit.

tests/spec/core/biblio-spec.js
```diff
@@ -92,26 +117,148 @@ describe("W3C — Bibliographic References", () => {
...
+    expect(ref.textContent.trim()).toEqual(finalString); 
+    //Make sure that the right things are hyperlinked
+    let anchors = ref.querySelectorAll("a");
+    expect(anchors.length).toEqual(2);
+    let [title, retrievedFrom] = [...anchors];
+    expect(title.href).toEqual("http://test.com/");
+    expect(title.firstElementChild.localName).toEqual("cite");
+    expect(title.firstElementChild.textContent).toEqual("\"Test ref title.\"");
+    expect(retrievedFrom.href).toEqual("http://test.com/");
+    ref = finalString = null;
```
marcoscaceres: As above, split into 3 or more tests

tests/spec/core/biblio-spec.js
```diff
@@ -92,26 +117,148 @@ describe("W3C — Bibliographic References", () => {
...
+    let ref = docAPA.querySelector("#bib-RefWithOnlyHref + dd");
+    expect(ref).toBeTruthy();
+    let finalString = "(2013-12-17). Retrieved from URL: http://test.com/";
+    expect(ref.textContent.trim()).toEqual(finalString); 
+    //Make sure that the right things are hyperlinked
+    let anchors = ref.querySelectorAll("a");
+    expect(anchors.length).toEqual(1);
+    [retrievedFrom] = [...anchors];
+    expect(retrievedFrom.href).toEqual("http://test.com/");
+    ref = finalString = null;
```
marcoscaceres: as above

tests/spec/core/biblio-spec.js
```diff
@@ -92,26 +117,148 @@ describe("W3C — Bibliographic References", () => {
...
+    let ref = docAPA.querySelector("#bib-RefWithOnlyHrefAndAuthor + dd");
+    expect(ref).toBeTruthy();
+    finalString = "William Shakespeare. (2013-12-17). Retrieved from URL: http://test.com/";
+    expect(ref.textContent.trim()).toEqual(finalString); 
+    //Make sure that the right things are hyperlinked
+    anchors = ref.querySelectorAll("a");
+    expect(anchors.length).toEqual(1);
+    [retrievedFrom] = [...anchors];
+    expect(retrievedFrom.href).toEqual("http://test.com/");
+    ref = finalString = null;
```
marcoscaceres: As above

OrionStar25:
@marcoscaceres all the docs (docDefault, docAPA, docMLA) show undefined even when there is an async/await placed there. :scream:

marcoscaceres:
These are starting to look good, @OrionStar25. I'll poke around a bit and see if I can get things unblocked.

marcoscaceres:
@OrionStar25, this is totally not your fault, but I'm looking at APA and MLA citations styles, and now I'm worried that we can't adequately support them for a number of reasons - and they are both not suitable for technical standards. 

With [APA](http://www.bibme.org/citation-guide/apa/website/), the structure is:

> Author, A. (Year, Month Date of Publication). Article title. Retrieved from URL
 
1. We currently don't convert authors into the correct format, that is - Last-name, First Initial. There are also a bunch of rules around 3-7 Authors, and having to put a "&" in there.
1. The publisher is missing. 
1. The document status is missing.

The last two are critical, because they establish how reputable a citation is in the standards world. Similarly with MLA, the author format we have currently implemented is incorrect. In both cases, the citation styles are too limited to cite standards.

Furthermore, we don't adhere to the rules for inline citations for either MLA or APA. That is, we support: `As stated in [SomeReference]`, but MLA and APA require: `As stated in (Lastname, 2012)` - so we would end up with two different systems. 

Thus, I think we should not support additional bibliographical styles. I'm sorry I asked you to do this work without investigating first if the citation formats were actually suitable to use. I'd assumed they were, as they are widely used. But sadly, it turns out they are not very good :(

OrionStar25:
Oh!! 🙈  Alright so we shouldn't go forward with this. That's cool, I
learnt a lot in this PR!

Well then can you please assign me another bug because I might choose
something way beyond my reach again? 😅

marcoscaceres:
>Oh!! 🙈  Alright so we shouldn't go forward with this. That's cool, I learnt a lot in this PR!

Yeah, will close. 

> Well then can you please assign me another bug because I might choose something way beyond my reach again? 😅

I don't think this was beyond your reach. You were doing great - and I also didn't realize this until super late. This happens sometimes... more often than not! I often end up throwing out a lot of code because I realize too late things are not going to work. 

I'll assign you another one.

OrionStar25:
Hey,

I just signed the petition "Students: Stop the rechristening of IIIT
Allahabad into IIIT Prayagraj" and wanted to see if you could help by
adding your name.

Our goal is to reach 5 signatures and we need more support. You can read
more and sign the petition here:

http://chng.it/Hw2scpLSpc

Thanks!
Niharika

PR #1566, opened 2018-03-10
fix(core/webidl): pad static methods

marcoscaceres:


src/core/webidl.js
```diff
@@ -672,13 +677,11 @@ function writeMethod(meth, max, indent) {
     });
   var params = paramObjs.join(", ");
   var len = idlType2Text(meth.idlType).length;
-  if (meth.static) len += 7;
   var specialProps = [
     "getter",
     "setter",
     "deleter",
-    "legacycaller",
-    "stringifier",
```
saschanaz: Shouldn't we keep `stringifier`? `legacycaller` is removed but `stringifier` is not.

marcoscaceres: Whoops!

marcoscaceres: yeah, will add a test!

marcoscaceres:
now with correct spelling 💯

marcoscaceres:
+ respec@20.0.4

PR #1569, opened 2018-03-12
fix(package): update command-line-usage to version 5.0.2

saschanaz:
Closes #1565.

I don't think we have anything to change.


saschanaz:
Oh wait, `node ./tools/respec2html.js` breaks.

saschanaz:
The test failure is because of GitHub API rate limit.

PR #1571, opened 2018-03-13
feat: show github issue's labels

### LINKED ISSUE #674: Show labels from github
For issues included from github, it might be nice to also include the labels.

himanish-star:
fixes #674 
Preview of changes:
image

### DIFF
--- a/js/ui/ui.css
+++ b/js/ui/ui.css
@@ -339,10 +339,12 @@
 
 .respec-label-dark {
   color: #fff;
+  background-color: #000;
 }
 
 .respec-label-light {
   color: #000;
+  background-color: #fff;
 }
 
 p + .respec-button-copy-paste {
--- a/js/ui/ui.css
+++ b/js/ui/ui.css
@@ -97,6 +97,8 @@
   padding: 5px;
   margin: 0 2px 0 2px;
   font-size: 10px;
+  text-transform: none;
+  text-decoration: none;
   font-weight: bold;
   border-radius: 4px;
   position: relative;
@@ -333,6 +335,14 @@
   -webkit-appearance: none;
   margin: 0px 127px;
   border-left: 0;
+}
+
+.respec-label-dark {
+  color: #fff;
+}
+
+.respec-label-light {
+  color: #000;
 }
 
 p + .respec-button-copy-paste {
--- a/js/ui/ui.css
+++ b/js/ui/ui.css
@@ -93,6 +93,16 @@
   font-size: 0.8em;
 }
 
+.respec-gh-label {
+  padding: 5px;
+  margin: 0 2px 0 2px;
+  font-size: 10px;
+  font-weight: bold;
+  border-radius: 4px;
+  position: relative;
+  bottom: 2px;
+}
+
 #respec-menu:not([hidden]) {
   display: block;
 }
--- a/js/ui/ui.css
+++ b/js/ui/ui.css
@@ -93,18 +93,6 @@
   font-size: 0.8em;
 }
 
-.respec-gh-label {
-  padding: 5px;
-  margin: 0 2px 0 2px;
-  font-size: 10px;
-  text-transform: none;
-  text-decoration: none;
-  font-weight: bold;
-  border-radius: 4px;
-  position: relative;
-  bottom: 2px;
-}
-
 #respec-menu:not([hidden]) {
   display: block;
 }
@@ -335,16 +323,6 @@
   -webkit-appearance: none;
   margin: 0px 127px;
   border-left: 0;
-}
-
-.respec-label-dark {
-  color: #fff;
-  background-color: #000;
-}
-
-.respec-label-light {
-  color: #000;
-  background-color: #fff;
 }
 
 p + .respec-button-copy-paste {
--- a/src/core/issues-notes.js
+++ b/src/core/issues-notes.js
@@ -118,7 +118,37 @@ function handleIssues($ins, ghIssues, conf) {
           }
         }
         $tit.find("span").text(text);
-        if (report.title) {
+        if (report.title && githubAPI) {
+          var labelListSpan = document.createElement('span');
+          Array.from(ghIssue.labels)
+            .forEach((label, index) => {
+              var labelSpan = document.createElement('span');
+              labelSpan.classList.add("respec-gh-label");
+              labelSpan.style.backgroundColor = `#${label.color}`;
+              labelSpan.style.textTransform = 'none';
+              labelSpan.id = `${dataNum}-label-${index}`;
+              labelSpan.innerText = label.name;
+              const color = label.color.substring(1);
+              const rgb = parseInt(color, 16);
+              const red = (rgb >> 16) & 0xff;
+              const green = (rgb >>  8) & 0xff;
+              const blue = (rgb >>  0) & 0xff;
+              const illumination = 0.2126 * red + 0.7152 * green + 0.0722 * blue;
+              if (illumination > 140) {
+                labelSpan.style.color = '#000000'
+              } else {
+                labelSpan.style.color = '#ffffff'
+              }
+              labelListSpan.appendChild(labelSpan);
+            });
+          $tit.append(
+            $(
+              "<span style='text-transform: none'>: " + report.title + labelListSpan.innerHTML
+              +"</span>"
+            )
+          );
+          $inno.removeAttr("title");
+        } else if (report.title) {
           $tit.append(
             $("<span style='text-transform: none'>: " + report.title + "</span>")
           );
@@ -163,7 +193,7 @@ async function fetchIssuesFromGithub({ githubAPI }) {
       if (!response.ok) {
         switch (response.status) {
           case 404:
-            throw new Error("Couldn't find issue on Github. Check if it exist?");
+            throw new Error("Couldn't find issue on Github. Check if it exists?");
           default:
             throw new Error("Network error. Github is down? or too many requests?");
         }
--- a/src/core/issues-notes.js
+++ b/src/core/issues-notes.js
@@ -131,7 +131,7 @@ function handleIssues($ins, ghIssues, conf) {
                 href: issuesURL.href,
               };
             })
-            .map(createLabels)
+            .map(createLabel)
             .reduce((frag, labelElem) => {
               frag.appendChild(labelElem);
               return frag;
@@ -183,7 +183,7 @@ function deriveTextColorClass(hexColor) {
   return illumination > 140 ? "light" : "dark";
 }
 
-function createLabels(label) {
+function createLabel(label) {
   const { color, href, name } = label;
   let textColorClass = "dark";
   try {
--- a/src/core/issues-notes.js
+++ b/src/core/issues-notes.js
@@ -178,14 +178,15 @@ function createLabels(label) {
   let textColorClass = "dark";
   try {
     textColorClass = deriveTextColorClass(color);
-  } catch(err) {
+  } catch (err) {
     console.error(err);
   }
   const cssClasses = `respec-gh-label respec-label-${textColorClass}`;
+  const style = `background-color: #${color}`;
   return hyperHTML`<a
-    class="${cssClasses}"
-    style="background-color: #${color}"
-    href="${url}">${name}</a>`;
+      class="${cssClasses}"
+      style="${style}"
+      href="${url}">${name}</a>`;
 }
 
 async function fetchIssuesFromGithub({ githubAPI }) {
--- a/src/core/issues-notes.js
+++ b/src/core/issues-notes.js
@@ -122,7 +122,7 @@ function handleIssues($ins, ghIssues, conf) {
         if (ghIssue && report.title && githubAPI) {
           const labelsGroup = Array.from(ghIssue.labels)
             .map(label => {
-              const issuesURL = new URL('issues/', conf.github + '/');
+              const issuesURL = new URL("issues/", conf.github + "/");
               issuesURL.searchParams.set("q", `is:issue is:open label:"${label.name}"`);
               return {
                 ...label,
--- a/src/core/issues-notes.js
+++ b/src/core/issues-notes.js
@@ -122,8 +122,8 @@ function handleIssues($ins, ghIssues, conf) {
         if (ghIssue && report.title && githubAPI) {
           const labelsGroup = Array.from(ghIssue.labels)
             .map(label => {
-              const issuesURL = new URL(`issues/`, conf.github);
-              issuesURL.searchParams.set("q", `is:issue is:open label:${label.name}`);
+              const issuesURL = new URL('issues/', conf.github + '/');
+              issuesURL.searchParams.set("q", `is:issue is:open label:"${label.name}"`);
               return {
                 ...label,
                 href: issuesURL.href,
--- a/src/core/issues-notes.js
+++ b/src/core/issues-notes.js
@@ -15,8 +15,6 @@ import css from "deps/text!core/css/issues-notes.css";
 import "deps/hyperhtml";
 export const name = "core/issues-notes";
 
-// const html =
-
 function handleIssues($ins, ghIssues, conf) {
   const { issueBase, githubAPI } = conf;
   var hasDataNum = $(".issue[data-number]").length > 0,
--- a/src/core/issues-notes.js
+++ b/src/core/issues-notes.js
@@ -121,7 +121,7 @@ function handleIssues($ins, ghIssues, conf) {
         if (report.title && githubAPI) {
           const labelsGroup = Array.from(ghIssue.labels)
             .map(createLabels)
-            .reduce((frag, labelElem) => { 
+            .reduce((frag, labelElem) => {
               frag.appendChild(labelElem);
               return frag;
             }, document.createDocumentFragment());
@@ -161,28 +161,31 @@ function handleIssues($ins, ghIssues, conf) {
 
 //derives the text-color-class based on the illumination score of the labelColor
 function deriveTextColorClass(hexColor) {
-  const rgb = parseInt(hexColor, 16);
-  const red = (rgb >> 16) & 0xff;
-  const green = (rgb >> 8) & 0xff;
-  const blue = (rgb >> 0) & 0xff;
-  const illumination = 0.2126 * red + 0.7152 * green + 0.0722 * blue;
-  return illumination > 140 ? "light" : "dark";
+  try {
+    const rgb = parseInt(hexColor, 16);
+    if (isNaN(rgb)) {
+      throw NaN;
+    }
+    const red = (rgb >> 16) & 0xff;
+    const green = (rgb >> 8) & 0xff;
+    const blue = (rgb >> 0) & 0xff;
+    const illumination = 0.2126 * red + 0.7152 * green + 0.0722 * blue;
+    return illumination > 140 ? "light" : "dark";
+  } catch (err) {
+    console.error(err);
+    return "dark";
+  }
 }
 
 function createLabels(label) {
-  const hexColor = label.color.substring(1);
-  var textColorClass = "dark";
-  var labelElement = document.createElement("a");
-  if (!isNaN(parseInt(hexColor, 16))) {
-    textColorClass = deriveTextColorClass(hexColor);
-    labelElement.style.backgroundColor = `#${label.color}`;
-  } else {
-    labelElement.style.backgroundColor = "#000";
-  }
+  const labelColor = label.color;
+  const labelElement = document.createElement("a");
+  const textColorClass = deriveTextColorClass(labelColor);
+  labelElement.style.backgroundColor = `#${labelColor}`;
   labelElement.classList.add("respec-gh-label");
+  labelElement.classList.add(`respec-label-${textColorClass}`);
   labelElement.href = label.url;
   labelElement.innerText = label.name;
-  labelElement.classList.add(`respec-label-${textColorClass}`);
   return labelElement;
 }
 
--- a/src/core/issues-notes.js
+++ b/src/core/issues-notes.js
@@ -15,6 +15,8 @@ import css from "deps/text!core/css/issues-notes.css";
 import "deps/hyperhtml";
 export const name = "core/issues-notes";
 
+// const html =
+
 function handleIssues($ins, ghIssues, conf) {
   const { issueBase, githubAPI } = conf;
   var hasDataNum = $(".issue[data-number]").length > 0,
@@ -119,8 +121,16 @@ function handleIssues($ins, ghIssues, conf) {
           }
         }
         $tit.find("span").text(text);
-        if (report.title && githubAPI) {
+        if (ghIssue && report.title && githubAPI) {
           const labelsGroup = Array.from(ghIssue.labels)
+            .map(label => {
+              const issuesURL = new URL(`issues/`, conf.github);
+              issuesURL.searchParams.set("q", `is:issue is:open label:${label.name}`);
+              return {
+                ...label,
+                href: issuesURL.href,
+              };
+            })
             .map(createLabels)
             .reduce((frag, labelElem) => {
               frag.appendChild(labelElem);
@@ -174,7 +184,7 @@ function deriveTextColorClass(hexColor) {
 }
 
 function createLabels(label) {
-  const { color, url, name } = label;
+  const { color, href, name } = label;
   let textColorClass = "dark";
   try {
     textColorClass = deriveTextColorClass(color);
@@ -184,9 +194,9 @@ function createLabels(label) {
   const cssClasses = `respec-gh-label respec-label-${textColorClass}`;
   const style = `background-color: #${color}`;
   return hyperHTML`<a
-      class="${cssClasses}"
-      style="${style}"
-      href="${url}">${name}</a>`;
+    class="${cssClasses}"
+    style="${style}"
+    href="${href}">${name}</a>`;
 }
 
 async function fetchIssuesFromGithub({ githubAPI }) {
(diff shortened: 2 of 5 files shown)


src/core/issues-notes.js
```diff
@@ -118,7 +118,37 @@ function handleIssues($ins, ghIssues, conf) {
...
+              const color = label.color.substring(1);
+              const rgb = parseInt(color, 16);
+              const red = (rgb >> 16) & 0xff;
+              const green = (rgb >>  8) & 0xff;
+              const blue = (rgb >>  0) & 0xff;
+              const illumination = 0.2126 * red + 0.7152 * green + 0.0722 * blue;
+              if (illumination > 140) {
+                labelSpan.style.color = '#000000'
+              } else {
+                labelSpan.style.color = '#ffffff'
```
himanish-star: This part of the code does this:
If the background color is too dark, the text color is set to white
else is set to black

tests/spec/core/issues-notes-spec.js
```diff
@@ -126,6 +126,30 @@ describe("Core — Issues and Notes", function () {
...
+    };
+    const doc = await makeRSDoc(ops);
+    const issueDiv = doc.getElementById('issue-1540');
+    expect(issueDiv).toBeTruthy();
+    const label0 = doc.getElementById('1540-label-0');
+    expect(label0).toBeTruthy();
+    expect(label0.innerText).toEqual("refactor");
+    const label1 = doc.getElementById('1540-label-1');
+    expect(label1).toBeTruthy();
+    expect(label1.innerText).toEqual("bug");
```
himanish-star: I guess there is no need to check for the colors of the label and yes should I also add the test for the issue with number 404?

src/core/issues-notes.js
marcoscaceres: Ah, that’s cute. Is this something you came up with?

marcoscaceres: Also, we should not set style directly. Set a css class instead, which sets the color to “white” or “black”.

src/core/issues-notes.js
```diff
@@ -118,7 +118,37 @@ function handleIssues($ins, ghIssues, conf) {
...
+          var labelListSpan = document.createElement('span');
+          Array.from(ghIssue.labels)
+            .forEach((label, index) => {
+              var labelSpan = document.createElement('span');
+              labelSpan.classList.add("respec-gh-label");
+              labelSpan.style.backgroundColor = `#${label.color}`;
+              labelSpan.style.textTransform = 'none';
+              labelSpan.id = `${dataNum}-label-${index}`;
+              labelSpan.innerText = label.name;
+              const color = label.color.substring(1);
```
marcoscaceres: All the color code should move to it’s own “deriveTextColorClass()” that returns “white” or “black”.

tests/data/issues/1540
```diff
@@ -1,5 +1,15 @@
 {
   "number": 1540,
   "title": "A mock open issue for testing",
-  "state": "open"
+  "state": "open",
+  "labels": [
```
marcoscaceres: 👌

src/core/issues-notes.js
```diff
@@ -118,7 +118,37 @@ function handleIssues($ins, ghIssues, conf) {
           }
         }
         $tit.find("span").text(text);
-        if (report.title) {
+        if (report.title && githubAPI) {
+          var labelListSpan = document.createElement('span');
```
marcoscaceres: If it’s a list, then we might want a `ul`, but I don’t think we want a list at all

src/core/issues-notes.js
```diff
@@ -118,7 +118,37 @@ function handleIssues($ins, ghIssues, conf) {
...
+                labelSpan.style.color = '#000000'
+              } else {
+                labelSpan.style.color = '#ffffff'
+              }
+              labelListSpan.appendChild(labelSpan);
+            });
+          $tit.append(
+            $(
+              "<span style='text-transform: none'>: " + report.title + labelListSpan.innerHTML
+              +"</span>"
```
marcoscaceres: The text-transform css should be in the  stylesheet. As a rule, never inline CSS.

src/core/issues-notes.js
```diff
@@ -118,7 +118,37 @@ function handleIssues($ins, ghIssues, conf) {
           }
         }
         $tit.find("span").text(text);
-        if (report.title) {
+        if (report.title && githubAPI) {
+          var labelListSpan = document.createElement('span');
+          Array.from(ghIssue.labels)
+            .forEach((label, index) => {
```
marcoscaceres: This “forEach” creates things (labels), so it’s a ` .map(createLabel)`.  And those labels get reduce()’d into the outter container.

src/core/issues-notes.js
marcoscaceres: If you want an arbitrary container, use a DocumentFragment.

tests/spec/core/issues-notes-spec.js
marcoscaceres: Yep

src/core/issues-notes.js
```diff
@@ -118,7 +118,37 @@ function handleIssues($ins, ghIssues, conf) {
...
+              var labelSpan = document.createElement('span');
```
marcoscaceres: Labels should be `a` elements, as the value of labels is that they group issues together. Clicking on a label should take you to github, so you can see other issues with the same label.

src/core/issues-notes.js
```diff
@@ -118,7 +118,37 @@ function handleIssues($ins, ghIssues, conf) {
...
+              const blue = (rgb >>  0) & 0xff;
+              const illumination = 0.2126 * red + 0.7152 * green + 0.0722 * blue;
+              if (illumination > 140) {
+                labelSpan.style.color = '#000000'
+              } else {
+                labelSpan.style.color = '#ffffff'
+              }
+              labelListSpan.appendChild(labelSpan);
+            });
+          $tit.append(
```
marcoscaceres: let's pull the actual HTML element out of `$tit`, so to avoid relying on more jQuery.

src/core/issues-notes.js
```diff
@@ -118,7 +118,37 @@ function handleIssues($ins, ghIssues, conf) {
...
+            $(
+              "<span style='text-transform: none'>: " + report.title + labelListSpan.innerHTML
```
marcoscaceres: Note here that we went to all the trouble of creating HTML objects, so it's a shame to have to change them to a string to add them to the container.

src/core/issues-notes.js
```diff
@@ -118,7 +118,37 @@ function handleIssues($ins, ghIssues, conf) {
...
-        if (report.title) {
+        if (report.title && githubAPI) {
+          var labelListSpan = document.createElement('span');
+          Array.from(ghIssue.labels)
+            .forEach((label, index) => {
+              var labelSpan = document.createElement('span');
+              labelSpan.classList.add("respec-gh-label");
+              labelSpan.style.backgroundColor = `#${label.color}`;
+              labelSpan.style.textTransform = 'none';
+              labelSpan.id = `${dataNum}-label-${index}`;
```
marcoscaceres: An issue can appear in the document more than once, so this would lead to duplicate IDs. That would make the document invalid as per the W3C Validator. 

We probably don't want to give these things IDs anyway (I can't think of any use case for having an id).

src/core/issues-notes.js
```diff
@@ -118,7 +118,37 @@ function handleIssues($ins, ghIssues, conf) {
...
         $tit.find("span").text(text);
-        if (report.title) {
+        if (report.title && githubAPI) {
+          var labelListSpan = document.createElement('span');
+          Array.from(ghIssue.labels)
+            .forEach((label, index) => {
+              var labelSpan = document.createElement('span');
+              labelSpan.classList.add("respec-gh-label");
+              labelSpan.style.backgroundColor = `#${label.color}`;
+              labelSpan.style.textTransform = 'none';
```
marcoscaceres: This should be in the style sheet instead.

src/core/issues-notes.js
```diff
@@ -118,7 +118,37 @@ function handleIssues($ins, ghIssues, conf) {
...
+              labelSpan.id = `${dataNum}-label-${index}`;
+              labelSpan.innerText = label.name;
+              const color = label.color.substring(1);
+              const rgb = parseInt(color, 16);
```
marcoscaceres: What happens if this is NaN?

src/core/issues-notes.js
```diff
@@ -118,7 +118,37 @@ function handleIssues($ins, ghIssues, conf) {
...
+              const red = (rgb >> 16) & 0xff;
+              const green = (rgb >>  8) & 0xff;
+              const blue = (rgb >>  0) & 0xff;
+              const illumination = 0.2126 * red + 0.7152 * green + 0.0722 * blue;
```
marcoscaceres: Please add a comment here that this works out the grayscale value.

tests/data/issues/1540
```diff
@@ -1,5 +1,15 @@
...
+      {
+        "name": "refactor",
+        "color": "47f441"
+      },
+      {
+        "name": "bug",
+        "color": "f4425c"
+      }
```
marcoscaceres: For completeness, we should add a `"color": ""` test case, and a `"color": "this is not a color"` test case.

src/core/issues-notes.js
```diff
@@ -163,7 +193,7 @@ async function fetchIssuesFromGithub({ githubAPI }) {
       if (!response.ok) {
         switch (response.status) {
           case 404:
-            throw new Error("Couldn't find issue on Github. Check if it exist?");
+            throw new Error("Couldn't find issue on Github. Check if it exists?");
```
marcoscaceres: Maybe send this fix as a different PR? We can get this one merged in quickly.

src/core/issues-notes.js
himanish-star: > This “forEach” creates things (labels), so it’s a .map(createLabel). And those labels get reduce()’d into the outer container.

Thanks for this, looks great to implement.

src/core/issues-notes.js
marcoscaceres: Actually, it should return “light” or “dark”.

himanish-star: Yep got it. I will update this in a while.

src/core/issues-notes.js
himanish-star: I was actually using the ids in testing :stuck_out_tongue:

tests/data/issues/1540
himanish-star: Couldn't understand this :disappointed:

marcoscaceres: A good option might be to have a fallback color for when things go wrong. E.g., white text, black background.

himanish-star: Ah fine :smiley:

himanish-star: I have pushed a part of the changes now. I am adding this feature now

src/core/issues-notes.js
```diff
@@ -118,7 +118,17 @@ function handleIssues($ins, ghIssues, conf) {
           }
         }
         $tit.find("span").text(text);
-        if (report.title) {
+        if (report.title && githubAPI) {
+          const labelsGroup = Array.from(ghIssue.labels)
+            .map(createLabels)
+            .reduce(groupLabels);
```
himanish-star: Is this what you wanted, I mean in the way I have used map and reduce

src/core/issues-notes.js
```diff
@@ -146,6 +156,40 @@ function handleIssues($ins, ghIssues, conf) {
...
+
+function createLabels(label) {
+  const textColorClass = deriveTextColorClass(label.color);
+  var labelElement = document.createElement("a");
+  labelElement.classList.add("respec-gh-label");
+  labelElement.style.backgroundColor = `#${label.color}`;
+  labelElement.style.textTransform = "none";
+  labelElement.href = label.url;
+  labelElement.innerText = label.name;
+  labelElement.classList.add(`respec-label-${textColorClass}`);
```
himanish-star: The CSS of this class has no role to play as it get's overwritten. How do I fix this? The UI looks very unpleasing without the CSS of this class.

src/core/issues-notes.js
```diff
@@ -118,7 +118,20 @@ function handleIssues($ins, ghIssues, conf) {
           }
         }
         $tit.find("span").text(text);
-        if (report.title) {
+        if (report.title && githubAPI) {
+          const labelsGroup = Array.from(ghIssue.labels)
+            .map(createLabels)
+            .reduce((frag, labelElem) => {
```
marcoscaceres: When in doubt: 

screenshot 2018-03-19 13 33 25

Where the sandwich is a `DocumentFragment` ;)

src/core/issues-notes.js
```diff
@@ -146,6 +159,33 @@ function handleIssues($ins, ghIssues, conf) {
...
+  const red = (rgb >> 16) & 0xff;
+  const green = (rgb >> 8) & 0xff;
+  const blue = (rgb >> 0) & 0xff;
+  const illumination = 0.2126 * red + 0.7152 * green + 0.0722 * blue;
+  return illumination > 140 ? "light" : "dark";
+}
+
+function createLabels(label) {
+  const hexColor = label.color.substring(1);
+  var textColorClass = "dark";
```
marcoscaceres: let

src/core/issues-notes.js
```diff
@@ -146,6 +159,33 @@ function handleIssues($ins, ghIssues, conf) {
...
+  var labelElement = document.createElement("a");
```
marcoscaceres: const

src/core/issues-notes.js
```diff
@@ -146,6 +159,33 @@ function handleIssues($ins, ghIssues, conf) {
...
+  if (!isNaN(parseInt(hexColor, 16))) {
```
marcoscaceres: parseInt is being done twice: here and in `deriveTextColorClass`. You should only do it in  `deriveTextColorClass()` and throw if NaN. Then here, you should deal with that exception.

src/core/issues-notes.js
himanish-star: Oh

src/core/issues-notes.js
himanish-star: Yep got it

himanish-star:
I have made the changes. Now we got to focus on the UI changes because based on this I'll be writing the tests which are now failing due to the proper CSS changes not reflecting
Currently the UI looks like this :stuck_out_tongue: 

image

src/core/issues-notes.js
```diff
@@ -118,7 +118,20 @@ function handleIssues($ins, ghIssues, conf) {
...
+        if (report.title && githubAPI) {
+          const labelsGroup = Array.from(ghIssue.labels)
+            .map(createLabels)
+            .reduce((frag, labelElem) => {
+              frag.appendChild(labelElem);
+              return frag;
+            }, document.createDocumentFragment());
+          $tit.append(
+            $(
+              "<span style='text-transform: none'>: " + report.title + "</span>"
```
himanish-star: This I had copied from the existing code, so in that this was inline styling. So, if you want me to make a class for this then I would have to make a class for this. If yes, suggest me a name for this class.

src/core/issues-notes.js
```diff
@@ -146,6 +159,36 @@ function handleIssues($ins, ghIssues, conf) {
...
+    if (isNaN(rgb)) {
+      throw NaN;
+    }
+    const red = (rgb >> 16) & 0xff;
+    const green = (rgb >> 8) & 0xff;
+    const blue = (rgb >> 0) & 0xff;
+    const illumination = 0.2126 * red + 0.7152 * green + 0.0722 * blue;
+    return illumination > 140 ? "light" : "dark";
+  } catch (err) {
+    console.error(err);
```
himanish-star: Maybe we can improve the text content of `console.error()`

marcoscaceres: Remember that console.error() is not for end-users, it's just for ReSpec maintainers so we can debug any issues. Generally, we only need `err` and the stack trace to figure out what happened.

src/core/issues-notes.js
```diff
@@ -146,6 +159,36 @@ function handleIssues($ins, ghIssues, conf) {
   }
 }
 
+//derives the text-color-class based on the illumination score of the labelColor
+function deriveTextColorClass(hexColor) {
+  try {
+    const rgb = parseInt(hexColor, 16);
+    if (isNaN(rgb)) {
+      throw NaN;
```
marcoscaceres: ProTip: never throw anything apart from an actual Error or something that inherits from Error (e.g., `TypeError`). Here we want a type error... 

Also, making the whole body of the function a try/catch is not a good idea. We don't want to mask programming errors, with input errors.  

Thus, this should just be:

```JS
function deriveTextColorClass(hexColor) {
  const rgb = parseInt(hexColor, 16);
  if (isNaN(rgb)) {
    throw new TypeError(`Invalid hex color: ${hexColor}`);
  }
  const red = (rgb >> 16) & 0xff;
  const green = (rgb >> 8) & 0xff;
  const blue = (rgb >> 0) & 0xff;
  const illumination = 0.2126 * red + 0.7152 * green + 0.0722 * blue;
  return illumination > 140 ? "light" : "dark";
}
```

src/core/issues-notes.js
marcoscaceres: if this is a `.issue`, then a `.issue .title, .note .title` rule would be good.

src/core/issues-notes.js
marcoscaceres: to be clear, you always want `Error`, because otherwise you don't get a stack trace.

src/core/issues-notes.js
```diff
@@ -146,6 +159,36 @@ function handleIssues($ins, ghIssues, conf) {
...
+    const blue = (rgb >> 0) & 0xff;
+    const illumination = 0.2126 * red + 0.7152 * green + 0.0722 * blue;
+    return illumination > 140 ? "light" : "dark";
+  } catch (err) {
+    console.error(err);
+    return "dark";
+  }
+}
+
+function createLabels(label) {
```
marcoscaceres: How about instead:

```JS
import "deps/hyperhtml"; // import this at the top

function createLabels(label) {
  const { color, url, name } = label;
  let textColorClass = "dark"; 
  try {
    textColorClass = deriveTextColorClass(color);
  } catch(err) {
     console.error(err);
  }
  const cssClasses = `respec-gh-label respec-label-${textColorClass}`;
  return hyperHTML`<a
    class="${cssClasses}"
    style="background-color: #${color}"
    href="${url}">${name}</a>`;
}
```

tests/data/issues/1540
```diff
@@ -1,5 +1,23 @@
 {
   "number": 1540,
   "title": "A mock open issue for testing",
-  "state": "open"
+  "state": "open",
+  "labels": [
+      {
+        "name": "refactor",
+        "color": "47f441"
```
marcoscaceres: Some of these should include a `url:` prop, as it's expected with `createLabels()`.

himanish-star: Yep sure

tests/spec/core/issues-notes-spec.js
```diff
@@ -126,6 +126,38 @@ describe("Core — Issues and Notes", function () {
...
+        <div class='issue' data-number='1540'>issue is open on github</div>
+        <div class='issue' id='this-is-404' data-number='404'>this is 404</div>
+        <section id='issue-summary'></section>
+        `,
+    };
+    const doc = await makeRSDoc(ops);
+    const issueDiv = doc.getElementById("issue-1540");
+    expect(issueDiv).toBeTruthy();
+    const labels = doc.getElementsByClassName("respec-gh-label");
+    expect(labels[0].innerText).toEqual("refactor");
```
marcoscaceres: nit: it's nice to name each label, so we can then more easily know what which is which. So:

```JS
const [refactorLabel, bugLabel, blankLabel] = doc.querySelectorAll(
  ".respec-gh-label"
);
expect(bugLabel.innerText).toEqual("bug");

```

tests/spec/core/issues-notes-spec.js
```diff
@@ -126,6 +126,38 @@ describe("Core — Issues and Notes", function () {
...
+    expect(labels[1].innerText).toEqual("bug");
```
marcoscaceres: Don't forget to check the `href` attribute for each.

js/ui/ui.css
```diff
@@ -93,6 +93,18 @@
   font-size: 0.8em;
 }
 
+.respec-gh-label {
```
marcoscaceres: Just realized that this CSS is in the wrong place.  It should be moved to:

```
js/core/css/issues-notes.css
```

The "js/ui/ui.css" is only for the ReSpec pill, and is removed from the document on save.

marcoscaceres: To get these to show up:

```CSS
.issue a.respec-gh-label {
  padding: 5px;
  margin: 0 2px 0 2px;
  font-size: 10px;
  text-transform: none;
  text-decoration: none;
  font-weight: bold;
  border-radius: 4px;
  position: relative;
  bottom: 2px;
}

.issue a.respec-label-dark {
  color: #fff;
  background-color: #000;
}

.issue a.respec-label-light {
  color: #000;
  background-color: #fff;
}

```

src/core/issues-notes.js
```diff
@@ -118,7 +119,20 @@ function handleIssues($ins, ghIssues, conf) {
           }
         }
         $tit.find("span").text(text);
-        if (report.title) {
+        if (report.title && githubAPI) {
+          const labelsGroup = Array.from(ghIssue.labels)
+            .map(createLabels)
```
marcoscaceres: nit: rename "createLabels" to "createLabel"

src/core/issues-notes.js
```diff
@@ -146,6 +160,35 @@ function handleIssues($ins, ghIssues, conf) {
...
+    throw new TypeError(`Invalid hex color: ${hexColor}`);
+  }
+  const red = (rgb >> 16) & 0xff;
+  const green = (rgb >> 8) & 0xff;
+  const blue = (rgb >> 0) & 0xff;
+  const illumination = 0.2126 * red + 0.7152 * green + 0.0722 * blue;
+  return illumination > 140 ? "light" : "dark";
+}
+
+function createLabels(label) {
```
marcoscaceres: This should be "createLabel" instead... as it only creates one at a time.

src/core/issues-notes.js
```diff
@@ -146,6 +160,35 @@ function handleIssues($ins, ghIssues, conf) {
...
+  const { color, url, name } = label;
```
marcoscaceres: Sorry, I misunderstood what `url` was for... that's not the URL we want :) we need to construct a different one above.

src/core/issues-notes.js
```diff
@@ -118,7 +119,20 @@ function handleIssues($ins, ghIssues, conf) {
           }
         }
         $tit.find("span").text(text);
-        if (report.title) {
+        if (report.title && githubAPI) {
```
marcoscaceres: here we actually want:
```
 if (ghIssue && report.title && githubAPI) {
```

himanish-star: Yeah sure

src/core/issues-notes.js
himanish-star: Yep

src/core/issues-notes.js
himanish-star: Ok

src/core/issues-notes.js
himanish-star: yep

js/ui/ui.css
himanish-star: Yeah sure

tests/data/issues/1540
himanish-star: Now I guess there is no need of the `url:` prop, as we are generating our own urls

src/core/issues-notes.js
```diff
@@ -146,6 +168,35 @@ function handleIssues($ins, ghIssues, conf) {
   }
 }
 
+//derives the text-color-class based on the illumination score of the labelColor
+function deriveTextColorClass(hexColor) {
+  const rgb = parseInt(hexColor, 16);
+  if (isNaN(rgb)) {
+    throw new TypeError(`Invalid hex color: ${hexColor}`);
```
himanish-star: Thanks for explaining this

src/core/issues-notes.js
```diff
@@ -118,7 +119,28 @@ function handleIssues($ins, ghIssues, conf) {
           }
         }
         $tit.find("span").text(text);
-        if (report.title) {
+        if (ghIssue && report.title && githubAPI) {
+          const labelsGroup = Array.from(ghIssue.labels)
+            .map(label => {
+              const issuesURL = new URL('issues/', conf.github + '/');
```
himanish-star: added this '/' otherwise the urls were like https://github.com/w3c/ instead of https://github.com/w3c/respec

tests/data/issues/1540
marcoscaceres: Correct!

src/core/issues-notes.js
marcoscaceres: No problem. Trust me, you'll be thanking me for years :) Keep an eye out for that from other developers. It can make code super difficult to debug, specially async code.

marcoscaceres:
Whoops, I accidentally merged without squashing :(

marcoscaceres:
+ respec@20.2.0 

Squashed on develop... because that's how we roll.

marcoscaceres:
We hit some bugs with this, so going to revert it.

himanish-star:
Oh

PR #1579, opened 2018-03-19
fix(w3c/templates/headers): doc.title must match conf.title

marcoscaceres:
This now sets the document title based on the `h1`, so it avoid having to declare the title of the spec twice. ReSpec gets upset when they are different.

### DIFF
--- a/src/w3c/templates/headers.js
+++ b/src/w3c/templates/headers.js
@@ -11,13 +11,15 @@ function getSpecTitleElem (conf) {
     specTitleElem.remove();
     conf.title = specTitleElem.textContent.trim();
   } else {
-    if (document.title !== conf.title) {
-      pub("warn", "The document's title and the `<title>` element differ.");
-    }
     specTitleElem.textContent = conf.title;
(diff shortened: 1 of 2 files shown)


src/w3c/templates/headers.js
```diff
@@ -11,13 +11,15 @@ function getSpecTitleElem (conf) {
...
   } else {
-    if (document.title !== conf.title) {
-      pub("warn", "The document's title and the `<title>` element differ.");
-    }
     specTitleElem.textContent = conf.title;
     specTitleElem.id = "title";
   }
   specTitleElem.classList.add("title", "p-name");
+  if (document.querySelector("title") === null) {
+    document.title = conf.title;
```
saschanaz: Will `document.title` remain alive after respec2html?

marcoscaceres: Yep, it literally adds a new title html element to the doc.

marcoscaceres: I'll add a test for this.

PR #1583, opened 2018-03-21
feat(core/issues-notes): show github labels

marcoscaceres:


src/core/issues-notes.js
```diff
@@ -146,38 +146,52 @@ function handleIssues($ins, ghIssues, conf) {
...
+    .filter(issueNumber => issueNumber)
+    .map(issueNumber => {
+      const issueURL = `${githubAPI}/issues/${issueNumber}`
+      const promiseToProcess = fetch(issueURL, {
         // Get back HTML content instead of markdown
         // See: https://developer.github.com/v3/media/
         Accept: "application/vnd.github.v3.html+json",
+      })
+      return promiseToProcess.then(response => {
+        return processResponse(response, issueNumber);
```
saschanaz: nit: `promiseToProcess.then(response => processResponse(response, issueNumber));` or even `const response = await fetch(...);` with `map(async issueNumber => { ... })`.

marcoscaceres:
Reapplied @himanish-star's patch. 

The network graph is now good (ignore that they are 404s! - just looking for a flat graph):

screen shot 2018-03-22 at 12 10 59 pm

marcoscaceres:
We might need to check if we need to batch-throttle the requests.

marcoscaceres:
@saschanaz, wdty?

src/core/issues-notes.js
```diff
@@ -146,50 +170,89 @@ function handleIssues($ins, ghIssues, conf) {
...
+  const specIssues = Array.from(document.querySelectorAll(".issue[data-number]"));
+  if (specIssues.length > MAX_GITHUB_REQUESTS) {
+    const msg = `Your spec contains ${specIssues.length} Github issues, `
+      + `but GitHub only allows ${MAX_GITHUB_REQUESTS} requests. Some issues might not show up.`;
+    pub("warning", msg);
+  }
+  const issuePromises = specIssues
+    .map(elem => Number.parseInt(elem.dataset.number, 10))
+    .filter(issueNumber => issueNumber)
+    .map(async issueNumber => {
```
marcoscaceres: this function, and processResponse should move to core/github, into a "export async function fetchIssue(number){...}"

src/core/issues-notes.js
```diff
@@ -146,50 +170,89 @@ function handleIssues($ins, ghIssues, conf) {
...
+  const green = (rgb >> 8) & 0xff;
+  const blue = (rgb >> 0) & 0xff;
+  const illumination = 0.2126 * red + 0.7152 * green + 0.0722 * blue;
+  return illumination > 140 ? "light" : "dark";
+}
+
+function createLabel(label) {
+  const { color, href, name } = label;
+  let textColorClass = "dark";
+  try {
```
saschanaz: I think this is an incorrect use of try-catch, we can just `const textColorClass = isNaN(color) ? "dark" : deriveTextColorClass(color);`.

src/core/issues-notes.js
```diff
@@ -146,50 +170,89 @@ function handleIssues($ins, ghIssues, conf) {
...
-      console.error(err);
-      const msg = `Error fetching issue #${issueNumber} from GitHub. ${err.message}. See developer console.`;
-      pub("error", msg);
-      issue = { title: "", number: issueNumber, state: "" };
-    }
-    issues.push([issueNumber, issue]);
-  }
-  return issues;
+      const issue = await processResponse(response, issueNumber);
+      return issue;
```
saschanaz: nit: `return await processResponse(response, issueNumber);` or even just `return processResponse(response, issueNumber);`.

src/core/issues-notes.js
```diff
@@ -146,50 +170,89 @@ function handleIssues($ins, ghIssues, conf) {
...
+      Object.assign(issue, await jsonPromise);
+    } catch (err) {
+      const msg = `Error JSON parsing issue #${issueNumber} from GitHub.`;
+      pub("error", msg);
+    }
+  }
+  return [issueNumber, issue];
+}
+
+async function fetchAndStoreGithubIssues(githubAPI) {
```
saschanaz: Does the `AndStore` part means the data will be stored in a map? If then we probably skip that part of the name, as it sounds the function stores data somewhere globally.

src/core/issues-notes.js
```diff
@@ -146,50 +170,89 @@ function handleIssues($ins, ghIssues, conf) {
...
+  const issue = { title: "", number: issueNumber, state: "" };
+  const jsonPromise = response.json();
+  if (!response.ok) {
+    const errJson = { message: "" };
+    try {
+      Object.assign(errJson, await jsonPromise);
+    } catch (err) {
+      errJson.message = `Error JSON parsing issue #${issueNumber} from GitHub.`;
+    } finally {
+      const msg = `Error fetching issue #${issueNumber} from GitHub. ${issue.message} (HTTP Status ${response.status}).`;
```
saschanaz: IMO we don't have to divide JSON parsing errors in two parts, should we?

tests/spec/core/issues-notes-spec.js
```diff
@@ -126,6 +126,51 @@ describe("Core — Issues and Notes", function () {
...
+        `,
+    };
+    const doc = await makeRSDoc(ops);
+
+    const issueDiv = doc.getElementById("issue-1540");
+    expect(issueDiv).toBeTruthy();
+
+    const issueDiv404 = doc.getElementById("this-is-404");
+    expect(issueDiv404).toBeTruthy();
+    expect(issueDiv404.innerText).toEqual("this is 404");
```
saschanaz: nit: `innerText` is a magic (it trims whitespaces for example) whereas `textContent` is (nearly) raw.

src/core/issues-notes.js
marcoscaceres: ok, yeah... so let's unwrap this and ensure `deriveTextColorClass` only takes numbers.

src/core/issues-notes.js
marcoscaceres: Yeah, I hated this too :( ... I was worried that "message" would have other semantics in the issue response.

src/core/issues-notes.js
marcoscaceres: @sidvishnoi is working on a IDBCache over at:
https://github.com/w3c/respec/pull/1555/files#diff-aa6009161df81293880ab4f7c52d9748

Idea is to land that, and then use it while we wait for support of the Cache API to land in Safari. Unfortunately, we won't be able to use service workers, because cross origin limitations. But we can at least throw responses into the cache.

src/core/issues-notes.js
```diff
@@ -119,7 +122,31 @@ function handleIssues($ins, ghIssues, conf) {
         }
       }
       $tit.find("span").text(text);
-      if (report.title) {
+      if (ghIssue && report.title && githubAPI) {
+        const labelsGroup = Array.from(ghIssue.labels)
+          .map(label => {
+            const issuesURL = new URL("issues/", conf.github + "/");
+            issuesURL.searchParams.set(
```
saschanaz: Using `searchParams` will be problematic in IE11 (which still has security support), can we ignore it?

src/core/issues-notes.js
```diff
@@ -147,59 +174,76 @@ function handleIssues($ins, ghIssues, conf) {
   }
 }
 
-async function fetchIssuesFromGithub({ githubAPI }) {
-  const issues = [];
-  const issueNumbers = [...document.querySelectorAll(".issue[data-number]")]
+async function fetchAndStoreGithubIssues(githubAPI) {
```
saschanaz: I still think we should at least write a comment about `AndStore` part before we actually implement it.

src/core/issues-notes.js
marcoscaceres: Yeah, we don't support IE.

src/core/issues-notes.js
marcoscaceres: Will do. We can't land this without caching regardless - as otherwise a few refreshes on some specs causes them to run out of Github request quota.

saschanaz: I hope GitHub people allow V4 API public access soon (as it will allow retrieving multiple issue data in a single request)...

marcoscaceres: oh, that's awesome to hear!

marcoscaceres:
@saschanaz, this should be good to go now :)

src/core/issues-notes.js
```diff
@@ -147,59 +175,83 @@ function handleIssues($ins, ghIssues, conf) {
...
-          default:
-            throw new Error(
-              "Network error. Github is down? or too many requests?"
-            );
-        }
+      try {
+        const response = await fetchAndCache(request);
+        return processResponse(response, issueNumber);
+      } catch(err){
+        return processResponse(err.response, issueNumber);
```
saschanaz: A fetch error won't always contain a `response` field as it can also be a native `TypeError`. I'm starting to think we are better to keep original `fetch` exception model here. (Throwing only when it couldn't even connect)

marcoscaceres: > A fetch error won't always contain a response field as it can also be a native TypeError.

True. 

> I'm starting to think we are better to keep original fetch exception model here. (Throwing only when it couldn't even connect)

Yeah, that might make more sense.

src/core/issues-notes.js
```diff
@@ -147,59 +175,79 @@ function handleIssues($ins, ghIssues, conf) {
...
-    } catch (err) {
-      console.error(err);
-      const msg = `Error fetching issue #${issueNumber} from GitHub. ${
-        err.message
-      }. See developer console.`;
-      pub("error", msg);
-      issue = { title: "", number: issueNumber, state: "" };
-    }
-    issues.push([issueNumber, issue]);
+      const response = await fetchAndCache(request);
```
marcoscaceres: @saschanaz, ok 404s etc, are now passed on. Exceptions go unhandled. Should we do anything else here?

PR #1584, opened 2018-03-22
Remove @@@ from sotd

### LINKED ISSUE #1494: What does handle @@@?
I found `@@@text@@@` forms in our SOTD template:

https://github.com/w3c/respec/blob/cb910fefe152db689659c6b16b319ffabcc1a239/js/w3c/templates/sotd.html#L37-L53

It seems they are not processed by handlebars or anything, what does handle them?

OrionStar25:
Fixes #1494 .
The url for team submission comment requires the submission year (not the published year) along with the comment number (serial no of what comment it was in that year).  Idk how to get the comment number.

### DIFF
--- a/src/w3c/templates/sotd.js
+++ b/src/w3c/templates/sotd.js
@@ -22,11 +22,8 @@ ${conf.isPreview
 ${conf.isUnofficial
     ? html`
   <p>
-    Do not attempt to implement this version of the specification. Do not reference this version as authoritative in any way.
-    ${conf.edDraftURI
-      ? html` Instead, see
-    <a href="${conf.edDraftURI}">${conf.edDraftURI}</a> for the Editor's draft. `
-      : ""}
+    This document is draft of a potential specification. It has no official standing of
+    any kind and does not represent the support or consensus of any standards organisation.
   </p>
   ${[conf.additionalContent]}
 `
@@ -55,8 +52,7 @@ ${conf.isUnofficial
         ${conf.isMemberSubmission
           ? html`
           <p>
-            By publishing this document, W3C acknowledges that <a href="${conf.thisVersion}">
-            the Submitting Members</a>
+            By publishing this document, W3C acknowledges that the <a href="${conf.thisVersion}">Submitting Members</a>
             have made a formal Submission request to W3C for discussion. Publication of this document
             by W3C indicates no endorsement of its content by W3C, nor that W3C has, is, or will be
             allocating any resources to the issues addressed by it. This document is not the product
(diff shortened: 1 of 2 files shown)


OrionStar25:
You mean for the comments? Yes that makes more sense!

OrionStar25:
What's conf.additionalContent?

marcoscaceres:
`additionalContent` seems to just be additional text content that can be added. You'll need to check where it is added tho.

OrionStar25:
I'm sorry I can't seem to figure out where the generated file's url is being stored.
isDraftURI  has a default of true, and hence even when it's not mentioned, an undefined spec link is created. Is that the link I want?

marcoscaceres:
Ok, let's take a step back and fix one thing at a time. I've reverted the changes. 

- [ ] Let's get the first commit to just remove the `teamSubmission`. 
- [ ] Figure out what the submission URL is. You need to go to the https://www.w3.org/Submission/ and have at how the URLs for submissions are structured.

marcoscaceres:
So, for example:

```
https://www.w3.org/Submission/2017/SUBM-epub-mediaoverlays-20170125/
```

We can see that these parts:

 1. base: https://www.w3.org/Submission/
 1. year: 2017 
 1. "SUBM" - {short-name} - {publicationDate} 

So, from the `conf`, you need to dig through it and find something that matches the `publicationDate`.  

You can do that by doing `console.log(conf)` and seeing what's available to you.

OrionStar25:
`teamSubmissiom` removed. Onto the next part.

OrionStar25:
`thisVersion`  happens to be the exact url that submitting members is hyperlinked with.  I can't seem to understand comment no though.

OrionStar25:
@marcoscaceres

marcoscaceres:
> thisVersion happens to be the exact url that submitting members is hyperlinked with.

Excellent! That makes it easy then. 

> I can't seem to understand comment no though.

@OrionStar25, not sure I understand what you are asking here? 

Regarding the failure on Travis:

```
  W3C — Headers
    Member-SUBM
      ✖ uses the right SoTD boilerplate for Member submissions
        HeadlessChrome 65.0.3325 (Linux 0.0.0)
      Expected 'foo' to match 'the Submitting Members have made a formal Submission request'.
          at <Jasmine>
          at UserContext.it (tests/spec/w3c/headers-spec.js:966:20)
          at <Jasmine>
```

Seems that there is a test that needs to be updated.

OrionStar25:
> not sure I understand what you are asking here?

Every W3C staff comment is of this form : https://www.w3.org/Submission/2015/03/Comment/

So:
it has a base + thisVersion year +  comment number in that year + "Comment".
How do I retrieve the comment number?

> Seems that there is a test that needs to be updated.

That test should actually work though, but I'll try to fix it.

OrionStar25:
Okay, so these are 2 differen specs both submitted in the same year:

https://www.w3.org/Submission/2017/01/Comment/
https://www.w3.org/Submission/2017/02/Comment/

> I think we need to derive it from person writing the spec

How would the spec writer know how many specs have been added before his in that year, unless he checks it up? I mean, is that what we want? For him to manually look up his order of spec in that year? That can lead to clashes too (If 2 concurrent specs are being written).

marcoscaceres:
@OrionStar25, about the comments, I think it's fine, because it's super rare. The member submissions are managed by W3C Team, and they know what they are doing :) This is just to help them a little bit.

OrionStar25:
@marcoscaceres `thisVersion` made me understand how I could retrive the year. Its working now. Also, added a `submissionCommentNumber`

marcoscaceres:
Ok, now we need to just add 2 tests:

- [ ] check to make sure the @@@ is replaced with a the expected URL. 
- [ ] check the comments link is as we expect it.

OrionStar25:
@marcoscaceres please review! Build fail is due to `run headless-timeout`.

OrionStar25:
@marcoscaceres please review!

OrionStar25:
Shouldnt we add a configuration option for submissionCommentNumber in the user guide?

marcoscaceres:
Yep, that should be added to the wiki.

marcoscaceres:
I need to work out what's happened with the formatting of the file before merging this. The original stod is formatted differently. 

See also the message I sent you on Slack. We should fix up your prettier configuration, as it's adding "," in various places.

marcoscaceres:
respec@20.3.0

marcoscaceres:
Documentation: https://github.com/w3c/respec/wiki/submissionCommentNumber

@OrionStar25 ^^^ please check. 👀

OrionStar25:
Perfect!

OrionStar25:
Need to also update this. https://github.com/w3c/respec/wiki/specStatus
`Team Submission` would be removed and `submissionCommentNumber` shall be added as necessary for member comment. (But then I'm thinking comments aren't compulsory).

PR #1586, opened 2018-03-26
Add an IndexedDB based key-value cache

sidvishnoi:
A promise based interface to a key-value based indexedDB cache

Status: WIP.
TODO: Tests

### DIFF
--- a/src/core/idb-cache.js
+++ b/src/core/idb-cache.js
@@ -75,11 +75,11 @@ export default class IDBCache {
   }
 
   /**
-   * @method @async remove a key-value pair from a given store
+   * @method @async delete a key-value pair from a given store
    * @param {string} key        object store record key
    * @param {string} storeName  store name
    */
-  async remove(key, storeName = this.defaultStore) {
+  async delete(key, storeName = this.defaultStore) {
     const db = await databases.get(this.name);
     const store = await getStore(db, storeName, "readwrite");
     return await getResponse(store.delete(key));
--- a/src/core/idb-cache.js
+++ b/src/core/idb-cache.js
@@ -20,10 +20,10 @@ export default class IDBCache {
       defaultStore,
       maxAge,
     } = {
-      version: 1,
-      defaultStore: null,
-      maxAge: 86400000, // 24 hours (in ms)
-    }
+        version: 1,
+        defaultStore: null,
+        maxAge: 86400000, // 24 hours (in ms)
+      }
   ) {
     this.name = name;
     if (!Array.isArray(stores) || stores.length < 1) {
@@ -46,17 +46,13 @@ export default class IDBCache {
         request.onsuccess = () => resolve(request.result);
       });
       databases.set(this.name, dbPromise);
+      // this.ready (read-only)
+      Object.defineProperty(this, "ready", {
+        get() {
+          return dbPromise;
+        }
+      });
     }
-  }
-
-  get ready() {
-    return new Promise(async (resolve) => {
-      const db = await databases.get(this.name);
-      if (db instanceof IDBDatabase) {
-        return resolve();
-      }
-      throw new Error(`The database ${this.name} is gone!`);
-    });
   }
 
   /**
--- a/src/core/idb-cache.js
+++ b/src/core/idb-cache.js
@@ -15,8 +15,15 @@ export default class IDBCache {
    */
   constructor(
     name,
-    stores,
-    { version, defaultStore } = { version: 1, defaultStore: null }
+    stores, {
+      version,
+      defaultStore,
+      maxAge,
+    } = {
+      version: 1,
+      defaultStore: null,
+      maxAge: 86400000, // 24 hours (in ms)
+    }
   ) {
     this.name = name;
     if (!Array.isArray(stores) || stores.length < 1) {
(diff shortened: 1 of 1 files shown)


src/core/idb-cache.js
```diff
@@ -0,0 +1,145 @@
...
+   *    @param {integer} version      database version
+   *    @param {string} defaultStore  name of default store
+   */
+  constructor(
+    name,
+    stores, {
+      version,
+      defaultStore,
+      maxAge,
+    } = {
```
marcoscaceres: Nit: less ugly to move the defaults to a `const` outside the function.

marcoscaceres:
@saschanaz, would really appreciate your thoughts on this.

saschanaz:
So this is going to be a new IndexedDB library... Do we need to maintain a new one, or can we just use existing libraries e.g. [localforage]`https://www.npmjs.com/package/localforage` from Mozilla?

marcoscaceres:
> So this is going to be a new IndexedDB library... Do we need to maintain a new one, or can we just use existing libraries e.g. localforage from Mozilla?

This is going to be somewhat temporary. Eventually we will move to using Cache API. I don't expect us to use this in so many places that we would need localForage.

saschanaz:
Hmm, yeah, localForage is fairly big, but then we have other smaller ones e.g. [idb-keyval](https://www.npmjs.com/package/idb-keyval).

sidvishnoi:
> we have other smaller ones e.g. idb-keyval

Oh my! `idb-keyval` looks a lot similar to what we've written!

As @marcoscaceres  said, we'll switch to Cache API when it becomes available in more browsers. We made the API interface as much as similar the Cache API one so that it'll be easier to migrate later when needed.

saschanaz:
>We made the API interface as much as similar the Cache API one so that it'll be easier to migrate later when needed.

Seems reasonable, but AFAIK we should write a new Service Worker script to use native Cache API, right?

marcoscaceres:
> Seems reasonable, but AFAIK we should write a new Service Worker script to use native Cache API, right? (edit: Although it's not strictly required)

We can't, unfortunately, because of same origin policy for service workers and the way specs are deployed (each spec would need to also host a SW script in its own repo... which would be complicated for many spec editors). However, we can use the Cache API independently of service workers: we can put things into, and pull things out of the cache directly from the window object, no worker needed. 

@sidvishnoi wrote:
> Oh my! idb-keyval looks a lot similar to what we've written!

We need to weight the pros and cons here... If we keep our current code: 

Pros: 
 * We control our code/we can quickly fix our code.
 * We have quality assured it against our use cases.
 * It's built for purpose - and it's comparatively small.
 
Cons:
 * it's not as "battle hardened", as idb-keyval.
 * it duplicates simple functionality. 
 * we are not as experienced as Jake (the idl-keyval author) in IDB, so he might have solved problems we didn't think of.
 * we plan to ditch our code in a few months, so "sunk costs". 

And pros and cons of adopting idb-keyval:

Pros:
 * battle hardened - has been around for 2 years. 
 * lean, and easy to use API. 
 * maintained by someone we know and trust (and responsive to bug fixes). 

Cons:
 * dependency on someone else's code. 
 * if things go bad, we might have trouble debugging, fixing, or getting help. 
 
I don't have a strong opinion and would be happy with either solution. @sidvishnoi, @himanish-star  wdty?  You both invested a lot into this.

marcoscaceres:
Ok, I've given this some thought over the last few days. I there might be a third option, and that's simply to just use the Cache API as a progressive enhancement. We are only a few weeks away from having Cache API [available across all browsers](https://jakearchibald.github.io/isserviceworkerready/#caches), with Cache being in beta in Safari. The cache API always serves as a fallback. 

Thus, let's abandon this and just go straight to implementing with the Cache API. 

Any objections?

saschanaz:
Great.

BTW, AFAIK Cache API receives Request object and this is best useful within Service Worker fetch event listeners. What's the plan for non-worker Cache use?

marcoscaceres:
Please see the first part of my response here (unfortunately, we can’t use service workers, but that’s ok): https://github.com/w3c/respec/pull/1586#comment-376742367

Basically, we fetch(), clone, add our own cache expire HTTP header. On subsequent loads, we pull from cache, check expiration, use if stale, otherwise we fetch fresh from network.

marcoscaceres:
*if NOT stale

marcoscaceres:
Actually, Safari update came out today on MacOS and iOS!!!! Was not expecting it to ship so soon. So yeah, Cache API 💯

sidvishnoi:
So we can use the cache api directly I guess. No need to create a wrapper. What do you say?

marcoscaceres:
Yep, no wrapper needed!

sidvishnoi:
I'll first try to implement that in caniuse feature :smile:

marcoscaceres:
Sounds good. Should be like 10 lines of code. Don’t forget response.clone() before putting things in the cache. That’s bitten me before when I was first learning the API.

sidvishnoi:
setting the custom headers is biting me more :laughing:

sidvishnoi:
Closed in favour of Cache API

PR #1587, opened 2018-03-27
Issue 670 fix gh issues

sauravjaiswalsj:


marcoscaceres:
Ah, forgot about this code. Should salvage that.

PR #1588, opened 2018-03-27
feat(core/highlight-vars): highlight vars in algos

### LINKED ISSUE #1529: Highlight variable matching in algorithm

marcoscaceres:
closes #1529 

## New Module core/highlight-vars
Highlights occurrences of a <var> within a section on click.
Set `conf.highlightVars = true` to enable.
removes highlights from `<var>` if clicked anywhere else.
All is done while keeping in mind that exported html stays clean (removeHighlight).


marcoscaceres:
Rebased the existing PR and squashed the history.

marcoscaceres:
@sidvishnoi, I added the entry to the wiki:
https://github.com/w3c/respec/wiki/highlightVars

Can you check if it's ok.

sidvishnoi:
> entry to the wiki

Yes. looks great to me. :smile:

src/core/highlight-vars.js
```diff
@@ -0,0 +1,95 @@
...
+  if (!el.getAttribute("style")) el.removeAttribute("style");
+  if (!el.getAttribute("class")) el.removeAttribute("class");
+}
+
+function highlightVars(target) {
+  const textContent = target.textContent.trim();
+  const parent = target.closest("section");
+  const highlightColor = getHighlightColor(target);
+
+  function highlightVar(elem) {
```
saschanaz: I would prefer moving this to the bottom of `highlightVars`, as currently expressions are divided into two parts by this function.

src/core/highlight-vars.js
```diff
@@ -0,0 +1,95 @@
...
+
+import { sub } from "core/pubsubhub";
+import hlVars from "deps/text!core/css/var.css";
+
+export const name = "core/highlight-vars";
+
+export function run(conf, doc, cb) {
+  if (!conf.highlightVars) {
+    return cb();
+  }
```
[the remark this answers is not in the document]
marcoscaceres: Good point... I even forget about this :(

src/core/highlight-vars.js
```diff
@@ -0,0 +1,93 @@
...
+function highlightVars(target) {
+  const textContent = target.textContent.trim();
+  const parent = target.closest("section");
+  const highlightColor = getHighlightColor(target);
+  [...parent.querySelectorAll("var")]
+    .filter(el => el.textContent.trim() === textContent)
+    .forEach(highlightVar);
+}
+
+function highlightVar(elem) {
```
marcoscaceres: @sidvishnoi, I moved this out, but broke it in the process 👍. However, this means something is not right with the design of the algorithm. It seems to have too many side effects. Can you look at refactoring this?

sidvishnoi: I made a little fix here: https://github.com/sidvishnoi/respec/commit/c1f263c0b4bb8f710137963bb3803261dd8d7ebe

sidvishnoi: Also, do have a look at this approach https://github.com/sidvishnoi/respec/commit/4154f8bdeb917d92d550e2fd7bd4f2dfc8cf9988

Did this because there was a common repetitive work going in `highlightVar` to set color status and splitting color to `[background, color]`. This comes at an expense of initialisation of a new variable `varsToHighlight` which II have found a little expensive.

marcoscaceres: 👹 says "Premature optimization is the root of all evil". But seriously, maintainability trumps any expense of initialization here.

sidvishnoi: If I understood it right, you meant the second one is more maintainable.
So I made another commit https://github.com/sidvishnoi/respec/commit/07c229054d4bb5380704a33b18e33028bd9f0c37 to separate out `addHighlight` and `removeHighlight` logic.

You might want to merge the [sidvishnoi/ISSUE_1529-alt](https://github.com/sidvishnoi/respec/tree/ISSUE_1529-alt/) branch into `w3c/sidvishnoi-ISSUE_1529` if you think it's right.

marcoscaceres: I think what's in sidvishnoi/respec@07c2290 should be ok.

marcoscaceres: @sidvishnoi, I think we are 99% done here. Can you bring in sidvishnoi/respec@07c2290?

src/core/highlight-vars.js
```diff
@@ -0,0 +1,102 @@
...
+export function run(conf) {
+  if (!conf.highlightVars) {
+    return;
+  }
+  const styleElement = document.createElement("style");
+  styleElement.textContent = hlVars;
+  styleElement.classList.add("removeOnSave");
+  document.head.appendChild(styleElement);
+
+  document.addEventListener("click", initHighlight);
```
marcoscaceres: I'm wondering if we shouldn't just attach listener on `<var>`s instead?

src/core/highlight-vars.js
```diff
@@ -0,0 +1,102 @@
...
+}
+
+function removeHighlight(el) {
+  // done so that only the respec classes are removed
+  el.classList.remove("respec-active");
+  el.style.removeProperty("--respec-background-color");
+  el.style.removeProperty("--respec-color");
+  // clean up empty class/style attributes so they don't come in export
+  if (!el.getAttribute("style")) el.removeAttribute("style");
+  if (!el.getAttribute("class")) el.removeAttribute("class");
```
marcoscaceres: Seems a little destructive to remove attributes here, because an author might have set their own. We should maybe just `classList.remove("respec-active")`. Then maybe if  classList is empty, we can remove the attribute too.

src/core/highlight-vars.js
marcoscaceres: Just thinking out loud here: 

```JS
[...document.querySelectorAll("var")].forEach(varElem =>
  varElem.addEventListener("click", e => {
    e.stopPropagation();
    highlightVars(e.target);
    const section = e.target.closest("section");
    section.addEventListener("click", resetHighlight, { once: true });
  })
);

function resetHighlight({ target: section }) {
  [...section.querySelectorAll("var.respec-active")].forEach(removeHighlight);
  [...HL_COLORS.keys()].forEach(key => HL_COLORS.set(key, true));
}
```

marcoscaceres: fixed code above... tested it, it works kinda nice :) @sidvishnoi, @saschanaz, thoughts?

marcoscaceres: Actually, need the reset listener on the body instead. Fixed above.

src/core/highlight-vars.js
```diff
@@ -0,0 +1,102 @@
...
+  const varsToHighlight = [...parent.querySelectorAll("var")].filter(
+    el => el.textContent.trim() === textContent
+  );
+
+  // update availability of highlight color
+  const colorStatus = varsToHighlight[0].classList.contains("respec-active");
+  HL_COLORS.set(highlightColor, colorStatus);
+
+  // highlight vars
+  const [background, color] = highlightColor.split(",");
```
marcoscaceres: I'm wondering if we should put HL_COLOR's into js/core/css/var.css, then we can name them:

```JS
const HL_COLORS = new Map([
  ["respec-hl-yellow", true],
  ["respec-hl-brown", true],
]);
```

And so on... then we don't need to deal foreground/background colors here. We just use the next CSS class available in the map.

marcoscaceres: Was looking for nice contrasting color palettes: quite like http://colorpalettes.net/color-palette-3694/

src/core/highlight-vars.js
saschanaz: >because an author might have set their own.

`if (!el.getAttribute("class"))` will check that situation, no? I agree using `.classList.length` would be cleaner though.

src/core/highlight-vars.js
```diff
@@ -0,0 +1,116 @@
...
+
+  // remove highlights, cleanup empty class/style attributes
+  sub("beforesave", outputDoc => {
+    [...outputDoc.querySelectorAll("var.respec-active")].forEach(
+      removeHighlight
+    );
+  });
+}
+
+function hightlightListener(ev) {
```
marcoscaceres: Did a little refactor here. Instead of querying the body for highlighted elements, we just unhighlight the ones that were highlighted.

src/core/highlight-vars.js
```diff
@@ -0,0 +1,118 @@
...
+  ev.stopPropagation();
+  const { target: varElem } = ev;
+  const hightligtedElems = highlightVars(varElem);
+  const resetListener = () => {
+    hightligtedElems.forEach(removeHighlight);
+    [...HL_COLORS.keys()].forEach(key => HL_COLORS.set(key, true));
+  };
+  if (hightligtedElems.length) {
+    document.body.addEventListener("click", resetListener, { once: true });
```
marcoscaceres: Had a small play on iOS... "click" only works on certain elements, but not the body... and "touchstart" catches all touches. Not sure what the best solution is here for mobile.

sidvishnoi: Can this be a solution too? : https://stackoverflow.com/questions/25333738/click-events-not-firing-on-ios (weird bug, weird fix)

js/core/css/var.css
```diff
@@ -3,13 +3,28 @@ var:hover {
   cursor: pointer;
 }
 
-var.respec-active {
-  background: var(--respec-background-color);
-  color: var(--respec-color);
+var.respec-hl-yellow {
+  background: #f6d51f;
```
marcoscaceres: We might need to set `color:` too for these, because we can get into this unfortunate situation: 

screenshot 2018-04-23 14 30 31

sidvishnoi: oh dear!

Need advice on color. If I use black, it look less visible to me but Chrome shows it as AAA accessible. If I pick white, it looks better to me but contrast ratio is weak.
Should I try complement colors?

marcoscaceres: > Should I try complement colors?

Might be worth a try.

sidvishnoi: complementary colors didn't turn out well. I'll look into better color schemes later, but for now, black text color.

marcoscaceres:
Ok, this is where we are now. And I can definitely work with that: 

screenshot 2018-04-25 16 00 24

@saschanaz, have your requested changes been addressed? Do you want to give this one last look over?

src/core/highlight-vars.js
```diff
@@ -0,0 +1,106 @@
...
+    hightligtedElems.forEach(el => removeHighlight(el, hlColor));
+    [...HL_COLORS.keys()].forEach(key => HL_COLORS.set(key, true));
+  };
+  if (hightligtedElems.length) {
+    document.body.addEventListener("click", resetListener, { once: true });
+  }
+}
+
+// availability of highlight colors. colors from var.css
+const HL_COLORS = new Map([
```
marcoscaceres: TODO: in the future, we should make these into configurable themes and "levels". So, instead of this, we just say `${conf.highlight.theme}-level-${X}`. 

Then we provide a bunch of themes that people may like. Let's see how popular this gets first.

sidvishnoi: I din't get what you mean by "level". Can you please clarify?

sidvishnoi:
This Codacy error looks weird. Can you check @marcoscaceres ?

marcoscaceres:
Looks like Codacy doesn’t know about css vars.

saschanaz:
>Looks like Codacy doesn’t know about css vars.

Probably this: https://github.com/CSSLint/csslint/issues/720

tests/spec/core/highlight-vars-spec.js
```diff
@@ -0,0 +1,64 @@
+"use strict";
+
+// TODO: add tests:
+// - right colors are applied
```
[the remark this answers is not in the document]
sidvishnoi: I hope so. Really not sure about how to test that. Will probably copy the `HL_COLORS` map here and check if classes are applied in right order. Any advice @marcoscaceres ?

src/core/highlight-vars.js
```diff
@@ -0,0 +1,107 @@
...
+function removeHighlight(el, highlightColor) {
+  el.classList.remove("respec-hl");
+  el.classList.remove(highlightColor);
+  // clean up empty class attributes so they don't come in export
+  if (el.classList.length == 0) el.removeAttribute("class");
+}
+
+function addHighlight(elem, highlightColor) {
+  elem.classList.add("respec-hl");
+  elem.classList.add(highlightColor);
```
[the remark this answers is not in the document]
sidvishnoi: Should we inline these `addHighlight` and `removeHighlight` functions?

tests/spec/core/highlight-vars-spec.js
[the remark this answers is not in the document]
marcoscaceres: I think as people start using this they will like ask us to change the colors, so I'm ok for us to wait while we "beta test" this.

Issue #1572, opened 2018-03-14
Linter should check for “.” in p and li


marcoscaceres:
Got a code review where I had forgotten full stops in a couple of places. That wastes reviewers time.

OrionStar25:
```
let var = doc.querySelector("p");
expect(var).toBeTruthy();
expect(var.textContent).toMatch($("p").text() + /./);

Is this what needs to be done essentially?
```

marcoscaceres:
something like that, but it would be a bit more involved: basically, I would like to asynchronously (via `requestIdleCallback()`, if available) check each `p.textContent.trim().endsWith(".")`.

OrionStar25:
Alright, excellent! I'd like to work on this :100:

marcoscaceres:
Ok, please take a look at:
https://github.com/w3c/respec/blob/develop/src/core/linter.js

And the linterRules directory: 
https://github.com/w3c/respec/blob/develop/src/core/

But let’s finish the biblio PR before starting on this one.

shikhar-scs:
@marcoscaceres can I please take up this issue if no one is working on it ?

marcoscaceres:
@shikhar-scs sure, but coordinate with @OrionStar25, as she also wanted to help. 

If you do start work on it, you might want to follow this one:
https://github.com/w3c/respec/blob/develop/src/core/linter-rules/no-http-props.js

Also, we should start with just the `p` elements, as I think they are less difficult than the `li`. The `li` checks would  be a separate linter rule - as some spec writers prefer to exclude "." from bulleted lists in certain situations). 

We should check this list of specs:
https://github.com/w3c/respec/blob/develop/tests/headless.js#L4

To see if there are odd edge cases where missing a "." might be ok.

shikhar-scs:
Awesome ! I'll get started with this :+1:

xfq:
Should we close this given that #1589 has been merged?

By the way, I think we should only check `.:!?` in documents with `lang=en`, because not all languages/scripts use the Latin punctuations, and we do have ReSpec documents in these languages. I can file another issue for this if needed.

Some examples of full stops:

* "。" in Chinese (`zh-hans` and `zh-hant`) and Japanese (`ja`)
* "।" in Hindi (`hi`)
* "۔" in Urdu (`ur`)

marcoscaceres:
True, let's leave this open then until we deal with the l10n issues.

marcoscaceres:
[...]

PR #1589, opened 2018-03-27
check period in p ; solves #1572

shikhar-scs:
Taking inspiration from `https://github.com/w3c/respec/blob/develop/src/core/linter-rules/no-http-props.js` and #1572 , some changes have been incorporated. 

Initially, when the logic was present in the file `no-https-props.js` file, it was working. However, it fails in a new file. @marcoscaceres could you please see to that. 

> https://github.com/w3c/respec/issues/1572#issuecomment-373238908

this will be taken care of once we the new file starts working. 
Info related to the same exists here https://developers.google.com/web/updates/2015/08/using-requestidlecallback
Please review :sweat_smile:


src/core/linter-rules/check-period-in-p.js
```diff
@@ -0,0 +1,47 @@
...
+  const result = {
+    name,
+    occurrences: offendingMembers.length,
+    ...meta[lang],
+  };
+  result.howToFix +=
+    offendingMembers.map(item => "`" + item + "`").join(", ") + ".";
+  return result;
+}
+console.log("k")
```
marcoscaceres: nit: delete...

src/w3c/defaults.js
```diff
@@ -52,6 +53,7 @@ const w3cDefaults = {
     "no-headingless-sections": true,
     "privsec-section": true,
     "no-http-props": true,
+    "check-period-in-p": true,
```
marcoscaceres: we would initially want false here... till we were sure it works well across a bunch of specs.

src/core/linter-rules/check-period-in-p.js
```diff
@@ -0,0 +1,47 @@
...
+ * Runs linter rule.
+ *
+ * @param {Object} config The ReSpec config.
+ * @param  {Document} doc The document to be checked.
+ */
+function lintingFunction(conf, doc) {
+  
+  const offendingMembers = Array.from(doc.querySelectorAll("p"))
+    // this gives us all paragraphs that don't end with a period
+    .filter(key => !key.textContent.trim().split("tests")[0].endsWith("."))
```
marcoscaceres: split("tests")[0] ?

Was expecting: 
```js
.filter(elem => elem.textContent.trim().endsWith("."))
```

src/core/linter-rules/check-period-in-p.js
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Runs linter rule.
+ *
+ * @param {Object} config The ReSpec config.
+ * @param  {Document} doc The document to be checked.
+ */
+function lintingFunction(conf, doc) {
+  
+  const offendingMembers = Array.from(doc.querySelectorAll("p"))
```
marcoscaceres: offendingMembers should be called "offendingElements"

src/core/linter-rules/check-period-in-p.js
shikhar-scs: I had initially written a patch at #1487 . Thus, when tests exist, the `textContent()` also contains the word `tests`. This I guess should not be a part of the innerText and thus, should be taken care of before analysing so as to avoid unnecessary cases.

shikhar-scs:
Thankyou for the review! I'll incorporate all of it ASAP

src/core/linter-rules/check-period-in-p.js
marcoscaceres: Should have included "!" in the code above, fixed in comment.

src/core/linter-rules/check-period-in-p.js
```diff
@@ -0,0 +1,47 @@
...
+    return [];
+  }
+  
+  const result = {
+    name,
+    occurrences: offendingMembers.length,
+    ...meta[lang],
+  };
+  result.howToFix +=
+    offendingMembers.map(item => "`" + item + "`").join(", ") + ".";
```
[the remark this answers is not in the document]
shikhar-scs: Yep they are !

shikhar-scs:
I have pushed some new commits which include the changes, Please have a look @marcoscaceres :sweat_smile:

src/core/linter-rules/check-period-in-p.js
```diff
@@ -0,0 +1,41 @@
...
+ * Runs linter rule.
+ *
+ * @param {Object} config The ReSpec config.
+ * @param  {Document} doc The document to be checked.
+ */
+function lintingFunction(conf, doc) {
+  
+  const offendingElements = Array.from(doc.querySelectorAll("p"))
+    // this gives us all paragraphs that don't end with a period
+    .filter(elem => !elem.textContent.trim().split("tests")[0].endsWith("."))
```
marcoscaceres: This points to a larger (known) issue that we need to fix... basically, the `<details>` for tests inside `p` are invalid HTML. Nevertheless, we should not special case "tests" here.

shikhar-scs: Umm okay. I'll remove that then.

marcoscaceres: Yeah, start it up!

shikhar-scs:
@marcoscaceres I've added the tests. Please verify if it covers everything or not.

tests/spec/core/linter-rules/check-period-in-p.js
```diff
@@ -0,0 +1,68 @@
+describe("Core Linter Rule - 'check-period-in-p'", () => {
+  const ruleName = "check-period-in-p";
+  const config = { lint: { [ruleName]: true } };
+  let rule;
+  beforeAll(async () => {
+    rule = await new romise(resolve => {
```
marcoscaceres: Nit: “romise”

tests/spec/core/linter-rules/check-period-in-p.js
```diff
@@ -0,0 +1,68 @@
...
+      doc.body.removeChild(doc.body.firstChild);
+    }
+  });
+  it("checks p ending without a period", async () => {
+    doc.body.innerHTML = `
+        <section>
+          <p id="no-fullstop">Should pass</p>
+          <p>Should fail.</p>
+          <p id="no-fullstop-at-end">Should. pass</p>
+          <p >Should. Fail.</p>
```
marcoscaceres: Nit: space after p

tests/spec/core/linter-rules/check-period-in-p.js
```diff
@@ -0,0 +1,68 @@
...
+  beforeAll(async () => {
+    rule = await new romise(resolve => {
+      require([`core/linter-rules/${ruleName}`], ({ rule }) => resolve(rule));
+    });
+  });
+  const doc = document.implementation.createHTMLDocument("test doc");
+  beforeEach(() => {
+    // Make sure every unordered test get an empty document
+    // See: https://github.com/w3c/respec/pull/1495
+    while (doc.body.firstChild) {
```
marcoscaceres: Nit: .innerHTML = “”

marcoscaceres: Actually, just create a fresh document as needed. No need for beforeEach.

tests/spec/core/linter-rules/check-period-in-p.js
```diff
@@ -0,0 +1,68 @@
...
+    expect(result.offendingElements[1]).toEqual(noFullStopAtEnd);
+    expect(result.offendingElements[2]).toEqual(unnecessaryTestsAfterFullStop);
+  });
+  it("checks error message for p ending without a period", async () => {
+    doc.body.innerHTML = `
+        <section>
+          <p id="no-fullstop">Should pass</p>
+          <p>Should fail.</p>
+          <p id="no-fullstop-at-end">Should. pass</p>
+          <p >Should. Fail.</p>
```
marcoscaceres: As above.

tests/spec/core/linter-rules/check-period-in-p.js
marcoscaceres: Bug actually;)

shikhar-scs: oops ok

shikhar-scs:
I've incorporated all the changes about the `nit` errors. 
> On phone, so will take a closer look tomorrow.

Yep sure :smile:

tests/spec/core/linter-rules/check-period-in-p.js
```diff
@@ -0,0 +1,62 @@
+fdescribe("Core Linter Rule - 'check-period-in-p'", () => {
```
marcoscaceres: nit: fdescribe.

marcoscaceres:
Ok, so, now we would need to check this against at least:

* https://w3c.github.io/html-media-capture/
* https://w3c.github.io/manifest/
* https://w3c.github.io/payment-request/
* https://w3c.github.io/resource-hints/
* https://w3c.github.io/wpub/
* https://wicg.github.io/web-share-target/

To see what edge cases come back. You will need to download those specs, and replace the `<script>` that loads ReSpec with your local version (e.g., from basic.html):

```HTML
    <script src='../js/deps/require.js' data-main='../js/profile-w3c-common' 
      async class='remove'></script>
```

That should give us some idea about the edge cases.

shikhar-scs:
Ok. I'll get back to it.

shikhar-scs:
@marcoscaceres I tried running all of them. And this is what I got : 
 - https://w3c.github.io/html-media-capture/ - 6 occurrences : 3 `<p>` ended with a semicolon, 2 had only `` tags & 1 for `Back to top`.

 - https://w3c.github.io/manifest/ :  29 occurrences : Semicolon & question Mark ended `<p>`.  Only this doc potentially contained some `<p>` ending without a period.

 - https://w3c.github.io/payment-request/ - 47 occurrences :Consisted of semicolon ended and `tests : {a number}` keyword, `Back-to-top` is present in all.

 - https://w3c.github.io/resource-hints/ - 10 occurrences : 9 had semicolon endings, 1 Back to top

 - https://w3c.github.io/wpub/ - didn't work, I dont know why.

 - https://wicg.github.io/web-share-target/ - 10 occurrences : `semicolon` endings


So, semicolons, question marks, back-to-top, `` tags & `tests` keyword

shikhar-scs:
:

marcoscaceres:
Ok, this is super interesting @shikhar-scs! So, it seems like we need to take colons (":") into consideration. We also need to filter out paragraphs that only have white space. 

And we need to ignore the "back to top" element... we can filter that out, as that's added by ReSpec.

marcoscaceres:
for the the query... `document.querySelectorAll("p:not(#back-to-top)")`

marcoscaceres:
I guess if we have "?", then we want to check for "!" too...

marcoscaceres:
Sorry to flood with comments... but I guess we need to rename this rule to "check-punctuation" or something similar.

marcoscaceres:
> https://w3c.github.io/wpub/

- [ ] We should remember to debug this. That's quite a good one to check, and it's interesting if it's not working.

shikhar-scs:
Yeah! No problem. I have college in a while so I might not be able to complete this now. I'll get it done today though. :+1: 
We need to consider `tests : {n}` too

shikhar-scs:
@marcoscaceres I've pushed up changes. Sorry for the delay, I had some important work yesterday :sweat_smile: .

src/core/linter-rules/check-period-in-p.js
```diff
@@ -0,0 +1,55 @@
...
+ * @param  {Document} doc The document to be checked.
+ */
+function lintingFunction(conf, doc) {
+  console.log("dsfdf")
+  
+  const allElements = Array.from(doc.querySelectorAll("p"));
+  console.log(allElements);
+  const offendingElements = Array.from(doc.querySelectorAll("p"))
+    // this gives us all paragraphs that don't end with a period
+    .filter(elem => !elem.textContent.trim().endsWith("."))
```
marcoscaceres: note that we are doing same operations over and over again here, which is not ideal (filter, access textContent, trim, check ending). 

We can combine the the punctuation checks with a regex like `[.!?:]\s*$` (ends with any these characters, and zero or more spaces).

src/core/linter-rules/check-period-in-p.js
```diff
@@ -0,0 +1,55 @@
...
+    .filter(elem => !elem.textContent.trim().endsWith(":"))
+    .filter(elem => !elem.textContent.trim().endsWith("?"))
+    .filter(elem => !elem.textContent.trim().endsWith("!"))
+    .filter(elem => (elem.textContent.trim() !== ""))
+    .filter(elem => !elem.textContent.trim().includes("tests:"))
```
marcoscaceres: nit, don't forget to remove this one... we can deal with this separately.

src/core/linter-rules/check-period-in-p.js
marcoscaceres: The regex needs to work multiline too.

shikhar-scs: Ohk. No problem. I'll try the regex :+1:

src/core/linter-rules/check-period-in-p.js
shikhar-scs: sure :+1:

src/core/linter-rules/check-period-in-p.js
shikhar-scs: @marcoscaceres I tried using the regex. I had never used it before so it took some time to understand. I guess its working now. Please have a look. :smile:

src/core/linter-rules/check-period-in-p.js
```diff
@@ -23,26 +23,12 @@ const lang = defaultLang in meta ? defaultLang : "en";
...
   const offendingElements = Array.from(doc.querySelectorAll("p"))
-    // this gives us all paragraphs that don't end with a period
-    .filter(elem => !elem.textContent.trim().endsWith("."))
-    .filter(elem => !elem.textContent.trim().endsWith(":"))
-    .filter(elem => !elem.textContent.trim().endsWith("?"))
-    .filter(elem => !elem.textContent.trim().endsWith("!"))
-    .filter(elem => (elem.textContent.trim() !== ""))
-    .filter(elem => !elem.textContent.trim().includes("tests:"))
-    
+    .filter((elem) => !(/[.!?:]$|^ *$/.test(elem.textContent.trim()))
```
marcoscaceres: As regexes are super annoying to read, always put a comment above describing what it does.

src/core/linter-rules/check-period-in-p.js
marcoscaceres: Ah yeah, regexes are the worst thing ever. But can be useful in situations like this. I live in:

https://www.regexpal.com

There are other useful sites that help unpack/write regexes.

src/core/linter-rules/check-period-in-p.js
marcoscaceres: So, in theory, we shouldn't need .trim() there either. 
```
.filter(({textContent}) => !(/[.!?:]$|^ *$/.test())
```

marcoscaceres: fixed mistake above

marcoscaceres: ok, now we also want multiline... so we need "m":

.filter(({textContent}) => !(/[.!?:]$|^\s*$/m.test(textContent))

src/core/linter-rules/check-period-in-p.js
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * Runs linter rule.
+ *
+ * @param {Object} config The ReSpec config.
+ * @param  {Document} doc The document to be checked.
+ */
+function lintingFunction(conf, doc) {
+  
+  const offendingElements = Array.from(doc.querySelectorAll("p"))
+    .filter((elem) => !(/[.!?:]$|^ *$/.test(elem.textContent.trim()))
```
marcoscaceres:
```JS
// punctuation marks OR zero or more space characters
const punctuationRegExp = /[.!?:]$|^\s*$/m
const offendingElements = Array.from(doc.querySelectorAll("p"))
  .filter(({textContent} ) => !(punctuationRegExp.test(textContent))
```

marcoscaceres: Something like the above... didn't test it.

shikhar-scs: Sure sure. I'll test it and push it right away :+1:

shikhar-scs: Done :grin: :grin: . And I've tested all edge cases too.

marcoscaceres: Keep those edge cases. Will be useful for actual testing.

shikhar-scs: Yeah Sure. Let's work on `{tests: n}` now. Tests are also running, I verified. Also we'll have to introduce new ones for the other punctuations and spaces . I'll include them too. 😁

shikhar-scs: @marcoscaceres any info here?
> Let's work on {tests: n}

shikhar-scs:
Hey @marcoscaceres lookup this whenever you're free

marcoscaceres:
Looks good. There are still a few unaddressed nits there from previous reviews. Main thing that needs to change is the name. Should be called “check-punctuation” everywhere that’s applicable.

tests/spec/core/linter-rules/check-period-in-p.js
```diff
@@ -0,0 +1,60 @@
...
+          <p id="no-fullstop">Should pass</p>
+          <p>Should fail.</p>
+          <p id="no-fullstop-at-end">Should. pass</p>
+          <p>Should. Fail.</p>
+          <p id="unnecessary-tests-after-fullstop">Should Pass. Tests</p>
+        </section>
+    `;
+  
+    const results = await rule.lint(config, doc);
+    const [result] = results;
```
marcoscaceres: Something is not right here... each offending element should have a unique result.

src/core/linter-rules/check-period-in-p.js
```diff
@@ -0,0 +1,43 @@
...
+ *   in the ReSpec config.
+ */
+import { lang as defaultLang } from "core/l10n";
+import LinterRule from "core/LinterRule";
+
+const name = "check-period-in-p";
+
+const meta = {
+  en: {
+    description: "`<p>` tags should end with a period",
```
marcoscaceres: This now needs to talk about missing  punctuation marks instead.

tests/spec/core/linter-rules/check-period-in-p.js
```diff
@@ -0,0 +1,60 @@
...
+    `;
+
+    const noFullStop = doc.getElementById("no-fullstop");
+    const noFullStopAtEnd = doc.getElementById("no-fullstop-at-end");
+    const unnecessaryTestsAfterFullStop = doc.getElementById("unnecessary-tests-after-fullstop");
+    
+    const results = await rule.lint(config, doc);
+    const [result] = results;
+    
+    expect(result.offendingElements.length).toEqual(2);
```
marcoscaceres: Make sure that offending elements contain the “respec-offending-element” class. It should have been added automatically, but we want to assure it was.

shikhar-scs:
Sure! I'll get back to all these changes :+1:

tests/spec/core/linter-rules/check-period-in-p.js
shikhar-scs: That unique result is contained in the result array which gets formed. On iterating through result, we can get teach of the results.

shikhar-scs:
@marcoscaceres I had pushed some changes the other day. I guess you forgot to have a look . I had changes the name & the respective nits.

src/core/linter-rules/check-punctuation.js
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * Runs linter rule.
+ *
+ * @param {Object} config The ReSpec config.
+ * @param  {Document} doc The document to be checked.
+ */
+function lintingFunction(conf, doc) {
+  
+  // ensures that either a string ends with one of [.!?:] or is empty
+  const punctuatingRegExp = /[.!?:]$|^ *$/m;
```
marcoscaceres: now we can do:

```JS
const punctuatingRegExp = new RegExp(`[${punctuationMarks.join("")}]$|^ *$`, "m");
```

src/core/linter-rules/check-punctuation.js
```diff
@@ -0,0 +1,44 @@
...
+ *   in the ReSpec config.
+ */
+import { lang as defaultLang } from "core/l10n";
+import LinterRule from "core/LinterRule";
+
+const name = "check-punctuation";
+
+const meta = {
+  en: {
+    description: "`<p>` tags should end with either of . | : | ! | ? ",
```
marcoscaceres: Apart from us 🤓s, it's a bit uncommon to use "|" to separate things. Let's use "," instead.  

```JS
description: "Found `<p>` element(s) without an ending punctuation mark.",
howToFix: `Add one of the following add one of the following punctuation marks to the offending element(s): ${punctuationMarks.join(", ")}`,
```

src/core/linter-rules/check-punctuation.js
```diff
@@ -0,0 +1,44 @@
+/**
+ * Linter rule "check-punctuation". Makes sure the there are no punctuations missing at the end of a <p>
+ *   in the ReSpec config.
+ */
+import { lang as defaultLang } from "core/l10n";
+import LinterRule from "core/LinterRule";
+
+const name = "check-punctuation";
+
```
marcoscaceres: here, let's add:

```JS
const punctuationMarks = [".", ":", "!", "?"];
```

That will allow us to more easily manage these throughout the code.

tests/spec/core/linter-rules/check-punctuation.js
```diff
@@ -0,0 +1,61 @@
...
+  beforeAll(async () => {
+    rule = await new Promise(resolve => {
+      require([`core/linter-rules/${ruleName}`], ({ rule }) => resolve(rule));
+    });
+  });
+  it("checks p ending without a punctuation", async () => {
+    const doc = document.implementation.createHTMLDocument("test doc");
+    doc.body.innerHTML = `
+        <section>
+          <p id="no-fullstop">Should pass</p>
```
marcoscaceres: Saying "pass" it's a bit confusing here, because this is about catching punctuation errors. This should maybe say:

```JS
should be caught by linter
```  

And variations of the above. 

And: 

```HTML
<p>This is fine.</p>
```

Also, we want to make sure we test a bit thoroughly by adding some whitespace:

```
<p>
  \t Should. be! caught. by? linter 
</p>
```

We also want to check for all the marks:

```
<p>
  This is fine.
</p>
<p>
  This is fine!
</p>
<p>
  This is fine?
</p>
<p>
  This is fine:
</p>
<p>
  !.?: This should be caught by linter
<p/>
```

tests/spec/core/linter-rules/check-punctuation.js
```diff
@@ -0,0 +1,61 @@
...
+    const results = await rule.lint(config, doc);
+    const [result] = results;
+    
+    const noFullStop = doc.getElementById("no-fullstop");
+    const noFullStopAtEnd = doc.getElementById("no-fullstop-at-end");
+    const unnecessaryTestsAfterFullStop = doc.getElementById("unnecessary-tests-after-fullstop");
+    
+    expect(result.name).toEqual(ruleName);
+    expect(result.occurrences).toEqual(2);
+    expect(result.description).toEqual("`<p>` tags should end with either of . | : | ! | ? ");
```
marcoscaceres: Here, we should just expect that the description `.toBeTruthy()`, as all we really care about is that some description is given. Same with `howToFix`.

tests/spec/core/linter-rules/check-punctuation.js
marcoscaceres: Updated one test case above.

src/w3c/defaults.js
```diff
@@ -52,6 +53,7 @@ const w3cDefaults = {
     "no-headingless-sections": true,
     "privsec-section": true,
     "no-http-props": true,
+    "check-punctuation": true,
```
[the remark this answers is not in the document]
shikhar-scs: Yep Sure ! I do it just for testing & then forget. Sorry :sweat_smile:

tests/spec/core/linter-rules/check-punctuation.js
```diff
@@ -0,0 +1,61 @@
+fdescribe("Core Linter Rule - 'check-punctuation'", () => {
```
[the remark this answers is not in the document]
shikhar-scs: Again. That's just for testing. I t skips off my mind before pushing :sweat_smile:

src/w3c/defaults.js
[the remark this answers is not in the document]
marcoscaceres: Coincidentally, that's actually a good test... should check that it's not enabled by default.

tests/spec/core/linter-rules/check-punctuation.js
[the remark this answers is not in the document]
marcoscaceres: No probs, happens to me all the time :) Keep meaning to write a git commit hook to check for it.

tests/spec/core/linter-rules/check-punctuation.js
shikhar-scs: Lolol :laughing: The test cases. I'll update them now :+1:

tests/spec/core/linter-rules/check-punctuation.js
shikhar-scs: Done :+1:

shikhar-scs:
@marcoscaceres I've pushed in the changes.  A strange thing is that the linter property now is not counting 
```html
          <p id="no-fullstop">Should be caught by linter</p>
```
the above as an error. I tried a lot of permutes but it doesn't anyhow. What can be the reason ?

src/core/linter-rules/check-punctuation.js
```diff
@@ -0,0 +1,45 @@
...
+ */
+import { lang as defaultLang } from "core/l10n";
+import LinterRule from "core/LinterRule";
+
+const name = "check-punctuation";
+const punctuationMarks = [".", ":", "!", "?"];
+
+const meta = {
+  en: {
+    description: "`<p>` tags should end with either of ., :, !, ? ",
```
marcoscaceres: See my previous comment about using the `punctuationMark` here via a template string. Otherwise, we would need to update this in multiple places.

src/core/linter-rules/check-punctuation.js
```diff
@@ -0,0 +1,45 @@
...
+ */
+function lintingFunction(conf, doc) {
+  
+  // ensures that either a string ends with one of [.!?:] or is empty
+  const punctuatingRegExp = new RegExp(`[${punctuationMarks.join("")}]$|^ *$`, "m");
+  
+  const offendingElements = Array.from(doc.querySelectorAll("p"))
+    .filter((elem) => !(punctuatingRegExp.test(elem.textContent)));
+  
+  offendingElements.splice(-1,1); // back-to-top is always the last element.
```
marcoscaceres: We want to filter this out via css (i.e., via “:not()”. This is probably causing the issue you mentioned below.

marcoscaceres: As a general rule, and even though not applicable here, never use `splice()`, alway `filter()`. You should treat all data sources (in this case “offendingElements”) as immutable.

marcoscaceres: See https://github.com/w3c/respec/pull/1589#comment-376754462

shikhar-scs: Thanks! I've pushed the new changes

shikhar-scs: @marcoscaceres look up the new commit whenever u are free

src/core/linter-rules/check-punctuation.js
```diff
@@ -0,0 +1,43 @@
...
+ */
+import { lang as defaultLang } from "core/l10n";
+import LinterRule from "core/LinterRule";
+
+const name = "check-punctuation";
+const punctuationMarks = [".", ":", "!", "?"];
+
+const meta = {
+  en: {
+    description: `<p> tags should end with either of ${punctuationMarks} `,
```
marcoscaceres: Please confirm this is rendering properly. The punctuationMarks will be missing spaces, no?

marcoscaceres:
Codacy spotted a few issues too.

marcoscaceres:
Also, test needs updating:

```

FAILED TESTS:
  W3C — Defaults
    ✖ sets sensible defaults for w3c specs
      HeadlessChrome 65.0.3325 (Linux 0.0.0)
    Expected object not to have properties
        check-punctuation: false
    Error: Expected object not to have properties
        at <Jasmine>
        at UserContext.it (tests/spec/w3c/defaults-spec.js:11:25)
```

src/core/linter-rules/check-punctuation.js
shikhar-scs: This is what is showing 

screenshot from 2018-04-03 07-49-31

shikhar-scs:
@marcoscaceres 
Upated the tests & codacy errors. In a previous comment I've also shown what gets rendered. :sweat_smile:

src/core/linter-rules/check-punctuation.js
marcoscaceres: Ok, I thought it might :) Fixing -. I forgot to mention that `description` and `howToFix` are markdown.

shikhar-scs:
That's awesome :tada:

shikhar-scs:
Yayyyyyy !!! Thanku @marcoscaceres 🎉🎉

Issue #879, opened 2016-07-20
Allow for figure references not to include the whole caption


r12a:
It's extremely useful to be able to automatically fill in the link text for pointers to figures. (Jlreq has over 300 figures, most of which are referred to from the text.)  However, it's a bit problematic when the link text becomes something like

`see Fig. 10: Example of face tsumegumi in horizontal writing mode. (The 1st line is the same text with solid setting, for comparison.).`

Note also that this tends to produce a double full stop at the end.

I'd like to be able to produce just the following instead

`see Fig. 10`

I think this should be the default actually.

(Note that i'm happy to continue filling section links with the section title – that gives a useful result.)

r12a:
In case it helps, i looked at the code. I'm not familiar with this style of js, but i guess the basic idea is as follows.

go to https://github.com/w3c/respec/blob/develop/js/core/figures.js

change
```
if (figMap[id]) {
        $a.addClass("fig-ref");
        if ($a.html() === "") $a.append(figMap[id].clone());
}
```

to append the following information created as part of a figure
```
.prepend($("<span class='figno'>" + num + "</span>"))
.prepend(doc.createTextNode(conf.l10n.fig))
```

rather than just cloning the whole figcaption.

r12a:
Still waiting for this to be done.  How can we move it forward?

halindrome:
I am uncomfortable just implementing this as it would significantly change the way the tool works for existing users.  I wouldn't mind this with an option I guess...   conf.figRefNumOnly = true or something.  @marcoscaceres what do you think?

r12a:
I'd prefer it to be a default, but i hear what you say, and a switch would be ok.

r12a:
Can we raise the priority of this?  I'm trying to prepare a doc for publication and hitting validation errors because the caption of the figure contains a tags linking to definitions.

```
<a href="#figure-letter_reh_shapes" class="fig-ref">Figure <span class="figno">6</span> 
<span class="fig-title">
<a href="#dfn-right-joining" class="internalDFN" data-link-type="dfn">Right-joining 
letters</a> only have two forms of <a href="#dfn-final" class="internalDFN" data-link-type="dfn">final</a> 
and <a href="#dfn-isolated" class="internalDFN" data-link-type="dfn">isolated</a>.
</span></a>
```

Which means that every time i generate the doc from respec, i need to track down and manually change the markup here (by removing the figure caption from the reference).  That's definitely broken!

[...]

PR #1593, opened 2018-03-28
feat(core/figures): anchor link texts to be shortened

himanish-star:
fixes: #879


src/core/figures.js
```diff
@@ -76,7 +76,7 @@ export function run(conf, doc, cb) {
     id = id.substring(1);
     if (figMap[id]) {
       $a.addClass("fig-ref");
-      if ($a.html() === "") $a.append(figMap[id].clone());
+      if ($a.html() === "") $a.append(figMap[id].slice(0, 2).clone());
```
marcoscaceres: To make this a bit more legible... 

```JS
if ($a.html() === "") {
  const $shortFigDescriptor = figMap[id].slice(0, 2).clone();
  $a.append(shortFigDescriptor);
}
```

himanish-star:
ah ok

himanish-star:
Now: 
image

Before:
image

marcoscaceres:
Looks good. For the `<a>`, we should add a title attribute containing the text we are now taking out. Then users can hover over “figure x” to see what it’s referring to exactly.

r12a:
This all looks very good. Thank you very much!

One question. Not having looked closely at the code, I hope the text 'Figure 1' will not appear in documents in other languages. For example, we have https://w3c.github.io/jlreq/ja/ where 'Figure 1' would be '図1'. I believe we have some localisation code in respec to handle that, and it should work here too.

himanish-star:
Sure thanks @marcoscaceres and @r12a for the review

himanish-star:
image

@r12a, this is ok now?

r12a:
Looks good, though it would be easier to spot details if you could show a picture of the jlreq doc at https://w3c.github.io/jlreq/ja/#kanji_hiragana_and_katakana. There is currently a problem in that references in the text to a figure seem to have a space added after. See the blue highlight in this picture:

screen shot 2018-03-28 at 08 07 37

There should be no spaces in the Japanese text. I'm not sure where this comes from: whether it's from the localization code or from the figure link insertion.

r12a:
(oops, wrong picture, i'm going to edit in place).

himanish-star:
Yeah sure

himanish-star:
image
Does this look right now @r12a?

himanish-star:
Also this (caption)
image

himanish-star:
@marcoscaceres, I have written down the tests too for this! :smile:

r12a:
>  Does this look right now @r12a?

I think it does indeed look correct now. Thanks.

himanish-star:
> I think it does indeed look correct now. Thanks.

Oh, that's nice. I have also set the title attribute of the anchor element to the figure caption

r12a:
> I have also set the title attribute of the anchor element to the figure caption

That's a nice feature.

src/core/figures.js
```diff
@@ -76,7 +76,12 @@ export function run(conf, doc, cb) {
     id = id.substring(1);
     if (figMap[id]) {
       $a.addClass("fig-ref");
-      if ($a.html() === "") $a.append(figMap[id].clone());
+      if ($a.html() === "") {
+        const $shortFigDescriptor = figMap[id].slice(0, 2).clone();
+        const longFigDescriptor = figMap[id].slice(2).clone().text();
+        $a.attr("title", longFigDescriptor.trim());
```
marcoscaceres: Small edge case here, but we should only set `title`, if the `title` attribute does not exist.

tests/spec/core/figures-spec.js
```diff
@@ -22,6 +22,42 @@ describe("Core - Figures", function() {
     expect(captions.item(0).textContent).toEqual("Figure 1 PREFIG");
     expect(captions.item(1).textContent).toEqual("Figure 2 IMGTIT");
   });
+
+  it("creates autolinks from the anchor to the figure", async () => {
+    const ops = {
+      config: makeBasicConfig(),
+      body:
+      makeDefaultBody() +
```
marcoscaceres: Nit: use multiline template strings, instead of concatenation with "+" ... makes code more easy to change later.

himanish-star: ok

src/core/figures.js
himanish-star: Oops! Didn't think of that!

tests/spec/core/figures-spec.js
```diff
@@ -22,6 +22,53 @@ describe("Core - Figures", function() {
...
+    const ops = {
+      config: makeBasicConfig(),
+      htmlAttrs: {
+        lang: "ja",
+      },
+      body:
+      makeDefaultBody() +
+      `<figure id='fig'> <img src='img' alt=''>
+        <figcaption>漢字と仮名のサイズの示し方</figcaption>
+       </figure>" +
```
marcoscaceres: nit: " +

marcoscaceres:
+ respec@20.4.0 

Great work @himanish-star!

PR #1595, opened 2018-03-28
fix(show-people): treat name as HTML

### LINKED ISSUE #1594: Language markup showing in Editors list
## Important info

* https://w3c.github.io/jlreq/
* ReSpec version: 20.3.0
* [ yes ] I did a "hard refresh", but it's still busted. 

## Description of problem

Expected: 

But what actually happened (the bug): Language-related markup showing through in Editors list.

screen shot 2018-03-28 at 08 07 37

marcoscaceres:
* fixes #1594


marcoscaceres:
We need to add a test for this, but ran out of time. If anyone want to be a hero and add one, would be greatly appreciated. Otherwise, will do it tomorrow.

OrionStar25:
I'll do it, I'm very free :smiley_cat:

src/w3c/templates/show-people.js
```diff
@@ -7,19 +7,20 @@ export default (conf, name, items = []) => {
...
+    const personName = html`${{html: p.name}}`;
     const editorid = p.w3cid ? parseInt(p.w3cid, 10): null;
     const dd = html`<dd class='p-author h-card vcard'
       data-editor-id='${editorid}'></dd>`;
       const span = document.createDocumentFragment();
     const contents = [];
     if (p.mailto) {
       contents.push(html`<a class='ed_mailto u-email email p-name'
-        href='${`mailto:${p.mailto}`}'>${p.name}</a>`);
+        href='${`mailto:${p.mailto}`}'>${personName}</a>`);
```
saschanaz: `${[p.name]}` should be enough to treat it as HTML.

https://viperhtml.js.org/hyperhtml/documentation/#essentials-8

>if it's an array of strings, they will be injected as explict opt-in for HTML

src/w3c/templates/show-people.js
```diff
@@ -7,19 +7,20 @@ export default (conf, name, items = []) => {
   return results;
 
   function getItem(p, i) {
+    const personName = html`${{html: p.name}}`;
```
saschanaz: How about `[p.name]`?

https://viperhtml.js.org/hyperhtml/documentation/#essentials-8

>if it's an array of strings, they will be injected as explict opt-in for HTML

marcoscaceres: Good suggestion!

src/w3c/templates/show-people.js
saschanaz: We probably should just do inline `${[p.name]}`, otherwise the same text node will just change its parent three times.

marcoscaceres: not sure I follow? Can you clarify?

marcoscaceres:
@OrionStar25, thanks for offering to help.

The file you want to add tests to is: tests/spec/w3c/headers-spec.js

When testing, please use this config:

```JS
const opts = {
  editors: [
    {
      name:
        "<span lang='ja'>阿南 康宏</span> (Yasuhiro Anan), (<span lang='ja'>第１版</span> 1st edition)",
      company: "Microsoft",
    },
  ],
  formerEditors: [
    {
      name:
        "<span lang='ja'>阿南 康宏</span> (Yasuhiro Anan), (<span lang='ja'>第１版</span> 1st edition)",
      company: "Microsoft",
    },
  ],
};
```

And, eventually:

```JS
expect(editor.textContent).toEqual("阿南 康宏 (Yasuhiro Anan), (第１版 1st edition) (Microsoft)")
expect(formerEditor.textContent).toEqual("阿南 康宏 (Yasuhiro Anan), (第１版 1st edition) (Microsoft)")
```

marcoscaceres:
Oh, and also check that the HTML element is there.

```
const firstSpan = editor.querySelector("span");
expect(firstSpan.lang).toBe("ja");
expect(firstSpan.textContent).toBe("阿南 康宏");

```

src/w3c/templates/show-people.js
saschanaz: Previous ``personName = html`${{html: p.name}}` `` creates a node, ``html`${personName}` `` adds it to another parent node, and then the same ``html`${personName}` `` moves the child node to a new parent (which means the previous parent loses a child), etc.

Now `personName` is not a node but an array so it's irrelevant 😁

marcoscaceres: Ah, yeah... got it. Whoops. Ok, yeah, so `[]` is all round better.

marcoscaceres:
@saschanaz, added tests. Awaiting your approval when you have time.

src/w3c/templates/show-people.js
marcoscaceres: See line 10 for `const personName = [p.name];`.

tests/spec/w3c/headers-spec.js
```diff
@@ -180,6 +181,33 @@ describe("W3C — Headers", function() {
...
+          },
+        ],
+      };
+      const ops = makeStandardOps(config);
+      const doc = await makeRSDoc(ops);
+      const dtElems = [...doc.querySelectorAll(".head dt")];
+      const dtElem = dtElems.find(findEditor);
+      const ddElem = dtElem.nextElementSibling;
+      const [personName, edition] = ddElem.querySelectorAll(
+        "span>span:nth-child(1), span>span:nth-child(2)"
```
saschanaz: IMO `span>span` should be enough, but not a blocker.

saschanaz:
It seems Codacy doesn't like the new single quote.

tests/spec/w3c/headers-spec.js
marcoscaceres: Good point. Will change it.

marcoscaceres:
Huge thanks to @OrionStar25 for helping with the tests.

marcoscaceres:
+ respec@20.5.1

PR #1600, opened 2018-04-02
fix(core/webidl): indent stringifiers

### LINKED ISSUE #1599: IDL stringifier is causing RangeError
## Important info

* URL to affected spec: https://w3c.github.io/selection-api/
* ReSpec version: 20.4.0.

## Description of problem

Expected:
Selection interface should be displayed correctly. 

But what actually happened (the bug):
The IDL below throws: 

```
RangeError: Invalid count value
    at String.repeat (<anonymous>)
    at pads (webidl.js:204)
    at Object.<anonymous> (webidl.js:105)
    at Object.main (templates.js:361)
    at main (handlebars.js:1145)
    at ret (handlebars.js:1148)
    at writeMethod (webidl.js:701)
    at webidl.js:604
    at Array.map (<anonymous>)
    at writeInterfaceDefinition (webidl.js:599)
```
Optional, steps to reproduce: 

Given: 
```IDL
interface Selection {
          stringifier DOMString ();
};
```

Seems this has regressed. It seems our friend "pads()" is again getting negative values:

```JS
function pads(num) {
  return " ".repeat(num);
}
```

marcoscaceres:
* fixes #1599

### DIFF
--- a/src/core/webidl.js
+++ b/src/core/webidl.js
@@ -589,6 +589,12 @@ function writeInterfaceDefinition(opt, fixes = {}) {
     } else if (it.type === "operation") {
       if (it.static) {
         len += "static ".length;
+      } else if (it.stringifier) {
+        len += "stringifier ".length;
+      } else if (it.getter) {
+        len += "getter ".length;
+      } else if  (it.setter) {
+        len += "setter ".length;
       }
       maxMeth = Math.max(len, maxMeth);
     } else if (it.type === "const") {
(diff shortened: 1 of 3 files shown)


saschanaz:
LGTM but a test is required.

marcoscaceres:
That's the weird things... we did have tests... ("MethBasic" in "tests/spec/core/webidl.html") :( Adding some more tho.

saschanaz:
The existing test currently only effectively tests the longest indentation (so only `static Promise&lt;RTCCertificate>` now). That's why the test couldn't catch this problem.

marcoscaceres:
Yeah, spotted it. Whoopsy. Incoming tests.

marcoscaceres:
Figures I'd add independent tests for each special type. Then we can more easily catch regressions if we ever deal types independently.

PR #1602, opened 2018-04-02
feat(core/utils): Utils fetch and cache

marcoscaceres:
Add fetchAndCache() utility, which fetches a request, caches the response in the Cache API. If the response is bad, and we have a state response we can use, we serve it.


marcoscaceres:
Let's treat this as part 1. Let's then integrate with caniuse.

tests/spec/core/utils-spec.js
```diff
@@ -8,6 +8,56 @@ describe("Core - Utils", () => {
...
+        expect(bodyText).toBe("PASS");
+        await clearCaches();
+      }
+    });
+
+    it("uses the origin as the cache key", async () => {
+      expect(await caches.keys()).toEqual([]);
+      const url = location.origin + "/tests/data/pass.txt";
+      await utils.fetchAndCache(url);
+      expect(await caches.keys()).toEqual([location.origin]);
```
sidvishnoi: maybe `Array.prototype.includes` would be clearer?

tests/spec/core/utils-spec.js
```diff
@@ -8,6 +8,56 @@ describe("Core - Utils", () => {
...
+    const keys = await caches.keys();
+    for (const key of keys) {
+      await caches.delete(key);
+    }
+  }
+
+  describe("fetchAndCache", () => {
+    beforeEach(clearCaches);
+    it("caches a requests of different type with a 1 day default", async () => {
+      const url = location.origin + "/tests/data/pass.txt";
```
sidvishnoi: We should add this file to `karma.conf.js`

Maybe merge in this: https://github.com/w3c/respec/blob/3c8defdee494ecc6d45e602d703c1e6ec51fa0f0/karma.conf.js#L50,L54

as 
``` js
     {
        pattern: "tests/data/**/*",
        included: false,
        served: true,
      },
```

tests/spec/core/utils-spec.js
```diff
@@ -8,6 +8,56 @@ describe("Core - Utils", () => {
...
+      for(const request of requests){
+        expect(await caches.match(request)).toBe(undefined);
+        await utils.fetchAndCache(url);
+        const cache = await caches.open(location.origin);
+        expect(cache).toBeTruthy();
+        const cachedResponse = await cache.match(url);
+        expect(cachedResponse).toBeTruthy();
+        const expires = new Date(cachedResponse.headers.get("Expires")).valueOf();
+        expect(expires).toBeGreaterThan(Date.now());
+        expect(expires).toBeGreaterThan(Date.now() + 86000000);
```
sidvishnoi: maybe add a comment here saying `86400000` is default `maxAge`

tests/spec/core/utils-spec.js
```diff
@@ -8,6 +8,56 @@ describe("Core - Utils", () => {
...
+
+  describe("fetchAndCache", () => {
+    beforeEach(clearCaches);
+    it("caches a requests of different type with a 1 day default", async () => {
+      const url = location.origin + "/tests/data/pass.txt";
+      const requests = [
+        url,
+        new URL(url),
+        new Request(url)
+      ]
```
sidvishnoi: Nit: simicolon

src/core/utils.js
```diff
@@ -447,3 +447,65 @@ export function runTransforms(content, flist) {
...
+    }
+    const msg = `Response not OK from ${url} (HTTP Status: ${response.status})`;
+    throw new NetworkError(msg, response);
+  }
+
+  // cache response
+  if (cache) {
+    const clonedResponse = response.clone();
+    const customHeaders = new Headers(response.headers);
+    const expiryDate = new Date(new Date().valueOf() + maxAge);
```
sidvishnoi: Lets replace this with following for consistency
``` js
const expiryDate = new Date(Date.now() + maxAge);
```

tests/spec/core/utils-spec.js
marcoscaceres: Good point.

tests/spec/core/utils-spec.js
```diff
@@ -8,6 +8,56 @@ describe("Core - Utils", () => {
     });
   });
 
+  async function clearCaches(){
```
sidvishnoi: should this function not go in describe block below?

tests/spec/core/utils-spec.js
marcoscaceres: Here we want to check that it’s exactly the expected array (as we’ve trashed all other caches, so there should be exactly one item).

tests/spec/core/utils-spec.js
marcoscaceres: Good point. Will move it.

marcoscaceres:
@saschanaz, anything else I should test for?

marcoscaceres:
@sidvishnoi, @saschanaz, I found an issue and could use your advice. Consider: 

```JS
      const url = location.origin + "/tests/data/pass.txt";
      await utils.fetchAndCache(url); // default max age
      const response = await utils.fetchAndCache(url, 0);
```

Right now, we return the response from the cache. We don't have a way of respecting that max time override. What should we do here?

saschanaz:
Hmm, we may recalculate expiration time using Date header if exists, but this way we would recalculate every time the page loads.

tests/spec/core/utils-spec.js
```diff
@@ -8,7 +8,7 @@ describe("Core - Utils", () => {
     });
   });
 
-  describe("fetchAndCache", () => {
+  fdescribe("fetchAndCache", () => {
```
saschanaz: typo?

sidvishnoi:
> We don't have a way of respecting that max time override. What should we do here?

Maybe instead of storing an absolute `Expire` time, we can store `max-age`?

saschanaz:
I think if we `fetchAndCache(url, maxAge)` then we always will `fetchAndCache(url, maxAge)`, not `fetchAndCache(url)`. Probably no need to save it then.

marcoscaceres:
Ok, so, I'm ok to go with what we got so far, as it's good enough for our simple purposes. We can then improve it over time. 

@sidvishnoi, @saschanaz, agreed?

saschanaz:
Agreed, but the Travis test is still failing for some reason. Is that still a GitHub API limit error?

marcoscaceres:
> Agreed, but the Travis test is still failing for some reason. Is that still a GitHub API limit error?

Yeah, I think dpub spec is maxing it out. I might remove it, as it does like 30+ pings for github issues.

PR #1603, opened 2018-04-02
fix(w3c/headers): head items should be lcase

marcoscaceres:
Editor should be in lower case, to match how other `dt` items are presented.


marcoscaceres:
Thanks for checking this, @sidvishnoi!

PR #1604, opened 2018-04-03
fix(w3c/headers): allow restrictive w3c license

### LINKED ISSUE #1598: The document license cannot be limited to the W3C Document License
The document license is either:

(a) the permissive license
(b) a dual license if `isCCBY` is true

and the [W3C Document License](https://www.w3.org/Consortium/Legal/2015/doc-license) cannot be selected alone.

This is AFAIK a change from earlier releases, and has a significant intellectual property impact.

Recommend adding an explicit option to select the [W3C Document License](https://www.w3.org/Consortium/Legal/2015/doc-license) alone.

marcoscaceres:
* fixes #1598


marcoscaceres:
@plehegar, see related discussion in #1598

PR #1607, opened 2018-04-03
fix(tools/respecDocWriter): stringify JSHandle objects from console

### LINKED ISSUE #1606: respec2html error is always `😱 ReSpec error: JSHandle@error`
For example: https://travis-ci.org/w3c/respec/jobs/361570561

We should be able to read JSHandle object wrapping Error, returned by puppeteer.

saschanaz:
Fixes #1606 🎉


marcoscaceres:
Well, at least now we can see that it’s definitely a network error. Those tests are kinda useless, as they test the currently deployed version of ReSpec, instead of the new version.

Issue #1416, opened 2017-10-26
Add "kitchen sink" test


marcoscaceres:
We've been using basic.html to exercise full spec generation - however, it's insufficient and ugly. We  should have a dedicated file to test all the things working together.

anweshknayak:
hey @marcoscaceres I would like to work on this one. Could you tell me how should I proceed?

marcoscaceres:
@anweshknayak, I think the easiest thing to do would be to got through the Wiki and copy/paste the examples into a new document. https://github.com/w3c/respec/wiki

you can copy/paste the "starter.html" in example to start with. I would start with a very minimal document and just start adding one one configuration property at a time.

Some things just need to be included into the `respecConfig`, like: `implementationReportURI`. 

Others, want to separate each test into a section. For example, a `note`:

```HTML
<section>
  <h2>Note test</h2>
  <div class="note" title="This is a note">
     <p>This is a note</p>
  </div>
</section>
```

anweshknayak:
@marcoscaceres so what I figured out is. I start by copy/pasting the examples from the **wiki** to a document. Something in `respec/tests/` would do? 
And do you mean like every configuration into one file?

Sorry If I am wrong.

marcoscaceres:
Yep, what you describe is correct. Note that some configuration options will conflict with each other, so just pick ones that don't (ReSpec should warn you if there is a conflict). 

The goal is to exercise as much of ReSpec as possible. 

However, as you start, keep the commits small (around ~100 lines is ideal)... otherwise it's going to be difficult for us to review. So just send things up as you add them, and I'll do my best to review things as quickly as possible. 

Thanks again!

marcoscaceres:
Sticking with current test set.

PR #1608, opened 2018-04-03
test(headless): reduce tested specs

marcoscaceres:


saschanaz:
So the failure happens if tests are frequent enough. Hmm.

marcoscaceres:
To do this properly, we need #1416.

marcoscaceres:
> I hope ultimately we only use GitHub API minimally in CI tests, but okay for now.

Agree.

PR #1610, opened 2018-04-04
chore(travis): protect against fdescribe and fit

### LINKED ISSUE #622: Write a pre-commit hook that checks that `fdescribe` and `fit` are not present in tests
A `fdescribe()` slipped through, which focuses the test suite :( My bad. Will add a test for this, initially as a pre-commit hook.

marcoscaceres:
* closes  #622
Guard against checking in `fdescribe(` and `fit(` in tests.


saschanaz:
`fdescribe` should trigger test failure because it's undefined, no? It seems it doesn't, that is strange.

marcoscaceres:
fdescribe is defined, it's a jasmine thing: "focus describe", meaning run only these tests.

saschanaz:
Oh, you're right. Sorry.

I neither am familiar with `grep` thing, how about this more javascript-y way by [`eslint-plugin-jasmine`](https://www.npmjs.com/package/eslint-plugin-jasmine)?

`"extends": ["eslint:recommended", "plugin:jasmine/recommended"]` on `.eslintrc.json` will check there is no focused tests.

saschanaz:
AFAIK Codacy already [internally supports `eslint-plugin-jasmine`](https://github.com/codacy/codacy-eslint/blob/99775f683ff190d781b99aa8759c1823ef40187b/build.sbt#L104), it's just that we are not using it. I might be wrong, though...

marcoscaceres:
Ok... this could work.

tests/spec/core/issues-notes-spec.js
```diff
@@ -197,16 +205,22 @@ describe("Core — Issues and Notes", function() {
...
+
     expect(issueDiv3).toBeTruthy();
     const summarySection = doc.getElementById("issue-summary");
+
     expect(summarySection).toBeTruthy();
-    const { textContent } = summarySection.querySelector("[href='#issue-1540']");
-    expect(textContent).toBe("Issue 1540");
+    const { innerText } = summarySection.querySelector("[href='#issue-1540']");
+
+    expect(innerText).toBe("Issue 1540");
```
saschanaz: textContent became innerText?

marcoscaceres: odd... maybe came from the merged branch.

tests/spec/core/biblio-db-spec.js
```diff
@@ -192,7 +200,9 @@ describe("Core - biblioDB", () => {
       var p2 = biblioDB.isAlias(null);
       var p3 = biblioDB.isAlias("");
       await Promise.all([
-        p1.catch(err => expect(err instanceof TypeError).toBe(true)),
+        p1.catch(err =>
+
+        expect(err instanceof TypeError).toBe(true)),
```
saschanaz: This looks strange, what eslint rule forces this?

marcoscaceres: new-line-before-expect | 1 |

marcoscaceres: yeah, that's the "buggy" one... the test could be rewritten to use async/await tho, in which case it would look right. Or could be refactored to use `expect().throws` (can't remember the exact Jasmine method).

tests/spec/w3c/linter-rules/privsec-section-spec.js
```diff
@@ -26,69 +28,89 @@ describe("W3C Linter Rule - 'privsec-section'", () => {
...
     const elem = doc.createElement("h2");
     elem.innerHTML = "Privacy and Security Considerations";
     doc.body.appendChild(elem);
+
     expect((await rule.lint(config, doc)).length).toEqual(0);
     elem.innerHTML = "Security and Privacy Considerations";
+
     expect((await rule.lint(config, doc)).length).toEqual(0);
     elem.innerHTML = "Considerations Security Privacy";
+
```
saschanaz: The newline-before-expect rule also looks weird here... should we just exclude it?

marcoscaceres: I kinda like that rule, tbh. What might be better is to revise the test so it makes more sense.  

Where I would like to get to is for Prettier and Eslint to agree, and so they don't trash each other's changes.

saschanaz: Okay, so that would be a future work 😁

marcoscaceres:
will send again.

PR #1611, opened 2018-04-04
refactor(core/utils): use value instead of getter

saschanaz:
It's a constant value so we can use `value` instead.

### DIFF
--- a/src/core/utils.js
+++ b/src/core/utils.js
@@ -451,7 +451,7 @@ export function runTransforms(content, flist) {
 class NetworkError extends Error {
   constructor(message, response) {
     super(message);
-    Object.defineProperty(this, "response", { get() { return response; } });
+    Object.defineProperty(this, "response", { value: response });
   }
 }
 


src/core/utils.js
```diff
@@ -451,7 +451,7 @@ export function runTransforms(content, flist) {
 class NetworkError extends Error {
   constructor(message, response) {
     super(message);
-    Object.defineProperty(this, "response", { get() { return response; } });
+    Object.defineProperty(this, "response", { value: response });
```
marcoscaceres: as it's readonly, would we want `writable: false` also?

saschanaz: Yup, and it's false by default 👍

marcoscaceres: ah, sorry, I'd assumed it was true by default (that would make no sense). Thanks for clarifying!

marcoscaceres: Given the discussion in the other bug about just using standard fetch error handling, maybe we don't need this at all.

saschanaz: Agreed, closing in favor of standard error handling.

PR #1612, opened 2018-04-04
fix(core/base-runner): extend plugin life

marcoscaceres:
wpub loads a lot of stuff, so the issues and notes plugin can time out.


marcoscaceres:
yeah:(

saschanaz:
Oh, but shouldn't #1583 remedy this with concurrent requests?

marcoscaceres:
Somewhat, but timing depends on client too: Noticed this on our network here which was being slow. Also would affect someone looking at a spec over 3G.  So lots of things at play.

Issue #923, opened 2016-08-23
Add natural width and height to images


marcoscaceres:
Images in spec sometimes lack the natural width and height - which means that when they load, you get layout thrashing and huge performance issues during load. 

We should warn when images are missing width and height attributes, and add them if they are missing on save.

PR #1616, opened 2018-04-05
feat(core/images): add natural width and height to images(#923)

himanish-star:
fixes: #923 

@marcoscaceres, I decided to create a new module name `core/images` as I think the code fits into no other existing modules. I may be wrong though

### DIFF
--- a/src/core/figures.js
+++ b/src/core/figures.js
@@ -60,7 +60,10 @@ export function run(conf, doc, cb) {
       .prepend(doc.createTextNode(conf.l10n.fig));
     figMap[id] = $cap.contents();
     var $tofCap = $cap.clone();
-    $tofCap.find("a").renameElement("span").removeAttr("href");
+    $tofCap
+      .find("a")
+      .renameElement("span")
+      .removeAttr("href");
     tof.push(
       $("<li class='tofline'><a class='tocxref' href='#" + id + "'></a></li>")
         .find(".tocxref")
@@ -79,8 +82,11 @@ export function run(conf, doc, cb) {
       $a.addClass("fig-ref");
       if ($a.html() === "") {
         const $shortFigDescriptor = figMap[id].slice(0, 2).clone();
-        if(!$a[0].hasAttribute("title")) {
-          const longFigDescriptor = figMap[id].slice(2).clone().text();
+        if (!$a[0].hasAttribute("title")) {
+          const longFigDescriptor = figMap[id]
+            .slice(2)
+            .clone()
+            .text();
           $a.attr("title", longFigDescriptor.trim());
         }
         $a.append($shortFigDescriptor);
@@ -118,9 +124,12 @@ export function run(conf, doc, cb) {
 }
 
 function normalizeImages(doc) {
-  [...doc.querySelectorAll(":not(picture)>img:not([width]):not([height]):not([srcset])")]
-    .forEach(img => {
-      img.height = img.naturalHeight;
-      img.width = img.naturalWidth;
-    });
+  [
+    ...doc.querySelectorAll(
+      ":not(picture)>img:not([width]):not([height]):not([srcset])"
+    ),
+  ].forEach(img => {
+    img.height = img.naturalHeight;
+    img.width = img.naturalWidth;
+  });
 }
--- a/src/core/figures.js
+++ b/src/core/figures.js
@@ -10,6 +10,7 @@ import { pub } from "core/pubsubhub";
 export const name = "core/figures";
 
 export function run(conf, doc, cb) {
+  normalizeImages(doc);
   // Move old syntax to new syntax
   $(".figure", doc).each(function(i, figure) {
     var $figure = $(figure),
@@ -114,4 +115,12 @@ export function run(conf, doc, cb) {
     while (tof.length) $ul.append(tof.shift());
   }
   cb();
+}
+
+function normalizeImages(doc) {
+  [...doc.querySelectorAll(":not(picture)>img:not([width]):not([height]):not([srcset])")]
+    .forEach(img => {
+      img.height = img.naturalHeight;
+      img.width = img.naturalWidth;
+    });
 }
--- a/tests/spec/core/figures-spec.js
+++ b/tests/spec/core/figures-spec.js
@@ -81,4 +81,73 @@ describe("Core - Figures", function() {
     expect(figLinks.item(0).textContent).toEqual("Figure 1 PREFIG");
     expect(figLinks.item(1).textContent).toEqual("Figure 2 IMGTIT");
   });
+
+
+  describe("normalize images", () => {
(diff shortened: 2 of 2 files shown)


tests/spec/core/images-spec.js
```diff
@@ -0,0 +1,32 @@
...
+
+describe("Core - Images", () => {
+  it("adds natural width and height to images if not set", async () => {
+    const ops = {
+      config: makeBasicConfig(),
+      body:
+      `<section>
+         <img id="image-with-id" src="data/images/image1.jpg">
+         <img src="data/images/image2.jpg">
+         <img id="image-with-dimensions" height=100 width=200 src="data/images/image1.jpg">
```
himanish-star: For testing, I have used two images stored in `tests/data/images`. However, I don't think this is what you would prefer. So I'll wait for your review on this

himanish-star:
Well the tests pass locally but I guess there is something that I am missing due to which Travis check is failing :thinking:

src/core/images.js
```diff
@@ -0,0 +1,21 @@
...
+// This checks for images missing width and height attributes
+// and adds them if they are missing
+
+import { pub } from "core/pubsubhub";
+
+export const name = "core/images";
+
+export function run(conf, doc, cb) {
+  [...doc.querySelectorAll("img")]
+    .filter(img => !img.hasAttribute("width") || !img.hasAttribute("height"))
```
marcoscaceres: WE should probably just do "&&" here. If someone has set one, but not the other, they probably have a good reason. 

We should also filter for `srcset` being set - and if `img` is inside a `<picture>` element, we should probably filter it out too.

src/core/images.js
```diff
@@ -0,0 +1,21 @@
...
+    .forEach((img, i) => {
+      img.height = img.naturalHeight;
+      img.width = img.naturalWidth;
+      if (!img.id) {
+        img.id = `${i+1}-image-with-no-width-height`;
```
marcoscaceres: There is a `makeID()` function you can use in the `utils` module, I think... or it might be part of "jQuery enhanced". In any case, please use that instead.

src/core/images.js
```diff
@@ -0,0 +1,21 @@
...
+      }
+      pub("warn", `[img](#${img.id}) is missing natural width and height`);
```
marcoscaceres: Let's not bother warning. Let's just set it and forget it :)

tests/spec/core/images-spec.js
marcoscaceres: other option is to use a data URL generated from a small image (e.g., 3x5 pixels)... just enough to give you unique values for width and height.

src/core/images.js
himanish-star: cool

marcoscaceres:
(Still on vacation, so just quick comments) 

Should probably move this into core/figure, as it’s expected that images are always used as figures (so, the check should be done on the collection of figures). Also needs to make sure that the image is not a child of a picture element or that it doesn’t have a srcset attribute.

himanish-star:
Sure, on it!

himanish-star:
> Also needs to make sure that the image is not a child of a picture element or that it doesn’t have a srcset attribute.

Done!!

> Should probably move this into core/figure

Doing this now! :smile:

himanish-star:
Done :tada:

src/core/figures.js
```diff
@@ -10,6 +10,14 @@ import { pub } from "core/pubsubhub";
 export const name = "core/figures";
 
 export function run(conf, doc, cb) {
+  [...doc.querySelectorAll("img")]
```
marcoscaceres: move this to it's own  "function normalizeImages(){}", so it's easier to know what it's doing.

src/core/figures.js
```diff
@@ -10,6 +10,14 @@ import { pub } from "core/pubsubhub";
...
+    .filter(img => !img.hasAttribute("width") && !img.hasAttribute("srcset")
+      && img.parentNode.tagName!=="PICTURE" && !img.hasAttribute("height"))
```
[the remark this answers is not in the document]
marcoscaceres: You might even be able to do the attribute filtering in the CSS selector too... maybe/maybe not, but not sure. Worth exploring.

tests/spec/core/figures-spec.js
```diff
@@ -81,4 +81,28 @@ describe("Core - Figures", function() {
...
+         <img id="image-with-id" src="${imgDataURL}">
+         <img id="image-with-dimensions" height=100 width=200 src="${imgDataURL}">
+      </section>`
+    };
+
+    const doc = await makeRSDoc(ops);
+
+    const image1 = doc.getElementById("image-with-id");
+    expect(image1).toBeTruthy();
+    expect(image1.height).toBe(3);
```
marcoscaceres: this test is not quite correct... we want to check that `hasAttribute("height")` and `hasAttribute("height")` and check that the attribute value is as expected.

tests/spec/core/figures-spec.js
```diff
@@ -81,4 +81,28 @@ describe("Core - Figures", function() {
...
   });
+
+  it("adds natural width and height to images if not set", async () => {
+    const imgDataURL = "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAUAAAADCAIAAADUVFKvAAAACXBIWXMAAAsTAAALEwEAmpwYAAAAB3RJTUUH4gQKACEWdS72PwAAABl0RVh0Q29tbWVudABDcmVhdGVkIHdpdGggR0lNUFeBDhcAAAAMSURBVAjXY2AgDQAAADAAAceqhY4AAAAASUVORK5CYII=";
+    const ops = {
+      config: makeBasicConfig(),
+      body:
+        `<section>
+         <img id="image-with-id" src="${imgDataURL}">
+         <img id="image-with-dimensions" height=100 width=200 src="${imgDataURL}">
```
marcoscaceres: We need two more here:

```HTML 
<img height=100>
<img width=100> 
```

And check that they are not touched. 

We also need tests for `<picture>` and `srcset`.

himanish-star:
Sure! Doing them now

src/core/figures.js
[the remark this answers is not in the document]
himanish-star: Sure

himanish-star:
Done with all changes. @marcoscaceres I hope you like them :smiley:

tests/spec/core/figures-spec.js
```diff
@@ -81,4 +81,63 @@ describe("Core - Figures", function() {
     expect(figLinks.item(0).textContent).toEqual("Figure 1 PREFIG");
     expect(figLinks.item(1).textContent).toEqual("Figure 2 IMGTIT");
   });
+
+  it("adds natural width and height to images if not set", async () => {
```
marcoscaceres: let's change this do `describe("normalize images", async() => {})`. You need to check if `describe` takes an async function or not. I can't recall. If not, then create the `doc` in `beforeAll()`. 

Then inside the `describe()`,  wrap each test with an `it("checks something", async ()=>{})`

tests/spec/core/figures-spec.js
```diff
@@ -81,4 +81,72 @@ describe("Core - Figures", function() {
...
+      expect(image.hasAttribute("height")).toBeTruthy();
+      expect(image.hasAttribute("width")).toBeFalsy();
+      expect(image.height).toBe(100);
+    });
+    it("doesn't change height and width for images with width only", async () => {
+      const image = doc.getElementById("image-with-width-only");
+      expect(image).toBeTruthy();
+      expect(image.hasAttribute("height")).toBeFalsy();
+      expect(image.hasAttribute("width")).toBeTruthy();
+      expect(image.width).toBe(200);
```
marcoscaceres: Here we want to also change the height didn't change.

PR #1617, opened 2018-04-06
feat(core/linter-rules/punctuation): Log offending elements : Fixes #1605

### LINKED ISSUE #1605: punctuation checker not working correctly
Testing with: 
https://w3c.github.io/payment-request/

## Description of problem

- [ ] The punctuation checker claims to have found a bad `p`, but doesn't actually help the Editor find it. 
- [ ] The description is still getting a bit messed up, seems to be generating two `<p>`.   

Expected:
For the linter to show me where the error is. 
For the description to render correctly. 

Optional, steps to reproduce: 

Add the following to the config of the spec:

```
      lint: {
        "no-headingless-sections": true,
        "privsec-section": true,
        "no-http-props": true,
        "check-punctuation": true,
      },
```

shikhar-scs:
As mentioned by @marcoscaceres at #1605 the changes have been made

### DIFF
--- a/src/core/linter-rules/check-punctuation.js
+++ b/src/core/linter-rules/check-punctuation.js
@@ -31,6 +31,9 @@ function lintingFunction(conf, doc) {
   const offendingElements = Array.from(doc.querySelectorAll("p:not(#back-to-top)"))
     .filter((elem) => !(punctuatingRegExp.test(elem.textContent)));
   
+  //To help the Editor reach he offending element at once
+  console.log(offendingElements);
+  
   const result = {
     name,
     offendingElements,


src/core/linter-rules/check-punctuation.js
```diff
@@ -31,6 +31,9 @@ function lintingFunction(conf, doc) {
   const offendingElements = Array.from(doc.querySelectorAll("p:not(#back-to-top)"))
     .filter((elem) => !(punctuatingRegExp.test(elem.textContent)));
   
+  //To help the Editor reach he offending element at once
+  console.log(offendingElements);
```
marcoscaceres: So, the linter already does this for us here:
https://github.com/w3c/respec/blob/develop/src/core/linter.js#L60

The actual bug is to figure out why the Linter is not doing this right now. I suspect there is something wrong with what this linter rule is returning.

shikhar-scs: I guess its working. Is this not what you are talking about?

screenshot from 2018-04-06 09-27-00

marcoscaceres: Ah, ok, thanks for confirming. Wonder why that didn't come up when I was trying it out.

marcoscaceres: I guess what is then left is to figure out why the error dialog is not rendering the markdown properly.

shikhar-scs: > I guess what is then left is to figure out why the error/warning dialog is not rendering the markdown properly.

yep! I'll look into that.

> . Wonder why that didn't come up when I was trying it out.

Maybe because no punctuation error existed. The ones present here are the ones I have put from the test files.

marcoscaceres: Ah, if it's showing an error when there are no invalid `p`s, then that's a bug. We should have added a test for that.

marcoscaceres: Oh, those `</p>` are what are screwing everything up. My editor must have added them by accident.

marcoscaceres: oh wait, I got confused. We want everything to show up like the other linter warnings (on one line).

marcoscaceres: > As we already have an open <p> tag in this statement, the next time we add a <p> tag it first closes the pre-existing <p> tag. And thus, it justifies </p><p>. I guess his might be one reason.

We shouldn't have any stray `<p>`, so the bug is that we didn't escape them :)

src/core/linter-rules/check-punctuation.js
```diff
@@ -10,7 +10,7 @@ const punctuationMarks = [".", ":", "!", "?"];
 const humanMarks = punctuationMarks.map(mark => `"${mark}"`).join(", ");
 const meta = {
   en: {
-    description: "`<p>` tags should end with a punctuation mark.",
+    description: "\`<p>\` tags should end with a punctuation mark.",
```
marcoscaceres: this should fix it.

src/core/linter-rules/check-punctuation.js
shikhar-scs: How do we escape ? 
```
en: {
    description: "`\<p\>` tags should end with a punctuation mark.",
    howToFix: `Please make sure \`\<p\>\` tags end with one of: ${humanMarks}.`,
  }
```

This isn't working :(

src/core/linter-rules/check-punctuation.js
shikhar-scs: Nope :cry:

marcoscaceres: How odd... what's the markup that's being generated now?

shikhar-scs:
```
const meta = {
    en: {
      description: "\`<p>\` tags should end with a punctuation mark.",
      howToFix: `Please make sure \`<p>\` tags end with one of: ${humanMarks}.`,
    },
};
```
This is still rendering :- 
screenshot from 2018-04-10 07-24-26

marcoscaceres:
We will need to dig into the markdown parse to see why it’s doing that.

shikhar-scs:
Yep Maybe

shikhar-scs:
Woah !! Thanks a lot @marcoscaceres :tada:

marcoscaceres:
np, thanks doing the base work on it. Just needed a couple of little things. Will deploy soon.

PR #1619, opened 2018-04-06
fix(core/caniuse): export as "caniuse.com"

### LINKED ISSUE #1618: caniuse should export as "caniuse.com"
When exporting caniuse to HTML, it should only export as "canisue.com" with the link to the corresponding feature.

sidvishnoi:
Fixes #1618
- Remove table from export
- Show only link on export
- Remove caniuse from exported config
- Remove stylesheet on export
- Minor update: Remove unused `id` from `.caniuse-title`

Regarding tests: 
1. How do I test the exported html? 
2. Do we need tests for this?


marcoscaceres:
For 1, import:
https://github.com/w3c/respec/blob/develop/src/ui/save-html.js#L166

For 2, always helps :)

src/core/caniuse.js
```diff
@@ -68,20 +70,25 @@ export async function run(conf) {
...
+    <dt class="caniuse-title">Can I Use this API?</dt>
     <dd class="caniuse-stats">${{
       any: contentPromise,
       placeholder: "Fetching data from caniuse.com...",
     }}</dd>`;
   headDlElem.appendChild(definitionPair);
   await contentPromise;
+
+  // remove from export
+  pub("amend-user-config", { caniuse: undefined });
```
marcoscaceres: Maybe set it to { caniuse: feature } ? - no need to totally trash it. Or is causing problems?

sidvishnoi: not really causing problems but I didn't find it useful there. We need it?

{caniuse: feature} would be good. Keeping whole configuration there seems, so, unnecessary.

sidvishnoi:
Hi
I'm having trouble in importing `ui/save-html` in tests. It says

```
ERROR: 'There is no timestamp for /base/ui/save-html.js!'
WARN [web-server]: 404: /base/ui/save-html.js
```

and I can't wrap my head around it!

marcoscaceres:
> I'm having trouble in importing ui/save-html in tests. 

Karma is really annoying :( I'll take a look.

src/core/caniuse.js
marcoscaceres: Let's go with `{caniuse: feature}`

PR #1623, opened 2018-04-16
chore(package): update dev deps

marcoscaceres:


saschanaz:
Failing tests... but nice to see actual messages rather than previous "maybe API limit?" one. 😍

marcoscaceres:
This should help us put an end to tge false positives https://github.com/w3c/spec-generator/pull/77

In this repo, we just need to test for generating specs over file:// URL.

marcoscaceres:
Err... false negatives!

saschanaz:
I'm not sure I understand correctly, are the false negatives just moved to spec-generator?

marcoscaceres:
> are the false negatives just moved to spec-generator?

No, because spec-generator would actually test the deployed version of ReSpec. The process is:
1. we run `./tools/release.js`, which does `npm publish`, 
1. ReSpec is deployed on W3C servers (it mirrors gh-pages branch). 
1. Greenkeeper tells Spec-Generator about new version, which triggers TravisCI.  
1. Spec-Generator then tests the correct deployed version (though it depends on how quickly ReSpec's release is deployed, but it generally takes around 1 minute). 

Right now, we are testing the "previously deployed" version of ReSpec (which is somewhat useful, but not ideal).

saschanaz:
IMO It's weird that (some) ReSpec tests is now on a different repo, but probably okay for now.

marcoscaceres:
> IMO It's weird that (some) ReSpec tests are now on a different repo, but probably okay for now.

Agree, I need to work on #1416 :(

Issue #1614, opened 2018-04-05
caniuse layout and click issues on mobile 


marcoscaceres:
## Important info

* URL to affected spec: https://w3c.github.io/payment-request/
* ReSpec version: Click the ReSpec pill (e.g., 20.7.0).
* [x] I did a "hard refresh", but it's still busted. 

## Description of problem

what actually happened (the bug):

The boxes are shown ontop of the copyright line. 

The boxes are not responding to touch/clicks. 

Expected:
For boxes to be above copyright line and boxes to be clickable. 

Optional, steps to reproduce: 

Load on a phone.

marcoscaceres:
Will provide a screenshot when I get to work.

saschanaz:
Not a phone, any touchable devices are enough to repro this. Tapping should lock the popup state.

marcoscaceres:
hover on small screen currently looks like this:

![screenshot 2018-04-05 12 51 59](https://user-images.githubusercontent.com/870154/38345101-9048a0ca-38d0-11e8-8bf1-79bf595db487.png)

Screenshot from iPhone: 
![screenshot 2018-04-05 12 58 22](https://user-images.githubusercontent.com/870154/38345215-3751400c-38d1-11e8-883a-597e05c8c425.png)

marcoscaceres:
No biggy tho. It should be fairly straight forward to fix. Part of "beta" testing ☺️

marcoscaceres:
I've removed this from Payment Request spec - but will re-enable it once we get this fixed.

sidvishnoi:
This would become a lot easier if we could use nested `ul` for the content that is hidden (submenu like, basically wrapping hidden items in another container). Any chance that we may do that instead?

marcoscaceres:
Nested ul should be ok. I thought that’s what we had already. Can you provide a quick example of what you are thinking here?

sidvishnoi:
Was thinking of something similar. 
`<button>` is something I didn't think of.
Do we also want hover support or just want to toggle on click/touch?

PR #1627, opened 2018-04-22
Fix: caniuse layout and click issues on mobile

sidvishnoi:
Fix: #1614 
- use nested ul for dropdown
- add support for touch by button:focus
- update tests for same


marcoscaceres:
Much better, but I'm still seeing a few small issues. There is a miss-alignment with the rendering and the margin top margin doesn't seem to be working on the first element (see 66-67 are shown as a single box):

screenshot 2018-04-23 14 11 28

Same happens on desktop:

screenshot 2018-04-23 14 12 07

Also, when tapping on the phone, it should be possible to deselect a currently selected dropdown. Otherwise, you can't, for instance, select Edge, after you select Firefox, without having to click on Chrome, etc.

sidvishnoi:
> see 66-67 are shown as a single box

actually it's the green background of Safari that makes it look it 66-67 of Chrome are one box :laughing: . I'll add a white background to it's `<ul>` to make sure this doesn't happen.

> misalignment

will fix. are you sure about using `cm`? Why not `em`?

> deselect

i'll look into how can we do that :)

marcoscaceres:
> will fix. are you sure about using cm? Why not em?

`em`'s are relative to base font, while cm are... fixed to a size. I personally like "cm" because I know what I'm going to get... with pixels... it's always weird because pixels are weird.

marcoscaceres:
@sidvishnoi, I sent through another set of changes (should include a `<button>`). Could you please merge them in? 🙏

js/core/css/caniuse.css
```diff
@@ -24,13 +28,24 @@
...
+  border: none;
+  font-size: 90%;
+  height: 0.8cm;
+  margin-right: 1px;
   margin-top: 0;
+  min-width: 3cm;
+  overflow: visible;
+  position: relative;
+  line-height: 0.79cm;
+  display: flex;
```
marcoscaceres: If the parent has display flex, does this need to be display flex?

sidvishnoi: Yes, I think so. As we need to center content in the current element (`<button>`)

With an aim for pixel perfect CSS, have a look at this little video clip: https://streamable.com/lye93

marcoscaceres: Thanks for clarifying!

marcoscaceres:
Showing in Safari as: 

screenshot 2018-04-30 12 23 52

marcoscaceres:
Same in Chrome.

src/core/caniuse.js
```diff
@@ -186,14 +187,16 @@ function createTableHTML(conf, stats) {
       .sort(semverCompare)
       .slice(-numVersions)
       .reverse();
-
     const { support, title } = getSupport(latestVersion);
+    const cssClass= `caniuse-cell ${support}`;
     return hyperHTML`
-      <ul class="caniuse-browser">
-        <li class="${"caniuse-cell " + support}" title="${title}">
+      <div class="caniuse-browser">
```
marcoscaceres: Can you help me understand why we need this div?

sidvishnoi: In order to make hover work properly (and avoid layout shift), we added `position: absolute` to the `ul`. But in order to show `ul` in it's browser area, we wrap it in `.caniuse-browser` which has `position: relative`.
I am not sure of some other way of doing this. Do you want me to look into it?

sidvishnoi:
> Same in Chrome.

Can't reproduce this. Working fine here for me in Chrome and Firefox.
screenshot from 2018-04-30 13-46-16


Can you confirm you have your local branch in sync?

src/core/caniuse.js
marcoscaceres: It's fine. Thanks for clarifying.

js/core/css/caniuse.css
```diff
@@ -1,16 +1,14 @@
...
   flex-wrap: wrap;
+  justify-content: flex-start;
+  align-items: baseline;
   cursor: pointer;
-  text-align: center;
 }
 
 button.caniuse-cell {
-  margin-top: 1px;
+  margin: 1px 1px 0 0;
```
marcoscaceres: This was the main problem in iOS... buttons have margins in Safari for some reason.

sidvishnoi: > buttons have margins in Safari for some reason

could `-webkit-appearance` have helped?

marcoscaceres: Not sure, tbh. With *-appearance, you never know what you will get because it's platform specific (as far as I understand it).

PR #1628, opened 2018-04-23
fix(core/issue-notes): give issues an unique ID

### LINKED ISSUE #1626: don't create a new id when an issue is referenced more than once
## Important info

* URL to affected spec: [Resource Timing L2](https://github.com/w3c/resource-timing/commit/bcb09cc55ba958616dd9959a6b92581f29269841)
* ReSpec version: 20.7.1.
* [Y ] I did a "hard refresh", but it's still busted. 

## Description of problem

Expected: don't create a new id when an issue is referenced more than once

But what actually happened (the bug): there was a duplicate ID (issue-82) in the generated document

Optional, steps to reproduce: put `data-number="**"` twice in the same document

marcoscaceres:
fixes #1626

### DIFF
--- a/src/core/issues-notes.js
+++ b/src/core/issues-notes.js
@@ -115,11 +115,16 @@ function handleIssues($ins, ghIssues, conf) {
         }
         if (report.number !== undefined) {
           // Add entry to #issue-summary.
-          var id = "issue-" + report.number,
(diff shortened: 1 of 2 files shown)


marcoscaceres:
Fixed commit message.

marcoscaceres:
This needs more work. Will resend it.

PR #1633, opened 2018-04-24
feat(core/utils): add flatten() util

marcoscaceres:


src/core/utils.js
```diff
@@ -522,3 +522,23 @@ export async function fetchAndCache(request, maxAge = 86400000) {
...
+// --- COLLECTION/ITERABLE HELPERS ---------------
+/**
+ * Spreads one iterable into another.
+ *
+ * @param {Iterable} collector
+ * @param {any|Iterable} item
+ * @returns {Array}
+ */
+export function flatten(collector, item) {
+  const isObject = typeof item === "object";
```
saschanaz: Oh, this is more advanced than my expectation:

```js
function flatten(arrayOfArray) {
  return [].concat(...arrayOfArray);
}
```

Deep flattening is great but I think we are not actually using it?

marcoscaceres: So, I would still like it to be .reduce() compatible in signature, as that's where we would most naturally want to use it. 

I can try to track down a few other places where we flatten things. 

The implementation is very loosely based on the ES spec (as I'm, I skimmed the spec... lol'ed... rolled my own 🤣).

src/core/utils.js
```diff
@@ -522,3 +522,23 @@ export async function fetchAndCache(request, maxAge = 86400000) {
...
+  const isIterable =
+    Object(item)[Symbol.iterator] && typeof item.values === "function";
+  const items = isObject
+    ? isIterable
+      ? [...item.values()].reduce(flatten, [])
+      : Object.values(item)
+    : [item];
```
saschanaz: How about:

```js
const items =
  !isObject ? [item] :
  isIterable ? [....item.values()].reduce(flatten, []) :
  Object.values(item);
```

This allows ladder-like format.

marcoscaceres: Good suggestion. Will include.

PR #1638, opened 2018-04-26
Remove IDL legacy things

marcoscaceres:
Based on https://github.com/heycam/webidl/pull/550 removes LegacyArrayClass. 

Also noticed the following had been removed from WebIDL:

-  legacycaller
-  legacyiterable


saschanaz:
Can we ignore DOMRectList?

marcoscaceres:
Yeah, it's in a BikeShed-based spec, so should be fine. No other spec depends on it.

PR #1641, opened 2018-04-27
Refactor save html

sidvishnoi:
basically moved export code to a separate file and made a little changes.

Reasons:
 - Couldn't test exported html without including unneeded modules (core/ui etc)

Can test easily now as:

``` js
  fit("lets access exported html", async() => {
    const doc = await makeRSDoc(makeStandardOps());
    await doc.respecIsReady;

    const { exportDocument } = await new Promise(resolve => {
      require(["core/exporter"], resolve);
    });
   const exportedHtml = await exportDocument("html", "text/html", doc);
   // .. tests  
  });
``` 

Please let me know if i broke something in between and the parameters that refactored functions need. :)

### DIFF
--- a/src/core/exporter.js
+++ b/src/core/exporter.js
@@ -1,0 +1,101 @@
+// module: exporter
+// used in ui/save-html
+
+import { removeReSpec } from "core/utils";
+import { pub } from "core/pubsubhub";
+
+export function serialize(format = "html") {
+  const cloneDoc = document.cloneNode(true);
+  cleanup(cloneDoc);
+  let result = "";
+  switch (format) {
+    case "xml":
+      result = new XMLSerializer().serializeToString(cloneDoc);
+      break;
+    default: {
+      if (cloneDoc.doctype) {
+        result += new XMLSerializer().serializeToString(cloneDoc.doctype);
+      }
+      result += cloneDoc.documentElement.outerHTML;
+    }
+  }
+  return result;
+}
+
+export async function exportDocument(format, mimeType) {
+  await document.respecIsReady;
+  const dataURL = toDataURL(serialize(format), mimeType);
+  const encodedString = dataURL.replace(/^data:\w+\/\w+;charset=utf-8,/, "");
+  const decodedString = decodeURIComponent(encodedString);
+  return decodedString;
+}
+
+export function toDataURL(data, mimeType = "text/html") {
+  const encodedString = encodeURIComponent(data);
+  return `data:${mimeType};charset=utf-8,${encodedString}`;
+}
+
+// Create and download an EPUB 3 version of the content
+// Using (by default) the EPUB 3 conversion service set up at labs.w3.org/epub-generator
+// For more details on that service, see https://github.com/iherman/respec2epub
+export function makeEPubHref() {
+  const url = new URL(
+    "https://labs.w3.org/epub-generator/cgi-bin/epub-generator.py"
+  );
+  const params = new URLSearchParams({
+    type: "respec",
+    url: document.location.href,
+  });
+  return `${url}?${params}`;
+}
+
+function cleanup(cloneDoc) {
+  const { head, body, documentElement } = cloneDoc;
+  cleanupHyper(documentElement);
+  Array.from(
+    cloneDoc.querySelectorAll(".removeOnSave, #toc-nav")
+  ).forEach(elem => elem.remove());
+  body.classList.remove("toc-sidebar");
+  removeReSpec(documentElement);
+  const insertions = cloneDoc.createDocumentFragment();
+  // Move meta viewport, as it controls the rendering on mobile.
+  const metaViewport = cloneDoc.querySelector("meta[name='viewport']");
+  if (metaViewport && head.firstChild !== metaViewport) {
+    insertions.appendChild(metaViewport);
+  }
+  // Move charset to near top, as it needs to be in the first 512 bytes.
+  let metaCharset = cloneDoc.querySelector(
+    "meta[charset], meta[content*='charset=']"
+  );
+  if (!metaCharset) {
+    metaCharset = cloneDoc.createElement("meta");
+    metaCharset.setAttribute("charset", "utf-8");
+  }
+  insertions.appendChild(metaCharset);
+  // Add meta generator
+  const metaGenerator = cloneDoc.createElement("meta");
+  metaGenerator.name = "generator";
+  metaGenerator.content =
+    "ReSpec " + window.respecVersion || "Developer Channel";
+  insertions.appendChild(metaGenerator);
+  head.insertBefore(insertions, head.firstChild);
+  pub("beforesave", documentElement);
+}
+
+function cleanupHyper(node) {
+  // collect first, or walker will cease too early
+  const comments = [...walkTree(
+    node, NodeFilter.SHOW_COMMENT,
+    comment => comment.textContent.startsWith("_hyper")
+  )];
+  for (const comment of comments) {
+    comment.parentNode.removeChild(comment);
+  }
+}
+
+function* walkTree(...args) {
+  const walker = document.createTreeWalker(...args);
+  while (walker.nextNode()) {
+    yield walker.currentNode;
+  }
+}
--- a/src/core/exporter.js
+++ b/src/core/exporter.js
@@ -4,8 +4,8 @@
 import { removeReSpec } from "core/utils";
 import { pub } from "core/pubsubhub";
 
-export function serialize(format = "html") {
-  const cloneDoc = document.cloneNode(true);
+export function serialize(format = "html", doc = document) {
+  const cloneDoc = doc.cloneNode(true);
   cleanup(cloneDoc);
   let result = "";
   switch (format) {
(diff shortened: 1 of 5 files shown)


src/ui/save-html.js
```diff
@@ -27,155 +26,52 @@ const downloadLinks = [
...
 
-export function exportDocument(format, mimeType) {
-  // Bug in babel won't export async function for some reason?
-  return document.respecIsReady.then(() => {
-    const dataURL = toDataURL(serialize(format), mimeType);
-    const encodedString = dataURL.replace(/^data:\w+\/\w+;charset=utf-8,/, "");
-    const decodedString = decodeURIComponent(encodedString);
-    return decodedString;
-  });
+const supportsDownload = "download" in document.createElement("a");
```
marcoscaceres: I think we need to keep `export function exportDocument(...)`, even if it calls into the new exporter module, as I think “respec2html” relies on this one. We can console.warn this one is deprecated.

src/core/exporter.js
```diff
@@ -0,0 +1,101 @@
...
+  let metaCharset = cloneDoc.querySelector(
+    "meta[charset], meta[content*='charset=']"
+  );
+  if (!metaCharset) {
+    metaCharset = cloneDoc.createElement("meta");
+    metaCharset.setAttribute("charset", "utf-8");
+  }
+  insertions.appendChild(metaCharset);
+  // Add meta generator
+  const metaGenerator = cloneDoc.createElement("meta");
```
marcoscaceres: While we are here, should we hyperHTML this?

src/ui/save-html.js
marcoscaceres: Not sure why my comment ended up here, it should be up where `exportDocument()` is. In any case, make sure you also update `evaluateHTML()` in `tools/respecDocWriter.js`.

sidvishnoi: can we import it from `exporter.js` and export it here? or move it here altogether? like:
``` js 
import { exportDocument } from "./exporter";
export exportDocument;
```

Saying this because I think `exportDocument` logically belongs to `exporter.js`

I didn't realize there were modules outside `src` folder that rely on these modules. Will check it and fix :slightly_smiling_face:

src/core/exporter.js
```diff
@@ -55,26 +55,29 @@ function cleanup(cloneDoc) {
...
+
   // Add meta generator
-  const metaGenerator = cloneDoc.createElement("meta");
-  metaGenerator.name = "generator";
-  metaGenerator.content =
-    "ReSpec " + window.respecVersion || "Developer Channel";
+  const respecVersion = window.respecVersion || "Developer Channel";
+  const metaGenerator = hyperHTML`
+    <meta name="generator" content="${"ReSpec " + respecVersion}">`;
+
```
marcoscaceres: Make sure you confirm this works... we I don't know what the owner document is here, and if adopting this node won't throw.

sidvishnoi: I exported it using the ui-pill. It had the `meta generator` and others. I'll confirm again

sidvishnoi: Update: works (exported html source: https://i.imgur.com/bKcE9To.png)

sidvishnoi: The owner document would be `cloneDoc` as we are appending `meta` into `insertions` which gets inserted into `cloneDoc.head`. Right?

marcoscaceres:
Ok, just quickly confirm in dev tools that .ownerDocument is indeed correct after inserting it (I guess it would throw if it wasn’t, but just wanted to be sure).

sidvishnoi:
> Ok, just quickly confirm in dev tools that .ownerDocument is indeed correct after inserting it (I guess it would throw if it wasn’t, but just wanted to be sure).

`metaGenerator.ownerDocument`:
screenshot from 2018-04-30 16-37-29

`document`:
screenshot from 2018-04-30 16-38-06

Confirm by seeing `<meta generator>` in exported one.

Is this what you were asking?

src/core/exporter.js
```diff
@@ -0,0 +1,104 @@
...
+export function toDataURL(data, mimeType = "text/html") {
+  const encodedString = encodeURIComponent(data);
+  return `data:${mimeType};charset=utf-8,${encodedString}`;
+}
+
+// Create and download an EPUB 3 version of the content
+// Using (by default) the EPUB 3 conversion service set up at labs.w3.org/epub-generator
+// For more details on that service, see https://github.com/iherman/respec2epub
+export function makeEPubHref() {
```
marcoscaceres: I don't think this is very useful as an export, tbh. This is very closely tied to the Save UI and I don't expect us to use it in other places. Thus, this should probably go back  into ui/save-html

src/core/exporter.js
```diff
@@ -0,0 +1,104 @@
+// module: exporter
+// used in ui/save-html
```
marcoscaceres: We should probably remove this comment, as things used solely by ui/save-html should like in that module... more below.

src/core/exporter.js
```diff
@@ -0,0 +1,104 @@
+// module: exporter
```
marcoscaceres: Nit: expand this out a little bit, so it's a bit more clear what this module does (i.e., exports a ReSpec document as either HTML or XML).

src/core/exporter.js
```diff
@@ -0,0 +1,104 @@
...
+      if (cloneDoc.doctype) {
+        result += new XMLSerializer().serializeToString(cloneDoc.doctype);
+      }
+      result += cloneDoc.documentElement.outerHTML;
+    }
+  }
+  return result;
+}
+
+export async function exportDocument(format, mimeType, doc = document) {
```
marcoscaceres: We should take this opportunity to document this function.

src/ui/save-html.js
```diff
@@ -1,15 +1,20 @@
...
-import { xmlEscape, removeReSpec } from "core/utils";
-import { pub } from "core/pubsubhub";
 import { ui } from "core/ui";
 import { l10n, lang } from "core/l10n";
+import { pub } from "core/pubsubhub";
+import {
+  toDataURL,
+  makeEPubHref,
+  serialize,
+  exportDocument as newExportDoc,
```
marcoscaceres: We should make `exportDocument()` smarter, we shouldn't need to import `toDataURL`, or `serialize` here. Below, we should just be able to say:

```JS
 get href(){
  return newExportDoc("data-url");
 }
```

And get back the equivalent of `return toDataURL(serialize());`

src/core/exporter.js
marcoscaceres: The more I think about it, `exportDocument()` should be the only thing we export from this module.

sidvishnoi: yes. we'll make `exportDocument` smarter

src/ui/save-html.js
sidvishnoi: I thought of this:
``` js
const mimeTypes = new Map([
  ["text/html", "html"],
  ["application/xml", "xml"],
]);
export async function exportDocument(mimeType = "text/html", doc = document) {
  const format = mimeTypes.get(mimeType);
  const dataURL = toDataURL(serialize(format, doc), mimeType);
 // ...
}
```

and we get use it as:
``` js
    // newExportDoc === exportDocument
    get href() {
      return newExportDoc(this.type);
    },
```

marcoscaceres: Nice!

tests/spec/core/exporter-spec.js
```diff
@@ -0,0 +1,51 @@
+"use strict";
+
+fdescribe("Core - exporter", () => {
```
marcoscaceres: fdescribe :)

tests/spec/core/exporter-spec.js
```diff
@@ -0,0 +1,51 @@
...
+    const doc = await makeRSDoc(ops);
+    await doc.respecIsReady;
+    const parser = new DOMParser();
+    const docString = decodeURIComponent(rsDocToDataURL("text/html", doc))
+      .replace("data:text/html;charset=utf-8,", "");
+    const exportedDoc = parser.parseFromString(docString, "text/html");
+    return exportedDoc;
+  }
+
+  it("should remove .removeOnSave elements", async () => {
```
marcoscaceres: nit: s/should remove/removes/

tests/spec/core/exporter-spec.js
```diff
@@ -0,0 +1,51 @@
...
+    const ops = makeStandardOps();
+    ops.body = `<div class="removeOnSave" id="this-should-be-removed">this should be removed</div>`;
+    const doc = await getExportedDoc(ops);
+
+    expect(doc.getElementById("this-should-be-removed")).toBeFalsy();
+    expect(doc.querySelectorAll(".removeOnSave").length).toBe(0);
+  });
+
+  it("should cleanup hyperHTML comments", async () => {
```
marcoscaceres: nit: cleans up

src/ui/save-html.js
```diff
@@ -43,139 +80,16 @@ if (supportsDownload) {
...
-      ui.freshModal(l10n[lang].save_snapshot, div, button);
-    });
-  },
-};
-
 export function exportDocument(format, mimeType) {
-  // Bug in babel won't export async function for some reason?
-  return document.respecIsReady.then(() => {
-    const dataURL = toDataURL(serialize(format), mimeType);
-    const encodedString = dataURL.replace(/^data:\w+\/\w+;charset=utf-8,/, "");
```
marcoscaceres: Whoops, we missed this bit.

sidvishnoi: oops. my bad. sorry about this. hopefully not too much damage

marcoscaceres: Not too much :) we took out the ability to generate specs for a little while. We have add a small test for this. Basically, we need to check if the exported file starts with `<html>`.

marcoscaceres: (also, not your fault... we both worked on it together, and I also missed it in review - it happens and doing hot fixes at 1:30am is what I live for 🤓🤘)

PR #1645, opened 2018-05-01
Remove irrelevant config for caniuse feature

sidvishnoi:
I'm not sure about the way I wrote tests. Please review.

### DIFF
--- a/tests/spec/core/include-config-spec.js
+++ b/tests/spec/core/include-config-spec.js
@@ -47,4 +47,24 @@ describe("Core — Include config as JSON", function() {
       expect(text).toEqual(expected);
     }).then(done);
   });
+
+  it("should remove irrelevant config for caniuse feature",async () => {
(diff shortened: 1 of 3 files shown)


tests/spec/core/include-config-spec.js
```diff
@@ -47,4 +47,24 @@ describe("Core — Include config as JSON", function() {
       expect(text).toEqual(expected);
     }).then(done);
   });
+
+  it("should remove irrelevant config for caniuse feature",async () => {
```
marcoscaceres: As it's a self contained test with its own document, let's move this to the caniuse-spec instead.

marcoscaceres: Nit: don't forget to run prettier on it! :)

marcoscaceres: nit: `it("removes irrelevant..."`

PR #1647, opened 2018-05-01
Use GitHub issueBase if wgPublicList is not given

### LINKED ISSUE #1636: Comment filing in SOTD does not match W3M practice
It looks like W3M has started pointing to github instead of a mailing list as a means of submitting comments.

See SOTD at https://www.w3.org/TR/2018/REC-ttml-imsc1.0.1-20180424/

It does not look like this is supported by respec.

sidvishnoi:
Fixes #1636
- use GitHub `issueBase` if `wgPublicList` is not given


src/w3c/templates/cgbg-sotd.js
```diff
@@ -39,7 +39,9 @@ ${!conf.sotdAfterWGinfo ? [conf.additionalContent] : ""}
       href='${`https://lists.w3.org/Archives/Public/${conf.wgPublicList}/`}'>archives</a>)${conf.subjectPrefix ? html`
     with <code>${conf.subjectPrefix}</code> at the start of your
     email's subject` : ""}.</p>
-  ` : ""}
+  ` : html`
+   <p>If you wish to make comments regarding this document, please raise them as <a href="${conf.issueBase}">GitHub issues</a>.</p>
```
marcoscaceres: Note that this is conditional on `conf.issueBase` actually being there. Same below. A lot of specs won't have `conf.issueBase`, because they don't declare `respecConfig.github` (or manually set `conf.issueBase`).

sidvishnoi: Oh. I thought `respecConfig.github` was a required config variable. 
Will fix this.

sidvishnoi:
I think I'll create a wrapper function to handle all such cases (to keep code clean).

Please let me know if you think a case needs to be handled differently.

1. no `github`, no `wgPublicList` : null
2. `wgPublicList`, no github:  mailing list
3. `github`, no `wgPublicList`: `issueBase` or github repo (needs clarification. `issueBase` is generated based on `github`)
4. `wgPublicList`, `github`: Both?

marcoscaceres:
> I think I'll create a wrapper function to handle all such cases (to keep code clean).

Normalization of W3C related things happens here:
https://github.com/w3c/respec/blob/develop/src/w3c/defaults.js

PR #1652, opened 2018-05-04
Make RespecDocWriter work with older versions of ReSpec

tidoust:
Version 20.10 introduced the `core/exporter` module, which is now used in `RespecDocWriter`. Problem is the spec on which RespecDocWriter runs may still reference an older version of ReSpec that does not have that module.

An example of a spec that has this problem today is [Media Capture and Streams](https://w3c.github.io/mediacapture-main/) which uses version 20.1.0. Running `RespecDocWriter` on it makes it crash badly (badly because the `Promise` that wraps the code to `require` does not get properly rejected when the module cannot be found, so the execution stays locked there with the error being reported to the console by RequireJS).

The writer checks the version of ReSpec and refuses to run on very old versions (lower than version 18). It still accepts to run on versions between 18 and 20.10 (which is a good thing :)). The old `ui/save-html` module needs to be used in such cases, which is what this update does.

Notes:
1. I copied the logic from `getVersion` to `evaluateHTML`. Not sure if there's a way to call that function directly from within `evaluateHTML`. I suppose not.
2. FWIW, I noticed the problem because Reffy uses `RespecDocWriter` to generate the HTML of all Respec specs it knows.

### DIFF
--- a/tools/respecDocWriter.js
+++ b/tools/respecDocWriter.js
@@ -189,13 +189,13 @@ async function evaluateHTML() {
       // Document references an older version of ReSpec that does not yet
       // have the "core/exporter" module. Try with the old "ui/save-html"
       // module.
-      const { exportDocument } = await new Promise((resolve, reject) => {
-        require(["ui/save-html"], resolve, reject);
+      const { exportDocument } = await new Promise(resolve => {
+        require(["ui/save-html"], resolve);
       });
       return exportDocument("html", "text/html");
     } else {
-      const { rsDocToDataURL } = await new Promise((resolve, reject) => {
-        require(["core/exporter"], resolve, reject);
+      const { rsDocToDataURL } = await new Promise(resolve => {
+        require(["core/exporter"], resolve);
       });
       const dataURL = rsDocToDataURL("text/html");
       const encodedString = dataURL.replace(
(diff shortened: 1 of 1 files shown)


marcoscaceres:
ReSpec at 20.10.3

tools/respecDocWriter.js
```diff
@@ -189,13 +189,13 @@ async function evaluateHTML() {
       // Document references an older version of ReSpec that does not yet
       // have the "core/exporter" module. Try with the old "ui/save-html"
       // module.
-      const { exportDocument } = await new Promise((resolve, reject) => {
-        require(["ui/save-html"], resolve, reject);
```
tidoust: There is some guarantee now that the modules do exist so this won't fail in practice but, unless the `require` here is not the one from RequireJS, I note it [does accept an error callback](http://requirejs.org/docs/api.html#errbacks).

Not rejecting the promise was precisely what made RespecDocWriter crash badly initially: the promise stayed in pending mode forever, and `evaluateHTML` simply never returned, so there was no way to tell programmatically that the generation had failed (except with a timeout). That behavior is easy to reproduce, just require a module here that does not exist!

marcoscaceres:
Ah, the RequireJS docs are inconsistent. Will add the reject back in, but want to check the error object is actually an Error. 

Thanks for clarifying!

PR #1654, opened 2018-05-07
fix(tools/respecDocWriter): reject on require load errors

marcoscaceres:


tools/respecDocWriter.js
```diff
@@ -186,16 +186,24 @@ async function evaluateHTML() {
...
+          `Your version: ${window.respecVersion}.`
+      );
       // Document references an older version of ReSpec that does not yet
       // have the "core/exporter" module. Try with the old "ui/save-html"
       // module.
-      const { exportDocument } = await new Promise(resolve => {
-        require(["ui/save-html"], resolve);
+      const { exportDocument } = await new Promise((resolve, reject) => {
+        require(["ui/save-html"], resolve, err => {
+          reject(new Error(err.message));
```
sidvishnoi: maybe we can reject directly without creating a new `Error` object?

sidvishnoi:
I ran respec2html. It rejected as required.

On a sidenote, (not sure if it matters much) the error message when `respecDocWriter` rejects says:
```
Error: Error: Evaluation failed: undefined
```

I think we can get a more meaningful message (instead of `undefined`) if we throw the entire error object instead of `err.stack` (in `evaluateHTML()`), that is:
```
Error: Error: Evaluation failed: Error: Script error for "core/exporte"
http://requirejs.org/docs/errors.html#scripterror
```

tools/respecDocWriter.js
marcoscaceres: I got w better stack in the browser by wrapping it. The error passed by requirejs didn’t seem very helpful. We want these lines to show up in the stack.

marcoscaceres:
Has a play. The problem is that "undefined" is actually a different bug. So, I'm inclined to go with this for now. I also don't want us getting into the habit of versioning things - we certainly don't want to support multiple version of ReSpec - specially for more than a few months. We want to give people enough time to upgrade and then move on.

PR #1657, opened 2018-05-10
Fix caniuse hover show/hide

sidvishnoi:
Fix the following: when button loses hover, it hides the browsers. 
Intended behaviour: hide active browser list when list is out of hover.


sidvishnoi:
> In hindsight, I think this is the right behavior.... the dropdown isn't a menu, and the user can't click on those things... so it's kinda correct that they should vanish.

But then a user won't be able to see the `title` text (Supported/not supported etc)

marcoscaceres:
> But then a user won't be able to see the title text (Supported/not supported etc)

Ah, good point. Got ya. ok :)

PR #1660, opened 2018-05-11
chore(package): update deps

marcoscaceres:


package.json
```diff
@@ -89,9 +89,10 @@
     "hyperhtml": "^2.10.0",
     "loading-indicator": "^2.0.0",
     "marked": "^0.3.19",
+    "npm": "^6.0.1",
```
saschanaz: Should we do this, or `"npm": ">=5"` in `engines` field instead?

saschanaz: Or just skip this line? I'm not sure installing npm as a dependency is right here...

marcoscaceres: agree.

marcoscaceres:
package-lock.json has messy merge conflicts... sending again.

PR #1664, opened 2018-05-11
(feat/lint) Warns about href's that link to nonexistent id's in a spec

### LINKED ISSUE #1643: Warn about href's that link to nonexistent id's
The linter should warn where there is a local reference (e.g., "#foo") but there is no element with id foo in the document.

sidvishnoi:
Fixes #1643

Need advice on:
1. rule name
2. linter message

I am presently using this as linter message:
```
    description: "Broken local reference found in document.",
    howToFix: "Please fix the href attributes of the links mentioned.",
    help: "See developer console.",
```
I do not find this very helpful but not able to think of a better one.

### DIFF
--- a/src/core/linter-rules/warn-broken-refs.js
+++ b/src/core/linter-rules/warn-broken-refs.js
@@ -1,0 +1,48 @@
+/**
+ * Linter rule "warn-local-ref".
+ * Warns about href's that link to nonexistent id's in a spec
+ */
+import {
+  lang as defaultLang
+} from "core/l10n";
+import LinterRule from "core/LinterRule";
+
+const name = "warn-broken-refs";
+
+const meta = {
+  en: {
+    description: "Broken local reference found in document.",
+    howToFix: "Please fix the href attributes of the links mentioned.",
+    help: "See developer console.",
+  },
+};
+
+// Fall back to english, if language is missing
+const lang = defaultLang in meta ? defaultLang : "en";
+
+const isBrokenReference = ({ href, ownerDocument: doc }) => {
+  const { hash } = new URL(href);
+  return !doc.getElementById(hash.substring(1));
+};
+
+/**
+ * Runs linter rule.
+ * @param {Object} config The ReSpec config.
+ * @param  {Document} doc The document to be checked.
+ */
+function linterFunction(conf, doc) {
+  const offendingElements = [...doc.querySelectorAll("a[href^='#']")].filter(
+    isBrokenReference
+  );
+  if (!offendingElements.length) {
+    return [];
+  }
+  return {
+    name,
+    offendingElements,
+    occurrences: offendingElements.length,
+    ...meta[lang],
+  };
+}
+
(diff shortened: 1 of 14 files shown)


src/w3c/defaults.js
```diff
@@ -54,6 +61,7 @@ const w3cDefaults = {
     "privsec-section": true,
     "no-http-props": true,
     "check-punctuation": false,
+    "warn-broken-refs": true,
```
marcoscaceres: Can we confirm with, say: 5-10 specs that this is fine.

src/core/linter-rules/warn-broken-refs.js
```diff
@@ -0,0 +1,46 @@
...
+ */
+import { lang as defaultLang } from "core/l10n";
+import LinterRule from "core/LinterRule";
+
+const name = "warn-broken-refs";
+
+const meta = {
+  en: {
+    description: "Broken local reference found in document.",
+    howToFix: "Please fix the href attributes of the links mentioned.",
```
marcoscaceres: "fix the href attributes" might not be possible, because it might be ReSpec creating the href.

marcoscaceres: Maybe, "fix the links mentioned" instead,

src/w3c/defaults.js
marcoscaceres: Let's make this `false`... check with URLs quickly... and then turn it to true.

tests/spec/core/linter-rules/warn-broken-refs-spec.js
```diff
@@ -0,0 +1,52 @@
+"use strict";
+describe("Core Linter Rule - 'warn-broken-refs'", () => {
+  const ruleName = "warn-broken-refs";
```
marcoscaceres: Nit: I think we should inline this. Doesn't seem to add much by having it as a variable.

src/w3c/defaults.js
```diff
@@ -7,8 +7,15 @@ import { rule as noHeadinglessSectionsRule } from "core/linter-rules/no-headingl
...
 import { rule as checkPunctuation } from "core/linter-rules/check-punctuation";
+import { rule as warnBrokenRefs } from "core/linter-rules/warn-broken-refs";
 
-linter.register(noHttpPropsRule, privsecSectionRule, noHeadinglessSectionsRule, checkPunctuation);
+linter.register(
+  noHttpPropsRule,
+  privsecSectionRule,
+  noHeadinglessSectionsRule,
+  checkPunctuation,
+  warnBrokenRefs
```
marcoscaceres: nit: missing "," ... maybe prettier didn't get run on this file?

marcoscaceres:
@sidvishnoi going to try this disabled first, and report back if there are any issues.

marcoscaceres:
@sidvishnoi we should start a new section in the wiki to document these things. Can you start with this one. I can file a bug to document the other ones.

PR #1671, opened 2018-05-14
fix(tools/respecDocWriter): support Puppeteer 1.4 error reporting

marcoscaceres:
Because of Puppeteer 1.4 (https://github.com/GoogleChrome/puppeteer/issues/1939), `respecDocWriter` is incorrectly exiting with: 

```
Failed to load file:////respec/js/core/css/var.css: Cross origin requests are only supported for protocol schemes: http, data, chrome, https.
```

For every css file.

### DIFF
--- a/tools/respecDocWriter.js
+++ b/tools/respecDocWriter.js
@@ -249,9 +249,20 @@ function makeConsoleMsgHandler(page) {
    */
   return function handleConsoleMessages(whenToHalt) {
     page.on("console", async message => {
+      const text = message.text();
       const type = message.type();
-      const args = await Promise.all(message.args().map(stringifyJSHandle));
-      const text = args.join(" ");
+      if (
+        type === "error" &&
+        !message.args().length && // browser errors have no arguments
+        text.startsWith("Failed to load")
+      ) {
+        // Since Puppeteer 1.4 reports _all_ errors, including CORS
+        // violations. Unfortunately, there is no way to distinguish these errors
+        // from other errors, so using this ugly hack.
+        // https://github.com/GoogleChrome/puppeteer/issues/1939
+        return;
+      }
+
       const abortOnWarning = whenToHalt.haltOnWarn && type === "warn";
       const abortOnError = whenToHalt.haltOnError && type === "error";
       const output = `ReSpec ${type}: ${colors.debug(text)}`;
@@ -274,7 +285,4 @@ function makeConsoleMsgHandler(page) {
   };
 }
 
-async function stringifyJSHandle(handle) {
-  return await handle.executionContext().evaluate(o => String(o), handle);
-}
 exports.fetchAndWrite = fetchAndWrite;
--- a/tools/respecDocWriter.js
+++ b/tools/respecDocWriter.js
@@ -248,12 +248,14 @@ function makeConsoleMsgHandler(page) {
    * @return {Void}
    */
   return function handleConsoleMessages(whenToHalt) {
-    page.on("console", message => {
-      const text = message.text();
+    page.on("console", async message => {
+      const args = await Promise.all(message.args().map(stringifyJSHandle));
+      const msgText = message.text();
+      const text = args.filter(msg => msg !== "undefined").join(" ");
       const type = message.type();
       if (
         type === "error" &&
-        text && // browser errors have text
+        msgText && // browser errors have text
         !message.args().length // browser errors have no arguments
       ) {
         // Since Puppeteer 1.4 reports _all_ errors, including CORS
@@ -282,6 +284,10 @@ function makeConsoleMsgHandler(page) {
       }
     });
   };
+}
+
+async function stringifyJSHandle(handle) {
+  return await handle.executionContext().evaluate(o => String(o), handle);
 }
 
 exports.fetchAndWrite = fetchAndWrite;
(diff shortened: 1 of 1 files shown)


marcoscaceres:
Updated issue description. Somewhat high priority because specs can't generate when updated to Puppeteer 1.4.

tools/respecDocWriter.js
```diff
@@ -249,9 +249,20 @@ function makeConsoleMsgHandler(page) {
    */
   return function handleConsoleMessages(whenToHalt) {
     page.on("console", async message => {
+      const text = message.text();
```
saschanaz: This was replaced by stringifyJSHandle in 6a4b0eabaf413cd8d0a532d692b66ab35359ce7c because message.text() just returns `JSHandle@error` when the argument is `Error`.

marcoscaceres: I don't know if they changed that too, but right now the old code returns nothing :( Give it a quick try:
```
node ./tools/respec2html.js -e --timeout 30 --src examples/basic.html --out /dev/null
```

tools/respecDocWriter.js
```diff
@@ -249,9 +249,20 @@ function makeConsoleMsgHandler(page) {
...
       const type = message.type();
-      const args = await Promise.all(message.args().map(stringifyJSHandle));
-      const text = args.join(" ");
+      if (
+        type === "error" &&
+        !message.args().length && // browser errors have no arguments
```
saschanaz: >browser errors have no arguments

IMO that means browser errors are actually distinguishable, no?

tools/respecDocWriter.js
marcoscaceres: I also tried `map()`ing the JSHandles too for real errors, and got basically the same as `.text()`. Would you mind having a quick look?

tools/respecDocWriter.js
marcoscaceres: In some sense, yes... but if someone calls `console.error()` they won't be. That's why I decided to check for the string.

tools/respecDocWriter.js
saschanaz: >but right now the old code returns nothing

Maybe because browser errors have no arguments, so `args.join(" ");` returns an empty string. We may ignore console messages which:

1. `message.args()` is empty
2. `message.text()` is not empty

What do you think?

marcoscaceres: Yeah, I think that works

saschanaz: I think we should brink it back, otherwise we'll have to reopen #1606.

marcoscaceres: no probs, will bring it back.

saschanaz:
>Failed to load file:////respec/js/core/css/var.css: Cross origin requests are only supported

Out of curiousity, does that error also happens in windowed Chrome?

saschanaz:
I'm starting to think that we should just resolve those errors rather than ignoring them... Currently we have some lines like `import css from deps/text!core/css.bp.css`, and this causes XHR-ing those CSS files. Instead we should be able to:

```js
hyperHTML`<link rel="stylesheet" href="${new URL("../core/css.bp.css", module.uri)}" />`
```

What do you think?

tools/respecDocWriter.js
marcoscaceres: Ok, so, unwrapping via `stringifyJSHandle` still get's me to, for example, this array:

```
[ 'Document on track but no previous version: Add `previousMaturity`, and `previousPublishDate` to ReSpec\'s config.',
  'undefined' ]
```

However, `.text()` still gives us back the correct `text()`. I think Puppeteer must have themselves fixed what as causing #1606 (i.e., we don't need to unwrap).

marcoscaceres: Oh, and it seems they changed "warn" to "warning" for type.

marcoscaceres:
I agree, we should get rid of text imports for CSS and add them properly. Let's do that as a new bug tho.

sidvishnoi:
@saschanaz That would require serving CSS files separately, right? Currently they are embedded inside the bundled file.

marcoscaceres:
yeah, we would need to serve them ... hmm... that might be complicated, specially as multiple standards orgs rely on ReSpec (thus we would need to deal with hosting those).

tools/respecDocWriter.js
```diff
@@ -248,18 +248,28 @@ function makeConsoleMsgHandler(page) {
    * @return {Void}
    */
   return function handleConsoleMessages(whenToHalt) {
-    page.on("console", async message => {
+    page.on("console", message => {
+      const text = message.text();
```
marcoscaceres: @saschanaz, just want to be clear. This line gives back the correct text for errors and warning, and not "JSHandle@error" (so no need to unwrap anymore, it seems). 

Here a sample of an error:  
screenshot 2018-05-15 00 41 47
 
Here is a sample of a warning: 
screenshot 2018-05-15 00 42 18

saschanaz: That's because the specific console message argument is string, but we have several `catch (err) { console.error(err); }` there.

saschanaz: Maybe we should remove those `console.error(err)` lines...

marcoscaceres: Ah, my bad - thanks for clarifying. Ok, I’d like to combine them and keep text and will use “args” as “details”.

marcoscaceres:
Ok, thinks we are back to were we were @saschanaz. Thanks again for your help!

PR #1682, opened 2018-05-25
feat(core/dfn): add automatic pluralization support for <dfn>

### LINKED ISSUE #1351: Investigate automatic pluralization of defined terms
Would be nice to not have to do `<dfn data-lt="things">thing</dfn>` in specs. 

https://github.com/blakeembrey/pluralize

sidvishnoi:
Fixes #1351 
See preview or read the test cases :)
screenshot from 2018-05-25 14-30-15
screenshot from 2018-05-25 14-28-57
screenshot from 2018-05-25 14-30-03


sidvishnoi:
Would it be better to do pluralisation on only those  `<dfn>` where `data-lt` is defined?

Adding automatic pluralisation on all has more chances of creating meaningless pluralisation (like on "html"). Currently we've options - enable all, disable all, enable all except where `data-lt-noPlural` is defined.

I think it might be better to enable by default but only add pluralisation on those that have `data-lt` attributes. The selective disable option remains valid, which keeps `data-lt` as it is, if any. 

This will reduce workload on adding manual pluralisation while also reducing unwanted ones.

What do you think?

marcoscaceres:
So, without yet having looked at the code, I think the right approach is as follows. The aim is to not blindly insert plurals, but to only insert plurals that are needed because the Editor actually used them in the document... thus: 

 1. find all `dfn:not([data-lt-noPlural])`, for each `dfn`:
 1. let `plural` be plural for dnf's textContent. If dfn-dt has `plural` return. 
 1. Find if the `plural` matches an `<a>` in the document. If no, return.   
 1. Check `plural` does not match any other dfn or any dfn's "data-lt". 
 1. Append `plural` to `dfn-lt`.

.gitignore
```diff
@@ -8,6 +8,7 @@ bevy.json
 .idea
 js/deps/*
 !js/deps/highlight.js
+!js/deps/pluralize.js
```
marcoscaceres: pluralize.js should be ignored, not added.

.gitattributes
```diff
@@ -0,0 +1 @@
+js/deps/pluralize.js binary
```
marcoscaceres: We probably don't need this, as we don't actually want to ship it... just copy it.

sidvishnoi:
> 1. find all dfn:not([data-lt-noPlural]), for each dfn

should we do it as a separate passs or in the `document.querySelectorAll("dfn").forEach(dfn => {` pass itself?

> 2. let plural be plural for dnf's textContent

only for `textContent`? and not for `data-lt` terms?

marcoscaceres:
> should we do it as a separate passs or in the document.querySelectorAll("dfn").forEach(dfn => { pass itself?

Not sure... whatever you think is best. 

> only for textContent? and not for data-lt terms?

Yeah, I think only for text content.

src/core/dfn.js
```diff
@@ -32,3 +46,27 @@ export function run(conf) {
...
+
+function autoPluralizeDfns() {
+  const links = [...document.querySelectorAll("a")];
+  const dfns = [...document.querySelectorAll("dfn")];
+
+  return function shouldPluralize(dfn) {
+    const plural = pluralize(norm(dfn.textContent));
+    if (
+      // has overridden to not do pluralization
+      dfn.hasAttribute("data-lt-noPlural") ||
```
marcoscaceres: this should move up to the query for dfns (i.e., ":not([data-lt-noPlural])")

src/core/dfn.js
```diff
@@ -32,3 +46,27 @@ export function run(conf) {
...
+      // no <a> references the plural term
+      !links.find(({ textContent }) => norm(textContent) === plural) ||
```
marcoscaceres: So, I think we want to reverse the logic here, because all "or"s needs to all be evaluated... what we want is "and" ... "and"... `return true`, so we do less work. Make sense?

src/core/dfn.js
```diff
@@ -32,3 +46,27 @@ export function run(conf) {
...
+      // there is any dfn with plural term in data-lt or textContent
+      dfns.find(
+        ({ textContent, dataset }) =>
+          norm(textContent) === plural ||
```
marcoscaceres: is there some way to reuse this check from line 60 here? `norm(textContent) === plural` is being done twice.

src/core/dfn.js
```diff
@@ -18,8 +23,17 @@ export function run(conf) {
     // Also, we should probably use weakmaps and weaksets here
     // to avoid leaks.
     const $dfn = $(dfn);
-    $dfn
-      .getDfnTitles({ isDefinition: true })
+    let dfnTitles = $dfn.getDfnTitles({ isDefinition: true });
```
marcoscaceres: nit const

src/core/dfn.js
sidvishnoi: not sure here. I'm doing it in one pass on all `dfn`.
If I add another pass, there might be those race conditions again. What do you think?

marcoscaceres: It shouldn't, as noPlural is never added dynamically.

src/core/dfn.js
```diff
@@ -32,3 +46,27 @@ export function run(conf) {
       }, $dfn);
   });
 }
+
+function autoPluralizeDfns() {
+  const links = [...document.querySelectorAll("a")];
+  const dfns = [...document.querySelectorAll("dfn")];
+
+  return function shouldPluralize(dfn) {
+    const plural = pluralize(norm(dfn.textContent));
```
marcoscaceres: Line 54: 
```
const normText = norm(dfn.textContent);
```

src/core/dfn.js
```diff
@@ -23,12 +29,17 @@ export function run(conf) {
...
     // to avoid leaks.
     const $dfn = $(dfn);
-    let dfnTitles = $dfn.getDfnTitles({ isDefinition: true });
+    const dfnTitles = $dfn.getDfnTitles({ isDefinition: true });
 
     // add automatic pluralization to `data-lt` attributes
     // see https://github.com/w3c/respec/pull/1682
-    if (conf.pluralize === true && shouldPluralize(dfn)) {
-      const plural = pluralize(norm(dfn.textContent));
+    const normText = norm(dfn.textContent);
```
sidvishnoi: these should be lower cased maybe?

src/core/dfn.js
```diff
@@ -32,3 +57,28 @@ export function run(conf) {
       }, $dfn);
   });
 }
+
+function autoPluralizeDfns() {
+  const links = new Set();
+  document
+    .querySelectorAll("a:not([href])")
+    .forEach(({ textContent, dataset }) => {
+      links.add(norm(textContent));
```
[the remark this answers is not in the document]
marcoscaceres: To be clear, if noDefault, then don't add text content.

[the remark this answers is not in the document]
marcoscaceres: Sorry, the above comment applies below  (to `dfn`), not here!

src/core/dfn.js
```diff
@@ -32,3 +57,28 @@ export function run(conf) {
       }, $dfn);
   });
 }
+
+function autoPluralizeDfns() {
+  const links = new Set();
+  document
+    .querySelectorAll("a:not([href])")
```
marcoscaceres: Make sure you test for this in the tests. I didn't see a test case that covered this.

src/core/dfn.js
```diff
@@ -18,8 +29,22 @@ export function run(conf) {
...
     const $dfn = $(dfn);
-    $dfn
-      .getDfnTitles({ isDefinition: true })
+    const dfnTitles = $dfn.getDfnTitles({ isDefinition: true });
+
+    // add automatic pluralization to `data-lt` attributes
+    // see https://github.com/w3c/respec/pull/1682
+    const normText = norm(dfn.textContent);
+    if (
+      conf.pluralize === true &&
```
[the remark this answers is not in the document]
sidvishnoi: not sure if I follow this one. `shouldPluralize` is a function. Do you want me to check existence of function definition?

src/core/dfn.js
```diff
@@ -18,8 +29,22 @@ export function run(conf) {
...
+      !dfn.hasAttribute("data-lt-noPlural") &&
+      shouldPluralize(normText)
+    ) {
+      const plural = pluralize(normText);
+      dfnTitles.push(plural);
```
sidvishnoi: we are calling `pluralize` twice (again in `shouldPluralize()`). What about `shouldPluralize()` returns the `plural` term if valid, otherwise null otherwise so we do it with one call to `pluralize()`?

src/core/dfn.js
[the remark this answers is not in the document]
marcoscaceres: it's truthy :) So, yes:

```
if( shouldPluralize && !dfn.hasAttribute("data-lt-noPlural") && shouldPluralize(normText))
```

src/core/dfn.js
marcoscaceres: Probably just return the empty string instead of null.

src/core/dfn.js
```diff
@@ -33,15 +33,17 @@ export function run(conf) {
...
     // add automatic pluralization to `data-lt` attributes
     // see https://github.com/w3c/respec/pull/1682
-    const normText = norm(dfn.textContent);
+    const normText = norm(dfn.textContent).toLowerCase();
     if (
-      conf.pluralize === true &&
+      pluralizeDfn &&
       !dfn.hasAttribute("data-lt-noPlural") &&
-      shouldPluralize(normText)
+      !dfn.hasAttribute("data-lt-noDefault")
```
sidvishnoi: This is required as:
``` html
<dfn data-lt="baz" data-lt-noDefault>bar</dfn> can be referenced
        as <a>baz</a>
        but not as <a>bar</a> or <a>bars</a>
```

without `!dfn.hasAttribute("data-lt-noDefault")`, `links.has(plural) && !dfns.has(plural) === true`
Took me a while to realise this, so reporting :)

src/core/dfn.js
```diff
@@ -18,8 +29,24 @@ export function run(conf) {
...
+    // add automatic pluralization to `data-lt` attributes
+    // see https://github.com/w3c/respec/pull/1682
+    const normText = norm(dfn.textContent).toLowerCase();
+    if (
+      pluralizeDfn &&
+      !dfn.hasAttribute("data-lt-noPlural") &&
+      !dfn.hasAttribute("data-lt-noDefault")
+    ) {
+      const plural = pluralizeDfn(normText);
+      if (plural !== "") {
```
marcoscaceres: nit: `if (plural) {`

src/core/dfn.js
```diff
@@ -18,8 +29,24 @@ export function run(conf) {
...
-    $dfn
-      .getDfnTitles({ isDefinition: true })
+    const dfnTitles = $dfn.getDfnTitles({ isDefinition: true });
+
+    // add automatic pluralization to `data-lt` attributes
+    // see https://github.com/w3c/respec/pull/1682
+    const normText = norm(dfn.textContent).toLowerCase();
+    if (
+      pluralizeDfn &&
+      !dfn.hasAttribute("data-lt-noPlural") &&
```
marcoscaceres: this should be "data-lt-no-plural" (`noDefault` was badly named, but we shouldn't continue the bad practice)

src/core/dfn.js
```diff
@@ -18,8 +29,24 @@ export function run(conf) {
...
+      !dfn.hasAttribute("data-lt-noDefault")
+    ) {
+      const plural = pluralizeDfn(normText);
+      if (plural !== "") {
+        dfnTitles.push(plural);
+        dfn.setAttribute("data-lt", dfnTitles.join("|"));
```
marcoscaceres: nit: `dfn.dataset.lt = dfnTitles.join("|")`

src/core/dfn.js
```diff
@@ -32,3 +59,27 @@ export function run(conf) {
...
+  document.querySelectorAll("a:not([href])").forEach(el => {
+    links.add(norm(el.textContent).toLowerCase());
+    if (el.dataset.lt) {
+      links.add(el.dataset.lt);
+    }
+  });
+
+  const dfns = new Set();
+  document.querySelectorAll("dfn:not([data-lt-noDefault])").forEach(dfn => {
+    dfns.add(norm(dfn.textContent).toLowerCase());
```
marcoscaceres: line is doing too much work:
```
const normText = norm(dfn.textContent).toLowerCase();
dfns.add(normText);
```

src/core/dfn.js
```diff
@@ -32,3 +59,27 @@ export function run(conf) {
...
+    if (dfn.dataset.lt) {
+      dfn.dataset.lt.split("|").forEach(lt => dfns.add(lt));
```
marcoscaceres: nit: dfn.dataset.lt.split("|").reduce((dfns, lt) => dfns.add(lt), dfns);

tests/spec/core/dfn-spec.js
```diff
@@ -46,6 +46,183 @@ describe("Core — Definitions", function() {
...
+          as <a>baz</a>
+          or <a>bar</a>
+          or <a>bars</a>
+          but not as <a id="link1">bazs</a>
+          or <a id="link2" href="PASS">bar</a>
+        </section>
+      `;
+      const ops = makeStandardOps({ pluralize: true }, body);
+      const doc = await makeRSDoc(ops);
+      await doc.respecIsReady;
```
marcoscaceres: don't need this, I think ... if we do, we have bigger problems :)

PR #1687, opened 2018-05-30
fix(w3c/defaults): allow lint rules override

sidvishnoi:
Update: Should we do a deep copy on entire object?
(Was focused on lint part right now)
Update: I think not needed.


src/w3c/defaults.js
```diff
@@ -95,6 +95,10 @@ export function run(conf) {
     ...w3cDefaults,
     ...conf,
   });
+  Object.assign(conf.lint, {
```
marcoscaceres: I guess it's nice that the original object reference doesn't get trashed.

sidvishnoi:
> Yeah, might be best to just do it properly now.

Use this https://github.com/KyleAMathews/deepmerge 
or create our own?

marcoscaceres:
using deep merge might be good... it's pretty tiny... but does mean another dependency

sidvishnoi:
Deep merging may not be needed. The `w3cDefaults` object doesn't have many nested objects (I could find only `lint`) that need to be deep copied and I see none of them need a copy with depth more than 1.

If there are more objects, we can use a simpler (depth 1) approach like this:
``` js
["lint", "someOtherKey"].forEach(key => {
  Object.assign(conf[key], { ...w3cDefaults[key], ...conf[key] })
});
```

Am I missing the big picture?

marcoscaceres:
I’m ok for this to go in as is.

marcoscaceres:
@sidvishnoi, please merge if you think this is ready.

PR #1689, opened 2018-05-31
perf(core/location-hash): avoid recalculating styles

marcoscaceres:
The costly check only needs to happen if the hash is truthy.


sidvishnoi:
I tried to run it. Weirdly, it's slower. Can you verify?

src/core/location-hash.js
```diff
@@ -14,11 +14,12 @@ export function run(conf, doc, cb) {
...
     hash = decodeURIComponent(window.location.hash).substr(1);
   } catch (err) {
-    hash = "";
+    console.warn(err);
   }
   // Only scroll to the hash if the document hasn't been scrolled yet
   // this ensures that a page refresh maintains the scroll position
-  if (!hash && !document.documentElement.scrollTop) {
+  const { documentElement, body } = document;
+  if (!hash || documentElement.scrollTop || body.scrollTop) {
```
marcoscaceres: Safari scrolls `body`, while other browsers scroll `documentElement`... the web 👍

src/w3c/style.js
```diff
@@ -10,15 +10,15 @@ import { pub, sub } from "core/pubsubhub";
...
   const script = doc.createElement("script");
-  script.addEventListener(
-    "load",
-    function() {
-      if (window.location.hash) {
+  if (location.hash) {
+    script.addEventListener(
+      "load",
+      () => {
         window.location = window.location;
```
saschanaz: What does this line do?

src/core/location-hash.js
sidvishnoi: Also: we were doing it wrong earlier (logical error). The `||` was required whereas we were doing `&&`

src/w3c/style.js
sidvishnoi: >What does this line do? 

forces page to jump to hash after page load.

Not sure if this `eventListener` is needed. If a URL has hash, browser itself should trigger the jump.

marcoscaceres: Unfortunately, this one is needed because fixup.js moves stuff around after ReSpec finishes (screwing up the scroll position). This should just set the hash tho.

src/core/location-hash.js
marcoscaceres: TIL `scrollingElement`

marcoscaceres:
Updated code now checks if the user starts scrolling while ReSpec is processing. If the user scrolls, we avoid jumping to location `.hash` (because user is in control now!). 

If the user doesn't scroll (lets ReSpec do its thing), then when ReSpec is done, the document naturally jumps to its `location.hash` in the post-processed document.

sidvishnoi:
> If the user scrolls, we avoid jumping to `location.hash` (because user is in control now!).

Can you verify this is working?
Not working here for me: https://streamable.com/0q8uz (See the scroll change)

src/core/location-hash.js
```diff
@@ -2,38 +2,35 @@
...
-  const isLegacyFrag = /\W/.test(hash);
-  if (!hasLink && isLegacyFrag) {
-    const id = hash
-      .replace(/[\W]+/gim, "-")
-      .replace(/^-+/, "")
-      .replace(/-+$/, "");
-    if (document.getElementById(id)) {
-      hash = id;
+  const scrollWatcher = new ScrollWatcher();
+  Promise.race([document.respecIsReady, scrollWatcher.promise]).then(() => {
```
sidvishnoi: lit :fire::1st_place_medal:

PR #1690, opened 2018-06-02
perf: dont create arrays unnecessarily

marcoscaceres:
as all browsers now treat NodeList as `@@iterable`, don't need to convert to array in some cases.


marcoscaceres:
Also includes some small prettier fixes.

PR #1694, opened 2018-06-18
refactor(core/link-to-dfn) split into internal and external linking modules

sidvishnoi:
Why? In upcoming xref feature, we only need to find references to those `<a>` which have no `dfn` locally. So splitting it into internal linking (runs first) and external linking (to run after core/xref and core/data-cite) modules.

TODO:
- [x] Split
- [ ] DRY
- [ ] Remove jQuery


sidvishnoi:
Regarding git history:
I've renamed `link-to-dfn` to `link-to-dfn-external` so the file history remains. 
If I were to edit `link-to-dfn-external` now and we squash-merge the edits & rename, will the git history of this file be fine? Maybe we should merge it in two parts to keep history, otherwise it's all fine.

js/profile-w3c-common.js
```diff
@@ -49,10 +49,11 @@ define(
     "core/best-practices",
     "core/figures",
     "core/webidl",
+    "core/link-to-dfn-internal",
```
marcoscaceres: Unfortunately, I don’t think we can’t rename this, as it would break other profiles.

sidvishnoi: Alternatives? We do need to separate out linking of internal and external definitions, as I believe.

marcoscaceres: doing the separation is fine, but this module should remain as is (and retain backwards compatibility). The new external xrefs can just live in “*-external.js”

sidvishnoi: > doing the separation is fine, but this module should remain as is

Didn't quite get that. "remain same" in what sense?

I actually want to separate it into two parts as:
- I do not want `link-to-dfn` to emit `"warn", Linking error: not matching <dfn>` while it has just looked for internal dfns

marcoscaceres: That’s ok, but that shouldn’t require the module to be renamed. The external xref module could be imported by this module, for instance. It doesn’t necessarily also need to be a separate plugin.

marcoscaceres: > Didn't quite get that. "remain same" in what sense?

To be clear, in the sense that the external reference code should be a progressive enhancement, and not a breaking change.

marcoscaceres: I’ll try to be online in about 2.5 hours so we can discuss... still 😴 🛏  here (yay jet lag!).

sidvishnoi:
Closing this in favour of new discussed strategy

PR #1698, opened 2018-06-18
refactor(core/jquery-enhanced): migrate $.allTextNodes to vanilla js

sidvishnoi:
As a part of https://github.com/w3c/respec/issues/1699

### DIFF
--- a/src/core/utils.js
+++ b/src/core/utils.js
@@ -582,3 +582,27 @@ export function addId(elem, pfx = "", txt = "", noLC = false) {
   elem.id = id;
   return id;
 }
+
+/**
+ * Returns all the descendant text nodes of an element.
+ * @param {Node} elem
+ * @param {Array:String} exclusions node types (localName) to exclude
+ * @returns {Array:String}
+ */
+export function allTextNodes(elem, exclusions) {
+  const excl = new Set(exclusions);
+  const textNodes = [];
+  getTextNodes(elem);
+  return textNodes;
+
+  function getTextNodes(node) {
+    if (node.nodeType === 1 && excl.has(node.localName.toLowerCase())) return;
+    if (node.nodeType === 3) {
+      textNodes.push(node);
+    } else {
+      for (const child of node.childNodes) {
+        getTextNodes(child);
+      }
+    }
+  }
+}
--- a/src/core/utils.js
+++ b/src/core/utils.js
@@ -586,18 +586,19 @@ export function addId(elem, pfx = "", txt = "", noLC = false) {
 /**
  * Returns all the descendant text nodes of an element.
  * @param {Node} elem
- * @param {Array:String} exclusions node types (localName) to exclude
+ * @param {Set:String} exclusions node localName to exclude
  * @returns {Array:String}
  */
 export function allTextNodes(elem, exclusions) {
-  const excl = new Set(exclusions);
   const textNodes = [];
   getTextNodes(elem);
   return textNodes;
 
   function getTextNodes(node) {
-    if (node.nodeType === 1 && excl.has(node.localName.toLowerCase())) return;
-    if (node.nodeType === 3) {
+    if (node.nodeType === Node.ELEMENT_NODE && exclusions.has(node.localName)) {
+      return;
+    }
+    if (node.nodeType === Node.TEXT_NODE) {
       textNodes.push(node);
     } else {
       for (const child of node.childNodes) {
--- a/src/core/utils.js
+++ b/src/core/utils.js
@@ -585,25 +585,26 @@ export function addId(elem, pfx = "", txt = "", noLC = false) {
 
 /**
  * Returns all the descendant text nodes of an element.
- * @param {Node} elem
- * @param {Set:String} exclusions node localName to exclude
+ * @param {Node} el
+ * @param {Array:String} exclusions node localName to exclude
  * @returns {Array:String}
  */
-export function allTextNodes(elem, exclusions) {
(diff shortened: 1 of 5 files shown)


src/core/utils.js
```diff
@@ -582,3 +582,27 @@ export function addId(elem, pfx = "", txt = "", noLC = false) {
...
 }
+
+/**
+ * Returns all the descendant text nodes of an element.
+ * @param {Node} elem
+ * @param {Array:String} exclusions node types (localName) to exclude
+ * @returns {Array:String}
+ */
+export function allTextNodes(elem, exclusions) {
+  const excl = new Set(exclusions);
```
marcoscaceres: should move this outside, me thinks... no point in creating it over and over again.

src/core/utils.js
```diff
@@ -582,3 +582,27 @@ export function addId(elem, pfx = "", txt = "", noLC = false) {
...
+  const textNodes = [];
+  getTextNodes(elem);
+  return textNodes;
+
+  function getTextNodes(node) {
```
marcoscaceres: we probably don't need this function. Can just just inline all this.

src/core/utils.js
sidvishnoi: make user call with set instead of array?

src/core/utils.js
```diff
@@ -582,3 +582,27 @@ export function addId(elem, pfx = "", txt = "", noLC = false) {
...
+    if (node.nodeType === 1 && excl.has(node.localName.toLowerCase())) return;
```
marcoscaceres: while we are here, we should replace the numbers with `Node.WHATEVER`, so we actually know what each number is.

marcoscaceres: isn't `node.localName` always in lower case?

src/core/utils.js
marcoscaceres: oh, sorry, I didn't see that it was recursive! Never mind.

sidvishnoi: can make it iterative though

src/core/utils.js
sidvishnoi: right!

src/core/utils.js
marcoscaceres: This all still seems silly tho... would this work? 

```
function getTextNodes(el, exclusions) {
  const textNodes = [];
  const walker = document.createTreeWalker(
    el,
    NodeFilter.SHOW_TEXT,
    null,
    false
  );
  let node;
  while ((node = walker.nextNode())) {
    const isExcluded =
      exclusions.includes(node.parentElement.localName) === false;
    textNodes.push(node);
  }
  return textNodes;
}
```

marcoscaceres: fixed typo above.

sidvishnoi: event better!
*@sidvishnoi always forgets about TreeWalker API*

marcoscaceres: How about: 

```JS
function getTextNodes(el, exclusions = []) {
  const acceptNode = node => {
    return exclusions.includes(node.parentElement.localName)
      ? NodeFilter.FILTER_REJECT
      : NodeFilter.FILTER_ACCEPT;
  };
  const nodeIterator = document.createNodeIterator(
    el,
    NodeFilter.SHOW_TEXT,
    { acceptNode },
    false
  );
  const textNodes = [];
  let node;
  while ((node = nodeIterator.nextNode())) {
    textNodes.push(node);
  }
  return textNodes;
}
```

sidvishnoi: _@sidvishnoi goes to read the docs_

marcoscaceres: Maybe the above... completely untested 🤪

sidvishnoi: it does work.
any reason to use NodeIterator instead of TreeWalker?
``` js
  // with TreeWalker
  while (walker.nextNode()) {
    textNodes.push(walker.currentNode);
  }
```

marcoscaceres: > NodeIterators and TreeWalkers are two different ways of representing the nodes of a document subtree and a position within the nodes they present. A NodeIterator presents a flattened view of the subtree as an ordered sequence of nodes, presented in document order. Because this view is presented without respect to hierarchy, iterators have methods to move forward and backward, but not to move up and down. Conversely, a TreeWalker maintains the hierarchical relationships of the subtree, allowing navigation of this hierarchy. In general, TreeWalkers are better for tasks in which the structure of the document around selected nodes will be manipulated, while NodeIterators are better for tasks that focus on the content of each selected node.

So, tl;dr: `NodeIterator` is flat, which is what we want.... don't worry, TIL me too!

marcoscaceres: > someone should add this example to MDN docs! 

Thanks for volunteering! 🤣

marcoscaceres:
Should probably land each of these one at a time. WDTY?

sidvishnoi:
> Tests

A test for this method exists:
https://github.com/w3c/respec/blob/develop/tests/spec/core/jquery-enhanced-spec.js#L98

core/inlines uses this method and it works.

> Should probably land each of these one at a time. WDTY?

Sure. I'll add the test to utils-specs. And migrate all usage of $.allTextNodes in ReSpec.

Will add deprecation notice to jquery-enhanced when all are done. WDYT?

marcoscaceres:
> Will add deprecation notice to jquery-enhanced when all are done. WDYT?

Sounds good.

sidvishnoi:
ReSpec v21.1.0 now?

PR #1703, opened 2018-06-20
fix(core/biblio): normalize references

sidvishnoi:
- normalize based on resolved `id` from spec ref
- fix terms not being found in database (do not lowercase)
- refactor

### DIFF
--- a/src/core/biblio.js
+++ b/src/core/biblio.js
@@ -140,22 +140,34 @@ function bibref(conf) {
       </section>`;
     addId(sec);
 
-    refs.sort((a, b) =>
-      a.toLocaleLowerCase().localeCompare(b.toLocaleLowerCase())
-    );
-    const refObjects = refs.map(getRefContent);
+    const refObjects = { good: [], bad: [] };
+    refs.map(getRefContent).reduce((refObjects, ref) => {
+      const refType = ref.refcontent ? "good" : "bad";
+      refObjects[refType].push(ref);
+      return refObjects;
+    }, refObjects);
 
-    const aliases = refObjects.reduce((aliases, { key, ref }) => {
+    const aliases = refObjects.good.reduce((aliases, ref) => {
+      const key = ref.refcontent.id;
       const keys = !aliases.has(key)
         ? aliases.set(key, []).get(key)
         : aliases.get(key);
-      if (!keys.includes(ref.toLowerCase())) keys.push(ref);
+      keys.push(ref.ref);
       return aliases;
     }, new Map());
-    const refsToAdd = refObjects.reduce((refsToAdd, ref) => {
-      if (!refsToAdd.find(r => r.key === ref.key)) refsToAdd.push(ref);
-      return refsToAdd;
-    }, []);
+
+    const uniqueRefs = refObjects.good.reduce((uniqueRefs, ref) => {
+      if (!uniqueRefs.has(ref.refcontent.id)) {
+        // the condition ensures that only the first used [[TERM]]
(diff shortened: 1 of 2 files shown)


src/core/biblio.js
```diff
@@ -140,22 +140,34 @@ function bibref(conf) {
       </section>`;
     addId(sec);
 
-    refs.sort((a, b) =>
-      a.toLocaleLowerCase().localeCompare(b.toLocaleLowerCase())
-    );
-    const refObjects = refs.map(getRefContent);
+    const refObjects = { good: [], bad: [] };
+    refs.map(getRefContent).reduce((refObjects, ref) => {
```
marcoscaceres: nit, this would be more consistent with the rest of the code as: 
```js
const { goodRefs, badRefs } = refs.map(getRefContent).reduce((refObjects, ref) => {
   // 
}, {goodRefs: [], badRefs: []});
```

src/core/biblio.js
```diff
@@ -140,22 +140,34 @@ function bibref(conf) {
...
-      if (!keys.includes(ref.toLowerCase())) keys.push(ref);
+      keys.push(ref.ref);
       return aliases;
     }, new Map());
-    const refsToAdd = refObjects.reduce((refsToAdd, ref) => {
-      if (!refsToAdd.find(r => r.key === ref.key)) refsToAdd.push(ref);
-      return refsToAdd;
-    }, []);
+
+    const uniqueRefs = refObjects.good.reduce((uniqueRefs, ref) => {
```
marcoscaceres: After this, `uniqueRefs` is only ever used as a an array (by spreading the values), so I wonder if you should just convert it to an array here.

src/core/biblio.js
```diff
@@ -140,22 +140,34 @@ function bibref(conf) {
...
+      if (!uniqueRefs.has(ref.refcontent.id)) {
+        // the condition ensures that only the first used [[TERM]]
+        // shows up in #references section
+        uniqueRefs.set(ref.refcontent.id, ref);
+      }
+      return uniqueRefs;
+    }, new Map());
+
+    const refsToAdd = [...uniqueRefs.values(), ...refObjects.bad].sort((a, b) =>
```
marcoscaceres: then here: `uniqueRefs.concat(badRefs).sort()`

tests/spec/core/biblio-spec.js
```diff
@@ -145,4 +146,17 @@ describe("W3C — Bibliographic References", () => {
     expect(first).toMatch("[a]");
     expect(last).toMatch("[Zzz]");
   });
+
```
marcoscaceres: I wonder if we need a test also that clears the biblio database?

sidvishnoi: not sure what and how would we test using that :confused:

marcoscaceres: so, you could just add the check to the "normalizes aliases" test above... but just clear the db, and create a new ReSpec doc.

src/core/biblio.js
```diff
@@ -140,22 +140,42 @@ function bibref(conf) {
       </section>`;
     addId(sec);
 
-    refs.sort((a, b) =>
-      a.toLocaleLowerCase().localeCompare(b.toLocaleLowerCase())
+    const refObjects = refs.map(getRefContent).reduce(
```
marcoscaceres: Nit: I still don't think we need "refObjects", we can destructure it away:

```
const { badRefs, goodRefs } = ... 
```

PR #1701, opened 2018-06-19
chore(package): update webidl2 to version 14.0.0

marcoscaceres:
Closes #1675

Issue #1559, opened 2018-03-06
RangeError: Invalid count value exception in idn()


Orphis:
When rendering the [webrtc-pc](https://w3c.github.io/webrtc-pc/) spec, we get this exception:
```
respec-w3c-common.js:1 RangeError: Invalid count value
    at String.repeat (<anonymous>)
    at M (respec-w3c-common.js:414)
    at Object.<anonymous> (respec-w3c-common.js:414)
    at Object.main (respec-w3c-common.js:1)
    at c (respec-w3c-common.js:1)
    at i (respec-w3c-common.js:1)
    at respec-w3c-common.js:414
    at respec-w3c-common.js:414
    at Array.map (<anonymous>)
    at Q (respec-w3c-common.js:414)
```

I tracked it down to `idn()` calls with a negative number. Not quite sure why, but it shouldn't throw any exception. Adding a `Math.max(0, lvl)` to `idn()` seems to fix the issue, but I didn't try to see why it would have a negative padding in the first place.

sidvishnoi:
I've tracked down the bug to 
`src/core/webidl.js#L677 ---  if (meth.static) len += 7;` and 
`src/core/webidl.js#L693 --- var pad = max - len;`.

What is this 7?
What should be done? Should we increase `max` by 7 also? or make `pad = Math.max(pad,0)`

saschanaz:
>What is this 7?

`"static ".length === 7`.

Currently the webidl module is slightly a mess... Anyway, the right fix is to do the same `+= 7` in line 594.

https://github.com/w3c/respec/blob/23001a604799c4c7c0d97b01b970128c8cfd6042/src/core/webidl.js#L588-L595

saschanaz:
@marcoscaceres, do we want to keep this formatter-like feature after we do https://github.com/w3c/webidl2.js/issues/125?

marcoscaceres:
No, I’d like to get rid of the formatter in ReSpec.

marcoscaceres:
@Orphis, should be good now, but you update ReSpec in the Web RTC repo.

marcoscaceres:
fixed typo above.

Orphis:
Thanks!

alvestrand:
@vivienlacourba can you take care of updating webrtc-pc and friends?

PR #1710, opened 2018-06-24
Update to webidl2.js 14

saschanaz:
...and replace `opt.ws` with the new trivia feature. It means that we lose the IDL formatting feature, and @marcoscaceres agreed to remove it in https://github.com/w3c/respec/issues/1559#issuecomment-371998009.

(This PR does not really support full whitespace conservation as the purpose here is to update without test breakage.)

Closes #1675 
Closes #1684 
Closes #1701

### DIFF
--- a/js/core/templates/webidl-contiguous/attribute.html
+++ b/js/core/templates/webidl-contiguous/attribute.html
@@ -1,3 +1,3 @@
 <span class='idlAttribute' id="{{obj.idlId}}" data-idl data-title='{{obj.name}}'>{{extAttr obj
-}}{{qualifiers}}{{obj.trivia.base}}attribute<span class='idlAttrType'>{{idlType
+}}{{{qualifiers}}}{{{trivia obj.trivia.base}}}attribute<span class='idlAttrType'>{{idlType
 obj}}</span> <span class='idlAttrName'>{{#tryLink obj}}{{obj.escapedName}}{{/tryLink}}</span>;</span>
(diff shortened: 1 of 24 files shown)


js/core/templates/webidl-contiguous/attribute.html
```diff
@@ -1,3 +1,3 @@
-<span class='idlAttribute' id="{{obj.idlId}}" data-idl data-title='{{obj.name}}'>{{extAttr obj indent
-}}{{idn indent}}{{qualifiers}}attribute <span class='idlAttrType'>{{idlType obj}}</span> {{pads
-pad}}<span class='idlAttrName'>{{#tryLink obj}}{{escapeAttributeName obj.name}}{{/tryLink}}</span>;</span>
+<span class='idlAttribute' id="{{obj.idlId}}" data-idl data-title='{{obj.name}}'>{{extAttr obj
+}}{{{qualifiers}}}{{{trivia obj.trivia.base}}}attribute<span class='idlAttrType'>{{idlType
+obj}}</span> <span class='idlAttrName'>{{#tryLink obj}}{{obj.escapedName}}{{/tryLink}}</span>;</span>
```
marcoscaceres: double checked that the new line at EOF is not needed, right? These can be super annoying.

marcoscaceres: Same with the ones below... they look ok, given that they are spans... but just triple check. It's gotten me before /shakes fist at handlebars.

src/core/webidl.js
```diff
@@ -141,36 +137,44 @@ function registerHelpers() {
     }
     return a.outerHTML;
   });
+  hb.registerHelper("trivia", writeTrivia);
```
marcoscaceres: I can't tell exactly here, but it looks like you are registering the "trivia" helper  inside "tryLink"? Is that correct? or should it be outside like the rest?

marcoscaceres:
🎉+ respec@21.3.0

PR #1712, opened 2018-06-25
 refactor(core/biblio): split into pre-processing and rendering parts

sidvishnoi:
Notes:
- splits into render-specific and pre-processing parts and make a single module do lesser work
- **Change**: always use latest biblio refs
- Refactor triggered in preparation of upcoming external references feature as the feature will dynamically modify `data-cite` attributes

### DIFF
--- a/src/core/biblio.js
+++ b/src/core/biblio.js
@@ -9,6 +9,9 @@ import { biblioDB } from "core/biblio-db";
 import { createResourceHint } from "core/utils";
 import { pub } from "core/pubsubhub";
 
+// for backward compatibity
+export { wireReference, stringifyReference } from "core/render-biblio";
+
 export const name = "core/biblio";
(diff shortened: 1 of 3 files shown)


src/core/biblio.js
```diff
@@ -1,14 +1,17 @@
...
 
 /*jshint jquery: true*/
 /*globals console*/
 import { biblioDB } from "core/biblio-db";
-import { createResourceHint, addId } from "core/utils";
+import { createResourceHint } from "core/utils";
 import { pub } from "core/pubsubhub";
 
+// for backward compatibity
+export { wireReference, stringifyReference } from "core/render-biblio";
```
sidvishnoi: Do we know where are they used externally? Do we really need to export them? Exporting won't hurt though.

PR #1719, opened 2018-06-27
feat(core/xref): Add support for automatically linking external refs

sidvishnoi:
Adds support to link unique externally defined terms.

Currently disabled by default.
To Enable:
``` js
respecConfig = {
  xref: {
    url: "URL-TO-JSON-DATA-SOURCE",
  }
}
```

Sample data:
See `/tests/data/xref.json`


src/core/xref.js
```diff
@@ -0,0 +1,134 @@
+// Automatically adds external refernces
```
marcoscaceres: typo refernces

marcoscaceres: We probably want a bit more detail of what this module actually does.

src/core/xref.js
```diff
@@ -0,0 +1,134 @@
...
+// https://github.com/w3c/respec/issues/1662
+
+import { norm } from "core/utils";
```
marcoscaceres: nit: `{ norm as normalize }` maybe?

src/core/xref.js
```diff
@@ -0,0 +1,134 @@
...
+import { pub } from "core/pubsubhub";
+
+/**
+ * main external reference driver
+ * @param {Object} conf respecConfig
+ * @param {Array:Elements} elems possibleExternalLinks
+ */
+export async function run(conf, elems) {
+  const xrefMap = createXrefMap(elems);
+
```
marcoscaceres: nit: probably don't need this new line

sidvishnoi: let me send the one with API end-point first.

src/core/xref.js
```diff
@@ -0,0 +1,134 @@
...
+function createXrefQuery(xrefs) {
+  const queryKeys = [...xrefs.keys()].reduce((queryKeys, term) => {
+    return queryKeys.add(JSON.stringify({ term }));
+  }, new Set());
+  return { keys: [...queryKeys].map(JSON.parse) };
+}
+
+// fetch from network
+async function fetchXrefs(query, url) {
+  return await simulateShepherd(query, url);
```
marcoscaceres: I like the idea of arbitrary end points that happen to respond with our preferred data structure. 

So here, we can build the actual thing without needing to say "simulateShepherd"... because we are doing the real thing :)

src/core/xref.js
```diff
@@ -0,0 +1,134 @@
...
+    });
+  }
+}
+
+// disambiguate fetched results based on context
+function disambiguate(data, context, term) {
+  const { elem } = context;
+
+  if (!data || !data.length) {
+    const msg = `No external reference data found for term \`${term}\`.`;
```
marcoscaceres: Make sure you give guidance to the Editor. What action can they take to fix this?

src/core/xref.js
```diff
@@ -0,0 +1,134 @@
...
+    showError(msg, elem);
+    return null;
+  }
+
+  if (data.length === 1) {
+    return data[0]; // unambiguous
+  }
+
+  const msg = `Ambiguity in data found for term \`${term}\`.`;
```
marcoscaceres: As above.

sidvishnoi:
Wait I messed up commits.

tests/data/xref.json
```diff
@@ -0,0 +1,1739 @@
+{
+  "data": [
```
marcoscaceres: We should probably just include what we are testing.

src/core/xref.js
```diff
@@ -0,0 +1,119 @@
...
+  showError(msg, elem);
+  return null;
+
+  function showError(msg, elem) {
```
marcoscaceres: nit: `showInlineError()` ... we should maybe move this to utils, as we do this a lot in the code base.

tests/data/xref.json
sidvishnoi: Now that it's in commit history, I'll do it :)

Issue #1683, opened 2018-05-26
Add CSS for :target


sidvishnoi:
For example, on https://w3c.github.io/payment-request/#bib-payment-method-basic-card, one cannot find `[payment-method-basic-card]` easily as it's on bottom of page where it can't be scrolled to based on URL fragment. One has to make eye movement all the way down to bottom to find it.

It would be nice if there is a little CSS on `:target` which lets one focus easily where the hash link lead to.

<img width="400" src="https://user-images.githubusercontent.com/8426945/40579055-4e457a0a-613d-11e8-896c-82fbd776caf7.png">

_It's not easy to find `[payment-method-basic-card]` in this image_

marcoscaceres:
Agree, some kind of emphasis would be nice. I think Wikipedia does something similar, so maybe we can look at their style.

PR #1720, opened 2018-06-28
feat(core/css/respec): add :target highlight #1683

Jigar3:
#1683 
Added a little CSS on :target which lets one focus easily where the hash link lead to.
Here is the screenshot.

screenshot from 2018-06-28 23-56-07

### DIFF
--- a/js/core/css/respec2.css
+++ b/js/core/css/respec2.css
@@ -39,11 +39,8 @@ a.bibref {
   text-decoration: none;
 }
 
-a.bibref:target {
-  background: #e2e2e2;
-  box-shadow: 0 0 5px;
-  padding: 5px;
-  margin-bottom: 5px;
+:target {
+  background: #eaf3ff;
 }
 
 cite .bibref {
--- a/js/core/css/respec2.css
+++ b/js/core/css/respec2.css
@@ -39,7 +39,7 @@ a.bibref {
   text-decoration: none;
 }
 
-#references:target {
+#references :target {
   background: #eaf3ff;
 }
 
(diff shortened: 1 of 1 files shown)


js/core/css/respec2.css
```diff
@@ -39,6 +39,13 @@ a.bibref {
   text-decoration: none;
 }
 
+a.bibref:target {
```
sidvishnoi: We may want the `:target` css to be applied to more places, not just `a.bibref`

Jigar3: so i should just change a.bibref:target to target??

sidvishnoi: @marcoscaceres WDYT?

sidvishnoi: Also, we may want to consider Wikipedia like `:target` styling. It's simpler :slightly_smiling_face:

Jigar3: hey just set the background-color to #eaf3ff :smile:

sidvishnoi: Lets do that :smiley:

Jigar3: :+1:  on all :target

Jigar3:
Updated Screenshot 
screenshot from 2018-06-29 00-48-15

sidvishnoi:
Oops. found an issue (entire section getting higlighted).
image

What do you think we should do? Use some JS? @marcoscaceres

Jigar3:
@sidvishnoi how to test my changes?. Actually i was just changing styles in dev tools.

js/core/css/respec2.css
```diff
@@ -39,6 +39,10 @@ a.bibref {
   text-decoration: none;
 }
 
+#references:target {
```
sidvishnoi: expected selector: `#references :target`
typo? :slightly_smiling_face:

PR #1722, opened 2018-06-29
fix(core/xref): Improve error messages

sidvishnoi:
- Updates error messages

One present issue that needs to be resolved is that the multiple warnings are emitted for same term. It also tells that our `disambiguate` algorithm is executing for each term (even when they are same). So, that fix is coming up next.

image

### DIFF
--- a/src/core/utils.js
+++ b/src/core/utils.js
@@ -293,11 +293,20 @@ export function removeReSpec(doc) {
   });
 }
 
-export function showInlineError(elem, msg, title) {
-  elem.classList.add("respec-offending-element");
-  elem.setAttribute("title", title || msg);
+/**
+ * Adds error class of each element while emitting a warning
+ * @param {Element|Array:Elements} elems
+ * @param {String} msg message to show in warning
+ * @param {String} title error message to add on each element
+ */
+export function showInlineError(elems, msg, title) {
+  if (!Array.isArray(elems)) elems = [elems];
+  elems.forEach(elem => {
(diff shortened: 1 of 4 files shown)


sidvishnoi:
Update: 
image

src/core/utils.js
```diff
@@ -293,11 +293,20 @@ export function removeReSpec(doc) {
   });
 }
 
-export function showInlineError(elem, msg) {
-  elem.classList.add("respec-offending-element");
-  elem.setAttribute("title", msg);
-  pub("warn", msg + " See develper console for details.");
-  console.warn(msg, elem);
+/**
+ * Adds error class of each element while emitting a warning
```
sidvishnoi: noticed nit 
s/of/to

src/core/xref.js
```diff
@@ -19,12 +19,12 @@ export async function run(conf, elems) {
   const { xref } = conf;
   const xrefMap = createXrefMap(elems);
   const query = createXrefQuery(xrefMap);
-  const apiURL = new URL(xref.url, location.href) || API_URL;
+  const apiURL = (xref.url && new URL(xref.url, location.href)) || API_URL;
```
marcoscaceres: Here you want:

```JS
const apiURL = xref.url ? new URL(xref.url, location.href).href : API_URL;
```

sidvishnoi: `.href`? need some consistency here
so, which one should change? API_URL or new URL here?

PR #1723, opened 2018-06-29
feat(core/xref): use data-cite to disambiguate

sidvishnoi:
- presently only supports parent's `data-cite`, not self `data-cite`

**Update**: Also supports self `data-cite` as:
- if the term exists, url resolves to url fragment
- else: url resolves to spec itself. no warning is emitted

### DIFF
--- a/src/core/xref.js
+++ b/src/core/xref.js
@@ -35,8 +35,10 @@ export async function run(conf, elems) {
 function createXrefMap(elems) {
   return elems.reduce((map, elem) => {
     const term = normalize(elem.textContent);
+    const datacite = elem.closest("[data-cite]");
+    const specs = datacite ? datacite.dataset.cite.split(" ") : [];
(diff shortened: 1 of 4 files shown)


src/core/xref.js
```diff
@@ -47,9 +54,15 @@ function createXrefMap(elems) {
  * @returns {Object} { keys: [{ term }] }
  */
 function createXrefQuery(xrefs) {
-  const queryKeys = [...xrefs.keys()].reduce((queryKeys, term) => {
-    return queryKeys.add(JSON.stringify({ term }));
-  }, new Set());
+  const queryKeys = [...xrefs.entries()].reduce(
+    (queryKeys, [term, entries]) => {
+      entries.forEach(({ specs }) => {
```
marcoscaceres: nit: maybe do `for of ` in here... having `.forEach` inside something you are already seems out of place.

marcoscaceres: Fixed typo above.

PR #1732, opened 2018-07-03
fix(core/xref): use empty data-cite for local references

sidvishnoi:


tests/spec/core/xref-spec.js
```diff
@@ -65,14 +65,24 @@ describe("Core — xref", () => {
         <p>As <a data-cite="!infra">ASCII uppercase</a> is valid dfn, it resolves to fragment. a local data-cite (infra) overrides parent's datacite.</p>
 
         <p>As <a data-cite="!infra">ASCII uppercasing</a> doesn't exist in INFRA, it resolves to spec only.</p>
+
+        <!-- local links with empty parent data-cite -->
```
marcoscaceres: This test might be getting a little big... would be good to split it, but have both conditions tested: that is, local link, and external link inside the section.

src/core/link-to-dfn.js
```diff
@@ -125,10 +127,17 @@ export async function run(conf, doc, cb) {
     }
   });
 
+  const additionaExternalLinks = [
```
marcoscaceres: A comment here would be nice, saying what "additionExternalLinks" are.

src/core/link-to-dfn.js
```diff
@@ -125,10 +127,17 @@ export async function run(conf, doc, cb) {
...
+    ...document.querySelectorAll(
+      "a[data-cite]:not([data-cite='']):not([data-cite*='#'])"
+    ),
+  ].filter(el => {
```
marcoscaceres: Looks like this is mapping? 

.map(el=> el.closest("[data-cite]")).filter(closest = > !closest || closest.dataset.cite !== "")

sidvishnoi: No. Not a mapping here.
The present one takes into account links of type `<section data-cite="foo"><a>bar</a></section>` and `<section><a data-cite="foo">bar</a></section>`. 
What we need in `additionalLinks` is the `el` itself, and not `el.closest()`

marcoscaceres: ok, cool.

Issue #1662, opened 2018-05-11
[META] Implement cross-references feature


sidvishnoi:
Today, creating a cross reference in ReSpec requires Editors to manually search for the `id` of a term by going into another spec and copying the URL. This means Editors that Editors need to write, For example:

```HTML
<a data-cite="HTML/webappapis.html#eventhandler">EventHandler</a>
```

For most Editors, this can be quite labor intensive and error prone. Also, the `id` being referenced might be removed or changed, causing cross references to break.

Ultimately, spec Editors just should be able to just write markup:

```HTML
<a>EventHandler</a>
```

And ReSpec, together with some hints, should know what the user means (i.e., the `EventHandler` typedef, as defined in the HTML specification).  

---

This project aims to make it easy for Editors to link terms defined in other specs. ReSpec should be able to understand itself where `EventHandler` is defined and automatically add the link (and citation of the linked spec) to it.

This feature will extend the present `data-cite` attribute. In the beginning, this feature will link to `dfn` and IDL fragments only. (later stages may support headings and other components).

In case ReSpec is not able to find a reference, or if there is some ambiguity, it will inform Editors so they can take appropriate action.

<h1 id="markup">Markup</h1>

``` html
<a>EventHandler</a>

<!-- really mean the IDL -->
<a data-lt="EventHandler">event handler</a>

<!-- all references are normative by default. -->
<!-- explicit non-normative references: -->
<a class="informative">EventHandler</a>
<a class="informative" data-lt="EventHandler">event handler</a>

<!-- shorthand syntax. (in future) -->
The {{EventHandler}} typedef.
```

One should keep in mind that `EventHandler` (type = `typedef`) is not as same as `event handler` (type = `dfn`).

The above will also be applicable to `<dfn>` (except the shorthand syntax).

All references are normative by default, unless used inside some non-normative parent (like being nested in some `.note`, `.ednote`, `figure`, `.example`) or having `.informative` themselves.

<h1 id="handling-ambiguity">Handling Ambiguity</h1>

In case of ambiguities, the spec writer should be able to resolve it by providing additional information.

``` html
<a>URL parser</a>
<!--
  is ambiguous, as defined at:
  - https://www.w3.org/TR/appmanifest/#dfn-parse
  - https://html.spec.whatwg.org/multipage/infrastructure.html#url-parser
  - https://url.spec.whatwg.org/#concept-url-parser
-->
```

The following markup may be used, to provide additional information:

``` html
[...]

PR #1733, opened 2018-07-04
feat(core/xref): support external dfn

sidvishnoi:
Part of https://github.com/w3c/respec/issues/1662

### DIFF
--- a/src/core/link-to-dfn.js
+++ b/src/core/link-to-dfn.js
@@ -132,10 +132,8 @@ export async function run(conf, doc, cb) {
     // The `possibleExternalLinks` above doesn't include references that
     //   match selectors like `a[data-cite="spec"]`
     ...document.querySelectorAll(
-      [
-        "a[data-cite]:not([data-cite='']):not([data-cite*='#'])",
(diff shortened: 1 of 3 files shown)


src/core/link-to-dfn.js
```diff
@@ -132,7 +132,10 @@ export async function run(conf, doc, cb) {
     // The `possibleExternalLinks` above doesn't include references that
     //   match selectors like `a[data-cite="spec"]`
     ...document.querySelectorAll(
-      "a[data-cite]:not([data-cite='']):not([data-cite*='#'])"
+      [
```
sidvishnoi: Recommendation: Array vs backticks vs `+` here

marcoscaceres: If all we are doing here is trying to avoid a long line, then just:
``
...document.querySelectorAll(
  "a[data-cite]:not([data-cite='']):not([data-cite*='#'])," +
  "dfn:not([data-cite='']):not([data-cite*='#'])"
);
```

Then no need for the array.

PR #1734, opened 2018-07-04
fix(core/biblio): prevent SpecRef request if there are no refs

### LINKED ISSUE #1726: Prevent SpecRef request if there are no refs
If a request to SpecRef is sent that has no values in `refs`, such as `https://specref.herokuapp.com/bibrefs?refs=`, we gets lots of errors (like `Legacy SpecRef entries are not supported`) as SpecRef returns data that we don't want (or handle) and respec hangs.

TODO: 
1. Investigate why and when does it happen.
2. As a fix, if there are no `refs`, don't send a request.

sidvishnoi:
Fixes #1726

### DIFF
--- a/src/core/biblio.js
+++ b/src/core/biblio.js
@@ -48,6 +48,7 @@ export async function updateFromNetwork(
   options = { forceUpdate: false }
 ) {
   // Update database if needed, if we are online
+  refs = refs.filter(ref => ref.trim());
   if (!refs.length || navigator.onLine === false) {
     return;
   }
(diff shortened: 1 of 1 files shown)


src/core/biblio.js
```diff
@@ -48,6 +48,7 @@ export async function updateFromNetwork(
   options = { forceUpdate: false }
 ) {
   // Update database if needed, if we are online
+  refs = refs.filter(ref => ref.trim());
```
marcoscaceres: nit: maybe move this above the comment?

marcoscaceres: oh, also, maybe make a new variable (cleanRefs)... best not to trash variables.

PR #1739, opened 2018-07-05
refactor(core/pluralize): separate pluralization from data-lt

sidvishnoi:
- Adds `core/pluralize` as plugin
- Separates `data-lt` and `data-plurals`
- **Uses first `data-lt` for dfn `id` if available, otherwise uses `textContent`**
- **Doesn't use `data-plural` for `id`**


src/core/pluralize.js
```diff
@@ -0,0 +1,68 @@
...
+
+import { plural as pluralOf } from "deps/pluralize";
+import { norm as normalize } from "core/utils";
+
+export const name = "core/pluralize";
+
+export function run(conf) {
+  if (!conf.pluralize) return;
+
+  if (!conf.hasOwnProperty("definitionMap")) {
```
marcoscaceres: we should probably move this now to base-runner.js, because we are doing this check twice now.

In base runner, we want:

```JS
const respecDefaults = {
  conf.definitionMap: Object.create(null);
}

// and in runAll()
const conf = { ...respecDefaults, ...window.respecConfig };
```

marcoscaceres: fixed typo.

marcoscaceres: And then we can remove this from check from src/core/dfn.js

src/core/pluralize.js
```diff
@@ -0,0 +1,68 @@
...
+        plurals.reduce((defMap, plural) => {
+          if (!defMap[plural]) defMap[plural] = [];
+          defMap[plural].push($(dfn));
+          return defMap;
+        }, conf.definitionMap);
+      }
+    });
+}
+
+function pluralizer() {
```
marcoscaceres: pluralizer is a noun... can we make this a verb?

sidvishnoi: `pluralizer` returns a function which causes pluralization. So, `pluralizer` isn't actually doing the pluralization itself.

We can rename it to `getPluralizer` maybe?

src/core/base-runner.js
```diff
@@ -48,6 +52,7 @@ function toRunnable(plug) {
 }
 
 export async function runAll(plugs) {
+  Object.assign(respecConfig, respecDefaults);
```
marcoscaceres: oh, this will trash things... we want it the other way

Issue #1742, opened 2018-07-06
Another somewhat unhelpful error


marcoscaceres:
## Important info

* ReSpec version: Click the ReSpec pill (e.g., 20.0.1).
* [x] I did a "hard refresh", but it's still busted. 

## Description of problem

What happened (e.g., it crashed)?: 

<img width="510" alt="screenshot 2018-07-06 14 14 43" src="https://user-images.githubusercontent.com/870154/42394131-0631fa72-8127-11e8-9811-25b81e0651a9.png">

But what I was expecting?:
A little bit more info... like, we know it's an operation, so `()` would be nice... 

Optional, steps to reproduce:
 1. write an operation in IDL (e.g., `void foo()`), but don't define it.

PR #1744, opened 2018-07-06
fix(core/webidl): improve warning msg for operation

sidvishnoi:
Fixes https://github.com/w3c/respec/issues/1742
image


src/core/webidl.js
```diff
@@ -817,9 +817,13 @@ function findDfn(parent, name, definitionMap, type, idlElem) {
       name &&
       idlElem.classList.contains("no-link-warnings") === false;
     if (showWarnings) {
-      var msg = `No \`<dfn>\` for ${type} \`${originalName}\`${originalParent
-        ? " in `" + originalParent + "`"
-        : ""}`;
+      let msg = `No \`<dfn>\` for ${type} `;
```
marcoscaceres: So, would really like this to be like:

```
const thingName = type === "operation" ? `${originalName}()` : originalName;
`Missing \`<dfn>\` for \`${originalParent}`\'s ${thingName} ${type}`;
```

sidvishnoi: why didn't I think like that?!

sidvishnoi: Is `originalParent` always present? Or I need to handle it?

Issue #1699, opened 2018-06-19
Refactor all jquery-enhanced utils to vanilla js while keeping backward compatibility


sidvishnoi:
Refactor all core/jquery-enhanced.js utils to vanilla js (in core/utils) while keeping backward compatibility and use vanilla js DOM utils in ReSpec.
- Modules:
  - [x] $.makeID -> addId (https://github.com/w3c/respec/pull/1691)
  - [x] $.allTextNodes -> getTextNodes (https://github.com/w3c/respec/pull/1698)
  - [x] $.renameElement -> renameElement (https://github.com/w3c/respec/pull/1806)
  - [x] $.getDfnTitles -> getDfnTitles (https://github.com/w3c/respec/pull/1746)
  - [x] $.linkTargets -> getLinkTargets (https://github.com/w3c/respec/pull/1785)
- [x] Move tests from `jquery-enhanced-spec.js` to `utils-spec.js` (https://github.com/w3c/respec/pull/1838)
- [ ] Add deprecation notice to jquery-enhanced

marcoscaceres:
Closing as done. Looking at making a new W3C profile that excludes jQuery.

PR #1746, opened 2018-07-09
refactor(core/jquery-enhanced): move getDfnTitles to vanilla js

sidvishnoi:
As a part of https://github.com/w3c/respec/issues/1699.

Do we still need to keep the backward compatibility for `dfn-type => data-dfn-type`, `lt => data-lt` etc? 

Presently just moved code and removed jquery. Will try to refactor this function more next.


saschanaz:
>Do we still need to keep the backward compatibility for dfn-type => data-dfn-type, lt => data-lt etc?

I think we already removed it but not sure...

marcoscaceres:
> Do we still need to keep the backward compatibility for dfn-type => data-dfn-type, lt => data-lt etc?

As @saschanaz, I think we removed that. But if there are some strays, it's ok to remove them.

src/core/utils.js
```diff
@@ -626,3 +626,60 @@ export function getTextNodes(el, exclusions = []) {
...
+ *   <dfn> element (but can apply to other as well).
+ * if args.isDefinition is true, then the element is a definition, not a
+ *   reference to a definition. Any @title or @lt will be replaced with
+ *   @data-lt to be consistent with Bikeshed / Shepherd.
+ * This method now *prefers* the data-lt attribute for the list of
+ *   titles. That attribute is added by this method to dfn elements, so
+ *   subsequent calls to this method will return the data-lt based list.
+ * @param {Element} elem
+ * @param {Object} args
+ * @returns {Array:String} array of title strings
```
saschanaz: Maybe just `string[]` according to the docs: http://usejsdoc.org/tags-type.html

src/core/utils.js
```diff
@@ -626,3 +626,60 @@ export function getTextNodes(el, exclusions = []) {
...
+  if (elem.dataset.lt) {
+    // prefer @data-lt for the list of title aliases
+    titleString = elem.dataset.lt.toLowerCase();
+    if (normText !== "" && !titleString.startsWith(`${normText}|`)) {
+      // Use the definition itself, so to avoid having to declare the definition twice.
+      titleString = titleString + "|" + normText;
+    }
+  } else if (
+    elem.childNodes.length === 1 &&
+    [...elem.children].filter(el => el.matches("abbr, acronym")).length === 1 &&
```
saschanaz: I would recommend `elem.querySelectorAll` or even `elem.getElementsByTagName`.

`acronym` is an obsolete element in HTML5, do we still need it? @marcoscaceres

src/core/utils.js
```diff
@@ -626,3 +626,60 @@ export function getTextNodes(el, exclusions = []) {
...
+    elem.querySelector(":first-child").getAttribute("title")
```
saschanaz: `elem.children[0]`?

src/core/utils.js
```diff
@@ -626,3 +626,60 @@ export function getTextNodes(el, exclusions = []) {
...
+  ) {
+    titleString = elem.querySelector(":first-child").getAttribute("title");
```
saschanaz: Same here.

src/core/utils.js
```diff
@@ -626,3 +626,60 @@ export function getTextNodes(el, exclusions = []) {
...
+  } else {
+    titleString =
+      elem.textContent === '""' ? "the-empty-string" : elem.textContent;
```
sidvishnoi: @saschanaz I don't understand this `'""'` test. Isn't that simply testing empty string? Or are there cases when `textContent === '""'`

marcoscaceres: The intention was to test if it's exactly what's there (i.e., a pair of quotes).

marcoscaceres: screenshot 2018-07-10 08 13 28

sidvishnoi: So, one may write `<a>""</a>`?

marcoscaceres: Yep.

src/core/utils.js
sidvishnoi: I found some instances of `acronym` in some `respecConfig`s and tests/specifcations in ReSpec. Not sure if we should remove its support. WDYT @marcoscaceres?

marcoscaceres:
Re: acronym, please kill it if you haven't already!

src/core/jquery-enhanced.js
```diff
@@ -50,57 +50,7 @@ window.$.fn.renameElement = function(name) {
 // This method will publish a warning if a title is used on a definition
 // instead of an @lt (as per specprod mailing list discussion).
 window.$.fn.getDfnTitles = function(args) {
```
saschanaz: @sidvishnoi @marcoscaceres Why do we have to keep this extension? Is there any compat issue?

saschanaz: Removing this will require fixing tests, though.

saschanaz: I think we should mark them as deprecated, as it's easy enough in this case.

marcoscaceres: Absolutely. We should console.warn and start seeing if there will be any damage in yanking support.

PR #1749, opened 2018-07-10
refactor(core/webidl): remove dead code, unjquery

marcoscaceres:


src/core/webidl.js
```diff
@@ -13,34 +13,33 @@ import { normalizePadding } from "core/utils";
...
 
-var idlAttributeTmpl = tmpls["attribute.html"];
-var idlCallbackTmpl = tmpls["callback.html"];
-var idlConstTmpl = tmpls["const.html"];
-var idlDictionaryTmpl = tmpls["dictionary.html"];
-var idlDictMemberTmpl = tmpls["dict-member.html"];
-var idlEnumItemTmpl = tmpls["enum-item.html"];
-var idlEnumTmpl = tmpls["enum.html"];
-var idlExtAttributeTmpl = tmpls["extended-attribute.html"];
-var idlFieldTmpl = tmpls["field.html"];
```
saschanaz: Can we also remove field.html?

src/core/webidl.js
```diff
@@ -13,34 +13,33 @@ import { normalizePadding } from "core/utils";
...
-var unlinkable = new Set(["maplike", "setlike", "stringifier"]);
+const unlinkable = new Set(["maplike", "setlike", "stringifier"]);
 
 function registerHelpers() {
-  hb.registerHelper("extAttr", function(obj, indent) {
-    return extAttr(obj.extAttrs, indent, /*singleLine=*/ false);
+  hb.registerHelper("extAttr", function(obj) {
+    return extAttr(obj.extAttrs);
   });
   hb.registerHelper("extAttrInline", function(obj) {
```
saschanaz: Maybe we even can kill `extAttrInline` and replace it by `extAttr`.

src/core/webidl.js
```diff
@@ -533,17 +447,11 @@ function writeAttributeQualifiers(attr) {
...
-  var len = idlType2Text(attr.idlType).length;
-  var pad = max - len;
+function writeAttribute(attr) {
   var qualifiers = writeAttributeQualifiers(attr);
-  qualifiers += pads(maxQualifiers);
-  qualifiers = qualifiers.slice(0, maxQualifiers);
   return idlAttributeTmpl({
     obj: attr,
-    indent: indent,
     qualifiers: qualifiers,
```
saschanaz: nit: just `qualifiers`

tests/spec/core/webidl-spec.js
```diff
@@ -602,42 +602,32 @@ partial dictionary AnotherThing {
...
-    expect($target.find(".idlEnumID").text()).toEqual("EnumBasic");
-    expect($target.find(".idlEnumItem").length).toEqual(4);
-    expect(
-      $target
-        .find(".idlEnumItem")
-        .first()
-        .text()
-    ).toEqual('"one"');
+    var target = doc.getElementById("enum-basic");
+    var text = `
```
saschanaz: nit: `const`

src/core/webidl.js
marcoscaceres: Oh, yeah! Good catch!

src/core/webidl.js
marcoscaceres: extAttrInline was not actually used anywhere at all... so killing it.

marcoscaceres: Actually, ignore that... it was used in a template.

src/core/webidl.js
```diff
@@ -809,9 +683,9 @@ function findDfn(parent, name, definitionMap, type, idlElem) {
...
   if (dfns.length > 1) {
-    const msg = `Multiple \`<dfn>\`s for \`${originalName}\` ${originalParent
-      ? `in \`${originalParent}\``
-      : ""}`;
+    const msg = `Multiple \`<dfn>\`s for \`${originalName}\` ${
+      originalParent ? `in \`${originalParent}\`` : ""
+    }`;
     pub("error", new Error(msg));
   }
   if (dfns.length === 0) {
```
sidvishnoi: Nit: _micro-optimization_: `else if`

src/core/webidl.js
```diff
@@ -731,7 +606,7 @@ function findDfn(parent, name, definitionMap, type, idlElem) {
   const originalName = name;
   parent = parent.toLowerCase();
   switch (type) {
-    case "operation":
+    case "operation": {
       // Overloads all have unique names
       if (name.search("!overload") !== -1) {
```
sidvishnoi: maybe use `name.includes("!overload")` here

src/core/webidl.js
```diff
@@ -310,77 +274,6 @@ const standardTypes = new Map([
...
-  "partial",
-  "required",
-  "setlike",
-  "setter",
-  "static",
-  "stringifier",
-  "typedef",
-  "unrestricted",
-]);
 var operationNames = {};
```
sidvishnoi: `const`

src/core/webidl.js
marcoscaceres: Going to leave the bits that are already there... as there is a lot of bad code in there that needs to change.

PR #1750, opened 2018-07-10
fix(core/xref): use context to disambiguate

sidvishnoi:
ReSpec was not taking full context into account.

``` html
<p data-cite="webidl"><a id="one">object</a></p>
<p data-cite="html"><a id="two">object</a></p>
```

Query sent to API looks like:
``` json
[
  { "term": "object", "specs": ["webidl"] },
  { "term": "object", "specs": ["html"] }
]
```

and the response:

``` json
"object": [
    {
      "spec": "html",
      "type": "element",
      "uri": "iframe-embed-object.html#the-object-element",
    },
    {
      "spec": "webidl",
      "type": "interface",
      "uri": "#idl-object",
    }
  ]
```

Which ReSpec treated as ambiguous as it only looked at term and not combination of term and specs (context).


tests/spec/core/xref-spec.js
```diff
@@ -197,6 +197,24 @@ describe("Core — xref", () => {
     expect(link.title).toEqual("Error: Linking an ambiguous dfn.");
   });
 
+  it("disambiguates response based on context", async () => {
+    // https://github.com/w3c/respec/pull/1750
+    const body = `
+      <p data-cite="webidl"><a id="one">object</a></p>
+      <p data-cite="html"><a id="two">object</a></p>
```
marcoscaceres: `
<p data-cite="html">
   <a id="three" data-cite="OTHER">object</a> 
   <a>object</a>
</p>
`

marcoscaceres: updated test case.

sidvishnoi: `OTHER` is intended to fail, right?

marcoscaceres: And maybe can add a expected fail case.

marcoscaceres:
```HTML
<a id="five" data-cite="">object</a>
<a id="six" data-cite="NOT-FOUND">object</a>
```

PR #1751, opened 2018-07-10
feat(core/xref): use inline references to provide context

sidvishnoi:
Uses nearby `[[SPEC]]` to provide context as `specs` to search for (as a part of https://github.com/w3c/respec/issues/1662)

Maybe this should be enabled via a flag, not always. Why? Consider:
``` html
<section data-cite="html">Uses [[WEBIDL]] to create context for <a>object</a></section>
```
Above, the query for this term will contain `specs: ["html", "webidl"]`, which results in an ambiguity that can't be resolved unless (either):
- User removes `data-cite="html"` from parent or remove the nearby `[[WEBIDL]]` 
- Instead of `specs.push(...inlineContextSpecs)`, we do `specs = inlineContextSpecs`

### DIFF
--- a/src/core/xref.js
+++ b/src/core/xref.js
@@ -48,11 +48,9 @@ function createXrefMap(elems) {
       : [];
     if (datacite !== elem) {
       // if element itself contains data-cite, we don't take inline context into account
-      const inlineContext = elem
-        .closest("section")
-        .querySelectorAll("a.bibref");
-      if (inlineContext) {
-        const inlineContextSpecs = [...inlineContext].map(el =>
+      const refs = elem.closest("section").querySelectorAll("a.bibref");
+      if (refs) {
+        const inlineContextSpecs = [...refs].map(el =>
           el.textContent.toLowerCase().replace(/^!/, "")
         );
         specs.push(...inlineContextSpecs);
--- a/src/core/xref.js
+++ b/src/core/xref.js
@@ -46,6 +46,14 @@ function createXrefMap(elems) {
           .replace(/!/g, "")
           .split(" ")
       : [];
+    const inlineContext = elem.parentElement.querySelectorAll("a.bibref");
+    if (inlineContext) {
+      const inlineContextSpecs = [...inlineContext].map(el =>
+        el.textContent.toLowerCase().replace(/^!/, "")
+      );
+      specs.push(...inlineContextSpecs);
+    }
+
     const xrefsForTerm = map.has(term) ? map.get(term) : [];
     xrefsForTerm.push({ elem, specs });
     return map.set(term, xrefsForTerm);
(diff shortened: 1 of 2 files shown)


marcoscaceres:
So, given:

```
<section data-cite="html">Uses [[WEBIDL]] to create context for <a>object</a></section>
```

I'm not sure I agree... that's clearly ambiguous, and the Editor should deal with.

marcoscaceres:
updated example above.

sidvishnoi:
So, if the element itself has `data-cite`, it wins. No inline context need to be taken into account. Right?

marcoscaceres:
> So, if the element itself has data-cite, it wins. No inline context need to be taken into account. Right?

Correct. `.closest()` wins.

src/core/xref.js
```diff
@@ -46,6 +46,17 @@ function createXrefMap(elems) {
           .replace(/!/g, "")
           .split(" ")
       : [];
+    if (datacite !== elem) {
+      // if element itself contains data-cite, we don't take inline context into account
+      const inlineContext = elem.parentElement.querySelectorAll("a.bibref");
```
marcoscaceres: discussed making this search wider... up to `section` level .

src/core/xref.js
```diff
@@ -46,6 +46,17 @@ function createXrefMap(elems) {
           .replace(/!/g, "")
           .split(" ")
       : [];
+    if (datacite !== elem) {
+      // if element itself contains data-cite, we don't take inline context into account
+      const refs = elem.closest("section").querySelectorAll("a.bibref");
+      if (refs) {
+        const inlineContextSpecs = [...refs].map(el =>
+          el.textContent.toLowerCase().replace(/^!/, "")
```
marcoscaceres: We should do the removal of the “!” when the inline references are converted. So, in core/inlines i think.

src/core/xref.js
```diff
@@ -46,6 +46,14 @@ function createXrefMap(elems) {
           .replace(/!/g, "")
           .split(" ")
       : [];
+    if (datacite !== elem) {
+      // if element itself contains data-cite, we don't take inline context into account
+      const refs = elem.closest("section").querySelectorAll("a.bibref");
```
marcoscaceres: might be cleaner to just do:
```
const refs = [...elem.closest("section").querySelectorAll("a.bibref")]
  .map(el => el.textContent.toLowerCase());
specs.push(...refs);
```

src/core/xref.js
```diff
@@ -46,6 +46,14 @@ function createXrefMap(elems) {
...
+      if (refs) {
```
marcoscaceres: note that `refs` is always true, because ref is an array object.

PR #1756, opened 2018-07-16
fix(core/webidl): unindentMarkup() should work with space-only lines

saschanaz:
The [Browser Extensions](https://browserext.github.io/browserext/) spec is broken because of this. (...which still has `implements`, BTW!)

### DIFF
--- a/src/core/webidl.js
+++ b/src/core/webidl.js
@@ -294,7 +294,7 @@ function unindentMarkup(text) {
   if (!text) {
     return text;
   }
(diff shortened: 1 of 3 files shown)


saschanaz:
Or should we use `trimRight` for every line?

marcoscaceres:
Don’t have a strong opinion about trimRight, whatever you think is best.

marcoscaceres:
+ respec@22.3.1

PR #1765, opened 2018-07-25
(core/xref): support inline IDL

sidvishnoi:
Supports expanding `{{{TERM}}}` to valid xrefs. As a part of #1662.


src/core/inlines.js
```diff
@@ -38,6 +40,7 @@ export function run(conf) {
   const rx = new RegExp(
     "(\\bMUST(?:\\s+NOT)?\\b|\\bSHOULD(?:\\s+NOT)?\\b|\\bSHALL(?:\\s+NOT)?\\b|" +
       "\\bMAY\\b|\\b(?:NOT\\s+)?REQUIRED\\b|\\b(?:NOT\\s+)?RECOMMENDED\\b|\\bOPTIONAL\\b|" +
+      "(?:{{3}.*}{3})|" +
```
marcoscaceres: make sure you add a comment here detailing what this one does.

src/core/xref.js
```diff
@@ -55,8 +57,11 @@ function createXrefMap(elems) {
       specs.push(...refs);
     }
 
+    const types = isIDL ? [elem.dataset.xrefType || "idl"] : ["dfn"];
+    const forContext = elem.dataset.xrefFor;
```
marcoscaceres: nit: `const {xrefFor: forContext} = elem.dataset;`

src/core/xref.js
```diff
@@ -55,8 +57,11 @@ function createXrefMap(elems) {
       specs.push(...refs);
     }
 
+    const types = isIDL ? [elem.dataset.xrefType || "idl"] : ["dfn"];
```
marcoscaceres: nit: ` const types = [ isIDL ? elem.dataset.xrefType || "idl" : "dfn" ]`

src/core/inlines.js
```diff
@@ -109,3 +123,58 @@ export function run(conf) {
...
 }
+
+function generateIDLMarkup(ref) {
+  const { base, attribute, method, args } = parseInlineIDL(ref);
+
+  if (!method && !attribute) {
+    return hyperHTML`<code><a class="respec-idl-xref">${base}</a></code>`;
+  }
+
+  const code = hyperHTML`<code></code>`;
```
marcoscaceres: nit: document.createElement("code");  // avoid "jquery"ing

src/core/inlines.js
```diff
@@ -109,3 +123,58 @@ export function run(conf) {
...
+
+  if (base) {
+    code.appendChild(hyperHTML`<a class="respec-idl-xref">${base}</a>`);
+    code.appendChild(document.createTextNode("."));
```
marcoscaceres: doesn't: 
```
code.appendChild(hyperHTML`<a class="respec-idl-xref">${base}</a>.`);
```

Work?

src/core/inlines.js
```diff
@@ -109,3 +123,58 @@ export function run(conf) {
...
+  if (attribute) {
+    // type: base.attribute
+    code.appendChild(hyperHTML`<a class="respec-idl-xref"
+      data-xref-type="attribute" data-xref-for="${base}">${attribute}</a>`);
+    return code;
+  }
+
+  if (method) {
+    // base.method(args)
+    const methodName = method.split("(", 1)[0];
```
marcoscaceres: nit: 
```
const [methodName] = method.split("(", 1);
```

src/core/inlines.js
```diff
@@ -109,3 +123,58 @@ export function run(conf) {
...
+    code.appendChild(hyperHTML`<a class="respec-idl-xref"
```
marcoscaceres: this part needs some attention.

I'd expected:
```JS
return hyperHTML`
  <a class="respec-idl-xref" data-xref-type="method" data-xref-for="${base}">${
   methodName
  }</a>(${args.map(toArgs)})
`
```

tests/spec/core/xref-spec.js
```diff
@@ -418,4 +456,113 @@ describe("Core — xref", () => {
       "[html],[infra],[local-1],[local-2],[local-3],[local-4],[webidl]"
     );
   });
+
+  describe("inline IDL references", () => {
+    it("ignores inlines starting with backslash", async () => {
+      const body = `<section><p id="test">{{{\\PASS }}}</p></section>`;
```
marcoscaceres: Does the space after PASS matter?

tests/spec/core/xref-spec.js
```diff
@@ -418,4 +456,113 @@ describe("Core — xref", () => {
...
+      expect(el.querySelector("code a")).toBeFalsy();
+      expect(el.textContent).toEqual("{{{PASS}}}");
+    });
+
+    it("links inline IDL references", async () => {
+      const body = `
+      <section id="test">
+        <p id="link1">{{{ Window }}}</p>
+        <p id="link2">{{{ [[query]] }}}</p>
+        <p id="link3">{{{ [[type]] }}} is ambiguous.</p>
```
marcoscaceres: type is an internal slot, no? so it shouldn't be ambiguous because it can only link internally?

tests/spec/core/xref-spec.js
sidvishnoi: It doesn't matter. I kept it to show it doesn't matter.

tests/spec/core/xref-spec.js
sidvishnoi: We are yet to do that.
Presently  (as discussed on Slack) - I was expecting `{{{ }}}` to work only for external IDL references. But we plan to support them locally also. (First look locally, if not found look external)

Unless: I'm missing some fact like "internal slots cannot be external references"

src/core/inlines.js
sidvishnoi: No it doesn't :disappointed: 
`"Failed to execute 'appendChild' on 'Node': parameter 1 is not of type 'Node'."` as there are two nodes here

marcoscaceres: Ah, right. I see what’s going on. On my phone now, but I think all this can be collapsed into the `<code>` above.

tests/spec/core/xref-spec.js
marcoscaceres: Internal slots can’t be external references :)

sidvishnoi: So, does this make sense?
``` html
<!-- external - link both Credential and [[CollectFromCredentialStore]] -->
<p>{{{ Credential.[[CollectFromCredentialStore]](options, sameOriginWithAncestors) }}}</p>

<!-- local -->
<p>{{{ [[CollectFromCredentialStore]](options, sameOriginWithAncestors) }}}</p>
```

src/core/inlines.js
sidvishnoi: I've done some edits. Will update :)

src/core/inlines.js
```diff
@@ -109,3 +125,71 @@ export function run(conf) {
     txt.parentNode.replaceChild(df, txt);
   }
 }
+
+function initInlineIdlParser() {
```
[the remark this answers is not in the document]
sidvishnoi: If we send it to its own module, it no longer requires to be a class. The wrapper her is just for encapsulation. We can export functions directly from module.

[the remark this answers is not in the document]
marcoscaceres: Imagine this as a generalized js module, published as it’s own npm package. Tho it’s certainly possible to just export a parse function, ergonomically it feels weird (to me) - specially if you have to go through the initialize function.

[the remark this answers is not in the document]
sidvishnoi: So, this (https://github.com/w3c/respec/pull/1765/commits/0b5977754872372f40a68d658789ddecb68d81a2) won't do?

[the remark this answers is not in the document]
sidvishnoi: Also, the markup generator should be in `core/inlines` I guess, right?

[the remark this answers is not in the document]
marcoscaceres: Can definitely live with just the function. If it needs a refactor in the future, should be trivial to do. 

Don’t have a strong opinion about where the markup generator should go.

src/core/inline-idl-parser.js
```diff
@@ -0,0 +1,39 @@
+/**
+ * Parses an IDL string and returns its components as:
+ *
```
marcoscaceres: Did you end up adding internal slot support: 

```
Foo[[bar]]
```

src/core/inlines.js
```diff
@@ -65,6 +67,19 @@ export function run(conf) {
           );
           // remember which ones were used
           conf.respecRFC2119[matched] = true;
+        } else if (conf.xref && matched.startsWith("{{{")) {
```
marcoscaceres: Does inline parser really need to depend on conf.xref?

src/core/inline-idl-parser.js
marcoscaceres: Ah, see it below. Should we document it here?

src/core/inlines.js
```diff
@@ -109,3 +124,47 @@ export function run(conf) {
     txt.parentNode.replaceChild(df, txt);
   }
 }
+
+/**
```
marcoscaceres: Maybe this should all live in inline-idl-parser.js ?

src/core/inline-idl-parser.js
sidvishnoi: `Foo[[bar]]` is also treated as internal slot? I have just added support for `[[bar]]` till now.

src/core/inlines.js
sidvishnoi: Just kept it to have a safe side on `conf.xref` enabled or not. Removing this :slightly_smiling_face:

src/core/inlines.js
sidvishnoi: moving again :slightly_smiling_face:

sidvishnoi:
TODO:
1. Test on payment-request spec
2. Clean up. Currently I've added a lot of additional attributes in markup to make things work. We need to clean them up.

src/core/inline-idl-parser.js
marcoscaceres: So yeah, `Foo[[bar]]` should be:

```
<a>Foo</a><a>[[\bar]]</a>
```

Internal slots are always for specific interfaces.

Issue #1760, opened 2018-07-17
Incorrect extended attribute in IDL causes silent failure


saschanaz:
## Important info

* URL to affected spec:
* ReSpec version: 22.3.1
* [x] I did a "hard refresh", but it's still busted. 

## Description of problem

Try inserting this in examples/basic.html:

```webidl
[CustomExtendedAttribute="wow"]
interface Foo {
  attribute Bar bar;
  void doTheFoo();
};
```

What happened (e.g., it crashed)?: The webidl module fails only with a console message.

But what I was expecting?: No failure

saschanaz:
Turns out that webidl2.js somehow supports the `[ExtAttr="stringliteral"]` syntactically-extended style but ReSpec doesn't know about it. Supporting it is easy, but should we?

saschanaz:
> but warn that it’s a bit non-standard?

We may do it via w3c/webidl2.js#138.

saschanaz:
@marcoscaceres 

So another important point here is that a silent failure happens in ReSpec. I think we may add `pub("fatal", err)` so that a user can be notified that a bug just happened. What do you think?

marcoscaceres:
Fatal could be good, so long as we don’t show the stack. Then we can console.error() for that

marcoscaceres:
We can now define `CustomExtendedAttribute` in the document, otherwise xref tries to link to it. I'm going to go ahead and close this, because the behavior seems good.

saschanaz:
Per https://github.com/w3c/respec/issues/1760#issuecomment-405912359 this is not just a linking problem. Did you try the actual behavior?

marcoscaceres:
I did, but maybe I missed something? It didn't cause any errors and everything looked ok. Please reopen if you think we should do more here.

saschanaz:
I tried this and you're right, it doesn't cause any errors. Probably because ReSpec replaced its own writer with webidl2.js one.

Issue #1741, opened 2018-07-06
Unhelpful error details 


marcoscaceres:
## Important info

* ReSpec version: Click the ReSpec pill (e.g., 20.0.1).
* [x] I did a "hard refresh", but it's still busted. 

## Description of problem
the first part is helpful, but the details are not:

<img width="532" alt="screenshot 2018-07-06 10 09 19" src="https://user-images.githubusercontent.com/870154/42383408-2249beec-8105-11e8-8aba-96c4d7465b34.png">

What happened (e.g., it crashed)?: 
It showed me JS errors... I don't care about those. 

But what I was expecting?:
To not see JS errors. 

Optional, steps to reproduce:
 1. Make some IDL attribute, and then define it twice.

sidvishnoi:
Should be caused on any `pub("error")` as:
https://github.com/w3c/respec/blob/b571192c7d7b9f09872e371c6245f740e9546511/src/core/pubsubhub.js#L63-L65

saschanaz:
I agree we can just remove it as AFAIK error stacks are visible via dev tools.

marcoscaceres:
Agree

ThobyV:
Would love to work on this issue @marcoscaceres. I'm pretty new, looking for work to get things started.

marcoscaceres:
Just checked... was actually fixed at some point.

PR #1766, opened 2018-07-25
refactor(core/webidl): modernize run()

saschanaz:
This:

1. removes deprecated `cb()` and an unused exposed promise `done`.
2. removes jQuery.

### DIFF
--- a/src/core/base-runner.js
+++ b/src/core/base-runner.js
@@ -62,7 +62,7 @@ export async function runAll(plugs) {
     try {
       await task(respecConfig);
     } catch (err) {
-      console.error(err);
+      pub("error", err);
     }
   }
   pub("plugins-done", respecConfig);
--- a/src/core/base-runner.js
+++ b/src/core/base-runner.js
@@ -62,7 +62,7 @@ export async function runAll(plugs) {
     try {
       await task(respecConfig);
     } catch (err) {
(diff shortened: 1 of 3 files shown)


src/core/base-runner.js
```diff
@@ -62,7 +62,7 @@ export async function runAll(plugs) {
     try {
       await task(respecConfig);
     } catch (err) {
-      console.error(err);
+      pub("error", err);
```
marcoscaceres: I don't think we want to do this, unfortunately: the reason is that we don't want to surface JS errors to end users, as they can't actually fix those. We only `pub("error")` for things users can fix.

marcoscaceres:
The rest looks great tho! thanks for doing that.

src/core/base-runner.js
saschanaz: I still think we should have a way to show that something failed in a fatal way. A silent failure is never helpful for both authors and devs.

saschanaz: Maybe something like `pub("fatal", err)`?

saschanaz: I'll revert that commit in this PR, we can discuss more in #1760.

sidvishnoi: Related: https://github.com/w3c/respec/issues/1741

PR #1767, opened 2018-07-25
refactor: remove domReady dependency

saschanaz:
Because it's 2018 😁

At first I tried using async function directly in `define()` but that failed. I absolutely have no idea why.

### DIFF
--- a/js/profile-w3c-common.js
+++ b/js/profile-w3c-common.js
@@ -21,72 +21,76 @@ require.config({
   deps: ["deps/hyperhtml", "deps/url-search-params"],
 });
 
-define(
-  [
-    // order is significant
-    "deps/domReady",
-    "core/base-runner",
-    "core/ui",
-    "core/location-hash",
-    "core/l10n",
-    "w3c/defaults",
-    "core/style",
-    "w3c/style",
-    "w3c/l10n",
-    "core/github",
-    "core/data-include",
-    "core/markdown",
-    "w3c/headers",
-    "w3c/abstract",
-    "w3c/conformance",
-    "core/data-transform",
-    "core/inlines",
-    "core/dfn",
-    "core/pluralize",
-    "w3c/rfc2119",
-    "core/examples",
-    "core/issues-notes",
-    "core/requirements",
-    "core/best-practices",
-    "core/figures",
-    "core/webidl",
-    "core/data-cite",
-    "core/biblio",
-    "core/webidl-index",
-    "core/link-to-dfn",
-    "core/render-biblio",
-    "core/contrib",
-    "core/fix-headers",
-    "core/structure",
-    "w3c/informative",
-    "w3c/permalinks",
-    "core/id-headers",
-    "core/caniuse",
-    "ui/save-html",
-    "ui/search-specref",
-    "ui/dfn-list",
-    "ui/about-respec",
-    "core/seo",
-    "w3c/seo",
-    "core/highlight",
-    "core/webidl-clipboard",
-    "core/data-tests",
-    "core/list-sorter",
-    "core/highlight-vars",
-    /*Linter must be the last thing to run*/
-    "core/linter",
-  ],
-  (domReady, runner, { ui }, ...plugins) => {
-    ui.show();
-    domReady(async () => {
-      try {
-        await runner.runAll(plugins);
-        await document.respecIsReady;
-      } catch (err) {
-        console.error(err);
-      } finally {
-        ui.enable();
-      }
-    });
+define([
+  // order is significant
+  "core/base-runner",
+  "core/ui",
+  "core/location-hash",
+  "core/l10n",
+  "w3c/defaults",
+  "core/style",
+  "w3c/style",
+  "w3c/l10n",
+  "core/github",
+  "core/data-include",
+  "core/markdown",
+  "w3c/headers",
+  "w3c/abstract",
+  "w3c/conformance",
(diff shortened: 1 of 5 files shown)


js/profile-w3c-common.js
```diff
@@ -76,9 +75,9 @@ define(
     /*Linter must be the last thing to run*/
     "core/linter",
   ],
-  (domReady, runner, { ui }, ...plugins) => {
+  (runner, { ui }, ...plugins) => {
```
marcoscaceres: I think we should just make this function `async` and inline the domReady() function here. That will clean up this quite a bit.

package.json
```diff
@@ -26,7 +26,6 @@
     "babel-preset-stage-3-without-async": "^1.0.0",
     "chai": "^4.1.2",
     "clipboard": "^2.0.1",
-    "domReady": "github:requirejs/domReady#d85bdc38a6df868ead52ab3e80715aaf60e68765",
```
marcoscaceres: So, other profiles will still depend on it, but the W3C profile won't. So we should leave it as a dependency, otherwise we will break others.

tools/copydeps.js
```diff
@@ -11,7 +11,6 @@ const unlink = promisify(fs.unlink);
 
 const srcDesMap = new Map([
   ["./node_modules/clipboard/dist/clipboard.js", "./js/deps/"],
-  ["./node_modules/domReady/domReady.js", "./js/deps/"],
```
marcoscaceres: As above. Let's leave it - it will just be ignored for W3C, which is great.

js/profile-w3c-common.js
saschanaz: As I said, it didn't work. I completely have no idea why... Both Edge and Firefox didn't even try to execute the callback if I use async function.

package.json
saschanaz: Hmm, do we have an example that uses `domReady`?

js/profile-w3c-common.js
saschanaz: This is what I tried:

```js
  async (runner, { ui }, ...plugins) => {
    ui.show();
    await domReady();
    try {
      await runner.runAll(plugins);
      await document.respecIsReady;
    } catch (err) {
      console.error(err);
    } finally {
      ui.enable();
    }
  }
```

saschanaz: Found the cause: https://github.com/requirejs/requirejs/pull/1729#issuecomment-382297882

OMG, seriously.

marcoscaceres: Argh! sorry you had to go through that.

js/profile-w3c-common.js
```diff
@@ -21,10 +21,9 @@ require.config({
   deps: ["deps/hyperhtml", "deps/url-search-params"],
 });
 
-define(
+require(
```
marcoscaceres: I don't know if this will build or not? I can't recall why it had to be define here.

marcoscaceres:
Strange IDL errors in Travis?

js/profile-w3c-common.js
marcoscaceres: Seems it builds ok, looking at the non-failing Travis tests, so ignore my comment above.

saschanaz:
Again I have no idea what's going on, but anyway reverting to `define()` fixed the errors.

package.json
marcoscaceres: ah, I was unaware that they had their own independent list of dependencies. I’d assumed they were using the same as us.

saschanaz: Any forks should also have their package.json forked, so IMO they shouldn't be affected, no?

marcoscaceres: The problem is merging https://github.com/Geonovum/respec/blob/develop/tools/copydeps.js

As it’s going to delete domRedady there, so things will break for them.

marcoscaceres: About package.json, no unfortunately :( merging still occurs, so deletions still happen. Only new files on their forks are unaffected.

marcoscaceres:
I’m hoping that by the end of the year we will be able to move to webpack or rollup. Should help make things better 🤞

saschanaz:
Any clue what Codacy wants? Does it really want removing all indentations?

marcoscaceres:
Lol, looks like it. It’s ok to ignore it.

package.json
saschanaz: Ah yes, deletions happen. I thought it will cause merge conflict for the profile but found that it's a separate one.

But even then they should have their CI on (or at least local test if not). Things can always be broken when merging, that's why we have tests...

saschanaz: 👍 for that reasoning.

Still we'll want to remove it after enough time, will be great if we can `mark as deprecated` or anything that will allow smooth migration. Otherwise we'll never be able to remove e.g. jQuery from package.json.

marcoscaceres: We might have lead by example, and point other fork maintainers to our code so they can copy what we did. Eventually, we can force a breaking change... but yeah, some might be impossible to ditch, like jQuery... but hopefully we will eventually fully remove it from core/ and w3c/.

saschanaz:
>I’m hoping that by the end of the year we will be able to move to webpack or rollup.

I was hoping ES modules instead, but that won't happen 😢

saschanaz:
Both Edge 18 preview and Firefox Nightly 63 still does not support dynamic module loading, so at least not in 2018.

Issue #1508, opened 2018-02-14
WebIDL name collision issue


danielpeintner:
We, the WoT Scripting API taskforce, run into a little issue with respec when having the same name (even though with different spelling) in WebIDL.

Please take a look at https://rawgit.com/w3c/wot-scripting-api/45666df86df76a7dc76a085c48d1e5d6183a5a42/index.html

In Section "5.7.1 The ThingPropertyInit dictionary" we have `boolean    observable = true;`.

In Section "6. Observables" we have `interface Observable ...`.

Having the same name causes respec error "`No <dfn> for interface Observable.`".

This problem disappears IF we rename the first snippet to `boolean    observableX = true;` for example.

I would think our use of WebIDL is correct? Hence I assume this is a bug or is there anything else we can do to work around it.

Thanks,

-- Daniel

dontcallmedom:
I hit the same issue recently - the workaround is to use `data-dfn-for=""` on the `<dfn>` used to markup the definition associated with the interface - see also https://github.com/w3c/webrtc-pc/commit/9403a1b3dc67db3243ba15f55e98fd6cac928175

dontcallmedom:
(leaving the issue open since avoiding the need for that workaround would be nice)

marcoscaceres:
@danielpeintner please also have a quick look at https://github.com/w3c/respec/wiki/WebIDL-Guide

We are working on trying to make ReSpec smarter as part of #1050.

saschanaz:
```html
<section data-dfn-for="Observable">
  <h3>The <dfn>Observable</dfn> interface</h3>
  ...
</section>
```

This is a weird situation, the parent `data-dfn-for` makes `<dfn>Observable</dfn>` be resolved like `observable.Observable`. Currently ReSpec first searches the top level `dfn`, then tries lower levels as a workaround.

https://github.com/w3c/respec/blob/29aebe1ebda0a83577eb1c829e0f69d148a4f399/src/core/webidl.js#L972-L976

But it didn't work because of the `boolean observable`.

saschanaz:
Currently the definition map is case insensitive, will there any problems if it becomes case sensitive? @marcoscaceres

marcoscaceres:
Not for IDL links, but yes for others... so we need to distinguish them.

marcoscaceres:
Closing as duplicate of #1050.

PR #1776, opened 2018-07-28
fix(core/idl): linking attribute id/type with same name

### LINKED ISSUE #1768: Multiple dfn with the same name
## Important info

* URL to affected spec: https://w3c.github.io/hr-time
* ReSpec version: 22.3.1

## Description of problem
The spec has a warning:

`Missing <dfn> for Performance interface.`

In the spec, we have:

- A  `<dfn>Performance</dfn>` in a section with `data-dfn-for="Performance"  data-link-for="Performance"`.

- A `<dfn>performance</dfn>` in a section with `data-dfn-for="WindowOrWorkerGlobalScope"  data-link-for="WindowOrWorkerGlobalScope"`.

It seems that the two dfns being the same name is causing the problem. In particular, there are two problems:

1.  "Performance" title in section 5 is not orange (but fortunately links to it do work).
2. The link to "Performance" in section 6 in `readonly    attribute Performance performance;` is to the attribute instead of the interface (even if I try to add a data-link-for for "Performance").

Not sure if this is an issue with ReSpec or if there's a way to fix this?

marcoscaceres:
closes #1768

This prevents the attribute's type from using the parent element's `data-link-for`, which was confusing the linker. Types always need to be defined at top level things (i.e., they are not "for" things), so this should work.

### DIFF
--- a/tests/spec/core/webidl-spec.js
+++ b/tests/spec/core/webidl-spec.js
@@ -323,7 +323,7 @@ describe("Core - WebIDL", function() {
   attribute Promise<DOMString> operation;
   // 5.0
   readonly attribute Performance performance;
-};`.trim();
+};`;
     expect(target.textContent).toEqual(text);
     const attrs = [...target.getElementsByClassName("idlAttribute")];
     expect(attrs.length).toEqual(9);
--- a/tests/spec/core/webidl-spec.js
+++ b/tests/spec/core/webidl-spec.js
@@ -336,9 +336,11 @@ describe("Core - WebIDL", function() {
   attribute FrozenArray<DOMString> alist;
   // 4.0
   attribute Promise<DOMString> operation;
+  // 5.0
+  readonly attribute Performance performance;
 };`.trim();
     expect($target.text()).toEqual(text);
-    expect($target.find(".idlAttribute").length).toEqual(8);
+    expect($target.find(".idlAttribute").length).toEqual(9);
     var $at = $target.find(".idlAttribute").first();
     expect($at.find(".idlAttrType").text()).toEqual(" DOMString");
     expect($at.find(".idlAttrName").text()).toEqual("regular");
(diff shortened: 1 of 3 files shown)


marcoscaceres:
Fixes methods too :) Adding a test.

saschanaz:
Does this also fix #1508?

tests/spec/core/webidl-spec.js
```diff
@@ -321,10 +321,12 @@ describe("Core - WebIDL", function() {
   attribute FrozenArray<DOMString> alist;
   // 4.0
   attribute Promise<DOMString> operation;
-};`;
+  // 5.0
+  readonly attribute Performance performance;
+};`.trim();
```
saschanaz: nit: this `trim()` doesn't seem to be needed

marcoscaceres:
> Does this also fix #1508?

I'm not sure, tbh. I'm going to say, "probably", but will need to do some more investigation/testing.

tests/spec/core/webidl-spec.js
```diff
@@ -352,6 +354,22 @@ describe("Core - WebIDL", function() {
         .find(c => c.textContent.includes("alist"))
         .querySelector(".idlAttrName a")
     ).toBeNull();
+
+    const performanceIterfaceLink = Array.from(
```
saschanaz: nit: `iterface` 😆

marcoscaceres: heh, almost was to keep it :)

marcoscaceres:
> Well I was wrong with "case sensitive definitionMap" as solution. Always a simpler way!

Definitely not wrong. I think we should, in the future, treat IDL things as case sensitive. IDL is case sensitive, so it makes sense to do so.

marcoscaceres:
@sidvishnoi, the rebase look good! Thanks for doing it!

PR #1778, opened 2018-07-28
fix(karma.conf.js): grep causes incorrect skips

saschanaz:
Currently only tests including "null" in their title are running because of this.


saschanaz:
Please merge ASAP after review to prevent any confusion.

PR #1779, opened 2018-07-28
refactor: use await for makeRSDoc()

saschanaz:
... and remove deprecated callbacks.

### DIFF
--- a/tests/spec/core/ui-spec.js
+++ b/tests/spec/core/ui-spec.js
@@ -1,48 +1,41 @@
 "use strict";
 
-describe("Core - Ui", function() {
+describe("Core - UI", () => {
   afterAll(flushIframes);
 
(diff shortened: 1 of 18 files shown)


tests/spec/core/ui-spec.js
```diff
@@ -1,48 +1,41 @@
...
-              expect(window.getComputedStyle(menu).display).toEqual("none");
-              done();
-            }, 500);
-          });
-        // give it time to fade out
-      },
-      null,
-      "display: block"
-    );
+        resolve();
```
marcoscaceres: Is this still testing that clicking makes the pill go away?

saschanaz: I think so, as it tests `display` after clicking repeatedly.

PR #1780, opened 2018-07-29
refactor: use getElementById whenever possible

saschanaz:
Maybe someone batch-converted jQuery into querySelector?

### DIFF
--- a/tests/spec/core/markdown-spec.js
+++ b/tests/spec/core/markdown-spec.js
@@ -61,7 +61,7 @@ describe("Core - Markdown", function() {
     ops.config.format = "markdown";
     const doc = await makeRSDoc(ops);
     expect(doc.querySelector("code")).toBeFalsy();
-    expect(doc.getElementById("foo").textContent).toBe("Foo");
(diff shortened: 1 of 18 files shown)


tests/spec/core/markdown-spec.js
```diff
@@ -61,7 +61,7 @@ describe("Core - Markdown", function() {
     ops.config.format = "markdown";
     const doc = await makeRSDoc(ops);
     expect(doc.querySelector("code")).toBeFalsy();
-    expect(doc.querySelector("#foo").textContent === "Foo");
+    expect(doc.getElementById("foo").textContent).toBe("1. Foo");
```
saschanaz: @marcoscaceres, is this expected?

marcoscaceres: It looks correct... which is odd, considering the old test.

saschanaz: I think the old test simply didn't work.

marcoscaceres: Ah, right. It just “expect()”s nothing.

PR #1781, opened 2018-07-29
refactor: use const and arrow functions

saschanaz:
Why are those two eslintrc files in different formats, one in JSON and one in YAML?


marcoscaceres:
No good reason - just me being sloppy 😰

saschanaz:
>No good reason

Which should be preferred? Personally I'll choose YAML as we already have .travis.yml.

saschanaz:
And BTW, are .jshintrc/.jscsrc needed? I don't think so.

PR #1787, opened 2018-07-30
refactor(core/examples): refactor and remove jQuery

sidvishnoi:


js/core/css/examples.css
```diff
@@ -3,8 +3,11 @@ div.example-title {
     min-width: 7.5em;
     color: #b9ab2d;
 }
-div.example-title span {
-    text-transform: uppercase;
+div.example-title span:nth-child(1) {
+  text-transform: uppercase;
+}
+div.example-title span:nth-child(2) {
```
saschanaz: This looks redundant... do we need this?

src/core/examples.js
```diff
@@ -6,69 +6,71 @@
...
+  document.head.insertBefore(
+    hyperHTML`<style>${css}</style>`,
+    document.querySelector("link")
+  );
+
+  let number = 0;
+  examples.forEach(example => {
+    const illegal = example.classList.contains("illegal-example");
+    const report = { number, illegal };
+    if (example.localName === "aside") {
```
saschanaz: TIL `localName`

src/core/examples.js
```diff
@@ -6,69 +6,71 @@
...
+      ++number;
+      const title = makeTitle(conf, example, number, report);
+      example.insertBefore(title, example.firstChild);
+      pub("example", report);
+    } else {
+      const inAside = !!example.closest("aside");
+      if (!inAside) ++number;
+
+      // reindent
```
saschanaz: So this reindentation algorithm is duplicated across files including `webidl.js`. I think I have seen another one elsewhere but I forgot where it was.

js/core/css/examples.css
sidvishnoi: Had to keep this because of base.css (I mainly translated this part so things work as they were working before)
https://github.com/w3c/tr-design/blob/f5285732ffcf13a7379c721758d7ea8baf8bf343/src/base.css#L659,L660

src/core/examples.js
sidvishnoi: Oh yes. We should move it to core/utils then, right? BTW, I found it only in webidl and here (no other place).

Also, I think the re-indent algorithm is slightly different (`leastIndent` in webidl vs indentation of `lines[0]` in examples). Which one should we use?

js/core/css/examples.css
saschanaz: I'm talking about the second `span:nth-child(2)` thing, I thank we can skip it.

For `.marker`, it seems tr-design prefers `.marker` over `.example-title`.

src/core/examples.js
saschanaz: I personally prefer `leastIndent` but both probably will work.

js/core/css/examples.css
sidvishnoi: `.marker` add `text-transform: uppercase`. So, that way, here `span:nth-child(1)` is redundant. So we can skip it but we need `span:nth-child(2)`. Earlier we added inline style for that.

saschanaz: >Earlier we added inline style for that.

That's because we didn't have `nth-child(1)`. Now we have it, so we don't need explicit uppercasing. (It should implicitly get `text-transform: none`.)

saschanaz: >.marker adds text-transform: uppercase.

Ah, I got it. Ignore the previous comment, sorry 😅

src/core/examples.js
```diff
@@ -6,69 +6,71 @@
...
-  }
-  $tit.addClass("marker");
-  return $tit;
+function makeTitle(conf, elem, num, report) {
+  report.title = elem.title;
+  if (report.title) elem.removeAttribute("title");
+  const number = num > 0 ? " " + num : "";
+
+  return hyperHTML`
+  <div class='marker example-title'><span>${conf.l10n.example}${number}</span>${
```
saschanaz: So `example-title` certainly looks redundant. How about slightly changing its purpose:

```css
.example-title {
  text-transform: none;
}
```

```html
<div class='marker'><span>${conf.l10n.example}${number}</span>${
  report.title ? hyperHTML`<span class='example-title'>: ${report.title}</span>` : ""
}</div>
```

sidvishnoi: We can inline the `text-transform: none;` though.

src/core/utils.js
```diff
@@ -286,6 +286,25 @@ export function normalizePadding(text = "") {
   return result;
 }
 
+/**
+ * Removes common indents across the IDL texts,
+ * so that indentation inside <pre> won't affect the rendered result.
+ * @param {string} text IDL text
+ */
+export function unindentMarkup(text) {
```
sidvishnoi: `unindentMarkup` as name doesn't seem very right. move to `reindentMarkup`?

saschanaz: IMO we don't need `Markup` to make it general, so maybe just `reindent`, `reduceIndent` or anything proper.

I used `unindent` because it basically does `Shift+Tab` 😁

sidvishnoi: `rebaseIndent` :laughing:

sidvishnoi:
Thank you @saschanaz for review and fixing the tests :slightly_smiling_face:

PR #1796, opened 2018-08-07
fix(core/webidl): only set pre id when needed

marcoscaceres:
We were accidentally setting the id of the `<pre>` element to `""`, which is causing the HTML validator to say the document is invalid (empty id values are invalid, apparently... which makes sense, I guess).

### DIFF
--- a/src/core/webidl.js
+++ b/src/core/webidl.js
@@ -759,7 +759,7 @@ export function run(conf) {
     }
     linkDefinitions(parse, conf.definitionMap, "", idlElement);
     const newElement = makeMarkup(parse);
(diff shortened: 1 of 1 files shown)


src/core/webidl.js
```diff
@@ -759,7 +759,7 @@ export function run(conf) {
     }
     linkDefinitions(parse, conf.definitionMap, "", idlElement);
     const newElement = makeMarkup(parse);
-    newElement.setAttribute("id", idlElement.id);
+    if (idlElement.id) newElement.setAttribute("id", idlElement.id);
```
saschanaz: nit: We also can use `newElement.id` directly.

PR #1806, opened 2018-08-20
refactor(core/jquery-enhanced): migrate $renameElement to vanilla js

sidvishnoi:
As a part of https://github.com/w3c/respec/issues/1699

### DIFF
--- a/src/core/utils.js
+++ b/src/core/utils.js
@@ -749,15 +749,8 @@ export function renameElement(elem, newName) {
 
   // copy attributes
   for (let i = 0, n = elem.attributes.length; i < n; i++) {
-    const attr = elem.attributes[i];
-    try {
-      newElement.setAttributeNS(attr.namespaceURI, attr.name, attr.value);
-    } catch (err) {
-      let msg = "Your HTML markup is malformed. Error in: \n";
-      msg += "```HTML\n" + elem.outerHTML + "\n```";
-      pub("error", msg);
-      break; // no point in continuing with this element
-    }
+    const { name, value } = elem.attributes[i];
+    newElement.setAttribute(name, value);
   }
 
   // copy child nodes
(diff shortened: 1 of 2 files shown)


src/core/utils.js
```diff
@@ -737,3 +737,35 @@ export function getLinkTargets(elem) {
...
+  const newElement = elem.ownerDocument.createElement(newName);
+
+  // copy attributes
+  for (let i = 0, n = elem.attributes.length; i < n; i++) {
+    const attr = elem.attributes[i];
+    try {
+      newElement.setAttributeNS(attr.namespaceURI, attr.name, attr.value);
+    } catch (err) {
+      let msg = "Your HTML markup is malformed. Error in: \n";
+      msg += "```HTML\n" + this.outerHTML + "\n```";
```
sidvishnoi: oops. s/this/elem/

marcoscaceres: oops. Yeah, please fix that.

src/core/utils.js
```diff
@@ -737,3 +737,35 @@ export function getLinkTargets(elem) {
...
+ * @returns {Element} new renamed element
+ */
+export function renameElement(elem, newName) {
+  const newElement = elem.ownerDocument.createElement(newName);
+
+  // copy attributes
+  for (let i = 0, n = elem.attributes.length; i < n; i++) {
+    const attr = elem.attributes[i];
+    try {
+      newElement.setAttributeNS(attr.namespaceURI, attr.name, attr.value);
```
marcoscaceres: Nit: I know this is copy/paste... but we can just jus `setAttribute`. We don't support XHTML anymore.

sidvishnoi: Removing this exception handling as we have `renameElement` as an internal function, and we rename only specific elements. So, we never get this exception :crossed_fingers:.

PR #1807, opened 2018-08-20
refactor(core/fix-headers): remove jQuery dependencies

sidvishnoi:
TODO: remove jQuery from tests

### DIFF
--- a/src/core/fix-headers.js
+++ b/src/core/fix-headers.js
@@ -2,17 +2,25 @@
 // Make sure that all h1-h6 headers (that are first direct children of sections) are actually
 // numbered at the right depth level. This makes it possible to just use any of them (conventionally
 // h2) with the knowledge that the proper depth level will be used
+import { renameElement } from "core/utils";
+
 export const name = "core/fix-headers";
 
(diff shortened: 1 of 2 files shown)


src/core/fix-headers.js
```diff
@@ -2,17 +2,25 @@
...
-  $secs.each(function() {
-    let depth = $(this).parents("section").length + 1;
-    if (depth > 6) depth = 6;
-    const h = "h" + depth;
-    if (this.localName.toLowerCase() !== h) $(this).renameElement(h);
-  });
-  cb();
+export function run() {
+  [...document.querySelectorAll("section:not(.introductory)")]
+    .map(sec => sec.querySelector("h1, h2, h3, h4, h5, h6"))
```
marcoscaceres: I recall `:first` being important.

sidvishnoi: `querySelector` will pick first one (either of h1, h2 ...) in every section here, right?

marcoscaceres: Ah, good point 🧐👍

PR #1810, opened 2018-08-23
Give examples IDs based on example number and title

gkellogg:
so that they may …be referenced.

I've received requests to add this to specs, but it makes the most sense to have ReSpec do this for you. Potentially, it could be included in the permalink logic as well.


gkellogg:
Don't really understand the unrelated Karma error in Travis.

marcoscaceres:
@gkellogg, I updated the underlying branch, as we had a regression that was causing Travis to fail.

sidvishnoi:
https://github.com/w3c/respec/blob/dad1eda2c985c4282377dda2011fcb9383602bd3/src/core/examples.js#L43-L47

maybe we need to `addId()` here also.

marcoscaceres:
Yeah, those asides could probably also use an id.

marcoscaceres:
I guess we want:
```
if (!example.hasAttribute("id")) addId(example, "ex-" + number, title);
```

marcoscaceres:
+ respec@22.6.0

PR #1817, opened 2018-09-05
fix(core/utils): generate ids without diacritical marks

### LINKED ISSUE #1813: Autogenerated (section) IDs omit unicode characters
## Important info

* ReSpec version: 22.5

## Description of problem

Autogenerated IDs of sections (and other parts of the document as well) based on headings omit unicode characters, creating nonsensical words in URLs.

I would expect the IDs to keep the words intact (i.e. do only lowercase, if that is necessary)

Optional, steps to reproduce:
1. Create a section with heading containing unicode characters, e.g. `Zkracování` and no `id` attribute
2. Observe that the generated id is e.g. `zkracov-n-` instead of `zkracování`

marcoscaceres:
* closes #1813

### DIFF
--- a/src/core/utils.js
+++ b/src/core/utils.js
@@ -599,7 +599,10 @@ export function addId(elem, pfx = "", txt = "", noLC = false) {
   }
   let id = noLC ? txt : txt.toLowerCase();
   id = id
-    .replace(/[\W]+/gim, "-")
+    .trim()
+    .normalize("NFD")
+    .replace(/[\s]+/gim, "-")
+    .replace(/[\u0300-\u036f]/g, "")
     .replace(/^-+/, "")
     .replace(/-+$/, "");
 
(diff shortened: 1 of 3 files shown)


src/core/utils.js
```diff
@@ -599,7 +599,10 @@ export function addId(elem, pfx = "", txt = "", noLC = false) {
   }
   let id = noLC ? txt : txt.toLowerCase();
   id = id
-    .replace(/[\W]+/gim, "-")
+    .trim()
+    .normalize("NFD")
+    .replace(/[\s]+/gim, "-")
```
sidvishnoi: `\W+` works as we're expecting it to work.
Any reason to switch to `\s`?

marcoscaceres:
Oh, I had issues with \W, but I guess maybe it was the ordering... dunno 🤷‍♂️ happy if the tests are passing

marcoscaceres:
> Nice. Learnt about string.normalize today 

Yeah, me too. Did you read the Stack Overflow link from the original bug?

sidvishnoi:
> Did you read the Stack Overflow link from the original bug?
Yes. And also a GitHub gist (Java based)

PR #1818, opened 2018-09-05
feat(core/caniuse): add more browser keys

marcoscaceres:


marcoscaceres:
Different problem now is that the boxes don't look great because some of the labels are a bit longer.

sidvishnoi:
> Different problem now is that the boxes don't look great because some of the labels are a bit longer.

Can add `width: 100%` on the `ul`.
But more consistent label names would be nicer.

marcoscaceres:
Actually, it doesn’t look too bad. Just had another look.

PR #1823, opened 2018-09-15
feat(core/examples): self-link examples

marcoscaceres:


tests/spec/core/examples-spec.js
```diff
@@ -48,4 +48,19 @@ describe("Core — Examples", () => {
...
+      <aside class="example" title="pass"></aside>
+    `;
+    const ops = makeStandardOps({}, body);
+    const doc = await makeRSDoc(ops);
+    const exampleLinks = doc.querySelectorAll("aside.example a.self-link");
+    expect(exampleLinks.length).toBe(3);
+    const [example1, example2, example3] = exampleLinks;
+    expect(example1.getAttribute("href")).toBe("#ex-1-example-1");
+    expect(example2.getAttribute("href")).toBe("#pass");
+    expect(example3.getAttribute("href")).toBe("#ex-2-example-3'");
```
sidvishnoi: should be `"#ex-3-pass"` here as we are passing `example.title` as `title` and:
https://github.com/w3c/respec/blob/cf1d2e930f9a32a81ed248fde16b0cb3672bbd04/src/core/utils.js#L597-L600

PR #1824, opened 2018-09-15
chore(package): update dep, add Sid as contributor

marcoscaceres:


package.json
```diff
@@ -58,7 +59,7 @@
     "requirejs": "^2.3.6",
     "text": "github:requirejs/text#d04de4ffd7bf5ba6cb80cdca2d40d4f6f52a1b1f",
     "uglify-es": "3.3.7",
-    "url-search-params": "^1.0.0",
+    "url-search-params": "^1.1.0",
```
sidvishnoi: Do we still need this polyfill?

marcoscaceres: Ah, nope! Yay https://caniuse.com/#feat=urlsearchparams

PR #1832, opened 2018-09-26
fix(core/example): duplicate ids

### LINKED ISSUE #1830: Duplicate ids are generated for sections with the same title in Markdown docs
## Important info

* URL to affected spec: (not live in production at the moment)
* ReSpec version: 23.0.2 based build of Geonovum, but generic issue

## Description of problem
In ReSpec documents using Markdown duplicate ids are generated on the header elements if sections have the same title. In HTML docs this does not occur.

But what I was expecting?
I expected unique ids, similar to an HTML doc where the same titles are used

Optional, steps to reproduce:
1. see the gist at https://gist.github.com/thijsbrentjens/43489b534ef9305881805076bcc35519
1. see the generated source or see snapshot.html
 1. there are 2 elements h3 with id="sectiontitle"

marcoscaceres:
closes #1830


src/core/examples.js
```diff
@@ -61,16 +61,20 @@ export function run(conf) {
 
       // wrap
       example.classList.remove("example", "illegal-example");
+      // relocate the id to the div
+      const id = example.id ? example.id : null;
+      if (id !== "") {
```
sidvishnoi: nit: assigning `null` but comparing to `""`

marcoscaceres: oh, good catch!

src/core/examples.js
```diff
@@ -63,7 +63,7 @@ export function run(conf) {
       example.classList.remove("example", "illegal-example");
       // relocate the id to the div
       const id = example.id ? example.id : null;
-      if (id !== "") {
+      if (id !== "" && id !== null) {
```
sidvishnoi: `""` isn't really needed I guess.  `example.id ? ... ` will be `true` only if `id` is non-empty

marcoscaceres: yeah, changed it.

marcoscaceres:
Thanks for the speedy review, @sidvishnoi!

PR #1833, opened 2018-09-27
fix(core/utils): renameElement throws unexpected error

marcoscaceres:


marcoscaceres:
The `do {} while` is causing the web manifest spec to throw when it hits an element with not inner content. So: `<a id="foo"><a>` throws.

src/core/utils.js
```diff
@@ -751,19 +751,17 @@ export function getLinkTargets(elem) {
 export function renameElement(elem, newName) {
   if (elem.localName === newName) return elem;
   const newElement = elem.ownerDocument.createElement(newName);
-
+  if (!elem.isConnected) throw new Error("Node must be attached to document.");
```
sidvishnoi: This seems to be not supported in Edge yet.
image

sidvishnoi: Can probably do `document.contains(elem)`

src/core/utils.js
```diff
@@ -751,19 +751,17 @@ export function getLinkTargets(elem) {
...
   }
-
   // copy child nodes
-  do {
+  while (elem.firstChild) {
     newElement.appendChild(elem.firstChild);
-  } while (elem.firstChild);
-
+  }
   // TODO: replace with ChildNode.replaceWith, when available in Safari
```
sidvishnoi: Nit: remove TODO

src/core/utils.js
marcoscaceres: good suggestion!

src/core/utils.js
```diff
@@ -751,19 +751,16 @@ export function getLinkTargets(elem) {
 export function renameElement(elem, newName) {
   if (elem.localName === newName) return elem;
   const newElement = elem.ownerDocument.createElement(newName);
-
+  if (document.contains(elem)) throw new Error("Node not in document.");
```
sidvishnoi: Typo: should be `!document.contains` 🙈

marcoscaceres: lol, yeah.

src/core/utils.js
```diff
@@ -751,19 +751,16 @@ export function getLinkTargets(elem) {
...
-  for (let i = 0, n = elem.attributes.length; i < n; i++) {
-    const { name, value } = elem.attributes[i];
+  for (const attribute of [...elem.attributes]) {
+    const { name, value } = attribute;
     newElement.setAttribute(name, value);
   }
-
   // copy child nodes
-  do {
+  while (elem.firstChild) {
```
sidvishnoi: Off-topic: Why does this loop - `while (elem.firstChild)` work? should it not be infinite loop? 👨‍🎓

marcoscaceres: ah, because `newElement.appendChild(elem.firstChild);` steals `elem.firstChild `, so eventually `elem.firstChild` becomes `null`.

sidvishnoi: steals? wow. never knew this. will read about it.
(gotta study for exam first 😆)

marcoscaceres: yeah, it relocates it in the tree. Approve before you run off!!!!

sidvishnoi: Test is failing though.

marcoscaceres: Tests! ppfffff! SHIP IT!

marcoscaceres: Oh, I'd assumed that nodes would have to be connected for it to work. But seems they don't, so removing.

marcoscaceres: ok, fixed... tests should pass now.

marcoscaceres:
Thanks for taking time to help me with this one! Great feedback as always.

PR #1835, opened 2018-10-01
feat(respec-worker): dynamically load highlight.js languages

marcoscaceres:
Adds the ability for users to dynamically load additional highlight languages into ReSpec's worker.  

Usage is, for example:

```JS 
async function loadSolidity() {
  const worker = await new Promise(resolve => {
    require(["core/worker"], ({ worker }) => resolve(worker));
  });
  const action = "highlight-load-lang";
  const langURL =
    "https://cdn.rawgit.com/pospi/highlightjs-solidity/b058d0f9d39065fde7303ab5b9a1005912cc3b6a/solidity.js";
  const propName = "hljsDefineSolidity";
  const lang = "solidity";
  worker.postMessage({ action, langURL, propName, lang });
  return new Promise(resolve => {
    worker.addEventListener("message", function listener({ data }) {
      const { action: responseAction, lang: responseLang } = data;
      if (responseAction === action && responseLang === lang) {
        worker.removeEventListener("message", listener);
        resolve();
      }
    });
  });
}
```

In ReSpec config:
```JS
 var respecConfig = {
        preProcess: [loadSolidity],
        /// other stuff 
 }
```


sidvishnoi:
Wow. Code is nice! I like it! 👌 💯  
It might be a good idea to add a `core/hljs-language-loader` plugin in near future, instead of `preprocess` maybe.

marcoscaceres:
> It might be a good idea to add a core/hljs-language-loader plugin in near future, instead of preprocess maybe.

Thought the same, but this is only come up once in the last few years. If a language gets popular enough, we can just bundle it in (as we do right now).... having said that, we could look at unbundling the current languages and loading them as needed using this mechanism (but in the future... someday... maybe 🤣 )

sidvishnoi:
> but in the future... someday... maybe

😄 

This feature needs to be documented though..

marcoscaceres:
> This feature needs to be documented though..

Agree. I'll do that.

chaals:
> > This feature needs to be documented though..
> 
> Agree. I'll do that.

Actually,  I had a go at it in https://github.com/w3c/respec/wiki/pre-and-code-elements

marcoscaceres:
Thanks for documenting it, @chaals. Looks great. Thanks also for adding the details about what languages we support out of the box!

karma.conf.js
```diff
@@ -27,6 +27,11 @@ module.exports = function(config) {
     // list of files / patterns to load in the browser
     files: [
       "js/deps/jquery.js",
+      {
```
sidvishnoi: can we merge these patterns?

tests/spec/core/highlight-spec.js
```diff
@@ -2,6 +2,12 @@
 describe("Core — Highlight", () => {
   afterAll(flushIframes);
 
+  it("highlights remote languages not bundled by default with ReSpec", async () => {
+    const doc = await makeRSDoc({}, "spec/core/highlight.html");
+    const span = doc.querySelector("pre.testlang span[class^=hljs]");
+    expect(span.innerText).toBe("funkyFunction");
```
sidvishnoi: is this test enough to ensure it works? 😕

marcoscaceres: I think so, as the matching `span` proves that it applied some highlight.

karma.conf.js
marcoscaceres: Ooops, didn’t mean to resolve that. The jQuery one gets included.

tests/spec/core/highlight-spec.js
sidvishnoi: My bad. IDK what I read!

karma.conf.js
sidvishnoi: I meant the ones below it. the `tests/` patterns.

marcoscaceres:
Gotta get up for a meeting in 5 hours 😭 but please feel free to merge after TravisCI is happy. Hopefully Codacy wakes up, but if not 🤷‍♂️

PR #1839, opened 2018-10-01
feat(core/inlines): change the way inline references are used

### LINKED ISSUE #1754: Change the way references are defined as informative or normative
The following modified syntax shall be used for defined whether an inline spec reference is informative or normative:
- `[[!spec]]`: normative.
- `[[?spec]]`: informative.
- `[[spec]]`: decided based on parent. If defined in a non-normative section, then informative, else normative.

sidvishnoi:
Closes #1754. I hope I've taken required cases into consideration.


sidvishnoi:
Looks like I forgot to update `biblio-spec` 🤷‍♂️

src/core/inlines.js
```diff
@@ -90,19 +90,26 @@ export function run(conf) {
...
-            else conf.informativeReferences.add(ref);
+
             df.appendChild(document.createTextNode("["));
             const refHref = `#bib-${ref.toLowerCase()}`;
-            df.appendChild(
+            const cite = df.appendChild(
               hyperHTML`<cite><a class="bibref" href="${refHref}">${ref}</a></cite>`
             );
+            if (illegal) {
+              const errorMsg = `Error: Normative reference \`${matched}\` in a non-normative section.`;
```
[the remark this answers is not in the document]
sidvishnoi: Fix as? It should be normative?
A linter rule would be unnecessary I think. Are there any reasons people might want to skip this warning?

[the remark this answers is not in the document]
marcoscaceres: > Fix as? It should be normative?

Yeah. 

> A linter rule would be unnecessary I think. Are there any reasons people might want to skip this warning?

It's more of a question about existing specs doing the wrong thing by accident. I don't want us to be bothering Editors with these - ReSpec should just do the right thing.

src/core/inlines.js
```diff
@@ -124,3 +124,31 @@ export function run(conf) {
...
+
+function isInformative(ref, parentNode) {
+  const informSelectors = ".informative, .note, figure, .example, .issue";
+  const closestInformative = parentNode.closest(informSelectors);
+
+  let informative = false;
+  if (closestInformative) {
+    // check if parent is not normative
+    if (!parentNode.closest(".normative")) {
+      informative = true;
```
marcoscaceres: The two statements inside the ifs could be combined into:

informative = stament1 || statement2; 

Wdyt? In either case, the second if is getting run even if informative is already true.

sidvishnoi: Yes. I was thinking it was cleaner (I do not enjoy multiple if statements. but looks like nested if's are also a no-go)

marcoscaceres: I guess the main thing is not to run the second if statement. It could be short circuited if the first if statement sets `informative` to true.

sidvishnoi: Yes. I missed this part earlier. Fix now

marcoscaceres:
These three PRs are shaping up to make a pretty awesome release!

sidvishnoi:
> These three PRs are shaping up to make a pretty awesome release!

Lets hope this one doesn't break the internet.

marcoscaceres:
“You can’t weave the web without breaking a few specs”, is what I always say.

PR #1841, opened 2018-10-02
feat(core/data-cite): support contextual citing

marcoscaceres:
Adds contextual citation to `data-cite` by reusing the inline contextual citation.

### DIFF
--- a/src/core/data-cite.js
+++ b/src/core/data-cite.js
@@ -66,15 +66,24 @@ function cleanElement(elem) {
     .forEach(attrName => elem.removeAttribute(attrName));
 }
 
+function makeComponentFinder(component) {
+  return key => {
+    const position = key.search(component);
+    if (position !== -1) {
+      return key.substring(position);
+    }
+    return "";
+  };
+}
+
 function citeDetailsConverter(conf) {
+  const findFrag = makeComponentFinder("#");
+  const findPath = makeComponentFinder("/");
   return function toCiteDetails(elem) {
     const { dataset } = elem;
-    let { cite: key, citeFrag: frag, citePath: path } = dataset;
-    let isNormative = key.startsWith("!");
-    const pathPosition = key.search("/");
-    const fragPosition = key.search("#");
+    const { cite: rawKey, citeFrag, citePath } = dataset;
     // The key is a fragment, resolve using the shortName as key
-    if (key.startsWith("#") && !frag) {
+    if (rawKey.startsWith("#") && !citeFrag) {
       // Closes data-cite not starting with "#"
       const closest = elem.parentElement.closest(
         `[data-cite]:not([data-cite^="#"])`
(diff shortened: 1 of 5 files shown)


marcoscaceres:
@sidvishnoi, this is just to set direction. Let know what you think. If you agree with the direction, I'll add/fix tests.

src/core/data-cite.js
```diff
@@ -70,7 +70,7 @@ function citeDetailsConverter(conf) {
   return function toCiteDetails(elem) {
     const { dataset } = elem;
     let { cite: key, citeFrag: frag, citePath: path } = dataset;
-    const isNormative = key.startsWith("!");
+    let isNormative = key.startsWith("!");
```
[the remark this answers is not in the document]
marcoscaceres: oh, because the actual `key` hasn't actually been established yet. At this point, key looks like `!foo`, which is not valid reference key.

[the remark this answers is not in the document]
sidvishnoi: Maybe I didn't get it right. But `!foo` looks valid. See https://github.com/w3c/respec/pull/1841/files/913f16dd20fa5f8846b03bd3aead4c6e858f3202#diff-47300e7ecd8989b6246221de88fc9a3cR780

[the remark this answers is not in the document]
marcoscaceres: Oh, you are right. Need to check if there is stuff on the end “!foo/whatever#bar”. Maybe doesn’t matter?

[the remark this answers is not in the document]
sidvishnoi: We already have them separated 🙂 
``` js
let { cite: key, citeFrag: frag, citePath: path }
```

src/core/data-cite.js
```diff
@@ -66,45 +66,40 @@ function cleanElement(elem) {
     .forEach(attrName => elem.removeAttribute(attrName));
 }
 
+function makeComponentFinder(component) {
```
sidvishnoi: move it to utils?

src/core/data-cite.js
```diff
@@ -66,45 +66,40 @@ function cleanElement(elem) {
...
-    if (path && path.startsWith("/")) {
-      path = path.substr(1);
-    }
-    return { key, isNormative, frag, path };
+    const frag = citeFrag ? "#" + citeFrag : findFrag(rawKey);
+    const path = citePath || findPath(rawKey).split("#")[0]; // path is always before "#"
+    const { type } = refTypeFromContext(rawKey, elem);
+    const isNormative = type === "normative";
+    // key is before "/" and "#" but after "!" or "?" (e.g., ?key/path#frag)
+    const key = rawKey.split(/[/|#]/)[0].substring(/^[?|!]/.test(rawKey));
```
sidvishnoi: use regex captures groups directly?

src/core/inlines.js
```diff
@@ -90,20 +90,30 @@ export function run(conf) {
...
+              cite.classList.add("respec-offending-element");
+              const msg =
+                "Normative references in informative sections are not allowed. " +
+                `Remove '!' from the start of the reference \`[[!${ref}]]\`. `;
+              pub(
+                "warn",
+                msg + "See developer console to find offending element."
+              );
+              cite.title = msg;
+              console.log(msg, cite);
```
sidvishnoi: `console.warn`?

src/core/data-cite.js
marcoscaceres: It's a bit specific for the micro syntax in data-cite. Might leave it here for now.

src/core/data-cite.js
marcoscaceres: Um... I tried... I... um... couldn't work it out 😭 RegEx is hard. Could you work out what it would be?

marcoscaceres: We can do a new PR.

sidvishnoi: Sure. I will.

PR #1848, opened 2018-10-08
feat(core/inlines): make variables context aware

marcoscaceres:
This is a work in progress


marcoscaceres:
@sidvishnoi would like to chat about this. I have some ideas and could use your help.

src/core/inline-idl-parser.js
```diff
@@ -9,75 +9,180 @@
...
-    result.method = splitted.pop();
-    result.args = result.method.match(methodRegex)[1].split(/,\s*/);
-  }
-  if (splitted.length > 1 && !result.method) {
-    result.attribute = splitted.pop();
+  const methodRegex = /(\w+)\((.*)\)$/;
+  const slotRegex = /$\[\[(\w+)\]\]$/;
+  const attributeRegex = /^(\w+)$/;
+  const enumRegex = /^(\w+)\["([\w ]+)"\]$/;
+  const tokens = str.split(".");
```
sidvishnoi: Lets split at `/\b\.\b/` to add support for variadic methods.

``` js
"foo.bar.baz(foo, ...bar)".split(".")
// ["foo", "bar", "baz(foo, ", "", "", "bar)"]
"foo.bar.baz(foo, ...bar)".split(/\b\.\b/)
// ["foo", "bar", "baz(foo, ...bar)"]
```

marcoscaceres: Unfortunately, that one doesn't work well for:
"foo.bar.[[test]].baz(foo, ...bar)".split(/\b\.\b/)

We might need to split on method, then split the rest? Or so some kind of fancy look ahead, the split behind.

sidvishnoi:
```js
const splitRegex = /(?<=\]\]|\b)\./
```
😎 🤓

sidvishnoi:
@marcoscaceres Removing `{{{ Dictionary["member"] }}}` syntax would work fine for referencing local IDL as we will be matching it up with local `pre.idl` (maybe a Map), but it would break the external references (in core/xref) as we would have no way to tell the xref API that we're looking for a `dict-member`. Or am I missing something?

marcoscaceres:
> @marcoscaceres Removing {{{ Dictionary["member"] }}} syntax would work fine for referencing local IDL as we will be matching it up with local pre.idl (maybe a Map), but it would break the external references (in core/xref) as we would have no way to tell the xref API that we're looking for a dict-member. Or am I missing something?

All IDL containers are uniquely named, so the distinction between dictionary and interface is kinda an optimization. So, if we make the the micro-syntax for both the same, then it only becomes a user detail that one is a dictionary and another is an interface. 

That is:

```
 {{{ DictionaryWhatever }}}
 {{{ DictionaryWhatever.memberWhatever }}}
```

src/core/xref.js
```diff
@@ -101,7 +101,17 @@ function createXrefMap(elems) {
       specs.push(...refs);
     }
 
-    const types = [isIDL ? elem.dataset.xrefType || "_IDL_" : "_CONCEPT_"];
+    const types = [];
+    if (isIDL) {
+      if (elem.dataset.xrefType) {
+        types.push(...elem.dataset.xrefType.split("|"));
```
sidvishnoi: this will let us support treating attribute and dict-member the same

tests/spec/core/xref-spec.js
```diff
@@ -547,7 +547,7 @@ describe("Core — xref", () => {
       expect(link2.href).toEqual(
         expectedLinks.get("PermissionStatus.[[query]]")
       );
-      expect(link2.textContent).toEqual("[[query]]");
+      expect(link2.textContent).toEqual("query");
```
sidvishnoi: need a thought here

src/core/inline-idl-parser.js
```diff
@@ -9,77 +9,196 @@
...
+  const { parent, identifier } = details;
+  const { idlType } = parent;
+  const html = hyperHTML`.<a
+      class="respec-idl-xref"
+      data-xref-type="attribute|dict-member"
+      data-link-for="${idlType}">${identifier}</a>`;
+  return html;
+}
+
+// Method: .identifier(arg1, arg2, ...)
```
sidvishnoi:
```suggestion
// Method: .identifier(arg1, arg2, ...), identifier(arg1, arg2, ...)
```

src/core/inline-idl-parser.js
```diff
@@ -9,77 +9,196 @@
...
+    class="respec-idl-xref"
+    data-xref-type="${type}"
+    data-link-for="${idlType}"
+    data-lt="${searchText}"
+    >${identifier}</a>(${[argsText]})`;
+  return html;
+}
+
+// Enum: "enum value"
+function renderEnum(details) {
```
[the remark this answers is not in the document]
marcoscaceres: hmm... we should support this.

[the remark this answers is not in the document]
marcoscaceres: {{{ "some value" }}}

marcoscaceres:
I think we should beta test this with Payment Request a bit.

src/core/inline-idl-parser.js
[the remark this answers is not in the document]
sidvishnoi: `const enumRegex = /^(\w+)\["([\w ]+)"\]$/;` seems to be matching something else

src/core/inline-idl-parser.js
```diff
@@ -9,77 +9,215 @@
...
 function parseInlineIDL(str) {
-  const result = Object.create(null);
-  const splitted = str.split(idlSplitRegex);
-  if (methodRegex.test(splitted[splitted.length - 1])) {
-    result.method = splitted.pop();
-    result.args = result.method.match(methodRegex)[1].split(/,\s*/);
-  }
-  if (splitted.length > 1 && !result.method) {
-    result.attribute = splitted.pop();
+  const methodRegex = /(\w+)\((.*)\)$/;
```
sidvishnoi: we should move these regex outside this function maybe

PR #1850, opened 2018-10-10
feat(core/data-type): auto assign IDL types to dfns

marcoscaceres:
Given:

```HTML
<pre class=idl>
interface Foo {
   attribute Bar bar;
   DOMString toString();
};
</pre>

<dfn>Foo</dfn>
<dfn data-dfn-for="Foo">bar</dfn>
<dfn data-dfn-for="Foo">toString()</dfn>
```

The `<dfn data-dfn-for="Foo">bar</dfn>` gains:

 * `data-type`: `Bar` 
 * `data-idl`: `attribute` 

The `<dfn data-dfn-for="Foo">bar</dfn>` gains:

 * `data-type`: `DOMString`, which is the return type.
 * `data-idl`: `operation` 

Having these bits of information allow us to do interesting things - like type checking, chained linking, cute pop-ups showing the type, etc.

### DIFF
--- a/src/core/webidl.js
+++ b/src/core/webidl.js
@@ -753,9 +753,19 @@ function findDfn(defn, parent, name, definitionMap, type, idlElem) {
       name.replace(/[()]/g, "").replace(/\s/g, "-");
     dfn.id = id;
   }
-  dfn.dataset.idl = "";
+  dfn.dataset.idl = type === "enum-value" ? type : defn.type;
   dfn.dataset.title = dfn.textContent;
   dfn.dataset.dfnFor = parent;
+  // Derive the data-type for dictionary members, interface attributes,
+  // and methods
+  switch (defn.type) {
+    case "operation":
+    case "attribute":
+    case "field":
+      dfn.dataset.type = getDataType(defn);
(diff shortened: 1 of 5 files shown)


src/core/webidl.js
```diff
@@ -753,9 +753,19 @@ function findDfn(defn, parent, name, definitionMap, type, idlElem) {
       name.replace(/[()]/g, "").replace(/\s/g, "-");
     dfn.id = id;
   }
-  dfn.dataset.idl = "";
+  dfn.dataset.idl = type === "enum-value" ? type : defn.type;
```
sidvishnoi: Why `"enum-value"` has a special case?

marcoscaceres: I think WebIDL2.js might (Not 💯 sure) be giving us the wrong type.

marcoscaceres: Will check, and leave a comment to fix if it is.

marcoscaceres: Confirmed. Webidl2.js has type “string” for enum values. Need to fix that, as its wrong.

PR #1857, opened 2018-10-14
fix(core/data-cite): always resolve path relative to reference URL

### LINKED ISSUE #1856: data-cite: Reference Subpage Resolution
## Important info

* URL to affected spec: http://w3c.github.io/sdw/proposals/geotagging/webvmt/#overview
* ReSpec version: 23.3.0

## Description of problem

I'm trying to use <code>data-cite</code> to link to a particular section of a subpage using <code>&lt;a&gt;</code>, as [described in the Respec Wiki](https://github.com/w3c/respec/wiki/data--cite#example-of-usage-with-a-element), e.g. <code>&lt;a data-cite="HTML51/subpage.html#section&gt;text&lt;/a&gt;</code>.

Either I've misunderstood how to use this or there's a bug, and I'd be appreciate help in either case.

1. Links without a subpage seem to correctly resolve, e.g. <code>&lt;a data-cite="HTML51&gt;text&lt;/a&gt;</code> correctly resolves to https://www.w3.org/TR/html51/.

1. Links with a subpage seem to incorrectly drop the path, so <code>&lt;a data-cite="HTML51/subpage.html#section"&gt;text&lt;/a&gt;</code> incorrectly resolves to https://www.w3.org/subpage.html#section instead of the expected https://www.w3.org/TR/html51/subpage.html#section.

1. There are examples of both of the above in section 4.1 at http://w3c.github.io/sdw/proposals/geotagging/webvmt/#overview.

Issue #1180 appears to give a nearly identical example, but I'm unsure of its eventual outcome. Please advise.

marcoscaceres:
* closes #1856

### DIFF
--- a/src/core/data-cite.js
+++ b/src/core/data-cite.js
@@ -37,7 +37,12 @@ function requestLookup(conf) {
       href = entry.href;
     }
     if (path) {
-      href = new URL(path, href).href;
+      // We resolve absolute paths relative to the spec
+      // so "html/foo.html#bar" becomes:
+      // "./foo.html", and base is "http://.../TR/html"
+      // giving us http://.../TR/html/foo.html"
+      const relPath = path.startsWith("/") ? `.${path}` : path;
+      href = new URL(relPath, href).href;
     }
     if (frag) {
       href = new URL(frag, href).href;
--- a/src/core/data-cite.js
+++ b/src/core/data-cite.js
@@ -37,10 +37,7 @@ function requestLookup(conf) {
       href = entry.href;
     }
     if (path) {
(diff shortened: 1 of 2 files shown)


src/core/data-cite.js
```diff
@@ -37,7 +37,12 @@ function requestLookup(conf) {
       href = entry.href;
     }
     if (path) {
-      href = new URL(path, href).href;
+      // We resolve absolute paths relative to the spec
```
sidvishnoi: maybe just adding a link to the issue (https://github.com/w3c/respec/issues/1856) would be enough?

PR #1866, opened 2018-11-02
Allow use of additionalCopyrightHolders for CG/BG documents

tidoust:
The option was ignored for CG/BG documents. However, it may be useful in some cases where a CG is a joint work with another organization (The Web Media API CG is an example of a joint group with CTA).

### DIFF
--- a/src/w3c/templates/cgbg-headers.js
+++ b/src/w3c/templates/cgbg-headers.js
@@ -133,7 +133,14 @@ export default conf => {
   }
   <p class='copyright'>
     <a href='https://www.w3.org/Consortium/Legal/ipr-notice#Copyright'>Copyright</a> &copy;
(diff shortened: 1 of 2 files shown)


marcoscaceres:
@tidoust would you mind adding the following test to tests/spec/w3c/headers-spec.js (at line 644, I think):

```JS
    it("takes additionalCopyrightHolders into account for CG drafts", async () => {
      const newProps = {
        specStatus: "CG-DRAFT",
        additionalCopyrightHolders: "XXX",
      };
      Object.assign(ops.config, newProps);
      expect($(".head .copyright", doc).text()).toMatch(/XXX\s+&\s+W3C/);
    });
```

And one test for "BG-DRAFT"

tidoust:
Added tests for `CG-DRAFT`, `BG-DRAFT`, `CG-FINAL` and `BG-FINAL`. Note the copyright is not `& W3C` in these cases but `& the Contributors to the...`

PR #1874, opened 2018-11-07
fix(core/data-cite): avoid including self in bibliography

marcoscaceres:


marcoscaceres:
GitHub is being weird.

PR #1879, opened 2018-11-09
refactor(core/webidl-index): remove callback

saschanaz:


marcoscaceres:
Oh, actually, I screwed something up when I did the last pretty. Will fix!

PR #1886, opened 2018-11-13
fix(package): puppeteer 1.10.0 causes hightlighter to pause

marcoscaceres:
Something in puppeteer 1.10.0 has caused a regression whereby the highlight worker doesn't respond. 

https://github.com/w3c/spec-generator/pull/135 

Although reproducible, it's unfortunately a heisenbug: If I enable puppeteers/chrome's devtools, the problem doesn't occur.  

Best solution appears to be to just roll back to 1.9.0.

### DIFF
--- a/examples/basic.built.html
+++ b/examples/basic.built.html
@@ -26,17 +26,25 @@
         This is required.
       </p>
     </section>
-    <section data-dfn-for="Foo" data-link-for="Foo">
+    <section data-dfn-for="Foo">
       <h2><dfn>Foo</dfn> interface</h2>
       <pre class="idl">
+      [Constructor]
       interface Foo {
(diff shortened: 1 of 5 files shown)


marcoscaceres:
It’s gonna be one those bugs 🕷...

marcoscaceres:
So, there are two problems: 

 1.  Importing the worker over file URL hits the security sandbox (`--disable-sandbox` doesn't help).
 1.  latest puppeteer exhibits the same error in either `file://` or `http:`, but 1.9.0 doesn't have the problem. 

So I switched to testing over HTTP instead - which reproduces the error, but works in puppeteer 1.9.0... locally at least... let's see what Travis thinks.

PR #1893, opened 2018-11-19
refactor: Remove unused CSS selectors

saschanaz:
Those are mostly for ancient non-contiguous IDL.


saschanaz:
@marcoscaceres Do we still need Codacy? I think eslint should be enough.

marcoscaceres:
Yeah, I guess it’s redundant if we are running linting on TravisCI.

PR #1895, opened 2018-11-21
refactor: modernize all plugins

saschanaz:
This removes all remaining callbacks from plugins and deprecates callbacks.

### DIFF
--- a/src/core/base-runner.js
+++ b/src/core/base-runner.js
@@ -32,7 +32,7 @@ function toRunnable(plug) {
           resolve();
         } else {
           console.warn(
-            `Plugin ${name} uses a deprecated callback signature. Return a Promise instead.`
+            `Plugin ${name} uses a deprecated callback signature. Return a Promise instead. Read more at: https://github.com/w3c/respec/wiki/Developers-Guide#plugins`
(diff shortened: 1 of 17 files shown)


src/core/base-runner.js
```diff
@@ -27,11 +27,13 @@ function toRunnable(plug) {
         performance.mark(name + "-start");
       }
       try {
-        // Modern plugins are async or normal functions, take zero or one argument (conf)
         if (plug.run.length <= 1) {
           await plug.run(config);
           resolve();
         } else {
+          console.warn(
+            `Plugin ${name} uses a deprecated callback signature. Return a Promise instead.`
```
marcoscaceres: We should probably link to: https://github.com/w3c/respec/wiki/Developers-Guide#plugins

saschanaz: Should we use `pub("warn")` for this?

marcoscaceres: hmm... It's really a developer error, not an Editor error.

marcoscaceres: (so, I say we just stick with console.warn)

PR #1904, opened 2018-11-22
refactor: split findDfn() as dfn-finder module

saschanaz:
findDfn() is an unreadable aircraft-carrier-sized function so this PR breaks it into pieces.

Some functions are still too long but should be more readable than before.

### DIFF
--- a/src/core/dfn-finder.js
+++ b/src/core/dfn-finder.js
@@ -1,0 +1,110 @@
+import { pub } from "./pubsubhub";
+import { wrapInner } from "./utils";
+
+const topLevelEntities = new Set([
+  "callback interface",
+  "callback",
+  "dictionary",
+  "enum",
+  "interface mixin",
+  "interface",
+  "typedef",
+]);
+
+export function findDfn_(
+  defn,
+  { parent, name, originalParent, originalName, definitionMap, type, idlElem }
+) {
+  let dfnForArray = definitionMap[name];
+  let dfns = [];
+  if (dfnForArray) {
+    // Definitions that have a title and [data-dfn-for] that exactly match the
+    // IDL entity:
+    dfns = dfnForArray.filter(dfn => dfn.closest(`[data-dfn-for="${parent}"]`));
+    // If this is a top-level entity, and we didn't find anything with
+    // an explicitly empty [for], try <dfn> that inherited a [for].
+    if (dfns.length === 0 && parent === "" && dfnForArray.length === 1) {
+      dfns = dfnForArray;
+    } else if (topLevelEntities.has(type) && dfnForArray.length) {
+      const dfn = dfnForArray.find(
+        dfn => dfn.textContent.trim() === originalName
+      );
+      if (dfn) dfns = [dfn];
+    }
+  }
+  // If we haven't found any definitions with explicit [for]
+  // and [title], look for a dotted definition, "parent.name".
+  if (dfns.length === 0 && parent !== "") {
+    const dottedName = parent + "." + name;
+    dfnForArray = definitionMap[dottedName];
+    if (dfnForArray !== undefined && dfnForArray.length === 1) {
+      dfns = dfnForArray;
+      // Found it: update the definition to specify its [for] and data-lt.
+      delete definitionMap[dottedName];
+      dfns[0].dataset.dfnFor = parent;
+      dfns[0].dataset.lt = name;
+      if (definitionMap[name] === undefined) {
+        definitionMap[name] = [];
+      }
+      definitionMap[name].push(dfns[0]);
+    }
+  }
+  if (dfns.length > 1) {
+    const msg = `Multiple \`<dfn>\`s for \`${originalName}\` ${
+      originalParent ? `in \`${originalParent}\`` : ""
+    }`;
+    pub("error", msg);
+  }
+  if (dfns.length === 0) {
+    const showWarnings =
+      type &&
+      idlElem &&
+      name &&
+      idlElem.classList.contains("no-link-warnings") === false;
+    if (showWarnings) {
+      const name = type === "operation" ? `${originalName}()` : originalName;
+      const parentName = originalParent ? ` \`${originalParent}\`'s` : "";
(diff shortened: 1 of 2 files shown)


src/core/dfn-finder.js
```diff
@@ -0,0 +1,222 @@
...
+// refers to a top-level entity. For top-level entities, <dfn>
+// elements that inherit a non-empty [dfn-for] attribute are also
+// counted as matching.
+//
+// When a matching <dfn> is found, it's given <code> formatting,
+// marked as an IDL definition, and returned. If no <dfn> is found,
+// the function returns 'undefined'.
+export function findDfn(defn, parent, name, definitionMap, idlElem) {
+  return (
+    findAttributeDfn(defn, parent, name, definitionMap, idlElem) ||
```
marcoscaceres: I think this really wants to be a `switch(defn.type)` statement.

src/core/dfn-finder.js
```diff
@@ -0,0 +1,222 @@
...
+    findOperationDfn(defn, parent, name, definitionMap, idlElem) ||
+    findGeneralDfn(defn, parent, name, definitionMap, idlElem)
+  );
+}
+
+function findAttributeDfn(defn, parent, name, definitionMap, idlElem) {
+  if (defn.type !== "attribute") {
+    return;
+  }
+  parent = parent.toLowerCase();
```
marcoscaceres: Maybe we should normalize parent in `findDfn()`?

src/core/dfn-finder.js
```diff
@@ -0,0 +1,222 @@
...
+    const dfn = findGeneralDfn(
+      defn,
+      parent,
+      lookupName,
+      definitionMap,
+      idlElem
+    );
+    if (!dfn) {
+      return; // try finding dfn using name, using normal search path...
+    }
```
marcoscaceres: s/normal/general

src/core/dfn-finder.js
```diff
@@ -0,0 +1,222 @@
...
+  "enum",
+  "interface mixin",
+  "interface",
+  "typedef",
+]);
+
+// TODO: make these linkable somehow.
+// https://github.com/w3c/respec/issues/999
+// https://github.com/w3c/respec/issues/982
+const unlinkable = new Set(["maplike", "setlike", "stringifier"]);
```
marcoscaceres: Maybe we should trash this. We could link to IDL spec for each of these.

src/core/dfn-finder.js
saschanaz: Both find(attr/op)Dfn wants to do a second try when no match, so the || chain. IMO a switch here will complicate the second try.

src/core/dfn-finder.js
saschanaz: I have no objection, but may I ask why?

src/core/dfn-finder.js
saschanaz: @marcoscaceres Wait, do we really want to link to Web IDL spec here? if `maplike` should be linked, why not `attribute`, `getter`, etc.? I think it shouldn't be linked because it's not a type but a syntax.

[Bikeshed does link `stringifier`](https://dom.spec.whatwg.org/#interface-range) but not to the IDL spec.

Can we do this in a separate PR? I think we need to discuss more.

src/core/dfn-finder.js
saschanaz: Reading more, the current setup needlessly run `findNormalDfn()` twice because of the implicit chain. It seems it should be explicitly called when needed and then we can use `switch()` in simpler way.

src/core/dfn-finder.js
```diff
@@ -0,0 +1,255 @@
...
+
+// This function looks for a <dfn> element whose title is 'name' and
```
marcoscaceres: Nit: let's merge these two sets of comments and get rid of the "//".

src/core/dfn-finder.js
```diff
@@ -0,0 +1,255 @@
...
+ * @param {string} name
+ * @param {Record<string, HTMLElement[]>} definitionMap
+ * @param {HTMLElement} idlElem
+ */
+function findAttributeDfn(defn, parent, name, definitionMap, idlElem) {
+  const parentLow = parent.toLowerCase();
+  const asLocalName = name.toLowerCase();
+  const asQualifiedName = parentLow + "." + asLocalName;
+  let dfn;
+  if (definitionMap[asQualifiedName] || definitionMap[asLocalName]) {
```
marcoscaceres: I wonder if this should be:

```JS
if (definitionMap[asQualifiedName]){ 
  dfn = findNormalDfn(defn, parent, asQualifiedName, definitionMap, idlElem);
} else if(definitionMap[asLocalName]){
  dfn = findNormalDfn(defn, parent, asLocalName, definitionMap, idlElem);
} 
if(!dfn) {
  return findNormalDfn(defn, parent, name, definitionMap, idlElem);
}
```

Or does it not matter?

saschanaz: `findNormalDfn()` internally does that in line 146 so maybe it's just okay.

marcoscaceres: Ok, cool. We probably **later** fix it so it's doesn't do things like `delete definitionMap[dottedName];` (old behavior)... that seems not great.

saschanaz: Should it be `definitionMap[dottedName] = undefined`?

saschanaz:
Was a long list of commits!

src/core/dfn-finder.js
marcoscaceres: IMO, it shouldn't trash definitions... the defintionMap should be treated as immutable.

marcoscaceres:
> Was a long list of commits!

s/long/good :)

saschanaz:
@marcoscaceres I feel we need a patch release now because we already have a fair amount of refactors. Piling more may potentially cause uncaught regressions.

marcoscaceres:
Sounds like a plan! I’ll pack it up now and get it out.

marcoscaceres:
Ok, pushed respec@23.7.1

PR #1910, opened 2018-11-24
 refactor(core/webidl): upgrade to webidl2.js 16 

saschanaz:
👍

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1418,7 +1418,7 @@
     },
     "array-equal": {
       "version": "1.0.0",
-      "resolved": "http://registry.npmjs.org/array-equal/-/array-equal-1.0.0.tgz",
+      "resolved": "https://registry.npmjs.org/array-equal/-/array-equal-1.0.0.tgz",
       "integrity": "sha1-jCpe8kcv2ep0KwTHenUJO6J1fJM="
     },
     "array-slice": {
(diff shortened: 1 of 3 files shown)


package-lock.json
```diff
@@ -1418,7 +1418,7 @@
     },
     "array-equal": {
       "version": "1.0.0",
-      "resolved": "http://registry.npmjs.org/array-equal/-/array-equal-1.0.0.tgz",
+      "resolved": "https://registry.npmjs.org/array-equal/-/array-equal-1.0.0.tgz",
```
saschanaz: back and forth 🤔

saschanaz: 6.4.1, right? So this is a mystery...

saschanaz:
Thank you for your quick review 👍

package-lock.json
marcoscaceres: 6.4.1... correct. Super odd.

saschanaz: Looks like this is a known issue:

https://npm.community/t/npm-install-downgrading-resolved-packages-from-https-to-http-registry-in-package-lock-json/1818
https://npm.community/t/some-packages-have-dist-tarball-as-http-and-not-https/285

The question is why I get HTTPS then? 🤔🤔🤔🤔🤔

PR #1915, opened 2018-11-26
refactor(core/dfn-finder): lt for empty enum value

saschanaz:
1. Remove `data-no-default` as nothing uses it
2. Remove the fourth argument from `addId` as nothing uses it.
3. Assign `data-lt` for empty enum value to handle it automatically

Currently we have three places to treat empty enum value in a special way (`util.getDfnTitles()`, `findNormalDfn()`, and `tryLink()`, respectively), and this PR removes one of them.

### DIFF
--- a/src/core/utils.js
+++ b/src/core/utils.js
@@ -588,18 +588,18 @@ export function flatten(collector, item) {
  * @param {Element} elem element
  * @param {String} pfx prefix
  * @param {String} txt text
- * @param {Boolean} noLC
+ * @param {Boolean} noLC do not convert to lowercase
  * @returns {String} generated (or existing) id for element
  */
-export function addId(elem, pfx = "", txt = "", noLC = false) {
+export function addId(elem, pfx = "", txt = "") {
   if (elem.id) {
     return elem.id;
   }
   if (!txt) {
     txt = (elem.title ? elem.title : elem.textContent).trim();
   }
-  let id = noLC ? txt : txt.toLowerCase();
-  id = id
+  let id = txt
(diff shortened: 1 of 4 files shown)


saschanaz:
I hope I'm not bothering you with frequent review requests 🙏

marcoscaceres:
No bother at all! This is great!

src/core/utils.js
```diff
@@ -588,18 +588,18 @@ export function flatten(collector, item) {
  * @param {Element} elem element
  * @param {String} pfx prefix
  * @param {String} txt text
- * @param {Boolean} noLC
+ * @param {Boolean} noLC do not convert to lowercase
  * @returns {String} generated (or existing) id for element
  */
-export function addId(elem, pfx = "", txt = "", noLC = false) {
+export function addId(elem, pfx = "", txt = "") {
```
marcoscaceres: any reason to trash the noLC from the args list?

marcoscaceres: Oh, you are always making it lcase.

marcoscaceres: I guess we should then remove the `@param` above

saschanaz: Because nothing uses it. Will this cause any compat issue?

marcoscaceres: Doing a quick check...

marcoscaceres: hmmm... I couldn't find any in other repos, but maybe we should leave it.  It's part of public API, and seems a lot of people use `makeID` in custom plugins.

PR #1916, opened 2018-11-26
fix(tools/builder): source map from terser

saschanaz:
Currently we chooses source map from requirejs and ignores terser, causing broken source map. This fixes the issue.


saschanaz:
@marcoscaceres Currently the comment from `respecJs` variable does not have any effect to the resulting file. `terser` suggests converting it as JSDoc-style comment, what do you think? Also, we may want to change the text.

marcoscaceres:
Changing it seems fine. We can even remove the comment if it helps.

saschanaz:
I decided to just remove it.

PR #1927, opened 2018-11-30
refactor: enforce eslint sort-imports

saschanaz:
Not sure we exactly want this, I just wanted to sort `import {c, b, a}` case and found that [eslint requires more](https://eslint.org/docs/rules/sort-imports).

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -18,6 +18,7 @@
     "indent": 0,
     "quotes": 0,
     "semi": ["error", "always"],
+    "sort-imports": 2,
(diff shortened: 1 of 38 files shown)


saschanaz:
The first commit is made automatically by --fix, the second is manual. 😭 

I just couldn't give up because the first one looks good. 😂

saschanaz:
So this may make people fix lint errors manually. That makes me worried...

marcoscaceres:
I think we should try it. If it’s too annoying we can always turn it off.

PR #1935, opened 2018-12-05
feat(core/webidl-index): exclude IDL from IDL index

### LINKED ISSUE #1880: Get a way to exclude certain IDL blocks from IDL index
Specs sometimes have IDLs only to give examples/templates but not for actual implementations.

https://drafts.csswg.org/cssom/#dom-cssstyledeclaration-_camel_cased_attribute

https://w3c.github.io/mediacapture-main/#interface-definition

Bikeshed has a way to exclude them from IDL index [by marking it as `extract`](https://tabatkins.github.io/bikeshed/#no-idl), but ReSpec has no equivalent feature.

Johanna-hub:
Fixes #1880

### DIFF
--- a/src/core/webidl-index.js
+++ b/src/core/webidl-index.js
@@ -29,7 +29,12 @@ export function run() {
     }
     idlIndexSec.insertAdjacentElement("afterbegin", header);
   }
-  if (!document.querySelector("pre.idl")) {
+  //filter out the IDL marked with class="idlExclude" and the IDL in non-normative sections
+  const idlIndex = Array.from(document.querySelectorAll("pre.def.idl"))
+    .filter(elem => !elem.classList.contains("idlExclude"))
+    .filter(elem => !elem.parentElement.classList.contains("informative"));
+
+  if (!document.querySelector("pre.idl") || idlIndex.length === 0) {
     const text = "This specification doesn't declare any Web IDL.";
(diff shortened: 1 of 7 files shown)


src/core/webidl-index.js
```diff
@@ -29,7 +29,23 @@ export function run() {
...
+      return false;
+    }
+    return true;
+  }
+
+  const idlIndex = Array.from(document.querySelectorAll("pre.def.idl")).filter(
+    idlFilter
+  );
+
+  if (!document.querySelector("pre.idl") || idlIndex.length === 0) {
```
marcoscaceres: I think you should be able to exclude/filter in this CSS query itself by using `:not(.exclude)`, then you might not need the filter above. 

Make sure you add some tests too :)

saschanaz: > then you might not need the filter above.

The exclude thing can be done with `:not()`, but can we use pure CSS selector to blacklist a parent?

marcoscaceres: @saschanaz is right, we can’t go to the parent. So at least `:not()` first then filter on `.closest()`.

saschanaz: (`:not(.informative) > pre.idl:not(.exclude)` will work if we are only to check direct ascendants, but probably `closest()` is better here.)

PR #1943, opened 2018-12-09
refactor(core/figure): remove some jQuery 

saschanaz:
Not super easy to replace everything in once.

### DIFF
--- a/src/core/figures.js
+++ b/src/core/figures.js
@@ -19,8 +19,9 @@ export function run(conf) {
     if (caption) {
       decorateFigure(fig, caption, i, conf);
       figMap[fig.id] = $(caption.childNodes);
-    } else
+    } else {
       showInlineWarning(fig, "Found a `<figure>` without a `<figcaption>`");
+    }
 
     tof.push(getTableOfFiguresListItem(fig.id, caption));
   });
--- a/src/core/figures.js
+++ b/src/core/figures.js
@@ -16,10 +16,10 @@ export function run(conf) {
   const tof = [];
   document.querySelectorAll("figure").forEach((fig, i) => {
     const caption = fig.querySelector("figcaption");
-    const title = caption ? caption.textContent : "";
-    const id = addId(fig, "fig", title);
 
     if (caption) {
+      const title = caption ? caption.textContent : "";
+      const id = addId(fig, "fig", title);
       // set proper caption title
       wrapInner(caption, hyperHTML`<span class='fig-title'>`);
       caption.prepend(
@@ -27,9 +27,9 @@ export function run(conf) {
         hyperHTML`<span class='figno'>${i + 1}</span>`,
         " "
       );
+      figMap[id] = $(caption.childNodes);
     } else pub("warn", "A `<figure>` should contain a `<figcaption>`.");
 
-    figMap[id] = $(caption ? [...caption.childNodes] : []);
     const tofCaption = caption.cloneNode(true);
     tofCaption.querySelectorAll("a").forEach(anchor => {
       renameElement(anchor, "span").removeAttribute("href");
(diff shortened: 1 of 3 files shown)


src/core/figures.js
```diff
@@ -12,33 +14,30 @@ export function run(conf) {
...
-  $("figure").each(function() {
-    const $fig = $(this);
-    const $cap = $fig.find("figcaption");
-    const tit = $cap.text();
-    const id = $fig.makeID("fig", tit);
-    if (!$cap.length)
-      pub("warn", "A `<figure>` should contain a `<figcaption>`.");
+  document.querySelectorAll("figure").forEach((fig, i) => {
+    const caption = fig.querySelector("figcaption");
+    const title = caption.textContent;
```
sidvishnoi: this should be moved down (inside `if (caption)`)

src/core/figures.js
```diff
@@ -12,33 +14,30 @@ export function run(conf) {
...
-      .removeAttr("href");
+    if (caption) {
+      // set proper caption title
+      wrapInner(caption, hyperHTML`<span class='fig-title'>`);
+      caption.prepend(
+        conf.l10n.fig,
+        hyperHTML`<span class='figno'>${i + 1}</span>`,
+        " "
+      );
+    } else pub("warn", "A `<figure>` should contain a `<figcaption>`.");
```
sidvishnoi: If we are emitting a warning here, maybe also add the relevant `<figure>` in `console.warn` to make it easier to identify

saschanaz: Maybe we can directly link in the warning list UI as link-to-dfn does:

https://github.com/w3c/respec/blob/c1a1f82e2bebd8e75bbcc809e8afebc7c1fee13c/src/core/link-to-dfn.js#L62-L67

I hope we have a general helper function for this.

saschanaz: The best would be pointing the line from the source, which is impossible 😂

src/core/figures.js
```diff
@@ -12,33 +14,30 @@ export function run(conf) {
...
-      .prepend($("<span class='figno'>" + num + "</span>"))
-      .prepend(document.createTextNode(conf.l10n.fig));
-    figMap[id] = $cap.contents();
-    const $tofCap = $cap.clone();
-    $tofCap
-      .find("a")
-      .renameElement("span")
-      .removeAttr("href");
+    if (caption) {
+      const title = caption ? caption.textContent : "";
```
sidvishnoi:
```suggestion
      const title = caption.textContent;
```

src/core/figures.js
saschanaz: Filed #1944

marcoscaceres: For now, maybe just do: 
```JS
else {
   const msg = "A `<figure>` should contain a `<figcaption>`".
   pub("warn", `${msg}. See developer console for offending element.`);
   elem.classList.add("respec-offending-element");
   console.warn(msg, elem);
}

```

src/core/figures.js
```diff
@@ -19,7 +19,8 @@ export function run(conf) {
     if (caption) {
       decorateFigure(fig, caption, i, conf);
       figMap[fig.id] = $(caption.childNodes);
-    } else showInlineWarning(fig, "Found a `<figure>` without a `<figcaption>`");
+    } else
+      showInlineWarning(fig, "Found a `<figure>` without a `<figcaption>`");
```
marcoscaceres: We should probably warp this "{ }"... we've not done unwrapped else before. Even though they are ok (and we have tooling to catch issues), after seeing bugs in other people's code because of unwrapped if/else statements, I'd prefer we wrap it (and yes, I know we increasingly doing unwrapped `if` nowadays in ReSpec, but the dangling else look a little odd).

saschanaz: `curly` is the rule but no option fully satisfies me. I'll just add the braces manually for now.

PR #1945, opened 2018-12-10
Add showInlineError/Warning helper functions

saschanaz:
Didn't know we already had `showInlineError()`! The function showed "warning" instead of "error" despite of its name, so this PR renames it as `showInlineWarning` and adds proper `showInlineError`.

### DIFF
--- a/src/core/data-cite.js
+++ b/src/core/data-cite.js
@@ -13,7 +13,7 @@
  * Usage:
  * https://github.com/w3c/respec/wiki/data--cite
  */
-import { refTypeFromContext, showInlineError, wrapInner } from "./utils";
+import { refTypeFromContext, showInlineWarning, wrapInner } from "./utils";
 import { resolveRef, updateFromNetwork } from "./biblio";
 import hyperHTML from "../deps/hyperhtml";
 export const name = "core/data-cite";
@@ -32,7 +32,7 @@ function requestLookup(conf) {
       const entry = await resolveRef(key);
       cleanElement(elem);
       if (!entry) {
-        showInlineError(elem, `Couldn't find a match for "${originalKey}".`);
+        showInlineWarning(elem, `Couldn't find a match for "${originalKey}".`);
         return;
       }
       href = entry.href;
--- a/src/core/data-cite.js
+++ b/src/core/data-cite.js
@@ -32,7 +32,7 @@ function requestLookup(conf) {
       const entry = await resolveRef(key);
       cleanElement(elem);
       if (!entry) {
-        showInlineWarning(elem, `Couldn't find a match for "${originalKey}".`);
+        showInlineWarning(elem, `Couldn't find a match for "${originalKey}"`);
         return;
       }
       href = entry.href;
--- a/src/core/inlines.js
+++ b/src/core/inlines.js
@@ -13,7 +13,7 @@
 //  - respecRFC2119: a list of the number of times each RFC2119
 //    key word was used.  NOTE: While each member is a counter, at this time
 //    the counter is not used.
-import { getTextNodes, refTypeFromContext } from "./utils";
+import { getTextNodes, refTypeFromContext, showInlineWarning } from "./utils";
 import hyperHTML from "../deps/hyperhtml";
 import { idlStringToHtml } from "./inline-idl-parser";
 import { pub } from "./pubsubhub";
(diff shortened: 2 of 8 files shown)


src/core/link-to-dfn.js
```diff
@@ -180,14 +176,12 @@ export async function run(conf) {
 
 function showLinkingError(elems) {
   elems.forEach(elem => {
-    elem.classList.add("respec-offending-element");
-    elem.title = "Linking error: not matching <dfn>";
-    pub(
-      "warn",
+    showInlineWarning(
```
sidvishnoi: We might need to collect these warnings together maybe.

image
See 2 warnings for two `<a>baz</a>` which should be warned together.

src/core/utils.js
```diff
@@ -315,22 +315,64 @@ export function removeReSpec(doc) {
...
+    addId(elem, "respec-offender");
+  }
+}
+
+/**
+ * @param {Element} element
+ * @param {number} i
+ */
+function generateMarkdownLink(element, i) {
+  return `[${i + 1}](#${element.id})`;
```
sidvishnoi: I'm not feeling very good about adding just numbers here (`i + 1`), even though I can't think of anything better.

src/core/data-cite.js
```diff
@@ -32,7 +32,7 @@ function requestLookup(conf) {
       const entry = await resolveRef(key);
       cleanElement(elem);
       if (!entry) {
-        showInlineError(elem, `Couldn't find a match for "${originalKey}".`);
+        showInlineWarning(elem, `Couldn't find a match for "${originalKey}".`);
```
sidvishnoi:
```suggestion
        showInlineWarning(elem, `Couldn't find a match for "${originalKey}"`);
```

src/core/inlines.js
```diff
@@ -99,16 +99,11 @@ export function run(conf) {
...
-              cite.classList.add("respec-offending-element");
-              const msg =
+              showInlineWarning(
+                cite,
                 "Normative references in informative sections are not allowed. " +
-                `Remove '!' from the start of the reference \`[[!${ref}]]\`. `;
-              pub(
-                "warn",
-                msg + "See developer console to find offending element."
+                  `Remove '!' from the start of the reference \`[[!${ref}]]\``
```
sidvishnoi: 😆 
``` html
<cite 
 class="respec-offending-element" 
 title="Normative references in informative sections are not allowed. Remove '!' from the start of the reference `[[!html]]`" 
 id="respec-offender-normative-references-in-informative-sections-are-not-allowed-remove-from-the-start-of-the-reference-html"
><a class="bibref" href="#bib-html">html</a></cite>
```
Should shorten it? But it maybe fine, as the `id` is there just for error linking.

saschanaz: I think it should be just okay, as it won't be used as an external link.

src/core/linter.js
```diff
@@ -41,28 +42,24 @@ const baseResult = {
...
+    const output = { ...baseResult, ...result };
+    const {
+      description,
+      help,
+      howToFix,
+      name,
+      occurrences,
+      offendingElements,
+    } = output;
+    const message = `Linter (${name}): ${description} ${howToFix} ${help} (x ${occurrences})`;
```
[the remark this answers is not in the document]
saschanaz: I replaced the part so that the sentence becomes: `Occured at: 1, 2` or `(Count: 2)` if no elements.

src/core/link-to-dfn.js
marcoscaceres: I don't feel strongly about the grouping. As an editor, as long as I can quickly get to what I need to fix, then I'm happy.

src/core/utils.js
marcoscaceres: Number are ok.

src/core/link-to-dfn.js
saschanaz: I think the suggested behavior is more complex to write, as it mixes the message part and the linking part. Maybe a separate PR would be good.

src/core/utils.js
saschanaz: >I can't think of anything better.

Neither can I. 😀

src/core/link-to-dfn.js
marcoscaceres: Yeah, agree (new PR). Let's see how this goes first :)

saschanaz:
I'm merging this as I don't see any strict blocker. Any suggestions here can be further discussed and opened as separate PRs 👍

saschanaz:
Special thanks for @sidvishnoi with the detailed review!

marcoscaceres:
Gonna publish this now :)

PR #1946, opened 2018-12-11
feat(core/idl-index): exclude IDL from IDL index

### LINKED ISSUE #1880: Get a way to exclude certain IDL blocks from IDL index

Johanna-hub:
Fixes #1880 
(I will close #1935)

Looking for feedback 🙂

- I have added extra part to the test, so it is testing for exclude class (now renamed) and for informative class in parent
- Now using :not and closest()


src/core/webidl-index.js
```diff
@@ -29,7 +29,20 @@ export function run() {
     }
     idlIndexSec.insertAdjacentElement("afterbegin", header);
   }
-  if (!document.querySelector("pre.idl")) {
+
+  //filter out the IDL marked with class="exclude" and the IDL in non-normative sections
+  function idlFilter(idl) {
+    if (idl.closest(".informative")) {
```
sidvishnoi: We can simply return it like this (without explicitly saying return `true` or `false`):
``` js
return !idl.closest(".informative");
```
Maybe we should inline it below in filter itself.

marcoscaceres: Agree, we can probably get rid of this function entirely and just inline: 

```JS
.filter(idl => !idl.closest(".informative"));
```

tests/spec/core/idl-index-spec.js
```diff
@@ -40,6 +40,42 @@ interface Bar {
...
+        };
+        </pre>
+      </section>
+      <section>
+        <pre class=idl>
+        interface Bar {
+          readonly attribute DOMString foo;
+        };
+        </pre>
+        </section>
```
marcoscaceres:
```suggestion
      </section>
```

tests/spec/core/idl-index-spec.js
```diff
@@ -40,6 +40,42 @@ interface Bar {
     expect(header.textContent).toEqual("1. IDL Index");
   });
 
+  it("generates an idl summary, but excludes things the editor doesn't want", async () => {
+    const body = `
+      ${makeDefaultBody()}
+      <section>
```
marcoscaceres: Nit: you might want to add a comment above each section describing what should be excluded. 
```
<!-- this should exclude because is has an "exclude" class declared -->
```

src/core/webidl-index.js
```diff
@@ -29,7 +29,13 @@ export function run() {
...
     idlIndexSec.insertAdjacentElement("afterbegin", header);
   }
-  if (!document.querySelector("pre.idl")) {
+
+  //filter out the IDL marked with class="exclude" and the IDL in non-normative sections
+  const idlIndex = Array.from(
+    document.querySelectorAll("pre.def.idl:not(.exclude)")
+  ).filter(idl => !idl.closest(".informative"));
+
+  if (!document.querySelector("pre.idl") || idlIndex.length === 0) {
```
saschanaz: Probably no need to query twice here.

Johanna-hub:
Classes to be excluded that I am certain about:
.note
.issue
.example

I am on the fence about .practice - seems that the best practices documents themselves are considered non-normative, but not sure about when they are included in the spec.  Have included it for now and can remove it obviously.

I also added in editor's notes to be excluded, not too sure if that's relevant so can also remove.

I have included tests for the new classes and also a test to make sure that the No IDL text is generated if all IDL is in non-normative sections or excluded.

src/core/webidl-index.js
```diff
@@ -29,16 +29,26 @@ export function run() {
...
     idlIndexSec.prepend(header);
   }
-  if (!document.querySelector("pre.idl")) {
+
+  //filter out the IDL marked with class="exclude" and the IDL in non-normative sections
+  const idlIndex = Array.from(
+    document.querySelectorAll("pre.def.idl:not(.exclude)")
+  ).filter(
+    idl =>
+      !idl.closest(".informative, .note, .issue, .example, .ednote, .practice")
```
saschanaz: This selector looks long enough, a separate variable would be cleaner.

tests/spec/core/idl-index-spec.js
```diff
@@ -40,6 +40,153 @@ interface Bar {
...
+      <section id="idl-index"></section>
+    `;
+    const ops = {
+      config: makeBasicConfig(),
+      body,
+    };
+    const doc = await makeRSDoc(ops);
+    const idlIndex = doc.getElementById("idl-index");
+    expect(idlIndex.innerText.includes("Foo")).toEqual(false);
+    expect(idlIndex.innerText.includes("Baz")).toEqual(false);
```
saschanaz: Would be great to also check `.toEqual(true)`.

marcoscaceres: Agree, it's good to check for the false positives too... to assure that what should be there is actually there.

Issue #1924, opened 2018-11-29
A duplicated IDL block unexpectedly wraps the former by `<code>`


saschanaz:
## Important info

* URL to affected spec: (local repo)/tests/spec/core/webidl.html
* ReSpec version: develop branch
* [x] I did a "hard refresh", but it's still busted. 

## Description of problem

What happened (e.g., it crashed)?:

![image](https://user-images.githubusercontent.com/3396686/49227270-0ede7b00-f42c-11e8-85c8-846f637bbc92.png)

But what I was expecting?:

This shouldn't happen.

~~I couldn't get a minimal repro yet.~~ Any duplicated IDL causes this.

```
<section>
<h2>Repro</h2>
<pre class="idl">
interface Foo {};
</pre>
<pre class="idl">
interface Foo {};
</pre>
</section>
```

marcoscaceres:
closing as invalid, as we already show an error if you have two IDL blocks with the same identifier. That's not allowed anyway, so anything occurring as consequence of falls under "undefined behavior".

saschanaz:
I think it's undefined when it's not a valid syntax, but it should be shown without bugs when it can be parsed.

marcoscaceres:
Ok, let's reopen. It's a bit of an edge case, but hopefully we get around to fixing it.

marcoscaceres:
This issue was filed in 2018 with ReSpec v20. The duplicated IDL block `<code>` wrapping has been addressed through numerous WebIDL processing improvements since then, including the v37 refactoring. If you can still reproduce this with the latest ReSpec, please reopen with a minimal reproduction.

marcoscaceres:
Reopening: closed prematurely without proper verification. Will re-investigate with /ai-skeptic validation before re-closing.

marcoscaceres:
Verified fixed. Current code at `src/core/webidl.js:351` uses `wrapInner(idlElement, document.createElement("code"))` which properly wraps each IDL block. Duplicate IDL blocks are handled by validation errors (lines 404-420) without the `<code>` wrapping side effect reported in 2018.

PR #1949, opened 2018-12-12
refactor(core/link-to-dfn): split functions

saschanaz:
Following #1948 

Adding jsconfig.json to correctly `@ts-check` latest ECMAScript native functions.

### DIFF
--- a/src/core/link-to-dfn.js
+++ b/src/core/link-to-dfn.js
@@ -236,7 +236,7 @@ function shouldWrapByCode(dfn, term) {
 function findExplicitExternalLinks() {
   const links = document.querySelectorAll(
     "a[data-cite]:not([data-cite='']):not([data-cite*='#']), " +
-    "dfn[data-cite]:not([data-cite='']):not([data-cite*='#'])"
+      "dfn[data-cite]:not([data-cite='']):not([data-cite*='#'])"
   );
   return [...links]
     .filter(el => {
--- a/src/core/link-to-dfn.js
+++ b/src/core/link-to-dfn.js
@@ -57,23 +57,8 @@ export async function run(conf) {
 
   showLinkingError(badLinks);
 
-  // These are additional references that need to be looked up externally.
-  // The `possibleExternalLinks` above doesn't include references that match selectors like
-  //   a[data-cite="spec"], dfn[data-cite="spec"], dfn.externalDFN
-  const additionalExternalLinks = [
-    ...document.querySelectorAll(
-      "a[data-cite]:not([data-cite='']):not([data-cite*='#']), " +
-        "dfn[data-cite]:not([data-cite='']):not([data-cite*='#'])"
-    ),
-  ]
-    .filter(el => {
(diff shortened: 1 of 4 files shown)


src/core/link-to-dfn.js
```diff
@@ -206,6 +230,22 @@ function shouldWrapByCode(dfn, term) {
...
+ * Find additional references that need to be looked up externally.
+ * Examples: a[data-cite="spec"], dfn[data-cite="spec"], dfn.externalDFN
+ */
+function findExplicitExternalLinks() {
+  const selector =
+    "a[data-cite]:not([data-cite='']):not([data-cite*='#']), " +
+    "dfn[data-cite]:not([data-cite='']):not([data-cite*='#'])";
+  return [...document.querySelectorAll(selector)]
+    .filter(el => {
+      const closest = /** @type {HTMLElement} */ (el.closest("[data-cite]"));
```
sidvishnoi: Maybe we should keep type definitions in their own line? Inline is reducing readability IMO.

saschanaz:
>But xref tests are failing in Travis (passing locally 😕)

Confirmed the cause of the failure. I will just revert the commit in this PR, maybe will open a separate PR later.

src/core/link-to-dfn.js
saschanaz: Inlining caused needless multi-lining:

```
const additionalExternalLinks = [
    ...document.querySelectorAll(
      "a[data-cite]:not([data-cite='']):not([data-cite*='#']), " +
        "dfn[data-cite]:not([data-cite='']):not([data-cite*='#'])"
    ),
  ]
```

4 lines vs 6 lines. Hmm.

saschanaz: Just tried some modification, hoping you like it 👍

sidvishnoi: I meant inside the filter. Sorry for misunderstanding

``` diff
- const closest = /** @type {HTMLElement} */ (el.closest("[data-cite]"));
+ /** @type {HTMLElement} */ 
+ const closest = el.closest("[data-cite]");
```

saschanaz:
>This is the most critical part of ReSpec, but also the least tested (I know, it's sad... but this is legacy code from before any of the new maintainers showed up).

Yes... I'm doing this to fix #1924 anyway 🤣

src/core/link-to-dfn.js
saschanaz: OMG, maybe a need some coffee.

That's unfortunately impossible because `@ts-check` will warn that `Element` cannot be assigned to `HTMLElement`. It's like:

```
var closest: HTMLElement = element; // type error!
```

saschanaz:
@marcoscaceres @sidvishnoi I just added two tests to make sure we are not breaking anything (as we currently only have three tests for this important core module...) Could you do a final check for them?

saschanaz:
I'll merge now, feel free to open new issues or PRs if anything looks wrong 😀

Issue #1878, opened 2018-11-09
data-lt generates duplicated definitions


saschanaz:
## Important info

* URL to affected spec: https://w3c.github.io/gamepad/extensions.html
* ReSpec version: 23.5.0
* [x] I did a "hard refresh", but it's still busted. 

## Description of problem

What happened (e.g., it crashed)?: 

`<dfn data-lt="pulse()>pulse</dfn>` becomes `data-lt="pulse()|pulse|gamepadhapticactuator.pulse()|gamepadhapticactuator.pulse|pulse()|pulse"` and causes an error saying "Duplicate definitions of 'pulse'".

But what I was expecting?:

`data-lt` shouldn't generate duplicated items.

Removing `data-lt` solves the issue but also removes `gamepadhapticactuator.pulse()|gamepadhapticactuator.pulse`. Is this expected?

marcoscaceres:
Hmm... need to check why the `data-lt`s are not being treated as a set. 

> Removing data-lt solves the issue but also removes `gamepadhapticactuator.pulse()|gamepadhapticactuator.pulse`. Is this expected?

No. We should fix that.

saschanaz:
Looks like the module `dfn` first generates `pulse()|pulse` and then registers them to `definitionMap`, and then `dfn-finder` (or `webidl` when writing the original post) again generates the remaining parts and then again registers them to `definitionMap`, causing the duplication error.

PR #1953, opened 2018-12-15
fix(core/dfn-finder): always enumerate every alternative form for operations

saschanaz:
This fixes https://github.com/w3c/respec/issues/1878#issuecomment-437748048:

>>Removing data-lt solves the issue but also removes gamepadhapticactuator.pulse()|gamepadhapticactuator.pulse. Is this expected?
>
>No. We should fix that.

### DIFF
--- a/src/core/dfn.js
+++ b/src/core/dfn.js
@@ -27,11 +27,11 @@ export function run(conf) {
  * @param {Record<string, HTMLElement[]>} definitionMap
  */
 export function registerDefinitionMapping(dfn, names, definitionMap) {
-  names.forEach(name => {
+  for (const name of names) {
     if (!definitionMap[name]) {
       definitionMap[name] = [dfn];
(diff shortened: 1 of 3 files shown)


src/core/dfn.js
```diff
@@ -16,22 +16,22 @@ export function run(conf) {
...
  * @param {Record<string, HTMLElement[]>} definitionMap
  */
-export function registerDefinitionMapping(dfn, name, definitionMap) {
-  if (!definitionMap[name]) {
-    definitionMap[name] = [dfn];
-  } else if (!definitionMap[name].includes(dfn)) {
-    definitionMap[name].push(dfn);
-  }
+export function registerDefinitionMapping(dfn, names, definitionMap) {
+  names.forEach(name => {
```
marcoscaceres: Nit: I wonder if we should should just `for ... of` here. Don't really mind tho.

PR #1957, opened 2018-12-16
refactor(core/structure): partially remove jQuery

saschanaz:
👍


saschanaz:
Looks like ReSpec implicitly normalizes section headers. Do we want to warn about this somehow?

PR #1962, opened 2018-12-18
fix(core/utils): Add non-normative classes

Johanna-hub:
As suggested in #1946 by Marcos, this creates a new const for non-normative classes for export


src/core/utils.js
```diff
@@ -97,6 +97,10 @@ const fetchDestinations = new Set([
   "",
 ]);
 
+//these classes mean a section is non-normative
```
marcoscaceres: As "section" is an element, maybe something like: 
```suggestion
 // CSS classes that indicate that a container element is non-normative.
```

src/core/utils.js
```diff
@@ -97,6 +97,10 @@ const fetchDestinations = new Set([
...
+export const nonNormativeClasses =
```
marcoscaceres: This being a CSS selector (and "classes" being ambiguous), I'd suggest: 

```suggestion
export const nonNormativeSelector =
```

src/core/utils.js
saschanaz: "a container element" somehow sounds like "a parent element" though.

marcoscaceres: True, we can probably drop "container".

marcoscaceres: Updated suggestion.

saschanaz: How about:

```diff
- that indicate that
+ indicating
```

as IMO doubled "that" looks weird 😀

src/core/utils.js
```diff
@@ -97,6 +97,10 @@ const fetchDestinations = new Set([
   "",
 ]);
 
+//CSS selector for matching elements that are non-normative
```
saschanaz: nit: space after `//`

Issue #1955, opened 2018-12-16
Refactor biblio-db with idb dependency


saschanaz:
We now have `idb-keyval` so we probably want to use it there, as the current code is a mixture of Promises and callbacks.

NJnisarg:
Hello, I am new to this project and am trying to contribute. Since this was marked as good first issue I was trying to understand it. Can you guide me to work on this?

What I understood is biblio-db is a module which stores the data related to some kind of references and it seems to be built on top of indexedDB API. If I am not wrong it has its own functions that work as wrappers to the lower level indexedDB API. And this issue is possibly trying to replace/modify this module with the package [idb-keyval](https://www.npmjs.com/package/idb-keyval) from nodejs/npm.

saschanaz:
@NJnisarg Hello, welcome to ReSpec! 🎉 

Yes, this issue is to replace direct calls to raw IndexedDB API with the library `idb-keyval`. AFAIK most of the things should be straightforward for one who knows IDB well, but one of my concern is that `idb-keyval` is just a key-value store so it does not support `onupgradeneeded` event. Maybe we should upgrade to https://www.npmjs.com/package/idb for this.

Feel free to ask questions if you find any.

cc: @sidvishnoi who added the dependency 👍

marcoscaceres:
@NJnisarg, just in case you haven't yet seen it, we have a dev guide to help people get started https://github.com/w3c/respec/wiki/Developers-Guide

NJnisarg:
@marcoscaceres and @saschanaz Thanks! I have gone through the Developer Guide and am trying to work with what the issue suggests. I am understanding the codebase a little as of now and would like to take this up as my first issue. I would post questions that are relevant to the issue here. Thanks again and looking forward to contribute.

sidvishnoi:
> Maybe we should upgrade to https://www.npmjs.com/package/idb for this.

Certainly if it helps. Presently, the `idb-keyval` is being used just in `core/xref` (and its tests) and it should be trivial to replace. I'll send a PR tomorrow for that.

NJnisarg:
@sidvishnoi Hello, by the way I am trying to work on this biblio-db issue, do you mind if I continue and generate a PR in a few days? I might reach out for help.

sidvishnoi:
@NJnisarg Sure feel free to ask wherever needed. Happy to help.
[...]

PR #1964, opened 2018-12-18
refactor(core/xref): use idb instead of idb-keyval

sidvishnoi:
This replaces `idb-keyval` package by `idb`. See https://github.com/w3c/respec/issues/1955#issuecomment-447921659


saschanaz:
Oh, can't we use it as a module anymore?

sidvishnoi:
> Oh, can't we use it as a module anymore?

Sadly it's exported only as:
``` js
  if (typeof module !== 'undefined') {
    module.exports = exp;
    module.exports.default = module.exports;
  } else {
    self.idb = exp;
  }
```

Thought of sending a PR for AMD export, but decided not to as we'll be (hopefully) moving to webpack soon. So, can keep it global till maybe.

saschanaz:
I think a PR will still be good for some other remaining AMD users.

src/core/xref.js
```diff
@@ -128,9 +130,9 @@ function collectKeys(xrefs) {
 // adds data to cache
 async function cacheResults(data, cache) {
   const promisesToSet = Object.entries(data).map(([term, results]) =>
-    IDB.set(term, results, cache)
+    idbSet(term, results, cache)
```
saschanaz: This should also be fixed, right?

sidvishnoi: Didn't get it. What exactly?

saschanaz: Oh, doesn't this line requires `return`? Otherwise the `promisesToSet` gets nothing.

src/core/xref.js
```diff
@@ -292,3 +296,17 @@ function disambiguate(fetchedData, context, term) {
   showInlineWarning(elem, msg, title);
   return null;
 }
+
+// utility functions for using `idb` as a key-value store
+function idbGet(key, db) {
```
saschanaz: Can we get a class that restores some of the previous functions in `utils` so that we can use it from other modules e.g. `biblio-db`? Something like `export class IDBKeyVal { constructor(idb); get(); set();  }`.

src/core/xref.js
sidvishnoi: It's using implicit return as it's an arrow function not surrounded by `{ }`

saschanaz: Ah right. My bad 😱

src/core/xref.js
sidvishnoi: Will do!

sidvishnoi: @saschanaz good now?

sidvishnoi:
Waiting on https://github.com/jakearchibald/idb/pull/76

marcoscaceres:
@sidvishnoi looks like jakearchibald/idb#76 was superseded by an ES module. Guessing this is good to merge now?

sidvishnoi:
@marcoscaceres upgraded to use latest idb release.

PR #1968, opened 2018-12-21
refactor: modularize definitionMap

### LINKED ISSUE #1954: Refactor `conf.definitionMap` as a module
`definitionMap` is not really a configuration, it's a collection of elements set in run-time. Maybe `import { definitionMap } from "./dfn";` should be preferred as currently passing `conf.definitionMap` to functions is a pain.

saschanaz:
Closes #1954

### DIFF
--- a/src/core/dfn-map.js
+++ b/src/core/dfn-map.js
@@ -1,0 +1,18 @@
+// @ts-check
+
+/** @type {Record<string, HTMLElement[]>} */
+export const definitionMap = Object.create(null);
+
+/**
+ * @param {HTMLElement} dfn A definition element to register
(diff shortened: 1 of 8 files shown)


saschanaz:
I have a concern that this may cause a conflict if we ever provide a JavaScript API to process documents. (such as `import { processDocument } from "respec";`)

src/core/dfn-map.js
```diff
@@ -0,0 +1,18 @@
+// @ts-check
+
+/** @type {Record<string, HTMLElement[]>} */
+export const definitionMap = Object.create(null);
```
marcoscaceres: My hope is that eventually this can be a real Map... in the future.

Issue #1846, opened 2018-10-04
(refactor): Remove jQuery from ReSpec


sidvishnoi:
Remove jQuery from following remaining modules:

- [x] core/requirements https://github.com/w3c/respec/pull/1845
- [x] [core/figures](https://github.com/w3c/respec/blob/develop/src/core/figures.js) https://github.com/w3c/respec/pull/1956
- [x] [core/structure](https://github.com/w3c/respec/blob/develop/src/core/structure.js) https://github.com/w3c/respec/pull/1957
- [x] [core/link-to-dfn](https://github.com/w3c/respec/blob/develop/src/core/link-to-dfn.js) #1897
- [x] [core/dfn](https://github.com/w3c/respec/blob/develop/src/core/dfn.js) #1898
- [x] [core/ui](https://github.com/w3c/respec/blob/develop/src/core/ui.js) #1930
- [x] [core/issue-notes](https://github.com/w3c/respec/blob/develop/src/core/issues-notes.js) #1969
- [x] [w3c/conformance](https://github.com/w3c/respec/blob/develop/src/w3c/conformance.js) #1926
- [x] tests: [~core/dfn-spec.js~](https://github.com/w3c/respec/blob/develop/tests/spec/core/dfn-spec.js), [~core/informative-spec.js~](https://github.com/w3c/respec/blob/develop/tests/spec/core/informative-spec.js), [~core/issues-notes-spec.js~](https://github.com/w3c/respec/blob/develop/tests/spec/core/issues-notes-spec.js), [~core/section-refs-spec.js~](https://github.com/w3c/respec/blob/develop/tests/spec/core/section-refs-spec.js), [~core/structure-spec.js~](https://github.com/w3c/respec/blob/develop/tests/spec/core/structure-spec.js), [~core/utils-spec.js~](https://github.com/w3c/respec/blob/develop/tests/spec/core/utils-spec.js), [~w3c/conformance-spec.js~](https://github.com/w3c/respec/blob/develop/tests/spec/w3c/conformance-spec.js), [~w3c/headers-spec.js~](https://github.com/w3c/respec/blob/develop/tests/spec/w3c/headers-spec.js) (https://github.com/w3c/respec/pull/1977, https://github.com/w3c/respec/pull/1979)

sidvishnoi:
Doing some of the quick ones (figures, structure, conformance) in [remove-jquery branch](https://github.com/w3c/respec/tree/remove-jquery).

[...]

PR #1969, opened 2018-12-21
refactor(core/issue-notes): remove jquery

devanshbatra04:
I did most of the refactoring. However, there is this small makeID function, as far as I understand, it's for tracking #issue on the page. It is linked to global jQuery object.
screenshot from 2018-12-22 02-54-18

But it's being used inside core/issues-notes.js. It further refers to another function somewhere in Utils. How do I proceed? @sidvishnoi @marcoscaceres 

refers to issue #1846

### DIFF
--- a/src/core/issues-notes.js
+++ b/src/core/issues-notes.js
@@ -10,7 +10,7 @@
 // numbered to avoid involuntary clashes.
 // If the configuration has issueBase set to a non-empty string, and issues are
 // manually numbered, a link to the issue is created using issueBase and the issue number
-import { addId, fetchAndCache } from "./utils";
+import { addId, fetchAndCache, parents } from "./utils";
 import css from "../deps/text!core/css/issues-notes.css";
 import hyperHTML from "../deps/hyperhtml";
 import { pub } from "./pubsubhub";
@@ -120,7 +120,7 @@ function handleIssues(ins, ghIssues, conf) {
             issueList.append(li);
           }
         }
-        tit.querySelector("span").innerHTML = text;
+        tit.querySelector("span").textContent = text;
         if (ghIssue && report.title && githubAPI) {
           if (ghIssue.state === "closed") div.classList.add("closed");
           const labelsGroup = Array.from(ghIssue.labels || [])
@@ -165,7 +165,7 @@ function handleIssues(ins, ghIssues, conf) {
           body = hyperHTML`${ghIssue.body_html}`;
         }
         div.append(body);
-        const level = getCount(tit);
+        const level = parents(tit, "section").length + 2;
         tit.setAttribute("aria-level", level);
       }
       pub(report.type, report);
@@ -176,7 +176,7 @@ function handleIssues(ins, ghIssues, conf) {
       issueSummaryElement.innerHTML = issueSummary.innerHTML;
   } else if (issueSummaryElement) {
     pub("warn", "Using issue summary (#issue-summary) but no issues found.");
-    issueSummaryElement.parentNode.removeChild(issueSummaryElement);
+    issueSummaryElement.remove();
   }
 }
 
@@ -237,16 +237,6 @@ function createLabel(label) {
     class="${cssClasses}"
     style="${style}"
     href="${href}">${name}</a>`;
-}
-
-function getCount(currentEl) {
-  let el = currentEl;
-  let count = 0;
-  while (el !== null) {
-    if (el.tagName === "SECTION") count++;
-    el = el.parentElement;
-  }
-  return count + 2;
 }
 
 async function processResponse(response, issueNumber) {
--- a/src/core/issues-notes.js
+++ b/src/core/issues-notes.js
@@ -86,11 +86,17 @@ function handleIssues(ins, ghIssues, conf) {
               text += " " + dataNum;
               // Set issueBase to cause issue to be linked to the external issue tracker
               if (!isFeatureAtRisk && issueBase) {
-                tit.querySelector(".issue-number").innerHTMl =
-                  "<a href='" + issueBase + dataNum + "'/>";
+                const span = tit.querySelector("span");
+                const a = hyperHTML`<a href='${issueBase + dataNum}'/>`;
+                span.parentNode.insertBefore(a, span);
+                a.append(span);
+                console.log(a.parentNode);
               } else if (isFeatureAtRisk && conf.atRiskBase) {
-                tit.querySelector(".issue-number").innerHTML =
-                  "<a href='" + conf.atRiskBase + dataNum + "'/>";
+                const span = tit.querySelector("span");
+                const a = hyperHTML`<a href='${conf.atRiskBase + dataNum}'/>`;
+                span.parentNode.insertBefore(a, span);
+                a.append(span);
+                console.log(a.parentNode);
               }
               tit.querySelector("span").classList.add("issue-number");
               ghIssue = ghIssues.get(Number(dataNum));
@@ -108,13 +114,13 @@ function handleIssues(ins, ghIssues, conf) {
             a.setAttribute("href", "#" + div.id);
             a.append(conf.l10n.issue + " " + report.number);
             if (report.title) {
-              li.appendChild(
+              li.append(
                 hyperHTML`<span style='text-transform: none'>: ${
                   report.title
                 }</span>`
               );
             }
-            issueList.appendChild(li);
+            issueList.append(li);
           }
         }
         tit.querySelector("span").innerHTML = text;
@@ -134,17 +140,18 @@ function handleIssues(ins, ghIssues, conf) {
             })
             .map(createLabel)
             .reduce((frag, labelElem) => {
-              frag.appendChild(labelElem);
+              frag.append(labelElem);
               return frag;
             }, document.createDocumentFragment());
-          tit.appendChild(
-            hyperHTML`<span style='text-transform: none'>: ${
-              report.title
-            }</span>`.append(labelsGroup)
-          );
+          const node = hyperHTML`<span style='text-transform: none'>: ${
+            report.title
+          }</span>`;
+          node.append(labelsGroup);
+
+          tit.append(node);
           inno.removeAttribute("title");
         } else if (report.title) {
-          tit.appendChild(
+          tit.append(
             hyperHTML`<span style='text-transform: none'>: ${
               report.title
             }</span>`
@@ -152,15 +159,16 @@ function handleIssues(ins, ghIssues, conf) {
           inno.removeAttribute("title");
         }
         tit.classList.add("marker");
-        div.appendChild(tit);
+        div.append(tit);
         let body = inno;
         inno.replaceWith(div);
         body.classList.remove(report.type);
         body.removeAttribute("data-number");
-        if (ghIssue && !body.innerHTML().trim()) {
-          body = ghIssue.body_html;
+        if (ghIssue && !body.innerHTML.trim()) {
+          body = hyperHTML`${ghIssue.body_html}`;
         }
-        div.appendChild(body);
+        console.log(body);
+        div.append(body);
         const level = getCount(tit);
         tit.setAttribute("aria-level", level);
       }
--- a/src/core/issues-notes.js
+++ b/src/core/issues-notes.js
@@ -117,7 +117,7 @@ function handleIssues(ins, ghIssues, conf) {
             const li = htmlToElement("<li><a></a></li>");
             const a = li.querySelectorAll("a")[0];
             a.setAttribute("href", "#" + div.id);
-            a.innerHtml = conf.l10n.issue + " " + report.number;
+            a.append(conf.l10n.issue + " " + report.number);
             if (report.title) {
               li.appendChild(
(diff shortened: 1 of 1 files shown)


sidvishnoi:
Hi @devanshbatra04, thanks for starting work on this.

> makeID function

Use the vanilla [addId](https://github.com/w3c/respec/blob/97defb86bcffdbee19e22a614303d8f5e893a29a/src/core/utils.js#L606) function defined in `core/utils`. You will need to import `addId` from `"./utils"` first as:

``` js
import { addId } from "./utils";
```

It can be used, for example as:
``` js
// before
$el.makeID("h", report.type);
// $el is jquery-ed element

// after
addId(el, "h", report.type):
// el is vanilla HTMLElement
```

I'll do a detail code review in day tomorrow.

src/core/issues-notes.js
```diff
@@ -244,6 +249,24 @@ function createLabel(label) {
     href="${href}">${name}</a>`;
 }
 
+function htmlToElement(html) {
```
sidvishnoi: Oh. We won't need to do this. We're using [hyperHTML](https://github.com/WebReflection/hyperHTML) in ReSpec. Have a look at other modules (like [core/examples](https://github.com/w3c/respec/blob/develop/src/core/examples.js)) to see how we are using it.

src/core/issues-notes.js
```diff
@@ -19,24 +19,25 @@ export const name = "core/issues-notes";
...
-    .filter((i, issue) => issue.parentNode)
-    .each((i, inno) => {
-      const $inno = $(inno);
-      const isIssue = $inno.hasClass("issue");
-      const isWarning = $inno.hasClass("warning");
-      const isEdNote = $inno.hasClass("ednote");
-      const isFeatureAtRisk = $inno.hasClass("atrisk");
-      const isInline = $inno[0].localName === "span";
-      const dataNum = $inno.attr("data-number");
+  const issueList = issueSummary.querySelectorAll("ul")[0];
```
sidvishnoi: We can use `querySelector` if we only need the first element.
```suggestion
  const issueList = issueSummary.querySelector("ul");
```

devanshbatra04:
Thank you @sidvishnoi I will work on the code a bit more.

src/core/issues-notes.js
devanshbatra04: Thanks! This was interesting.

devanshbatra04:
Look forward to the review! :)

src/core/issues-notes.js
```diff
@@ -10,33 +10,32 @@
...
+  Array.prototype.slice
+    .call(ins)
+    .filter(issue => issue.parentElement)
+    .forEach(inno => {
+      const isIssue = inno.classList.contains("issue");
+      const isWarning = inno.classList.contains("warning");
+      const isEdNote = inno.classList.contains("ednote");
+      const isFeatureAtRisk = inno.classList.contains("atrisk");
+      const isInline = inno.localName === "span";
+      const dataNum = inno.getAttribute("data-number");
```
marcoscaceres:
```suggestion
      const { number:dataNum } = inno.dataset;
```

src/core/issues-notes.js
```diff
@@ -77,31 +69,36 @@ function handleIssues(ins, ghIssues, conf) {
...
+                console.log(a.parentNode);
               } else if (isFeatureAtRisk && conf.atRiskBase) {
-                $tit
-                  .find("span")
-                  .wrap($("<a href='" + conf.atRiskBase + dataNum + "'/>"));
+                const span = tit.querySelector("span");
+                const a = hyperHTML`<a href='${conf.atRiskBase + dataNum}'/>`;
+                span.parentNode.insertBefore(a, span);
+                a.append(span);
+                console.log(a.parentNode);
```
marcoscaceres: stray console statement

src/core/issues-notes.js
```diff
@@ -146,42 +140,47 @@ function handleIssues(ins, ghIssues, conf) {
...
+        inno.replaceWith(div);
+        body.classList.remove(report.type);
+        body.removeAttribute("data-number");
+        if (ghIssue && !body.innerHTML.trim()) {
+          body = hyperHTML`${ghIssue.body_html}`;
         }
-        $div.append(body);
-        const level = $tit.parents("section").length + 2;
-        $tit.attr("aria-level", level);
+        console.log(body);
```
marcoscaceres: Please remove this one too :)

src/core/issues-notes.js
```diff
@@ -10,33 +10,32 @@
...
-      const isIssue = $inno.hasClass("issue");
-      const isWarning = $inno.hasClass("warning");
-      const isEdNote = $inno.hasClass("ednote");
-      const isFeatureAtRisk = $inno.hasClass("atrisk");
-      const isInline = $inno[0].localName === "span";
-      const dataNum = $inno.attr("data-number");
+  const issueSummary = hyperHTML`<div><h2>${
+    conf.l10n.issue_summary
+  }</h2><ul></ul></div>`;
+  const issueList = issueSummary.querySelector("ul");
```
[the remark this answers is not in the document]
saschanaz: @marcoscaceres `querySelector` does not return a NodeList 😁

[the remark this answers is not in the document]
marcoscaceres: Oh, woops! comment probably still applies for ` ins` tho.

src/core/issues-notes.js
devanshbatra04: Well that's embarrassing. Thanks!

src/core/issues-notes.js
devanshbatra04: Interesting. Thank you.

src/core/issues-notes.js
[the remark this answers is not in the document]
devanshbatra04: wow, thanks!

src/core/issues-notes.js
```diff
@@ -146,42 +137,46 @@ function handleIssues(ins, ghIssues, conf) {
...
-      $("#issue-summary").append($issueSummary.contents());
-  } else if ($("#issue-summary").length) {
+  const issueSummaryElement = document.getElementById("issue-summary");
+  if (document.querySelectorAll(".issue").length) {
+    if (issueSummaryElement)
+      issueSummaryElement.innerHTML = issueSummary.innerHTML;
+  } else if (issueSummaryElement) {
     pub("warn", "Using issue summary (#issue-summary) but no issues found.");
-    $("#issue-summary").remove();
+    issueSummaryElement.parentNode.removeChild(issueSummaryElement);
```
saschanaz: You can instead do `issueSummaryElement.remove()` even without jQuery 😀

src/core/issues-notes.js
```diff
@@ -146,42 +137,46 @@ function handleIssues(ins, ghIssues, conf) {
...
+        body.classList.remove(report.type);
+        body.removeAttribute("data-number");
+        if (ghIssue && !body.innerHTML.trim()) {
+          body = hyperHTML`${ghIssue.body_html}`;
         }
-        $div.append(body);
-        const level = $tit.parents("section").length + 2;
-        $tit.attr("aria-level", level);
+        div.append(body);
+        const level = getCount(tit);
```
saschanaz: We recently added `util.parents()` function to do this. So you may do `parents(tit, "section") + 2`.

src/core/issues-notes.js
```diff
@@ -112,26 +106,23 @@ function handleIssues(ins, ghIssues, conf) {
...
+                  report.title
+                }</span>`
               );
             }
-            $issueList.append($li);
+            issueList.append(li);
           }
         }
-        $tit.find("span").text(text);
+        tit.querySelector("span").innerHTML = text;
```
saschanaz: IMO `.textContent` should be preferred as we know it's a plain text.

src/core/issues-notes.js
devanshbatra04: lol, it gets confusing sometimes. Thanks!

src/core/issues-notes.js
devanshbatra04: gotcha. Thanks.

devanshbatra04:
Should be good to go. Learnt a lot in my first PR. Thanks a ton @sidvishnoi @marcoscaceres @saschanaz :)

src/core/issues-notes.js
```diff
@@ -10,33 +10,31 @@
...
-      const isEdNote = $inno.hasClass("ednote");
-      const isFeatureAtRisk = $inno.hasClass("atrisk");
-      const isInline = $inno[0].localName === "span";
-      const dataNum = $inno.attr("data-number");
+  const issueSummary = hyperHTML`<div><h2>${
+    conf.l10n.issue_summary
+  }</h2><ul></ul></div>`;
+  const issueList = issueSummary.querySelector("ul");
+  [...ins]
+    .filter(issue => issue.parentElement)
```
saschanaz: nit: We probably don't need this filter because any element cannot lack a parent. Not a blocker, though.

devanshbatra04: Yup. Makes sense. I was just rewriting the old code which used it. 
Updated now.

src/core/issues-notes.js
```diff
@@ -10,178 +10,169 @@
...
+      report.title = inno.getAttribute("title");
+      if (isIssue) {
+        if (hasDataNum) {
+          if (dataNum) {
+            text += " " + dataNum;
+            // Set issueBase to cause issue to be linked to the external issue tracker
+            if (!isFeatureAtRisk && issueBase) {
+              const span = tit.querySelector("span");
+              const a = hyperHTML`<a href='${issueBase + dataNum}'/>`;
+              span.parentNode.insertBefore(a, span);
```
sidvishnoi:
```suggestion
              span.before(a);
```

src/core/issues-notes.js
```diff
@@ -10,178 +10,169 @@
...
-      const isEdNote = $inno.hasClass("ednote");
-      const isFeatureAtRisk = $inno.hasClass("atrisk");
-      const isInline = $inno[0].localName === "span";
-      const dataNum = $inno.attr("data-number");
-      const report = {
-        inline: isInline,
-      };
-      report.type = isIssue
-        ? "issue"
+  const issueSummary = hyperHTML`<div><h2>${
```
sidvishnoi: Nit: This might be prettier:
``` js
const issueSummary = hyperHTML`
  <div>
    <h2>${conf.l10n.issue_summary}</h2>
    <ul></ul>
  </div>`
```

src/core/issues-notes.js
```diff
@@ -10,178 +10,169 @@
...
-        }
-        $tit.addClass("marker");
-        $div.append($tit);
-        $inno.replaceWith($div);
-        let body = $inno.removeClass(report.type).removeAttr("data-number");
-        if (ghIssue && !body.text().trim()) {
-          body = ghIssue.body_html;
+        if (report.number !== undefined) {
+          // Add entry to #issue-summary.
+          const li = hyperHTML`<li><a></a></li>`;
```
sidvishnoi: Lets use more of hyperHTML 😄 

``` js
const li = hyperHTML`
  <li>
    <a href="${"#" + div.id}">${conf.l10n.issue + " " + report.number}</a>
    ${report.title ? hyperHTML`<span style="text-transform: none">: ${report.title}</span>`  : ""}
  </li>`
issueList.append(li);
```

src/core/issues-notes.js
saschanaz: This ultimately needs a refactoring as it shouldn't predefine a ul and require querySelector to add items. So I would say it's okay for now...

src/core/issues-notes.js
saschanaz: AFAIK preceding whitespaces inside the string will generate a non-DOM object and will cause a failure.

sidvishnoi: (was just demonstrating we can use hyperHTML here, instead of direct DOM manipulation)
Lets leave these for another refactor.

src/core/issues-notes.js
```diff
@@ -10,178 +10,168 @@
...
+          const li = hyperHTML`
+  <li>
+    <a href="${"#" + div.id}">${conf.l10n.issue + " " + report.number}</a>
+    ${
+      report.title
+        ? hyperHTML`<span style="text-transform: none">: ${report.title}</span>`
+        : ""
+    }
+  </li>`;
+          issueList.append(li);
```
sidvishnoi:
```suggestion
```

sidvishnoi: How are the tests passing even when we have appended twice?!

saschanaz: >How are the tests passing even when we have appended twice?!

Because appending a child will automatically disconnect it from the previous parent. A magic 😲

sidvishnoi: 😮 🤯

marcoscaceres:
Thanks for your hard work @devanshbatra04! 

@saschanaz, will wait for your ✅ before merging.

saschanaz:
I'm merging this, 5, 4, 3,..

saschanaz:
Yay!

PR #1971, opened 2018-12-22
fix(core/structure): fix undefined property access error

saschanaz:
This change just removes a console error message without any behavior change. I have no idea how to add a test for this, because it's totally expected that nothing occurs when no sections.

Any ideas?


marcoscaceres:
Maybe let's leave it untested for now? I don't think this comes up as problem in practice?

saschanaz:
Makes sense, I'll merge this as it is.

PR #1973, opened 2018-12-22
refactor(core/issues-notes): split functions and add JSDocs

saschanaz:
This PR includes two fixes for regressions after #1969.

### DIFF
--- a/src/core/issues-notes.js
+++ b/src/core/issues-notes.js
@@ -26,7 +26,7 @@ const MAX_GITHUB_REQUESTS = 60;
  * @param {*} conf
  */
 function handleIssues(ins, ghIssues, conf) {
-  const { issueBase, githubAPI } = conf;
+  const { issueBase } = conf;
   const hasDataNum = !!document.querySelector(".issue[data-number]");
   let issueNum = 0;
   const issueSummary = hyperHTML`<div><h2>${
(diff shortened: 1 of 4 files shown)


src/core/issues-notes.js
```diff
@@ -159,23 +136,58 @@ function handleIssues(ins, ghIssues, conf) {
...
+}
+
+/**
+ * @param {GitHubLabel[]} labels
+ * @param {string} title
+ * @param {string} repoURL
+ */
+function createLabelsGroup(labels, title, repoURL) {
+  const labelsGroup = Array.from(labels || [])
+    .map(label => {
```
sidvishnoi: Move this into `createLabel` below?

src/core/issues-notes.js
```diff
@@ -78,79 +88,46 @@ function handleIssues(ins, ghIssues, conf) {
...
-          }
-        } else {
+        if (!hasDataNum) {
           text += " " + issueNum;
+        } else if (dataNum) {
+          text += " " + dataNum;
+          // Set issueBase to cause issue to be linked to the external issue tracker
+          if (!isFeatureAtRisk && issueBase) {
+            const a = hyperHTML`<a href='${issueBase + dataNum}'/>`;
+            title.before(a);
```
sidvishnoi: I'm not sure if we need to do this (`title.before(a);`)

src/core/issues-notes.js
```diff
@@ -78,79 +88,46 @@ function handleIssues(ins, ghIssues, conf) {
...
-        tit.append(node);
-        inno.removeAttribute("title");
-      } else if (report.title) {
-        tit.append(
-          hyperHTML`<span style='text-transform: none'>: ${report.title}</span>`
-        );
+      title.textContent = text;
+      if (report.title) {
         inno.removeAttribute("title");
+        let labels = [];
```
sidvishnoi: Nit:
```suggestion
        const labels = ghIssue && githubAPI ? ghIssue.labels : [];
```

saschanaz:
I'm locking this PR. Sorry for too many commits after requesting a review 😭

src/core/issues-notes.js
sidvishnoi: I think as we're just adding a `href: issuesURL.href` property here and using it just inside `createLabel`, it makes sense to move this into `createLabel` instead of using a `.map`. WDYT?

saschanaz: Please open a new conversation rather than continuing outdated conversation.

Anyway, done! 😁

PR #1974, opened 2018-12-22
feat(core/pluralize): add singularize capability

### LINKED ISSUE #1743: (core/pluralize): Add support for singularize
It would be nice if pluralization also works for singularizing terms. For example,
Instead of `<dfn data-lt="bar">bars</dfn>`, we could write `<dfn>bars</dfn>` but still refer to it as `<a>bar</a>` or `<a>bars</a>`.

Presently it works opposite way only. i.e.
`<dfn>bar</dfn>` can be refereed as `<a>bar</a>` or `<a>bars</a>`, but `<dfn>bars</dfn>` cannot be referred as `<a>bar</a>`.

devanshbatra04:
Initial commit for adding singularize capability. I am pretty sure more refactor is required. Reviews appreciated @sidvishnoi @saschanaz @marcoscaceres 

Closes #1743

### DIFF
--- a/src/core/ui.js
+++ b/src/core/ui.js
@@ -132,13 +132,8 @@ export const ui = {
   enable() {
     respecPill.removeAttribute("disabled");
   },
-  addCommand(label, module, keyShort, icon) {
+  addCommand(label, handler, keyShort, icon) {
     icon = icon || "";
-    const handler = function() {
-      require([module], mod => {
-        mod.show();
-      });
-    };
     const id = "respec-button-" + label.toLowerCase().replace(/\s+/, "-");
     const button = hyperHTML`<button id="${id}" class="respec-option" title="${keyShort}">
       <span class="respec-cmd-icon">${icon}</span> ${label}…
--- a/src/ui/about-respec.js
+++ b/src/ui/about-respec.js
@@ -10,7 +10,7 @@ const div = document.createElement("div");
 const render = hyperHTML.bind(div);
 const button = ui.addCommand(
   `About ${window.respecVersion}`,
-  "ui/about-respec",
+  show,
   "Ctrl+Shift+Alt+A",
   "ℹ️"
 );
@@ -81,5 +81,3 @@ function perfEntryToTR({ name, duration }) {
     </td>
   `;
 }
-
-export { show };
--- a/src/ui/dfn-list.js
+++ b/src/ui/dfn-list.js
@@ -7,7 +7,7 @@ import { ui } from "../core/ui";
 
 const button = ui.addCommand(
   l10n[lang].definition_list,
-  "ui/dfn-list",
+  show,
   "Ctrl+Shift+Alt+D",
   "📔"
 );
@@ -36,5 +36,3 @@ function show() {
   render`${definitionLinks}`;
   ui.freshModal(l10n[lang].list_of_definitions, ul, button);
 }
-
-export { show };
--- a/src/ui/save-html.js
+++ b/src/ui/save-html.js
@@ -74,13 +74,13 @@ let button;
 if (supportsDownload) {
   button = ui.addCommand(
     l10n[lang].save_snapshot,
-    "ui/save-html",
+    show,
     "Ctrl+Shift+Alt+S",
     "💾"
   );
 }
 
-export function show() {
+function show() {
   if (!supportsDownload) return;
   saveDialog.show(button);
 }
--- a/src/ui/search-specref.js
+++ b/src/ui/search-specref.js
@@ -7,7 +7,7 @@ import { wireReference } from "../core/biblio";
 
 const button = ui.addCommand(
   l10n[lang].search_specref,
-  "ui/search-specref",
+  show,
   "Ctrl+Shift+Alt+space",
   "🔎"
 );
@@ -160,5 +160,3 @@ function render({ state, results, timeTaken, query } = { state: "" }) {
   }</section>
   `;
 }
-
-export { show };
--- a/tests/spec/core/biblio-spec.js
+++ b/tests/spec/core/biblio-spec.js
@@ -150,11 +150,11 @@ describe("W3C — Bibliographic References", () => {
 
   it("makes sure that normative references win irrespective of case", () => {
     expect(doc.querySelectorAll("#bib-dom").length).toBe(1);
-    const domRef = doc.querySelector("#bib-dom");
+    const domRef = doc.getElementById("bib-dom");
     expect(domRef.closest("section").id).toBe("normative-references");
 
     expect(doc.querySelectorAll("#bib-fetch").length).toBe(1);
-    const fetchRef = doc.querySelector("#bib-fetch");
+    const fetchRef = doc.getElementById("bib-fetch");
     expect(fetchRef.closest("section").id).toBe("normative-references");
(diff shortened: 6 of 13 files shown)


src/core/pluralize.js
```diff
@@ -35,10 +41,20 @@ export function run(conf) {
...
         registerDefinition(dfn, uniquePlurals);
       }
+      if (singulars.length) {
+        const userDefinedPlurals = dfn.dataset.singulars
+          ? dfn.dataset.singulars.split("|")
+          : [];
+        const uniqueSingulars = [
+          ...new Set([...userDefinedPlurals, ...singulars]),
+        ];
+        dfn.dataset.singulars = uniqueSingulars.join("|");
```
sidvishnoi: We would prefer to not add new data attributes, so lets merge `uniquePlurals` and `uniqueSingulars` into `uniquePlurals` itself (and add to `dfn.dataset.plurals`)

src/core/pluralize.js
```diff
@@ -58,9 +74,16 @@ function getPluralizer() {
...
   // returns pluralized term if `text` needs pluralization, "" otherwise
-  return function pluralizeDfn(text) {
-    const normText = normalize(text.toLowerCase());
-    const plural = pluralOf(normText);
-    return links.has(plural) && !dfns.has(plural) ? plural : "";
+  return {
+    pluralizeDfn: text => {
+      const normText = normalize(text.toLowerCase());
+      const plural = pluralOf(normText);
+      return links.has(plural) && !dfns.has(plural) ? plural : "";
```
sidvishnoi: Lets keep a single function and return an array of two items from it?

devanshbatra04: Done, it seems I only ever needed to change 1 line.

src/core/pluralize.js
devanshbatra04: This isn't required if we use a single function as you suggested.

src/core/pluralize.js
```diff
@@ -60,7 +64,9 @@ function getPluralizer() {
   // returns pluralized term if `text` needs pluralization, "" otherwise
```
sidvishnoi:
```suggestion
  // returns pluralized/singularized term if `text` needs pluralization/singularization, "" otherwise
```

sidvishnoi:
Brilliant! We just need to add tests for this now.

devanshbatra04:
Yup, haven't ever written one. Will need to study up a bit and will send asap.

devanshbatra04:
@sidvishnoi look forward to test reviews :)

tests/spec/core/pluralize-spec.js
```diff
@@ -225,3 +208,49 @@ describe("Core - Pluralize", () => {
     expect(badLink.classList.contains("respec-offending-element")).toBeTruthy();
   });
 });
+
+it("doesn't mishandle <dfn> referring plural nouns without singulars", async () => {
```
sidvishnoi: Nit: Move the tests inside the `describe` block

tests/spec/core/pluralize-spec.js
```diff
@@ -189,23 +189,6 @@ describe("Core - Pluralize", () => {
     ).toBeTruthy();
   });
 
-  it("doesn't add pluralization when no <a> references plural term", async () => {
```
sidvishnoi: any reason to remove this test?

sidvishnoi: Lets just add a test `"doesn't add singularization when no <a> references singular term"` and keep this here

devanshbatra04: Doesn't make sense. Essentially now either pluralization or singularization will be applied ( both being called Pluralization). We will always have one form of pluralization, which is ensured in the first test itself.

devanshbatra04: My Bad, I seem to have made a very bad error, I accidentally pushed it out of the describe block. Correcting.

sidvishnoi: We still need to ensure that both paths in code are tested.
So, lets try these:
``` html
<dfn data-lt="baz">bar</dfn> can be referenced as <a>baz</a> or <a>bar</a> 
so dataset.plurals should be empty as we're not referencing "bazs" or "bars"
```

``` html
<dfn data-lt="tables">chairs</dfn> can be referenced as <a>chairs</a> or <a>tables</a> 
so dataset.plurals should be empty as we're not referencing "table" or "chair".
```

But in above, if we respectively refer to plural/singular terms also, we should have them in `dataset.plurals`. Makes sense?

devanshbatra04: Yes, I did get your point. Both path are technically the same, so the same test works for both. but I will add the test as a good security measure. Really appreciate going out of your way to make things clear :)

devanshbatra04:
@sidvishnoi could you please have a quick look here as well? just the test file :smile:

tests/spec/core/pluralize-spec.js
```diff
@@ -206,6 +235,23 @@ describe("Core - Pluralize", () => {
...
+      </section>
+    `;
+    const ops = makeStandardOps({ pluralize: true }, body);
+    const doc = await makeRSDoc(ops);
+
+    const dfn = doc.querySelector("#section dfn");
+    expect(dfn.id).toEqual("dfn-tables");
+    const dfnlt = dfn.dataset.lt.split("|").sort();
+    const expectedDfnlt = "chairs|tables".split("|"); // no "table" here
+    expect(dfnlt).toEqual(expectedDfnlt);
```
sidvishnoi: We should also test for value of `dfn.dataset.plurals`.

tests/spec/core/pluralize-spec.js
```diff
@@ -189,6 +189,35 @@ describe("Core - Pluralize", () => {
     ).toBeTruthy();
   });
 
+  it("doesn't mishandle <dfn> referring plural nouns without singulars", async () => {
```
sidvishnoi: I think we don't need this test as we don't need to test what `pluralize` package gives us. 
Unless `scissors` and `trousers` are names of potential APIs.

devanshbatra04:
@sidvishnoi I changed the second test a bit, adding plural and singular terms in the same case, just for more scrutiny.

tests/spec/core/pluralize-spec.js
```diff
@@ -3,40 +3,60 @@ describe("Core - Pluralize", () => {
...
       <section id="section">
-        <dfn data-lt="baz">bar</dfn> can be referenced
+        <dfn data-lt="baz">bars</dfn> can be referenced
         as <a>baz</a>
         or <a>bar</a>
         or <a>bars</a>
         or <a>bazs</a>
         but not as <a id="ignored-link" href="/PASS">bar</a>
       </section>
+      <section>
```
sidvishnoi: Nit: this section is redundant

tests/spec/core/pluralize-spec.js
sidvishnoi: Lets remove this test

devanshbatra04: done. thanks.

tests/spec/core/pluralize-spec.js
devanshbatra04: whoops.

PR #1978, opened 2018-12-28
refactor(tests/test-main): remove redundant lines

saschanaz:
👍

### DIFF
--- a/tests/test-main.js
+++ b/tests/test-main.js
@@ -7,11 +7,7 @@ const testFiles = Object.keys(window.__karma__.files)
   // Normalize paths to RequireJS module names.
   // If you require sub-dependencies of test files to be loaded as-is (requiring file extension)
   // then do not normalize the paths
-  .map(file => file.replace(/^\/base\//, "").replace(/\.js$/, ""))
-  .reduce((collector, file) => {
-    collector.push(file);
-    return collector;
-  }, []);
+  .map(file => file.replace(/^\/base\//, "").replace(/\.js$/, ""));
 
 require.config({
   // Karma serves files under /base, which is the basePath from your config file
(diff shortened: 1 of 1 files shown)


tests/test-main.js
```diff
@@ -7,11 +7,7 @@ const testFiles = Object.keys(window.__karma__.files)
   // Normalize paths to RequireJS module names.
   // If you require sub-dependencies of test files to be loaded as-is (requiring file extension)
   // then do not normalize the paths
-  .map(file => file.replace(/^\/base\//, "").replace(/\.js$/, ""))
-  .reduce((collector, file) => {
```
sidvishnoi: lol why did we even have this?! 🤣

tests/test-main.js
```diff
@@ -21,28 +17,7 @@ require.config({
...
-    "w3c/linter-rules/privsec-section":
-      "/base/js/w3c/linter-rules/privsec-section",
-    "w3c/l10n": "/base/js/w3c/l10n",
-    "core/LinterRule": "/base/js/core/LinterRule",
-    "core/biblio-db": "/base/js/core/biblio-db",
-    "core/jquery-enhanced": "/base/js/core/jquery-enhanced",
-    "core/pubsubhub": "/base/js/core/pubsubhub",
-    "core/utils": "/base/js/core/utils",
-    "deps/jquery": "/base/js/deps/jquery",
-    "deps/marked": "/base/js/deps/marked",
```
sidvishnoi: Question: Why `deps` aren't needed?

saschanaz: Because we don't directly refer `deps` inside tests. Some tests does refer `core/utils` and `w3c/linter` so we need them.

PR #1979, opened 2018-12-28
refactor(core/tests): remove jQuery from core/issues-notes and w3c/headers-spec

devanshbatra04:
So I am still to work on ```w3c/headers-spec``` However it has a lot of occurrences of ```:contain``` selectors (55 to be exact). Like ```$("dt:contains('Editor:')", doc)```. 

How should we handle these? completely rewrite these tests or define a new utils function for contains?
I found similar occurrences in a file earlier, @saschanaz suggested we could manually point to those elements for test purposes. Doing that here seems much of a chore at first. 

How should we go about it? @sidvishnoi @saschanaz


devanshbatra04:
@sidvishnoi @saschanaz @marcoscaceres Reviews would be great :) (might be tricky though)

tests/spec/core/issues-notes-spec.js
```diff
@@ -1,5 +1,12 @@
 "use strict";
 describe("Core — Issues and Notes", () => {
+  let utils;
+  beforeAll(done => {
```
marcoscaceres: I don't think you need this import. Instead, for example, you can just do the following below without `parents()`:

```JS
expect(spiss.closest("div")).toBe(null);
```

Also, "toEqual()" means slightly different thing to `.toBe()`. In general, use `.toBe()` because it means "===". Some nice details here:
https://stackoverflow.com/questions/22413009/jasmine-javascript-testing-tobe-vs-toequal

tests/spec/core/issues-notes-spec.js
```diff
@@ -43,36 +50,37 @@ describe("Core — Issues and Notes", () => {
...
+    const atr = doc.querySelector("div.atrisk");
+    const spiss = doc.querySelector("span.issue");
+    const spatr = doc.querySelector("span.atrisk");
+    const not = doc.querySelector("div.note");
+    const spnot = doc.querySelector("span.note");
 
-    expect($spiss.parent("div").length).toEqual(0);
-    expect($spatr.parent("div").length).toEqual(0);
-    expect($spnot.parent("div").length).toEqual(0);
+    expect(utils.parents(spiss, "div").length).toEqual(0);
```
marcoscaceres: Comment above applies to these checks.
```suggestion
    expect(spiss.closest("div")).toBeNull();
```
Same with the next two below...

tests/spec/core/issues-notes-spec.js
```diff
@@ -84,14 +92,13 @@ describe("Core — Issues and Notes", () => {
         "<p class='ednote' title='EDNOTE-TIT'>EDNOTE</p>",
     };
     const doc = await makeRSDoc(ops);
-    const $not = $("div.ednote", doc);
-    const $pnot = $not.find("p");
-    expect($not.find("div.ednote-title").length).toEqual(1);
-    expect($not.find("div.ednote-title").text()).toEqual(
+    const not = doc.querySelector("div.ednote");
```
marcoscaceres: nit: let fix the naming of the variable, call it `const edNote`.

tests/spec/core/issues-notes-spec.js
devanshbatra04: Gotcha. Thanks!

devanshbatra04:
Thanks for the review @marcoscaceres . Did you look at w3c/headers-spec.js too? cant believe it didn't have any changes. 
BTW Happy New Year! :smile:

tests/spec/w3c/headers-spec.js
```diff
@@ -998,27 +997,24 @@ describe("W3C — Headers", () => {
...
-          .next("dd")
-          .text()
-      ).toMatch(/http:\/\/LATEST/);
-      const $sotd = $("#sotd", doc);
-      expect($sotd.find("a[href='http://WG']").length).toEqual(1);
-      expect($sotd.find("a:contains(WGNAME)").length).toEqual(1);
+      expect(doc.querySelector(".head h2").textContent).toMatch(
+        /Final Business Group Report/
+      );
+      expect(doc.querySelectorAll("dd")[0].textContent).toMatch(
```
saschanaz: Could we additionally make sure that the preceding `dt` contains `"This version":`?

devanshbatra04: I just couldn't find any direct JS function, I thought the HTML structure wouldn't change. Is there a way to do that without manually iterating?
If not, I think I will make a util function for this file, as there are a couple more such cases in the file.

marcoscaceres: Unfortunately, you might need to add the util function.

devanshbatra04: We need to substitute for the `.next("...")` jquery function. It selects the next sibling which satisfies the selector passed. There are several `.next` occurences in the original `headers-spec.js` file. I remember doing something similar using `.nextElementSibling`, it didn't work out in some of them, as it could be any subsequent sibling, not necessarily the next. 

I think we do need the looping function.

saschanaz: `.next("...")` is actually looser because it doesn't check `<dd>` and `<dt>` are direct neighbor siblings. My thought:

```js
const terms = doc.querySelectorAll("dt");
expect(terms[0].textContent).toContain("...");
expect(terms[0].nextElementSibling.localName).toBe("dd");
expect(terms[0].nextElementSibling.textContent).toContain("...");
```

saschanaz: IMO `toMatch` should be replaced with `toContain` as we don't need regex here.

devanshbatra04: Yep it works, previously I had used the same approach using ```.nextSibling()``` My bad :sweat:

saschanaz:
Please remove build results (`builds/*`) as they are only updated when version bump.

tests/spec/core/issues-notes-spec.js
```diff
@@ -84,14 +83,13 @@ describe("Core — Issues and Notes", () => {
...
-    expect($not.find("div.ednote-title").text()).toEqual(
+    const edNote = doc.querySelector("div.ednote");
+    expect(edNote.querySelectorAll("div.ednote-title").length).toBe(1);
+    expect(edNote.querySelector("div.ednote-title").textContent).toBe(
       "Editor's note: EDNOTE-TIT"
     );
-    expect($pnot.attr("title")).toBeUndefined();
-    expect($pnot.text()).toEqual("EDNOTE");
+    expect(edNote.querySelector("p").getAttribute("title")).toBeNull();
+    expect(edNote.querySelector("p").textContent).toBe("EDNOTE");
```
saschanaz: Do we have a specific reason to remove variables and prefer inlining? Personally I would prefer the reverse way...

devanshbatra04: No specific reason. Should I rewrite them?

saschanaz: Repeated inline code is bad for readability, so yes for where the original had variables. Probably okay for where the original didn't, to ease reviewing 😁

devanshbatra04: Gotcha!

tests/spec/w3c/headers-spec.js
```diff
@@ -29,7 +32,9 @@ describe("W3C — Headers", () => {
       };
       Object.assign(ops.config, newProps);
       const doc = await makeRSDoc(ops);
-      expect($(".head h2", doc).text()).toMatch(/W3C Editor's Draft/);
+      expect(doc.querySelector(".head h2").textContent).toMatch(
```
saschanaz: nit: `toContain()`

tests/spec/w3c/headers-spec.js
```diff
@@ -42,16 +47,14 @@ describe("W3C — Headers", () => {
...
-      expect(
-        $("dt:contains('Latest published version:')", doc)
-          .next("dd")
-          .text()
-      ).toMatch(/\/TR\/xxx\//);
+
+      const terms = doc.querySelectorAll("dt");
+      expect(terms[0].textContent).toBe("This version:");
+      expect(terms[0].nextElementSibling.localName).toBe("dd");
+      expect(terms[0].nextElementSibling.textContent).toMatch(/\/REC-xxx-/);
```
saschanaz: Also here

tests/spec/w3c/headers-spec.js
```diff
@@ -42,16 +47,14 @@ describe("W3C — Headers", () => {
...
+      expect(terms[1].textContent).toBe("Latest published version:");
+      expect(terms[1].nextElementSibling.localName).toBe("dd");
+      expect(terms[1].nextElementSibling.textContent).toMatch(/\/TR\/xxx\//);
```
saschanaz: Here too

tests/spec/w3c/headers-spec.js
```diff
@@ -75,17 +78,21 @@ describe("W3C — Headers", () => {
...
+      expect(dd.querySelector("a[href='mailto:EMAIL']").textContent).toBe(
+        "NAME"
+      );
       // if `mailto` is specified in People, `url` won't be used
-      expect($dd.find("a[href='http://URI']").length).toEqual(0);
-      expect($dd.get(0).dataset.editorId).toEqual("1234");
-      expect($dd.text()).toMatch(/\(NOTE\)/);
+      expect(dd.querySelectorAll("a[href='http://URI']").length).toBe(0);
+      expect(dd.dataset.editorId).toBe("1234");
+      expect(dd.textContent).toMatch(/\(NOTE\)/);
```
saschanaz: Here too

tests/spec/w3c/headers-spec.js
```diff
@@ -103,11 +110,10 @@ describe("W3C — Headers", () => {
...
       Object.assign(ops.config, newProps);
       const doc = await makeRSDoc(ops);
-      expect($("dt:contains('Editors:')", doc).length).toEqual(1);
-      expect($("dt:contains('Editor:')", doc).length).toEqual(0);
-      const $dd = $("dt:contains('Editors:')", doc).next("dd");
-      expect($dd.text()).toEqual("NAME1");
-      expect($dd.next("dd").text()).toEqual("NAME2");
+      expect(contains(doc, "dt", "Editors:").length).toBe(1);
+      expect(contains(doc, "dt", "Editor:").length).toBe(0);
+      expect(doc.querySelectorAll("dd")[5].textContent).toBe("NAME1");
```
saschanaz: I think this line is to check `NAME1` is exactly after `Editors`, so probably using `nextElementSibling` is better.

tests/spec/w3c/headers-spec.js
```diff
@@ -328,13 +334,12 @@ describe("W3C — Headers", () => {
       };
       Object.assign(ops.config, newProps);
       const doc = await makeRSDoc(ops);
-      expect($("dt:contains('Authors:')", doc).length).toEqual(0);
-      expect($("dt:contains('Author:')", doc).length).toEqual(1);
-      const $dd = $("dt:contains('Author:')", doc).next("dd");
-      expect($dd.text()).toEqual("NAME1");
+      expect(contains(doc, "dt", "Authors:").length).toBe(0);
+      expect(contains(doc, "dt", "Author:").length).toBe(1);
+      expect(doc.querySelectorAll("dd")[6].textContent).toBe("NAME1");
```
saschanaz: Also `nextElementSibling` here.

tests/spec/w3c/headers-spec.js
```diff
@@ -741,7 +757,9 @@ describe("W3C — Headers", () => {
       };
       Object.assign(ops.config, newProps);
       const doc = await makeRSDoc(ops);
-      expect($(".head .copyright", doc).text()).toMatch(/1977-2012/);
+      expect(doc.querySelector(".head .copyright").textContent).toMatch(
+        /1977-2012/
```
saschanaz: `toContain`

tests/spec/w3c/headers-spec.js
```diff
@@ -752,7 +770,9 @@ describe("W3C — Headers", () => {
       };
       Object.assign(ops.config, newProps);
       const doc = await makeRSDoc(ops);
-      expect($(".head .copyright", doc).text()).not.toMatch(/2012-2012/);
+      expect(doc.querySelector(".head .copyright").textContent).not.toMatch(
```
saschanaz: `toContain`

tests/spec/w3c/headers-spec.js
```diff
@@ -840,34 +862,22 @@ describe("W3C — Headers", () => {
...
-          .first()
-          .attr("href")
-      ).toEqual("WGURI2");
-      expect(
-        $sotd
-          .find("a:contains('WGNAME2')")
-          .last()
-          .attr("href")
-      ).toEqual("WGPATENT2");
-      expect($sotd.find("a:contains('disclosures')").length).toEqual(2);
```
saschanaz: Looks like this line has no equivalent.

tests/spec/w3c/headers-spec.js
```diff
@@ -998,30 +1016,30 @@ describe("W3C — Headers", () => {
...
+        "Final Business Group Report"
+      );
+      const terms = doc.querySelectorAll("dt");
+      expect(terms[0].textContent).toBe("This version:");
+      expect(terms[0].nextElementSibling.localName).toBe("dd");
+      expect(terms[0].nextElementSibling.textContent).toMatch(/http:\/\/THIS/);
+      expect(terms[1].textContent).toBe("Latest published version:");
+      expect(terms[1].nextElementSibling.localName).toBe("dd");
+      expect(terms[1].nextElementSibling.textContent).toMatch(
+        /http:\/\/LATEST/
```
saschanaz: `toContain`

tests/spec/w3c/headers-spec.js
```diff
@@ -998,30 +1016,30 @@ describe("W3C — Headers", () => {
...
-      expect($sotd.find("a[href='http://WG']").length).toEqual(1);
-      expect($sotd.find("a:contains(WGNAME)").length).toEqual(1);
+      ).toBe(1);
+      expect(doc.querySelector(".head h2").textContent).toContain(
+        "Final Business Group Report"
+      );
+      const terms = doc.querySelectorAll("dt");
+      expect(terms[0].textContent).toBe("This version:");
+      expect(terms[0].nextElementSibling.localName).toBe("dd");
+      expect(terms[0].nextElementSibling.textContent).toMatch(/http:\/\/THIS/);
```
saschanaz: `toContain`

tests/spec/w3c/headers-spec.js
```diff
@@ -1026,11 +1033,11 @@ describe("W3C — Headers", () => {
       const terms = doc.querySelectorAll("dt");
       expect(terms[0].textContent).toBe("This version:");
       expect(terms[0].nextElementSibling.localName).toBe("dd");
-      expect(terms[0].nextElementSibling.textContent).toMatch(/http:\/\/THIS/);
+      expect(terms[0].nextElementSibling.textContent).toMatch("http://THIS");
```
saschanaz: Should be `toContain` to use string 🤣

Issue #1713, opened 2018-06-25
Add linter rule for internal slots


marcoscaceres:
Internal slots need to be preceded by a "."... so:

```HTML
<var>foo</var><a>[[\bar]]</a>.
```

Should make ReSpec upset.

marcoscaceres:
Note to self... something along the lines: 
```JS
[...document.querySelectorAll("var+a")].filter(a => a.textContent.startsWith("[[") && a.previousSibling.textContent !== "." )
```

saurabhmittal16:
Hi, I was interested in working on this, could you explain the issue and where to get started?
Thanks

sidvishnoi:
Hi @saurabhmittal16 👋 
First have a look at our [Developers Guide](https://github.com/w3c/respec/wiki/Developers-Guide). You'll need to create a lint rule in [core/linter-rules](https://github.com/w3c/respec/blob/develop/src/core/linter-rules/). 
For a reference on how to create a linter rule, have a look at [local-refs-exist linter rule](https://github.com/w3c/respec/blob/develop/src/core/linter-rules/local-refs-exist.js). It should provide plenty of insight. If not, we'll be here to help.

In this rule:

``` html
<!-- following should be valid -->
<var>foo</var>.<a>[[\bar]]</a>
<!-- but this should be invalid, hence provide a linter warning -->
<var>foo</var><a>[[\bar]]</a>
```

yatri1609:
@sidvishnoi @marcoscaceres I would like to work on this. Can I go ahead with this?

saurabhmittal16:
Hi @sidvishnoi, thanks for the information
I will update her as soon as possible

PR #1982, opened 2018-12-29
linter rule for internal slots

saurabhmittal16:
I am working on #1713. I created the linter rule following the other rule files. 
This `<var>bar</var>.<a>[[\Foo]]</a>` is valid but since the text inside the `<a>` does not match any `<dfn>`, warning of type `Found linkless <a> element with text "[[Foo]]" but no matching <dfn>` is thrown.

Could you review this file and suggest how to work on this issue? @saschanaz @sidvishnoi @marcoscaceres

### DIFF
--- a/src/core/linter-rules/internal-slot-warn.js
+++ b/src/core/linter-rules/internal-slot-warn.js
@@ -1,0 +1,50 @@
+/**
+ * Linter rule "internal-slot-warn".
+ */
+import LinterRule from "../LinterRule";
+import { lang as defaultLang } from "../l10n";
+
+const name = "internal-slot-warn";
+
+const meta = {
+  en: {
+    description: "Internal slots should be preceded by a '.'",
+    howToFix: "Add a '.' between the elements mentioned.",
+    help: "See developer console.",
+  },
+};
+
+// Fall back to english, if language is missing
+const lang = defaultLang in meta ? defaultLang : "en";
+
+/**
+ * Runs linter rule.
+ * @param {Object} config The ReSpec config.
+ * @param  {Document} doc The document to be checked.
+ */
+function linterFunction(conf, doc) {
+  const offendingElements = [...doc.querySelectorAll("var+a")].filter(
+    checkElement
+  );
+
+  if (!offendingElements.length) {
+    return [];
+  }
+
+  return {
+    name,
+    offendingElements,
+    occurrences: offendingElements.length,
+    ...meta[lang],
+  };
+}
+
+export const rule = new LinterRule(name, linterFunction);
+
+function checkElement(elem) {
+  return !(
+    elem.textContent.startsWith("[[") &&
+    elem.textContent.endsWith("]]") &&
+    elem.previousSibling.textContent === "."
+  );
+}
--- a/src/w3c/defaults.js
+++ b/src/w3c/defaults.js
@@ -3,6 +3,7 @@
  */
 export const name = "w3c/defaults";
 import { rule as checkPunctuation } from "../core/linter-rules/check-punctuation";
+import { rule as internalSlotWarning } from "../core/linter-rules/internal-slot-warn";
 import linter from "../core/linter";
 import { rule as localRefsExist } from "../core/linter-rules/local-refs-exist";
 import { rule as noHeadinglessSectionsRule } from "../core/linter-rules/no-headingless-sections";
@@ -14,7 +15,8 @@ linter.register(
   privsecSectionRule,
   noHeadinglessSectionsRule,
   checkPunctuation,
-  localRefsExist
+  localRefsExist,
+  internalSlotWarning
 );
 
 const cgbg = new Set(["BG-DRAFT", "BG-FINAL", "CG-DRAFT", "CG-FINAL"]);
@@ -62,6 +64,7 @@ const w3cDefaults = {
     "no-http-props": true,
     "check-punctuation": false,
     "local-refs-exist": true,
+    "internal-slot-warn": true,
   },
   pluralize: false,
   highlightVars: true,


src/w3c/defaults.js
```diff
@@ -62,6 +64,7 @@ const w3cDefaults = {
     "no-http-props": true,
     "check-punctuation": false,
     "local-refs-exist": true,
+    "internal-slot-warn": true,
```
marcoscaceres: We probably only want to enable this by default once we beta test it in a few specs.

src/core/linter-rules/internal-slot-warn.js
```diff
@@ -0,0 +1,50 @@
...
+    name,
+    offendingElements,
+    occurrences: offendingElements.length,
+    ...meta[lang],
+  };
+}
+
+export const rule = new LinterRule(name, linterFunction);
+
+function checkElement(elem) {
```
marcoscaceres: We can probably simplify this a bit by destructuring: 

```suggestion
function checkElement({textContent, previousSibling}) {
```

src/core/linter-rules/internal-slot-warn.js
```diff
@@ -0,0 +1,50 @@
...
+  return !(
+    elem.textContent.startsWith("[[") &&
+    elem.textContent.endsWith("]]") &&
```
marcoscaceres: I wonder if you need this `endsWith` check? Could it work without it?

marcoscaceres: It might be better to do this as a single RegExp... Something like: 

```js
const potentialSlot = previousSibling.textContent + textContent;
// starts with a dot, followed by [[ and a word. 
return  /^\.\[\[\w+/.test(potentialSlot);
```

saurabhmittal16:
Thanks, I'll work on this as per your suggestions

saurabhmittal16:
I changed the rule as per your suggestions. I mistakenly merged branch `develop` in my branch `lint-rule-slot` instead of rebasing. Hope that it's not an issue.
Also created the test, please review. Thanks.

Jonathanmarsh4:
This is all good to me

tests/spec/core/linter-rules/internal-slot-warn-spec.js
```diff
@@ -0,0 +1,49 @@
+"use strict";
+describe("Core Linter Rule - 'internal-slot-warn'", () => {
```
marcoscaceres:
```suggestion
describe("Core Linter Rule - 'check-internal-slots'", () => {
```

tests/spec/core/linter-rules/internal-slot-warn-spec.js
```diff
@@ -0,0 +1,49 @@
...
+  const ruleName = "internal-slot-warn";
```
marcoscaceres:
```suggestion
  const ruleName = "check-internal-slots";
```

tests/spec/core/linter-rules/internal-slot-warn-spec.js
```diff
@@ -0,0 +1,49 @@
...
+  const doc = document.implementation.createHTMLDocument("test doc");
+  beforeEach(() => {
+    // Make sure every unordered test get an empty document
+    // See: https://github.com/w3c/respec/pull/1495
+    while (doc.body.firstChild) {
+      doc.body.removeChild(doc.body.firstChild);
+    }
+  });
+
+  it("return error when no '.' between var and a for internal slots", async () => {
```
marcoscaceres: when writing "it()" functions always make sure the description reads like "it does X". So:

```suggestion
  it("only returns an error when there is no '.' between var and an internal slot", async () => {
```

tests/spec/core/linter-rules/internal-slot-warn-spec.js
```diff
@@ -0,0 +1,49 @@
...
+    expect(result.name).toEqual(ruleName);
+    expect(result.occurrences).toEqual(1);
+
+    const offendingElement = result.offendingElements[0];
+    const { previousSibling } = offendingElement;
+    // offending element's previous element won't have '.'
+    expect(previousSibling.textContent).not.toEqual(".");
+  });
+
+  it("no error when correct pattern is followed", async () => {
```
marcoscaceres:
```suggestion
  it("generates no error when the expected pattern is followed", async () => {
```

tests/spec/core/linter-rules/internal-slot-warn-spec.js
```diff
@@ -0,0 +1,49 @@
...
+    doc.body.innerHTML = `
+      <var>bar</var>.<a>[[foo]]</a>
+      <var>bar</var>.<a>[[foo]]</a>
```
marcoscaceres:
```suggestion
      <var>foo</var>.<a>[[bar]]</a>
```

tests/spec/core/linter-rules/internal-slot-warn-spec.js
marcoscaceres: this test is not different from the one above... i.e., `<var>bar2</var>.<a>[[foo]]</a>`)... so maybe we can remove this entirely. What we want to actually do here is maybe split the first test in two... one set of tests generates validation errors, and the other set of tests doesn't generate valuation  errors.

tests/spec/core/linter-rules/internal-slot-warn-spec.js
```diff
@@ -0,0 +1,49 @@
...
+    while (doc.body.firstChild) {
+      doc.body.removeChild(doc.body.firstChild);
+    }
+  });
+
+  it("return error when no '.' between var and a for internal slots", async () => {
+    doc.body.innerHTML = `
+    <var>bar1</var><a>[[foo]]</a>
+    <var>bar2</var>.<a>[[foo]]</a>
+    <var>bar3</var>.<a>[[foo]]</a>
```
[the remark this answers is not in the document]
saurabhmittal16:
```
<var>bar</var>.<a>[[foo</a>
```
According to the current rule, this case is not invalid. So should I change the RegExp to check for the closing ]] or is this a valid case? 
Thanks for the help.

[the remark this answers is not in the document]
marcoscaceres: yeah, let's go back to checking for "]]". Let's also assume that internal slots can't have spaces in the name for now.

src/core/linter-rules/check-internal-slots.js
```diff
@@ -0,0 +1,49 @@
...
+  },
+};
+
+// Fall back to english, if language is missing
+const lang = defaultLang in meta ? defaultLang : "en";
+
+/**
+ * Runs linter rule.
+ * @param {Object} config The ReSpec config.
+ * @param  {Document} doc The document to be checked.
```
marcoscaceres: Nit: whitespace..

```suggestion
 * @param {Document} doc The document to be checked.
```

tests/spec/core/linter-rules/check-internal-slots-spec.js
```diff
@@ -0,0 +1,61 @@
...
+    // Make sure every unordered test get an empty document
+    // See: https://github.com/w3c/respec/pull/1495
+    while (doc.body.firstChild) {
+      doc.body.removeChild(doc.body.firstChild);
+    }
+  });
+
+  it("only returns an error when there is no '.' between var and an internal slot", async () => {
+    doc.body.innerHTML = `
+    <var>bar</var><a>[[foo]]</a>
```
marcoscaceres: Small nit: could you kindly please indent these by two spaces? (similar to the ones below)

saurabhmittal16:
https://github.com/w3c/respec/blob/b9da7cb008e274c82765ecc314bacb5b64a481ed/tests/spec/w3c/defaults-spec.js#L4-L8
Should i update the two tests here since the Travis Build is failing due to these?

marcoscaceres:
> Should i update the two tests here since the Travis Build is failing due to these?

Ah, no, we just need to add your new rule name here:
https://github.com/w3c/respec/blob/b9da7cb008e274c82765ecc314bacb5b64a481ed/tests/spec/w3c/defaults-spec.js#L16

(defaulting to `false` for now)

saurabhmittal16:
Should I add the rule name? And how to test it on a real spec? Thanks for the help.

marcoscaceres:
> Should I add the rule name?

yep, just add `"check-internal-slots": false`. After line 16:
https://github.com/w3c/respec/blob/b9da7cb008e274c82765ecc314bacb5b64a481ed/tests/spec/w3c/defaults-spec.js#L16

That should make the test pass. 

> And how to test it on a real spec? 

For now, I'll just get you to manually confirm it. 
 
Easiest option is to edit the `examples/basic.html` file. There, inside one of the `<section>`s, add:

```HTML
<p>
  <dfn>[[\testSlot]]<dfn>
</p>

<p>
   <!-- this should generate a warning -->
   <var>foo</var><a>[[\testSlot]]</a>.
</p>
```

And just confirm the warning is there when you navigate to `localhost:5000/examples/basic.html` when you: 

```
npm start
```

After we release this, we check in the Payment Request spec.

saurabhmittal16:
I added the rule name in ```default-spec.js``` file and tested it on ```basic.html```. Everything seems to work fine. Thanks for the help.

marcoscaceres:
The override check was failing (given we added "check-internal-slots"), so I quickly added that for you. TravisCI should hopefully be happy now 🤞

marcoscaceres:
Released in respec@24.1.0. And beta tested it with Payment Request spec. 

However, unfortunately, it's not working as expected.

saurabhmittal16:
Thanks @marcoscaceres 
can i work on this to improve this?

marcoscaceres:
@saurabhmittal16, would appreciate it. If you clone:

https://github.com/w3c/payment-request/

And amend the `lint` configuration option at the top of `index.html` like so: 

```
lint: { 
  "check-internal-slots": true,
}
```

You can see that it overmatches on this linter rule. I didn't get a chance to figure out what was wrong - so if you can take a look, I'd really appreciate it.

saurabhmittal16:
@marcoscaceres 
Before filtering for offending elements, if we could filter the selected elements to only contain those which are internal slots (by checking for [[ and ]]) and then apply the checkElement function, we could avoid the extra elements. What do you think?

saurabhmittal16:
```
let x = document.querySelectorAll('var+a');
let y = Array.from(x);
let filtered = y.filter(
	elem => elem.innerText.startsWith('[[')
);
console.log(filtered.length);
offendingElements = filtered.filter(
	elem => elem.previousSibling.textContent !== '.'
);
```
I tried this on rough code on payment-request, return no offending elements and also tried by removing '.' in one of the slots, worked fine.

marcoscaceres:
The above sounds like a plan @saurabhmittal16!

marcoscaceres:
Just a quick JS tip (you might know this already): You can use the spread operator with query selector to convert the node list into an array, like so (which collapses the code right down): 

```JS
const offendingElements = [...document.querySelectorAll('var+a')].filter(
  elem =>  elem.innerText.startsWith('[[') && elem.previousSibling.textContent !== '.'
);
```

PR #1983, opened 2018-12-30
refactor(tests/spec/core): use getElementById

saschanaz:
Just a simple replacement.


saschanaz:
So this PR doesn't have the random highlighter problem. Good, or bad...

PR #1984, opened 2018-12-30
feat(core/data-cite): shows a warning on self citing

### LINKED ISSUE #1950: Warn when self-citing
## Important info

* URL to affected spec:
* ReSpec version: develop

## Description of problem

`USVString` links to http://localhost:8080/tests/spec/core/webidl.html#idl-USVString when locally opened.

But what I was expecting?:

It should always link to https://heycam.github.io/webidl/#idl-USVString.

CCing @sidvishnoi, any idea why? AFAIK this has been broken for a while, not sure since when.

devanshbatra04:
closes #1950 
Although I think we should also incorporate a way cite specs with such IDs conflicting with Short Name.


src/core/data-cite.js
```diff
@@ -26,6 +26,10 @@ function requestLookup(conf) {
     let href = "";
     // This is just referring to this document
     if (key.toLowerCase() === conf.shortName.toLowerCase()) {
+      showInlineWarning(
+        elem,
+        `couldn't cite external spec, reference ID "${key}" matches Short Name for current spec`
```
marcoscaceres:
```suggestion
        `Couldn't cite external document. The reference "${key}" is the same as the Short Name of current spec.`
```

marcoscaceres:
> Although I think we should also incorporate a way cite specs with such IDs conflicting with Short Name.

The way to do it is just: `<a href="#hi">hi</a>`

src/core/data-cite.js
devanshbatra04: Edited. Thank you :)

devanshbatra04:
Okay got it. Hope it's good to go now :)

marcoscaceres:
> tl;dr: This PR ultimately prevents shortName: "webidl" and I don't like it...

Understood, but Isn't "webidl" a special case? It's a foundational spec whose short name is already squatted.

saschanaz:
>It's a foundational spec whose short name is already squatted.

Right... I just feel bad to prevent any existing specs from using ReSpec. Maybe we can just log rather than warn so that we can know what happened when debugging.

marcoscaceres:
> Right... I just feel bad to prevent any existing specs from using ReSpec. Maybe we can just log rather than warn so that we can know what happened when debugging.

Just log is fine. I'm also a bit uncomfortable with not being able to self cite, tbh. The only thing that really matters here is that the (self) reference doesn't end up in the bibliography - which we have already solved for.

src/core/data-cite.js
```diff
@@ -26,6 +26,10 @@ function requestLookup(conf) {
     let href = "";
     // This is just referring to this document
     if (key.toLowerCase() === conf.shortName.toLowerCase()) {
+      showInlineWarning(
```
marcoscaceres: the suggestion is to replace this with just `console.warn()`

saschanaz: Could it be `console.log`, as respec2html still will potentially throw if warnings exist?

marcoscaceres: good point. Yeah, `console.log()` is better here.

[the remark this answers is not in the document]
marcoscaceres: updated this, so we can get this merged.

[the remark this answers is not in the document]
devanshbatra04: Sorry I have been recovering from a dislocated ankle, was in rehab the past few days. Will work on the other PRs as well today. Thanks again :)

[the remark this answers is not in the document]
marcoscaceres: oh, really sorry to hear that! No stress. Please take it easy and get well soon!

src/core/data-cite.js
```diff
@@ -27,6 +27,10 @@ function requestLookup(conf) {
     let href = "";
     // This is just referring to this document
     if (key.toLowerCase() === conf.shortName.toLowerCase()) {
+      console.log(
+        elem,
+        `Couldn't cite external document. The reference "${key}" is the same as the Short Name of current spec.`
```
saschanaz: The remaining thing is to revise the message as the requirement is changed. My suggestion:

```suggestion
        `The reference "${key}" is resolved into the current document per \`conf.shortName\`.`
```

saschanaz: Oops, already merged

saschanaz: @marcoscaceres Mind a quick fix? 😀

PR #1985, opened 2018-12-31
refactor(tests/spec/SpecHelper): avoid call()

saschanaz:
as `call()` makes code hard to read 🤔

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1429,6 +1429,7 @@
       "version": "0.4.2",
       "resolved": "https://registry.npmjs.org/boom/-/boom-0.4.2.tgz",
       "integrity": "sha1-emNune1O/O+xnO9JR6PGffrukRs=",
+      "optional": true,
       "requires": {
         "hoek": "0.9.x"
       }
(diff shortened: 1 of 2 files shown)


saschanaz:
This should also fix the problem in #1981, where ReSpec is incorrectly loaded twice.

Inserting ReSpec twice has not been a problem with requirejs as 1) previously ReSpec has been a set of AMD modules 2) defining an AMD module twice does not overwrite the previous one. With webpack ReSpec is not a set of modules anymore, so it becomes a problem.

tests/spec/SpecHelper.js
```diff
@@ -51,34 +51,38 @@ function decorateDocument(doc, opts) {
...
     Object.keys(loadAttr)
       .reduce(intoAttributes.bind(loadAttr), loader)
       .classList.add("remove");
-    this.appendChild(config);
+    doc.head.appendChild(loader);
+  }
+
+  function addRespecConfig(opts) {
+    const config = doc.createElement("script");
+    let configText = "";
```
sidvishnoi: Nit:
``` js
    const configText = opts.config
      ? `var respecConfig = ${JSON.stringify(opts.config || {})};`
      : "";
```

Issue #1905, opened 2018-11-23
Use @ts-check wherever appropriate


saschanaz:
It enables strict type check for JavaScripts on any TypeScript-based IDE (e.g. VSCode) so that inappropriate function call can be prohibited before running tests.

saschanaz:
Adding `//@ts-check` at the top of the file is all that needed.

saschanaz:
>we might eventually be able to compile some modules as WASM (🤞).

TIL [AssemblyScript](https://github.com/AssemblyScript/assemblyscript) exists 😮

devanshbatra04:
@saschanaz I am taking this up if that's fine. Oh and Happy New Year!! :smiley:

marcoscaceres:
@saschanaz, this can be closed, right?

saschanaz:
We still have files without `@ts-check`, so no. The ultimate goal is to add `checkJs: true` in `jsconfig.json` which will enable ts-check in every JS file.

marcoscaceres:
Oh, ok. Thanks for confirming.

saschanaz:
https://github.com/w3c/webidl2.js/issues/255 is currently a blocker.

marcoscaceres:
I need to play less computer games and spend some time learning ts. #lifegoals.

rahulpurohit29:
Can I do this?

saschanaz:
@rahulpurohit29 If you have some TypeScript and JSDoc experience, sure.

rahulpurohit29:
If I just add `"checkJs":true` to the src/jsconfig.json it will globally type check for all .js files, right?

saschanaz:
All `src/**/*.js` files, yes. You will then have to solve the type errors TS shows.

marcoscaceres:
That might be a good place to start - but @saschanaz will know more. Try it and see what errors are emitted.  For where the type checks fail, then I assume you would need to go and fix those errors.

saschanaz:
I've found too hard to do fix everything at once, so if you get too much errors then just solve some of them and open a PR. Then we can make things better progressively.

marcoscaceres:
Breaking up the problem is a good suggestion... @rahulpurohit29, try to do one file at a time. Any fixes you can make will be a significant contribution to the project. It will also make it easier for us to review the changes you make (and a good opportunity for you to understand the code).

isunitha98selvan:
If no one is working on this, I would like to take it up.

[...]

PR #1988, opened 2019-01-01
refactor: use package names for external dependencies

saschanaz:
Because currently external dependencies can only be automatically typed through their package names with TypeScript.

Importing with package name blocks migration to ES modules, but those dependencies are not ES module compatible anyway. There is a discussion [to enable import-by-package-name with native ES modules](https://github.com/domenic/import-maps) so I hope we get a fix soon.


saschanaz:
cc: @devanshbatra04 as you are taking #1905 up.

PR #1989, opened 2019-01-01
refactor: use @ts-check wherever applicable

devanshbatra04:
So I applied ```@ts-check``` on all files involving `@type` occurences in ```src/```
This gives a bunch of errors, the import package errors should be fixed by #1988 but there are also other strange ones like 
```error TS2339: Property 'querySelectorAll' does not exist on type 'Node'```
and 
``` error TS2339: Property 'title' does not exist on type 'Element' ```

### DIFF
--- a/src/core/best-practices.js
+++ b/src/core/best-practices.js
@@ -1,3 +1,4 @@
+// @ts-check
 // Module core/best-practices
 // Handles the marking up of best practices, and can generate a summary of all of them.
 // The summary is generated if there is a section in the document with ID bp-summary.
--- a/src/core/data-cite.js
+++ b/src/core/data-cite.js
@@ -1,3 +1,4 @@
+// @ts-check
 /**
  * Module core/data-cite
  *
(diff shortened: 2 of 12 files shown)

### CODE REVIEW

**Comment by devanshbatra04** [comment]
I think I should manually define the types where it fails. Analogous to https://stackoverflow.com/questions/31237237/property-queryselector-does-not-exist-on-type-node-in-typescript

File: src/core/dfn.js [hunk 1]
```diff
@@ -1,6 +1,7 @@
 // @ts-check
 // Module core/dfn
 // - Finds all <dfn> elements and populates conf.definitionMap to identify them.
+// @ts-check
```
saschanaz: This is redundant 🤣

saschanaz:
We should cast the types where appropriate rather than manually adding types.

src/core/dfn.js
marcoscaceres:
```suggestion
```

marcoscaceres:
As @saschanaz says, you will probably need to cast for:

> Property 'querySelectorAll' does not exist on type 'Node'

Which should be cast to `Element`. 

Likewise, `HTMLElement` has the title attribute:
https://html.spec.whatwg.org/multipage/dom.html#htmlelement

src/core/dfn.js
devanshbatra04: LOL, my bad :sweat_smile:

marcoscaceres:
Just want to say we appreciate the work you did here, @devanshbatra04. It motivated us to merge in there other PR, giving us really good TypeScript coverage.

devanshbatra04:
Yep, I will work on the other files! Didn't know PRs close that way.

PR #1992, opened 2019-01-04
chore(package): update deps + cleanup

marcoscaceres:


saschanaz:
Some of our tests prefer loading js/profile-w3c-common.js via requirejs and thus need requirejs/text.

marcoscaceres:
ok, right now trying to see what else is needed for us to get rid of require... so just playing around :)

PR #1920, opened 2018-11-29
refactor(core/webidl): migrate writer to webidl2.js

saschanaz:

PR #1993, opened 2019-01-04
Remove redundant files

marcoscaceres:
highlight.js is a generated dependency, so doesn't need to be in the tree. 

The permalink.css template is not needed anymore, as there is css for that in respec2.css.


saschanaz:
Is it now safe to drop handlebars build? We previously decided to keep it when we do #1920.

Also, can these files also be removed?

* .jscsrc - should have been replaced by ESLint
* .jshintrc - same as above
* yarn.lock - hasn't been updated for a year

marcoscaceres:
> Is it now safe to drop handlebars build?

yes, I think it's safe now as it was only doing .css anyway. 

Agree about the other files... will trash. We are not using yarn anyway, so that's also safe to trash.

marcoscaceres:
:feelsgood:

PR #2001, opened 2019-01-04
test(core/link-to-dfn): ensure data-lt is preferred

### LINKED ISSUE #510: data-lt attribute ignored on links when an IDL definition is found
The [Presentation API](http://w3c.github.io/presentation-api/) defines a `Presentation` interface but also uses the term "presentation" as alias for the "receiving browsing context". To avoid linking occurrences of "presentation" used as alias to the IDL definition, the spec uses `data-lt` attributes. However, that does not seem to work.

The following example, taken from the spec, contains both the definition of "receiving browsing context" and the term "presentation" used as an alias to it. ReSpec seems to ignore the `data-lt` attribute on the `<a>` tag and simply considers that "presentation" is a reference to the IDL definition:

```
<p>The <dfn data-lt="receiving browsing context|receiving browsing contexts">receiving browsing context</dfn> (or <a data-lt="receiving browsing context">presentation</a> for short) is the browsing context responsible for rendering to a <a>presentation display</a>.</p>
```

### LINKED ISSUE #1161: Referencing term using @data-lt can get confused
Consider the following:

      <p>
        Define <dfn>bar</dfn> and
        <dfn data-lt="foo-bar" data-lt-noDefault>bar</dfn>
        (which uses <code>@data-lt="foo-bar"</code> and <code>@data-lt-noDefault</code>)
        confuses a reference to <a>bar</a> and <a data-lt="foo-bar">bar</a> (which uses <code>@data-lt="foo-bar"</code>).
      </p>

The last reference to `bar` should get to that defined in `<dfn data-lt="foo-bar" data-lt-noDefault>bar</dfn>`, however, it ends up referencing `<dfn>bar</dfn>`. This becomes an issue when a term is defined within the main body of a document, and also used as a parameter in a WebIDL block using `@data-dfn-for`.

The problem seems to be in link-to-dfn around line 50, which doesn't seem to take into account `@data-lt`, although it could be because `target.for_` is not defined as expected. I'm a bit over my head in determining exactly what should change.

This is fighting the [ShEx spec](https://shexspec.github.io/spec/) [API](https://shexspec.github.io/spec/#the-application-programming-interface), where I'm forced to use an anchor body of `validate.schema` instead if simply `schema`, even though `@data-lt="ShExProcessor.validate.schema"`.

saschanaz:
Closes #510, closes #1161 

I found that the issue is already fixed, so just adding a test.

### DIFF
--- a/tests/spec/core/link-to-dfn-spec.js
+++ b/tests/spec/core/link-to-dfn-spec.js
@@ -109,7 +109,7 @@ describe("Core — Link to definitions", () => {
     expect(dfn.id).toBe("dfn-test2");
   });
 
-  it("should prefer data-lt", async () => {
+  it("prefers data-lt over text content", async () => {
     const bodyText = `
       <section>
         <h2>Test Section</h2>
--- a/tests/spec/core/link-to-dfn-spec.js
+++ b/tests/spec/core/link-to-dfn-spec.js
@@ -108,4 +108,27 @@ describe("Core — Link to definitions", () => {
     const [dfn] = doc.getElementsByTagName("dfn");
     expect(dfn.id).toBe("dfn-test2");
   });
+
+  it("should prefer data-lt", async () => {
+    const bodyText = `
+      <section>
+        <h2>Test Section</h2>
+        <pre class="idl">
+          interface Presentation {};
+        </pre>
+        <dfn>Foo</dfn>
+        <dfn>Bar</dfn>
+        <a id="testFoo" data-lt="Foo">Presentation</a>
+        <a id="testBar" data-lt="Bar">Foo</a>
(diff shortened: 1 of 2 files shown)


tests/spec/core/link-to-dfn-spec.js
```diff
@@ -108,4 +108,27 @@ describe("Core — Link to definitions", () => {
     const [dfn] = doc.getElementsByTagName("dfn");
     expect(dfn.id).toBe("dfn-test2");
   });
+
+  it("should prefer data-lt", async () => {
```
marcoscaceres:
```suggestion
  it("prefers data-lt over text content", async () => {
```

PR #2007, opened 2019-01-07
refactor: put assets into /assets

### LINKED ISSUE #1994: Move CSS out of js folder
It would be nice if we could move the css out of the js folder, and ultimately delete the js folder (as we use src/ for JS stuff).

saschanaz:
Fixes #1994


saschanaz:
Oops, something went wrong.

marcoscaceres:
Karma's path stuff is such a nightmare :(

marcoscaceres:
I seem to recall maybe having to set paths here:
https://github.com/w3c/respec/blob/develop/tests/test-main.js

saschanaz:
@marcoscaceres Can/should we merge fixup.js into ReSpec? Currently the fixup script is loaded asynchronously (always on Firefox) or synchronously (if cached, on Chrome), causing potential heisenbug like https://github.com/w3c/respec/pull/2007/commits/e3a371dcf0619dfc163fff19d69735f6c7faf987.

marcoscaceres:
> @marcoscaceres Can/should we merge fixup.js into ReSpec?

I'm not sure we can... the W3C Publication Rules checker expects it to be included "at the end of the document using `<script src=""></script>`". Maybe we can force it to always be sync or async?

saschanaz:
Alternative we may use onload event to wait consistently. (I'm not sure whether sub() waits async function call though.)

saschanaz:
>It still feels like it's going to be racy.

The fixup is added when `end-all` event fires so it shouldn't conflict with other modules. My problem is it races with `document.respecIsReady`.

marcoscaceres:
Right, I'm totally fine if you want to try different things here. As you showed with e3a371d and evident in w3c/tr-design#62, it's annoying.

Issue #2013, opened 2019-01-09
Stop polluting configuration object in w3c/headers.js


saschanaz:
Currently it pollutes the configuration object only to pass them to templates. They should be instead passed as a dictionary object.

marcoscaceres:
Agree! and we should identify other modules that do this. So 🤮

sidvishnoi:
I think [core/caniuse.js](https://github.com/w3c/respec/blob/develop/src/core/caniuse.js#L50) is another potential candidate

saschanaz:
@sidvishnoi It looks like core/caniuse just normalizes the user configurable fields, which should be okay.

PR #2014, opened 2019-01-09
refactor(w3c/headers): pass `additional*` as option arguments

### LINKED ISSUE #2012: add eol-last to our ES Lint config
It's generally nice to have a new line at the end of files, particularly when looking a git changes. As such, we should add add "eol-last: always" to our ES Lint config.

saschanaz:
Part of #2013.

Closes #2012


src/w3c/templates/cgbg-sotd.js
```diff
@@ -1,6 +1,6 @@
 import hyperHTML from "hyperhtml";
 
-export default conf => {
+export default (conf, opts) => {
```
saschanaz: Intentionally didn't use object spread because I think we'll migrate more and more options here that will generate too much lines.

.eslintrc.json
```diff
@@ -15,6 +15,7 @@
   "rules": {
     "prettier/prettier": ["error", { "endOfLine": "auto" }],
     "dot-notation": "error",
+    "eol-last": 2,
```
marcoscaceres: Oh, we have a dup https://github.com/w3c/respec/pull/2015 (preference is for new contributor ;) )

src/w3c/templates/sotd.js
```diff
@@ -1,6 +1,6 @@
 import hyperHTML from "hyperhtml";
 
-export default conf => {
+export default (conf, opts) => {
```
marcoscaceres: Just wondering, is conf used?

src/w3c/templates/cgbg-sotd.js
marcoscaceres: makes sense.

src/w3c/templates/sotd.js
saschanaz: Yes, `conf.isMemberSubmission`, `conf.sotdAfterWGinfo`, etc. ESLint warns for us when it's not used 👍

marcoscaceres: ok, makes sense. The thing is, I think `isMemberSubmission` and friends are all generated in headers.js? We could, in theory, start putting those into the `opts` object.

marcoscaceres: (Slight ramble, we don't need to do this now!)

`isMemberSubmission` is used in sotd.js also... previously, I'd thought attaching these things to w3c-defaults.js and making them proper getters, which can be computed.  So, imagine:

```JS
{
   get isMemberSubmission(){
      return this.specStatus === "Member-SUBM" 
   }
}
```

PR #2020, opened 2019-01-10
feat(w3c/defaults): enable pluralize by default

marcoscaceres:
We've been running with this for a few months in various specs and it's working pretty well, so feel confident enabling by default.

### DIFF
--- a/tests/spec/core/pluralize-spec.js
+++ b/tests/spec/core/pluralize-spec.js
@@ -106,6 +106,26 @@ describe("Core - Pluralize", () => {
     ).toBeTruthy();
   });
 
+  it("does nothing if conf.pluralize is not defined", async () => {
+    const body = `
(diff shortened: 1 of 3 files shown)


tests/spec/core/pluralize-spec.js
```diff
@@ -106,26 +106,6 @@ describe("Core - Pluralize", () => {
     ).toBeTruthy();
   });
 
-  it("does nothing if conf.pluralize is not defined", async () => {
```
marcoscaceres: Actually,  I wonder if I can do this differently without trashing the test... I might be able to set status to unofficial, and that should stop treating this as a w3c spec.

marcoscaceres: ok, should hopefully be good now.

PR #2022, opened 2019-01-10
fix(w3c/defaults): expose conf.definitionMap to w3c specs

### LINKED ISSUE #2019: replacement for conf.definitionMap?
## Important info

* URL to affected spec: https://w3c.github.com/json-ld-syntax
* ReSpec version: 24.1.0

## Description of problem

  #1954 removed `definitionMap` from the configuration. This is used in my specs to iterate over definitions and remove those that aren't referenced; it's now broken.

  Is there a replacement mechanism for me to do this using pubsub or something similar?

  Developer console complains `TypeError: undefined is not an object (evaluating 'respecConfig.definitionMap[item]')`

But what I was expecting?: A means to iterate over definitions. The code also still depends on jQuery for `makeID`.

Optional, steps to reproduce:
 1. Load, for example, https://w3c.github.com/json-ld-syntax
 2. Note developer console message

marcoscaceres:
closes #2019

### DIFF
--- a/src/w3c/defaults.js
+++ b/src/w3c/defaults.js
@@ -4,6 +4,7 @@
 export const name = "w3c/defaults";
 import { rule as checkInternalSlots } from "../core/linter-rules/check-internal-slots";
 import { rule as checkPunctuation } from "../core/linter-rules/check-punctuation";
+import { definitionMap } from "../core/dfn-map";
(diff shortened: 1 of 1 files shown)


marcoscaceres:
Contains the damage to just the W3C specs :)

PR #2027, opened 2019-01-11
fix(core/jquery-enhanced): set window.jQuery

saschanaz:
This is to fix https://github.com/w3c/aria/issues/871.

### DIFF
--- a/src/core/jquery-enhanced.js
+++ b/src/core/jquery-enhanced.js
@@ -9,7 +9,7 @@ import $ from "jquery";
 
 export const name = "core/jquery-enhanced";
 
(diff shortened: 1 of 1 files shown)


saschanaz:
Yes... we need to somehow mark jQuery as deprecated.

marcoscaceres:
Maybe least painless is to just make a "lean" profile (let's call it "respec-ultra" or "respec-ultimate", heh). The "fat" profile can keep jQuery and keep using the current file name. That will avoid breaking things and allow us to incrementally move the community over.... we use speed and size as a selling point.

PR #2028, opened 2019-01-13
fix(w3c/headers): suggest correct wpt url

saschanaz:
Tests are in https://github.com/web-platform-tests/wpt/ so the message should probably point it.

### DIFF
--- a/src/w3c/headers.js
+++ b/src/w3c/headers.js
@@ -165,6 +165,7 @@ const status2text = {
   "BG-FINAL": "Final Business Group Report",
 };
 const status2long = {
+  ...status2text,
   "FPWD-NOTE": "First Public Working Group Note",
   "LC-NOTE": "Last Call Working Draft",
 };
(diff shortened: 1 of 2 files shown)


src/w3c/headers.js
```diff
@@ -267,7 +268,7 @@ export function run(conf) {
       const msg =
         "Web Platform Tests have moved to a new Github Organization at https://github.com/web-platform-tests. " +
         "Please update your [`testSuiteURI`](https://github.com/w3c/respec/wiki/testSuiteURI) to point to the " +
-        `new tests repository (e.g., https://github.com/web-platform-tests/${
+        `new tests repository (e.g., https://github.com/web-platform-tests/wpt/${
```
marcoscaceres: Good catch!

src/w3c/headers.js
```diff
@@ -248,7 +249,7 @@ export function run(conf) {
   }
   conf.licenseInfo = licenses[conf.license];
   conf.isCGBG = cgbg.includes(conf.specStatus);
-  conf.isCGFinal = conf.isCGBG && /G-FINAL$/.test(conf.specStatus);
+  conf.isCGFinal = conf.isCGBG && conf.specStatus.endsWith("G-FINAL");
   conf.isBasic = conf.specStatus === "base";
```
marcoscaceres: Do we need this one given it’s in w3c/default.js?

saschanaz: Hmm, maybe those duplicated fields can be removed. I'll try.

saschanaz: I was to remove them from w3c/headers but instead removed from w3c/defaults, because the props are only used within headers.

marcoscaceres: Ok, that makes sense.

PR #2029, opened 2019-01-13
refactor: use eslint prefer-template 

saschanaz:
Some changes may look too aggressive but I think this is generally good.

### DIFF
--- a/tools/respecDocWriter.js
+++ b/tools/respecDocWriter.js
@@ -66,7 +66,7 @@ async function fetchAndWrite(
   whenToHalt,
   { timeout = 300000, disableSandbox = false, debug = false } = {}
 ) {
-  const userDataDir = await mkdtemp(os.tmpdir() + "/respec2html-");
+  const userDataDir = await mkdtemp(`${os.tmpdir()}/respec2html-`);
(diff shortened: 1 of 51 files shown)


tools/respecDocWriter.js
```diff
@@ -133,9 +132,9 @@ async function checkReSpecVersion(page) {
   const [mayor] = version;
   // The exportDocument() method only appeared in vesion 18.
   if (mayor < 18) {
-    let msg =
-      "👴🏽  Ye Olde ReSpec version detected! " +
-      colors.debug("Sorry, we only support ReSpec version 18.0.0 onwards.\n");
+    let msg = `👴🏽  Ye Olde ReSpec version detected! ${colors.debug(
```
marcoscaceres: Coincidentally, we should remove this in another PR.

saschanaz: Or should we set it as 20.10.0, as another line checks it to fallback to older module?

marcoscaceres: Nah, it's been over a year. I think that's an acceptable length of time/warning for anyone on.

PR #2031, opened 2019-01-14
refactor(tools/respecDocWriter): remove older version check

saschanaz:


marcoscaceres:
Looks like there is a `checkReSpecVersion` somewhere?

saschanaz:
Now it should be good.

PR #2032, opened 2019-01-14
Convert Editor's Notes to Normal Notes

Swapnilr1:
Note that I have just started removing CSS, just a starting PR for now. Lots of work remaining.

### DIFF
--- a/assets/issues-notes.css
+++ b/assets/issues-notes.css
@@ -36,11 +36,6 @@ span.note, span.ednote, span.issue, span.warning { padding: .1em .5em .15em; }
     text-decoration: line-through;
 }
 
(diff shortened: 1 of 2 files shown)


Swapnilr1:
I will modify tests later.

Swapnilr1:
Since we do not wish to remove Editor's Notes but only use common CSS, I have just modified the test for `ednote` to select elements with class `note` instead. Since only an editor's note was created in that test, I think my change is correct. 

Also, the `.warning` class still requires the CSS, so length of the CSS file remains same. 
I don't think there is anything in `base.css` for the auto-generated `div` and `span`, so I think they should be kept.

 Please review this.

PR #2040, opened 2019-01-21
fix(core/data-tests): Remove duplicates

### LINKED ISSUE #1834: Warn when test appears more than once
Given, for instance:

```HTML
<p data-tests="
  payment-request-show-method.https.html
  some.other.test.html
  payment-request-show-method.https.html">
          ...
</p>
```

ReSpec should warn about duplicate tests. 

Effectively, somewhere here, we need to check and warn of duplicates:
https://github.com/w3c/respec/blob/develop/src/core/data-tests.js#L86-L98

And also drop the duplicates.

Johanna-hub:
Fixes #1834


tests/spec/core/data-tests-spec.js
```diff
@@ -23,6 +23,13 @@ describe("Core — data-tests attribute", () => {
           >
              metadata tests
           </p>
+        </section>
+        <section>
+          <p id="duplicates" data-tests="
+            payment-request-show-method.https.html,
+            some-other-test.html,
+            payment-request-show-method.https.html">
```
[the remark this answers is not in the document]
Johanna-hub: I've added some more repeats in and variation on paths and absolute URLs, just realised I need to follow up with some extra non-repeats.

src/core/data-tests.js
```diff
@@ -88,6 +88,7 @@ export function run(conf) {
       const renderer = hyperHTML.bind(details);
       const testURLs = elem.dataset.tests
         .split(/,/gm)
+        .filter((links, i, self) => self.indexOf(links) === i)
```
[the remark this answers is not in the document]
Johanna-hub: I have put in the warning, but unsure exactly of the appropriate wording.

src/core/data-tests.js
```diff
@@ -98,12 +98,22 @@ export function run(conf) {
           }
           return href;
         });
+      const duplicates = testURLs.filter(
+        (links, i, self) => self.indexOf(links) !== i
+      );
+      if (duplicates) {
+        pub(
```
marcoscaceres: We really recently added a new `showInlineWarning` util function that might be good to use here instead. It's shows "inline warnings" on the element itself, which is kinda nice. 

For the message, maybe:
 
```JS
showInlineWarning(
  elem, 
  `Duplicate tests found.`
  `To fix, remove duplicates from "data-tests": ${duplicates.join(", ")}`
);
```

That should add a red squiggly underneath the offending element. See if the message makes sense when you hover over the element.

Johanna-hub:
@marcoscaceres I have made the changes, just one question.  When we display the inline warning, it is showing the generated URLs, this could be different to the original input and what they are looking for to remove.  i.e. they gave `some-other-test.html` and the URL is `https://w3c-test.org/some-API/some-other-test.html` Just checking we can assume they can work it out themselves?

tests/spec/core/data-tests-spec.js
```diff
@@ -138,4 +149,18 @@ describe("Core — data-tests attribute", () => {
...
   });
+  describe("duplicate test removal", () => {
+    it("checks for and removes duplicate links", () => {
+      const items = Array.from(
+        doc.querySelectorAll(
+          "#duplicates > .respec-tests-details > ul > li > a"
+        )
+      );
+      expect(items.length).toEqual(2);
+      expect(items[0].textContent.trim()).toEqual(
```
marcoscaceres:
```suggestion
      expect(items[0].textContent.trim()).toBe(
```

tests/spec/core/data-tests-spec.js
```diff
@@ -138,4 +149,18 @@ describe("Core — data-tests attribute", () => {
...
+        "payment-request-show-method"
+      );
+      expect(items[1].textContent.trim()).toEqual("some-other-test");
```
marcoscaceres:
```suggestion
      expect(items[1].textContent.trim()).toBe("some-other-test");
```

Issue #1976, opened 2018-12-26
New API for external plugins


saschanaz:
The current plugin API depends on AMD module system (via requirejs) and exposes internal modules. This blocks #1975 and potentially causes breaking changes when any internal module functions change.

We may introduce a new API looks something like webpack one:

```js
var respecConfig = {
   // ...
   plugins: [
     new MyPlugin()
   ]
};
```

saschanaz:
Looks like the vast majority of plugin users only depends on `core/pubhubsub`.

saschanaz:
Maybe we can set `window.require` that supports certain modules to keep backward compatibility. For example:

```js
if (!window.require) {
  window.require = function(deps, callback) {
    const modules = deps.map(dep => {
      if (!(dep in window.require.modules)) {
        throw new Error("Invalid dependency name");
      }
      return window.require.modules[dep];
    });
    callback(modules);
  };
  window.require.modules = {};
}
```

```js
// inside modules e.g. pubsubhub
if (window.require && window.require.modules) {
  window.require.modules["core/pubsubhub"] = { sub };
}
```

Will probably be longer than this, though... @sidvishnoi, does this idea look okay?

sidvishnoi:
That's a really cool idea. We won't need to bundle requireJS this way.

saschanaz:
`window.require` thing landed via #1981

sidvishnoi:
I'm not sure we want to support "tapping" into any/all core plugin/hook. How about something like:
```js
const myPlugin = {
  name: 'plugin-name',
  // some custom hook name instead of listening to pubsub events, or, "when": "after-markdown", "when": "before-markdown"
  after: 'markdown',
  // conf is respecConfig as usual.
  // utils is an object of methods as an abstraction over respec internals, 
  //  making us expose as little as required, reducing breaking changes in future.
  async run(conf, utils) {
    utils.showError(`${conf.unCool} option has invalid value.`)
  },
};
```

This follows pretty much same structure as present ReSpec plugins.

saschanaz:
It looks promising! Probably needs some experiment by converting existing specs to make sure that we have all required features.

PR #2042, opened 2019-01-23
feat(core/defaults): Add core/defaults

### LINKED ISSUE #2023: Add core defaults
It would be nice to have a "core/defaults" module, that sets the default core defaults that every specification or document is expected to have. Those can the be amended or overridden by specific profiles. 

For example, all specs should have a specStatus of "unofficial" by default. The license should probably be some CC license, and so on...  

Example of W3C defaults, on which this could be based:
https://github.com/w3c/respec/blob/develop/src/w3c/defaults.js

pradeepgangwar:
Adds core defaults to respec

Fixes #2023


pradeepgangwar:
@marcoscaceres I have added some of the defaults on basis of [this](https://github.com/w3c/respec/blob/develop/src/w3c/defaults.js) file. Since I am new to this project let me know if i missed something. 
Also from where I can get to know more default conf that can be added?

src/core/defaults.js
```diff
@@ -0,0 +1,48 @@
...
+    "check-internal-slots": false,
+  },
+  pluralize: true,
+  highlightVars: true,
+  license: "CC-BY",
+  addSectionLinks: true,
+};
+
+export function run(conf) {
+  if (conf.specStatus === "unofficial") return;
```
marcoscaceres: this line can go, as this will apply to all document, including "unofficial" ones.

marcoscaceres:
This is good @pradeepgangwar. Now, notice that this duplicates things in `w3c/defaults` - so things that are imported or set to default here shouldn't be `w3c/defaults`. 

That should allow us to start answering the question "from where I can get to know more default conf that can be added?"

We might then need to go and look at a few specs to get a sense of what people are using/enabling.

pradeepgangwar:
@marcoscaceres  This means that If I copy defaults from `w3c/defaults`  to `core/defaults` then I should remove them from `w3c/defaults`? Because they are already covered as core defaults.

pradeepgangwar:
@marcoscaceres I did some changes and pushed the code again. Please review again. :)

> We might then need to go and look at a few specs to get a sense of what people are using/enabling.

Also please let me know what do i need to do for this.

marcoscaceres:
> Also please let me know what do i need to do for this. ^

Let's get the foundational stuff done first, then we can do the above in a separate PR. The above will require us to go look at quite a few specs.

pradeepgangwar:
> Let's get the foundational stuff done first, then we can do the above in a separate PR. The above will require us to go look at quite a few specs.

Sure, Thanks @marcoscaceres . Let me know if this PR requires further changes. :)

src/w3c/defaults.js
```diff
@@ -2,35 +2,17 @@
...
-    "no-http-props": true,
-    "check-punctuation": false,
-    "local-refs-exist": true,
-    "check-internal-slots": false,
   },
-  pluralize: true,
-  highlightVars: true,
   doJsonLd: false,
   license: "w3c-software-doc",
   specStatus: "base",
```
marcoscaceres: This one can move to core/defaults instead. 

```suggestion

```

src/w3c/defaults.js
```diff
@@ -43,17 +25,18 @@ const w3cDefaults = {
...
     },
   ],
-  addSectionLinks: true,
 };
 
 export function run(conf) {
   if (conf.specStatus === "unofficial") return;
   // assign the defaults
   Object.assign(conf, {
+    ...coreDefaults,
```
marcoscaceres: As above, don't need this... 
```suggestion

```

src/w3c/defaults.js
```diff
@@ -43,17 +25,18 @@ const w3cDefaults = {
...
     ...w3cDefaults,
     ...conf,
   });
   Object.assign(conf.lint, {
+    ...coreDefaults.lint,
```
marcoscaceres: Don't need this either... 

```suggestion

```

src/core/defaults.js
```diff
@@ -0,0 +1,49 @@
+/**
+ * Sets the core defaults
+ */
+export const name = "core/defaults";
+import { rule as checkInternalSlots } from "./linter-rules/check-internal-slots";
+import { rule as checkPunctuation } from "./linter-rules/check-punctuation";
+import { definitionMap } from "./dfn-map";
```
marcoscaceres: Only import this one in the `w3c/defaults` (see the comment in that file). Please also remove the where definitionMap is assigned to conf.

marcoscaceres:
@pradeepgangwar, could you please revert the changes to the files in the "build/" directory? We only build as part of a release.

marcoscaceres:
@saschanaz, would really appreciate your input on this one. Do you think we should keep the current defaults, or turn everything off by default in  "core/defaults"?

src/core/defaults.js
```diff
@@ -0,0 +1,47 @@
...
+    "check-internal-slots": false,
+  },
+  pluralize: true,
+  specStatus: "base",
+  highlightVars: true,
+  license: "CC-BY",
+  addSectionLinks: true,
+};
+
+export default coreDefaults;
```
marcoscaceres: I'm not sure we should export these... or we should only export them if there is a good reason to do so (e.g., some other module asks for them in the future for some valid reason).

src/core/defaults.js
```diff
@@ -0,0 +1,47 @@
+/**
+ * Sets the core defaults
+ */
+export const name = "core/defaults";
```
saschanaz: I think nearly, if not all, every `name` exported constant is redundant as AFAIK no module uses it.

marcoscaceres: I've kept them only for debugging... when Uglyfied, and ReSpec throws or halts, the core runner prints the name of module as a hint as which module died.

marcoscaceres: https://github.com/w3c/respec/blob/develop/src/core/base-runner.js#L23 for example.

src/core/defaults.js
saschanaz: I think exporting anything without actual use should be avoided.

src/core/defaults.js
marcoscaceres: Also...
https://github.com/w3c/respec/blob/develop/src/core/base-runner.js#L46

What might be nice, once we have the new class-based plugins, is to build in performance timing directly into the sub class... and error reporting. That will clean up base-runner really nicely. Something for us to discuss elsewhere tho.

src/core/defaults.js
marcoscaceres: agree,

saschanaz:
>Do you think we should keep the current defaults, or turn everything off by default in "core/defaults"?

Keeping the current sounds okay to me if they are recommended.

marcoscaceres:
> Keeping the current sounds okay to me if they are recommended.

Ok, just wanted to double check I wasn't overlooking something obvious. Let's run with this current structure.

marcoscaceres:
@pradeepgangwar, two tests are failing:

```
    ✖ sets sensible defaults for w3c specs
    ✖ allows w3c defaults to be overridden
```

To figure out why, you will need to have a look at: 
https://github.com/w3c/respec/blob/develop/tests/spec/w3c/defaults-spec.js

And to run the tests locally:

```
npm start
```

And navigate to:

```
http://localhost:5000/tests/SpecRunner.html
```

You should then be able to debug the tests that are failing (e.g., by setting a `debugger` statement in the test to figure out what's different).

pradeepgangwar:
@marcoscaceres Looks like it is not taking into account the core/defaults while determining defaults for w3c in test file. Earlier it was working when I imported `core/defaults` in `w3c/defaults`. Is it like are automatically that core/defaults are automatically set and w3c/defaults overrides that.

Debugging statements how that w3c/defaults didn't had core/defaults in tests.

saschanaz:
Looks like we created core/defaults but not using it at all.

marcoscaceres:
> Looks like we created core/defaults but not using it at all, with no imports for the module.

marcoscaceres:
> Looks like we created core/defaults but not using it at all, with no imports for the module.

lol, yeah - you are right @saschanaz ... sorry @pradeepgangwar, I should spotted that. 

As this is "core", you need to `import` the "core/defaults" plugin directly into base-runner.js.  

Then you should *unshift()* the "core/defaults" into the `plugs` array here:
https://github.com/w3c/respec/blob/develop/src/core/base-runner.js#L53

saschanaz:
>Then you should unshift() the "core/defaults" into the plugs array here:

Or we may just explicitly call `run()` there. (I would change the function name if we go this way.)

marcoscaceres:
> Or we may just explicitly call run() there. (I would change the function name if we go this way.)

I think we should keep treating it as a plugin, just to be consistent. If we ever update it to be async, the base runner will know how to deal with it. Same with errors, and measuring perf stuff and so on. That will all work the same as with other plugins.

src/core/base-runner.js
```diff
@@ -4,6 +4,7 @@ import "./include-config";
 import "./override-configuration";
 import "./respec-ready";
 import "./jquery-enhanced"; // for backward compatibility
+import coreDefaults from "./defaults";
```
saschanaz:
```suggestion
import * as coreDefaults from "./defaults";
```

src/core/defaults.js
saschanaz: This still exists, should be removed.

marcoscaceres:
```suggestion

```

saschanaz:
Because core and w3c defaults shouldn't define the same field.

pradeepgangwar:
> Because core and w3c defaults shouldn't define the same field.

Oh okay! So should i remove that or there is any other way to override that?

marcoscaceres:
Right, so:
https://github.com/w3c/respec/blob/develop/tests/spec/w3c/defaults-spec.js#L6

Should probably change to:

```JS
config: { editors: [{ name: "foo" }], specStatus: "ED" },
```

Try that first.

saschanaz:
Shouldn't we remove the `license` default value from one of the default sets? Any duplication will cause it be ignored by the earlier default value.

marcoscaceres:
> Shouldn't we remove the license default value from one of the default sets? Any duplication will be ignored by the earlier default value.

Ah, good point. Ok, no license on `core/default`.

pradeepgangwar:
> Ah, good point. Ok, no license on `core/default`.

Done. Please review. Tests are now passing. :D

marcoscaceres:
Got another fail (see Travis). Pluralize should be off by default.

marcoscaceres:
Ok, cool. But we do want it on for W3C specs, so please add it to "w3c/defaults". If that test fails again, then we need to set the "specStatus" to "unofficial".

src/w3c/defaults.js
```diff
@@ -2,38 +2,19 @@
...
-    "check-punctuation": false,
-    "local-refs-exist": true,
-    "check-internal-slots": false,
   },
-  pluralize: true,
-  highlightVars: true,
   doJsonLd: false,
   license: "w3c-software-doc",
-  specStatus: "base",
+  pluralize: false,
```
marcoscaceres: On for this on :)
 
```suggestion
  pluralize: true,
```

pradeepgangwar:
@marcoscaceres sorry for repetitive commits. Actually I was editing that on my mobile. 😬

marcoscaceres:
> sorry for repetitive commits. Actually I was editing that on my mobile.

No problem at all. I do the same all the time 💃

src/core/defaults.js
```diff
@@ -0,0 +1,44 @@
...
+
+const coreDefaults = {
+  lint: {
+    "no-headingless-sections": true,
+    "no-http-props": true,
+    "check-punctuation": false,
+    "local-refs-exist": true,
+    "check-internal-slots": false,
+  },
+  pluralize: false,
```
saschanaz: No, this will cause the ignored value situation again.

pradeepgangwar: Okay. I will remove it from core/defaults then.

pradeepgangwar:
@marcoscaceres @saschanaz surprisingly Travis passed and synk failed. :/

saschanaz:
Just pushing an empty commit should fix the snyk issue.

pradeepgangwar:
> A tip for today: `git commit --allow-empty -m "wow an empty commit that changes nothing"`

Thanks @saschanaz. :D

saschanaz:
>default <- profile <- user-conf

You mean core-default <- w3c-default <- user-conf ?

marcoscaceres:
> You mean core-default <- w3c-default <- user-conf ?

Yes, sorry for not being clear. I was using "profile" generically, as I was imaging other ReSpec profiles using this too.

saschanaz:
We are adding missing conf fields in order of user-conf -> core-default -> w3c-default, thus causing any proceding w3c-default (aka profile) is ignored by preceding core default (because it's not *missing*).

We should instead effectively do this:

```js
respecConfig = { ...coreDefaults, ...w3cDefaults, ...respecConfig };
```

... which requires a profile should run *before* the `...respecConfig` part.

saschanaz:
Simpler way: Just put `core/defaults` in `js/profile-w3c-common.js` after `w3c/defaults`.

This way the "other ReSpec profiles" may miss the new module, which is why I don't like the current profile system.

marcoscaceres:
I'm happy with either solution. @pradeepgangwar, you ok to keep going with this? We are nearly there.

pradeepgangwar:
@marcoscaceres yes I will do the necessary changes. Please let me know which one you would prefer.

marcoscaceres:
My feeling is that we should do: 

 1. export `coreDefaults` from "core/default" - and remove the `run()` function. 
 1. import  `coreDefaults` into "w3c/defaults".
 1. In "w3c/defaults"'s `run()` function, do `Object.assign(conf, { ...coreDefaults, ...w3cDefaults, ...conf });`

@saschanaz wdty? That feels the least abhorrent for now. At least, it's quite explicit and predictable what's happening.

saschanaz:
Looks okay for now.

I would prefer the whole profile feature to be included in the future plugin API (#1976):

```
var respecConfig = {
   // ...
   plugins: [
     new W3CProfilePlugin() // sets defaults
   ]
};
```

marcoscaceres:
@saschanaz, ok, let's work up to that.

@pradeepgangwar please start with what I proposed above. That will allow us to merge this PR and will give us the core defaults.

marcoscaceres:
Travis is failing because `conf` or `lint` is `undefined`. You might need to poke at it a bit so make sure there is always an object there. Let me know if you need a hand.

src/core/defaults.js
```diff
@@ -0,0 +1,34 @@
...
+  specStatus: "base",
+  highlightVars: true,
+  addSectionLinks: true,
+};
+
+export default coreDefaults;
```
marcoscaceres: The bug is here, don't export `default`. 
```suggestion
export coreDefaults;
```

marcoscaceres: Actually, let's trash this one... and see above. 
```suggestion
export default coreDefaults;
```

src/core/defaults.js
```diff
@@ -0,0 +1,34 @@
...
+
+linter.register(
+  noHttpPropsRule,
+  noHeadinglessSectionsRule,
+  checkPunctuation,
+  localRefsExist,
+  checkInternalSlots
+);
+
+const coreDefaults = {
```
marcoscaceres: export here instead: 

```suggestion
export const coreDefaults = {
```

marcoscaceres:
🤞 fingers crossed that fixed it! 🤞

marcoscaceres:
Failing tests is a good thing: it means that the new defaults are being applied  :) We are changing some pretty fundamental stuff in ReSpec, so some tests are going to get upset.

marcoscaceres:
@pradeepgangwar, are you good to investigate these failures or would you prefer I have a look? Understand if it's starting to get more "in the weeds" than originally intended. However, if you want a bit of a challenge, it might be worth while seeing why those two tests are failing.

pradeepgangwar:
@marcoscaceres Sure I wil try to investigate the tests. Give me some time ... Maybe till tomorrow. I will let you know If I get into some issues.

src/core/defaults.js
```diff
@@ -0,0 +1,32 @@
...
+ */
+export const name = "core/defaults";
+import { rule as checkInternalSlots } from "./linter-rules/check-internal-slots";
+import { rule as checkPunctuation } from "./linter-rules/check-punctuation";
+import linter from "./linter";
+import { rule as localRefsExist } from "./linter-rules/local-refs-exist";
+import { rule as noHeadinglessSectionsRule } from "./linter-rules/no-headingless-sections";
+import { rule as noHttpPropsRule } from "./linter-rules/no-http-props";
+
+linter.register(
```
saschanaz: Potential problem: This runs before calling `run()`.

pradeepgangwar:
@marcoscaceres I tried it for but could not figure out actual problem. 🤕  Can you help me here.

marcoscaceres:
Will try to have a look soon.

saschanaz:
Okay, I found what's the problem for the test "resolve with the resulting respecConfig". Previously the code does not overwrite `conf.lint` if it's not an object, as it possibly can be `false` and `Object.assign(conf.lint, { ... })` does nothing then. Now we always rewrite `conf.lint` as object which caused the test failure.

saschanaz:
The problem with the test "uses inline references to provide context" is similar. We now overwrites on `lint: false` so it forcibly enabled all linters, which caused the warning count to be different.

src/w3c/defaults.js
```diff
@@ -43,19 +25,21 @@ const w3cDefaults = {
       url: "https://www.w3.org/",
     },
   ],
-  addSectionLinks: true,
 };
 
 export function run(conf) {
   if (conf.specStatus === "unofficial") return;
   // assign the defaults
+  const lint = {
```
saschanaz:
```suggestion
  const lint = conf.lint && {
```

src/w3c/defaults.js
```diff
@@ -43,19 +25,21 @@ const w3cDefaults = {
       url: "https://www.w3.org/",
     },
   ],
-  addSectionLinks: true,
 };
 
 export function run(conf) {
   if (conf.specStatus === "unofficial") return;
   // assign the defaults
+  const lint = conf.lint && {
```
saschanaz: Ah, sorry, we should support undefined `conf.lint` situation. So:

```suggestion
  const lint = conf.lint === false ? conf.lint : {
```

marcoscaceres:
Thanks all for your persistence on this. I think this moves us in a good new direction.

PR #2043, opened 2019-01-23
fix(core/caniuse): Add emojis for status in print stylesheet

### LINKED ISSUE #2041: Unclear if "can I use" when printing
## Important info

* URL to affected spec: https://w3c.github.io/payment-request/
* ReSpec version:  24.2.3.

## Description of problem
When a spec is printed (on paper/PDF), the "can I use" section in the header of doc doesn't actually show if something is supported or not:

screenshot 2019-01-22 16 25 27

But what I was expecting?:

To be able to distinguish them, maybe with ✅ and ❌ and ？ (or similar). 

Optional, steps to reproduce:
 1. Print first page of the spec.

CodHeK:
Fixes #2041

### DIFF
--- a/assets/caniuse.css
+++ b/assets/caniuse.css
@@ -22,10 +22,6 @@ button.caniuse-cell {
     display: block !important;
     text-align: center;
   }
-  .tick .cross .q_mark {
-    display: block !important;
-    text-align: center;
-  }
 }
 
 /* reset styles, hide old versions by default */
@@ -96,10 +92,6 @@ see https://github.com/Fyrd/caniuse/blob/master/CONTRIBUTING.md for stats */
 .caniuse-stats button:focus + ul,
 .caniuse-stats .caniuse-browser:hover > ul {
   display: block;
-}
-
-.tick .cross .q_mark {
-
 }
 
 .display_emoji {
--- a/assets/caniuse.css
+++ b/assets/caniuse.css
@@ -19,12 +19,12 @@ button.caniuse-cell {
 /* handle case when printing */
 @media print {
   .caniuse-cell.y::before {
-    content: " ✔️ ";
+    content: "✔️";
     padding: 0.5em;
   }
 
   .caniuse-cell.n::before{
-    content: " ❌ ";
+    content: "❌";
     padding: 0.5em;
   }
 
@@ -32,7 +32,7 @@ button.caniuse-cell {
   .caniuse-cell.a::before,
   .caniuse-cell.x::before,
   .caniuse-cell.p::before {
-    content: " ⚠️ ";
+    content: "⚠️";
     padding: 0.5em;
   }
 }
--- a/assets/caniuse.css
+++ b/assets/caniuse.css
@@ -21,11 +21,13 @@ button.caniuse-cell {
   .caniuse-cell.y::before {
     background: #8bc34a;
     content: "&#10004;";
+    padding: 0.5em;
   }
 
   .caniuse-cell.n::before{
     background: #e53935;
     content: "&#x2718;";
+    padding: 0.5em;
   }
 
   .caniuse-cell.d::before,
--- a/assets/caniuse.css
+++ b/assets/caniuse.css
@@ -19,14 +19,12 @@ button.caniuse-cell {
 /* handle case when printing */
 @media print {
   .caniuse-cell.y::before {
-    background: #8bc34a;
-    content: "&#10004;";
+    content: " ✔️ ";
     padding: 0.5em;
   }
 
   .caniuse-cell.n::before{
-    background: #e53935;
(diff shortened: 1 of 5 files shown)


examples/basic.html
```diff
@@ -19,6 +19,10 @@
         github: "https://github.com/w3c/some-API/",
         testSuiteURI: "https://w3c-test.org/some-API/",
         implementationReportURI: "https://w3c.github.io/test-results/some-API",
+        caniuse: {
```
marcoscaceres: Best to revert this. Tho, of course change it for your own local testing (it's also what I do!) :)

assets/caniuse.css
```diff
@@ -85,3 +93,8 @@ see https://github.com/Fyrd/caniuse/blob/master/CONTRIBUTING.md for stats */
 .caniuse-stats .caniuse-browser:hover > ul {
   display: block;
 }
+
+.display_emoji {
```
marcoscaceres: I think it should be possible to just use `@media print` rule without introducing a new emoji rule. As mentioned below, use "CSS Generated Content" to dynamically add the emojis without needing to touch the JS.

src/core/caniuse.js
```diff
@@ -200,8 +200,20 @@ function createTableHTML(conf, stats) {
       .reverse();
     const { support, title } = getSupport(latestVersion);
     const cssClass = `caniuse-cell ${support}`;
+    let emoji;
```
marcoscaceres: As I mentioned in the original bug, it should be possible to do this without modifying the caniuse.js file by using CSS Generated content: an example where an emoji is added https://www.viget.com/articles/css-generated-content/

assets/caniuse.css
```diff
@@ -16,6 +16,14 @@ button.caniuse-cell {
   position: relative;
 }
 
+/* printing */
+@media print {
+  .display_emoji {
+    display: block !important;
+    text-align: center;
+  }
```
marcoscaceres: Instead of the above, try:
```
.caniuse-cell.n::before {
  content: "❌";
}
```

CodHeK:
@marcoscaceres  This is how it looks now, used the css content property

screenshot 2019-01-23 at 7 50 39 pm

assets/caniuse.css
```diff
@@ -16,6 +16,30 @@ button.caniuse-cell {
   position: relative;
 }
 
+/* handle case when printing */
+@media print {
+  .caniuse-cell.y::before {
+    background: #8bc34a;
```
marcoscaceres: `background` doesn't show up in print mode, right? So maybe we should remove it.  
```suggestion

```

assets/caniuse.css
```diff
@@ -16,6 +16,30 @@ button.caniuse-cell {
...
+    content: "&#10004;";
```
marcoscaceres: It's preferable just to put the actual emoji here, otherwise it's difficult to know what's actually going to shown.

marcoscaceres: The above two comments apply below too.

assets/caniuse.css
```diff
@@ -16,6 +16,27 @@ button.caniuse-cell {
...
+  .caniuse-cell.n::before{
+    content: " ❌ ";
+    padding: 0.5em;
+  }
+
+  .caniuse-cell.d::before,
+  .caniuse-cell.a::before,
+  .caniuse-cell.x::before,
+  .caniuse-cell.p::before {
+    content: " ⚠️ ";
```
marcoscaceres: Good choice.

assets/caniuse.css
```diff
@@ -16,6 +16,27 @@ button.caniuse-cell {
   position: relative;
 }
 
+/* handle case when printing */
+@media print {
+  .caniuse-cell.y::before {
+    content: " ✔️ ";
```
marcoscaceres: Just wondering, do we need the white space at the start? I get why the whitespace after the emoji might be a good idea... but is the whitespace at the start not handled by the padding?

CodHeK: The content value gets trimmed anyway so, I just put the spaces for it to look clearly visible in the code.

marcoscaceres: Ah, ok. No, then let's remove the white space. I work on a browser, so I don't trust browsers :)

CodHeK: Done!

marcoscaceres:
Will merge once Travis is happy.

marcoscaceres:
> Great! I'd like to work on more issues...any particular issue that you might suggest ? :)

See any of "[Good first issue](https://github.com/w3c/respec/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22)" 

Some have been claimed, but any of those would be great.

PR #2049, opened 2019-01-28
fix(core/issues-notes.js): add ARIA role "note" to notes

### LINKED ISSUE #2045: div that are of class "note" should have a role="note"
Given:

```HTML
<div class="note">
  <p>I'm a note</p>
</div>
```

It would be great if the `<div>` got a [`role="note"`](https://www.w3.org/TR/wai-aria-1.1/#note) attribute also attached to it.

nkkumawat:
@marcoscaceres  This fixes #2045 (div that are of class "note" should have a role="note" )

### DIFF
--- a/src/core/issues-notes.js
+++ b/src/core/issues-notes.js
@@ -51,7 +51,7 @@ function handleIssues(ins, ghIssues, conf) {
     // wrap
     if (!isInline) {
       const div = hyperHTML`<div class='${type +
-        (isFeatureAtRisk ? " atrisk" : "")}'></div>`;
+        (isFeatureAtRisk ? " atrisk" : "")}' role=${(type === "note"? "note" :"")} ></div>`;
       const title = document.createElement("span");
       const titleParent = hyperHTML`
         <div role='heading' class='${`${type}-title marker`}'>${title}</div>`;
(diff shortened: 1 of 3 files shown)


src/core/issues-notes.js
```diff
@@ -311,4 +311,8 @@ export async function run(conf) {
     ednote.classList.remove("ednote");
     ednote.classList.add("note");
   });
+  const notes = document.querySelectorAll(".note");
```
marcoscaceres: This definitely get the job done, but it's rather inefficient, because we already did the query for `.note` a few lines above (line 295). 

I was expecting `handleIssues()` around the following line to add this attribute:

https://github.com/w3c/respec/blob/c98b2a7d76cb77a0c077a6619dd7c810eb720269/src/core/issues-notes.js#L53

But only if the `type === "note"`.

src/core/issues-notes.js
```diff
@@ -51,7 +51,7 @@ function handleIssues(ins, ghIssues, conf) {
     // wrap
     if (!isInline) {
       const div = hyperHTML`<div class='${type +
-        (isFeatureAtRisk ? " atrisk" : "")}'></div>`;
+        (isFeatureAtRisk ? " atrisk" : "")}' role=${(type === "note"? "note" :"")} ></div>`;
```
marcoscaceres: Might be a little easier on the eyes... also, setting it to `null` means that `role` attribute will not be there when not needed (instead of being `role=""`, which could mean something else):

```suggestion
        (isFeatureAtRisk ? " atrisk" : "")}'
        role='${type === "note"? "note" : null}'></div>`;
```

marcoscaceres:
@nkkumawat after you apply the above, TravisCI might still be fail because of ESLint. Please be sure to run `--fix` on your code, as suggested there. If you need a hand with that, let us know.

PR #2057, opened 2019-02-01
fix(tools/respecDocWriter): handle console messages earlier 

saschanaz:
Webpack-built ReSpec can finish things before `document.onload` fires, and this currently causes unexpected console error ignorance even with `-e -w`. Attaching event listener before navigation fixes the bug.

### DIFF
--- a/tests/headless.js
+++ b/tests/headless.js
@@ -1,4 +1,5 @@
 #!/usr/bin/env node
+// @ts-check
 /*eslint-env node*/
 "use strict";
 const port = 5000;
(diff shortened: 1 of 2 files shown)


tests/headless.js
```diff
@@ -9,17 +10,6 @@ const testURLs = [
 const colors = require("colors");
 const { exec } = require("child_process");
 const moment = require("moment");
-colors.setTheme({
```
marcoscaceres: 😢 ... 🤔🤗

marcoscaceres:
I wonder what kind of errors have slipped through. How did you find this?

saschanaz:
@marcoscaceres I ran respec2html against Media Source Extensions spec and found it shows nothing even when my browser showed several warnings.

PR #2068, opened 2019-02-05
chore(package): update deps

marcoscaceres:


marcoscaceres:
Rebased... got some weird bustage on develop.

marcoscaceres:
@saschanaz, VS Code got upset about "src/w3c/templates/sotd.js" not being "pretty" ... so I tried prettying it, but then tests started failing. Not sure why running prettier would affect the tests or what it could be doing. I reverted running prettier. Will need to investigate further.

marcoscaceres:
Yeah, it might be that... I think I have something set for VS Code which might be conflicting.

PR #2050, opened 2019-01-28
Reindent innerHTML of children of aside tag

Swapnilr1:
Fixes https://github.com/w3c/respec/issues/2034 
I have currently re-indented all children of `aside` tag. Please let me know if it is to be done only on the `pre` tags which are children of `aside`.

PR #2073, opened 2019-02-05
refactor: apply prettier changes

saschanaz:
Looks like this is intended change from prettier. prettier/prettier#5771

Not sure why Travis didn't say anything for other PRs.

### DIFF
--- a/src/w3c/templates/sotd.js
+++ b/src/w3c/templates/sotd.js
@@ -3,294 +3,235 @@ import html from "hyperhtml";
 export default (conf, opts) => {
   return html`
     <h2>${conf.l10n.sotd}</h2>
-    ${
-      conf.isPreview
-        ? html`
-            <details class="annoying-warning" open="">
-              <summary
-                >This is a
-                preview${
-                  conf.prUrl && conf.prNumber
-                    ? html`
-                        of pull request
-                        <a href="${conf.prUrl}">#${conf.prNumber}</a>
-                      `
-                    : ""
-                }</summary
-              >
-              <p>
-                Do not attempt to implement this version of the specification.
-                Do not reference this version as authoritative in any way.
-                ${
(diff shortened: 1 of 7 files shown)


saschanaz:
Okay, there is an error you mentioned earlier...

src/w3c/templates/sotd.js
```diff
@@ -3,308 +3,247 @@ import html from "hyperhtml";
...
                 `
-                                          : ""
-                                      }
-                                      ${
-                                        conf.isPER
-                                          ? html`
-                                              W3C Advisory Committee Members are
-                                              invited to send formal review
-                                              comments on this Proposed Edited
-                                              Recommendation to the W3C Team
```
saschanaz: The error is because previously it was wrapped and now is restored🤣

saschanaz: We should consider using `innerText` in this case.

marcoscaceres: Maybe in the test we can normalize checking without the spaces? I’m fine with any solution tbh.

marcoscaceres: Just merged in the other sotd related pull request. Might cause conflict.

saschanaz:
Sometimes eslint on Travis does not say the same thing as my local eslint says, but probably a temporary thing as my one didn't either hours ago...

marcoscaceres:
We might need to periodically run —fix over the entire tree.

saschanaz:
Wait, changes introduced by #2050 passed on Travis for 7 days? That's not a temporary thing. Something is wrong.

marcoscaceres:
🙈🙉🙊😭

saschanaz:
Pushed a reverting commit https://travis-ci.org/w3c/respec/builds/489101376 and it passes without any error. Confused 😕

PR #2077, opened 2019-02-06
refactor(core/linter-rules): add ts-check

### LINKED ISSUE #1670: Linter rules API is ... not great
The linter rule API returns either and object or an array... this is pretty terrible. It should return just one type.

saschanaz:
May help future linter contributions write correct types 😁

Closes #1670

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,7 +71,7 @@
     "build": "npm run snyk-protect && npm run build:components",
     "copydeps": "node ./tools/copydeps.js",
     "karma": "karma start --single-run",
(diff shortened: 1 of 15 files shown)


saschanaz:
@marcoscaceres Can we just fix it or are we currently exposing linter API?

marcoscaceres:
We can just fix it. I have a bug filed to do that... just a matter of finding time to do it.

saschanaz:
@marcoscaceres Had some changes, still look good?

PR #2082, opened 2019-02-07
fix: reindent only pre elements

### LINKED ISSUE #2064: ReSpec intercepting button onclick?
## Important info

* URL to affected spec: https://w3c.github.io/json-ld-syntax
* ReSpec version: Click the ReSpec pill (e.g., 20.0.1). 24.2.5

## Description of problem

The spec uses an "onclick" in buttons used for tabbing in examples (e.g., [example 3](https://w3c.github.io/json-ld-syntax/#example-3-sample-json-ld-document-using-full-iris-instead-of-terms). The onclick event is not detected for these buttons, although it was recently. Interestingly, if the spec is saved to HTML, the events are triggered, and tabbing works.

The code that does this is inline in the spec [here](https://github.com/w3c/json-ld-syntax/blob/master/index.html#L121-L134). You can see how it should work in the latest published spec [here](https://www.w3.org/TR/json-ld11/#ex-3-sample-json-ld-document-using-full-iris-instead-of-terms), or by saving the editors draft and trying there.

saschanaz:
Currently the feature is spread across core/examples and core/webidl, so this PR collects them as a single module.

Closes #2064


src/core/examples.js
```diff
@@ -33,13 +49,18 @@ export function run(conf) {
   if (!examples.length) return;
 
   document.head.insertBefore(
-    hyperHTML`<style>${css}</style>`,
+    html`
+      <style>
+        ${css}
+      </style>
+    `,
```
saschanaz: I don't like this but prettier now wants to to this 😕

marcoscaceres: could be worse 👍

tests/spec/core/examples-spec.js
```diff
@@ -39,10 +39,12 @@ describe("Core — Examples", () => {
     expect(markers.length).toBe(1);
 
     const [marker] = markers;
-    expect(marker.textContent).toBe("Example 1: EX");
+    expect(marker.textContent.trim()).toBe("Example 1: EX");
     expect(marker.querySelector(".example-title").textContent).toBe(": EX");
     expect(example.getAttribute("title")).toBeNull();
-    expect(example.textContent).toBe("Example 1: EX\n{\n  CONTENT\n}\n  ");
+    expect(example.textContent.trim()).toBe(
+      "Example 1: EX\n    \n{\n  CONTENT\n}"
```
marcoscaceres: Maybe we should normalize this and remove the spaces... we don't really care about the \n

marcoscaceres: (I guess if works for now, it's fine... otherwise, we will need to normalize the others)

saschanaz: I'll merge this first and do it on another PR.

PR #2083, opened 2019-02-08
fix(assets/bp.css): style best practices consistently

### LINKED ISSUE #2008: best practices boxes look weird
## Important info

* URL to affected spec: https://w3c.github.io/screen-orientation/#reading-the-screen-orientation
* ReSpec version: Click the ReSpec pill (e.g., 24.1.0).

## Description of problem
Looking at the screen orientation API spec, it contains a best practices box... but the box doesn't match the style of note, issue, etc. 

But what I was expecting?: for the best practices box to look like other things in ReSpec.

Johanna-hub:
closes #2008 

Now styled like examples, notes, webidl boxes but orange (open to other colour suggestions!)


marcoscaceres:
Could you please drag and drop a screenshot of what it looks like?

Johanna-hub:
screenshot 2019-02-08 at 16 51 57

marcoscaceres:
coincidently, the GitHub comment boxes are actually really nicely designed from both a contrast and color perspective... it would be nice to use colors similar to these boxes: 

screen shot 2019-02-10 at 7 18 30 pm

marcoscaceres:
one of the tests seems to be failing. @Johanna-hub, can you please take a look when you are back?

Johanna-hub:
Looks like this now, let me know your thoughts!
I've also fixed the test, it was expecting an extra space after the colon in the title, but I've done that spacing through CSS so took the space out in the test. I hope that's ok.

screenshot 2019-02-25 at 11 30 55

tests/spec/core/best-practices-spec.js
```diff
@@ -21,8 +21,8 @@ describe("Core — Best Practices", () => {
     const doc = await makeRSDoc(ops);
     const pls = doc.body.querySelectorAll("span.practicelab");
     const bps = doc.getElementById("bp-summary");
-    expect(pls[0].textContent).toEqual("Best Practice 1: BP1");
-    expect(pls[1].textContent).toEqual("Best Practice 2: BP2");
+    expect(pls[0].textContent).toEqual("Best Practice 1:BP1");
```
marcoscaceres: We probably want a space after the :

Johanna-hub: The space is there if you look in the example, but because of how I have styled it and split the title if I put a space there then the underline of the title goes out too far.  I can change that though if needed.

marcoscaceres: The problem is if someone copy/pastes the header, the space won’t be there. Thus,  I’d prefer the space.

Johanna-hub: ah ok, I understand! Have fixed

Johanna-hub:
Now styled like this:

screenshot 2019-02-25 at 13 24 03

marcoscaceres:
Looks really nice! Can you please check what TravisCI is upset about?

tests/spec/core/best-practices-spec.js
marcoscaceres: Just a nit, always use toBe(). To equal works, but it means something slightly different.

src/core/best-practices.js
```diff
@@ -21,7 +21,9 @@ export function run() {
       bp.textContent
     }</li>`;
     ul.appendChild(li);
-    bp.prepend(document.createTextNode(`Best Practice ${num}: `));
+    ul.classList.add("practicebox");
+    const title = hyperHTML`<span><span class="practicetitle">Best Practice ${num}</span>: </span>`;
```
marcoscaceres: Nested spans are probably not needed here... show you a trick 🎩 

```suggestion
    const title = hyperHTML`<span class="practicetitle">Best Practice ${num}</span>: `;
```

src/core/best-practices.js
```diff
@@ -21,7 +21,9 @@ export function run() {
...
+    bp.prepend(title);
```
marcoscaceres: And here I think this will work: 
```suggestion
    bp.prepend(...title.childNodes);
```

Issue #1640, opened 2018-04-27
[META] Support ES Modules


marcoscaceres:
When Firefox 60 lands in May, we will have ES module support across all target browsers. That means we can start considering using them in ReSpec. 

We should evaluate what that means for our current plugin design and for the dependencies (which are not ES Modules). With dynamic `import()` support, we can basically get rid of RequireJS. 

Ideally, we would then have ReSpec do this:

 1.  load a light stub (or some kind of master stub) for a plugin. These stubs intelligently determine if a plugin is needed: for example, "does this document require syntax highlighting? does this document use WebIDL?" and so on..
 1. For each "yes" stub, it would asynchronously kick off download of each needed module. These modules would have good cache control headers set (we could serve them from either the W3C server, or via respec.org, where we control the HTTP cache settings).
 1. Then we split the modules into group: needs further processing (e.g., anything that needs xrefs), or standalone (e.g., caniuse). 

Just some lose thoughts for now... but it's exciting because potentially this will make ReSpec super tiny, with each spec only loading just what it needs. Right now, we load everything, and that's super wasteful. The modules can be cached, and shared across specs... that may lead to interesting versioning issues, but we can deal with that. 

Thoughts?

saschanaz:
A downside is more round trips for the first visit, which is probably okay.

saschanaz:
Firefox 60 has landed 🎉

saschanaz:
>With dynamic import() support

It seems MSEdge on April 2018 update does not support dynamic import() yet.

saschanaz:
@marcoscaceres You recently said you are planning to use Webpack/Rollup. Do we still ultimately want to do ES modules then?

marcoscaceres:
Closing, as I think we will still want to bundle everything for the foreseeable future. 

We can always look at this again down the road... and, in fact, it kinda already works... we just don't deploy ReSpec this way.

PR #2085, opened 2019-02-11
chore(karma): added karma-coverage and coveralls

CodHeK:
UNTIL NOW:

- converted `tests/spec/core/linter-rules` to jest.
- applied `jest-codemods` to other test files, but needs many manual adjustments.
- config for jest is in `jest.config.js`


marcoscaceres:
Ok, step this right back :) only do the linter files and the required configuration. Trying to do all these files at once is going to be too much.

CodHeK:
@marcoscaceres I'll just checkout those files to match develop and push !

CodHeK:
@marcoscaceres restricted the changed to just the `linter-rules` , if you could please review the changes made.

.eslintrc.json
```diff
@@ -5,9 +5,10 @@
     "node": true,
     "jquery": true,
     "amd": true,
-    "worker": true
+    "worker": true,
+    "jest/globals": true
   },
-  "extends": ["eslint:recommended", "plugin:prettier/recommended"],
+  "extends": ["eslint:recommended", "plugin:prettier/recommended", "airbnb-base", "plugin:jest/recommended"],
```
marcoscaceres: What is the airbnb-base? Is it needed by jest?

.eslintrc.json
```diff
@@ -43,5 +44,5 @@
     "pickRandomsFromList": true,
     "respecConfig": true
   },
-  "plugins": ["prettier"]
+  "plugins": ["prettier", "import", "jest"]
```
marcoscaceres: What does import do here?

tests/spec/core/linter-rules/check-charset-spec.js
```diff
@@ -1,16 +1,12 @@
+import { rule } from  '../../../../src/core/linter-rules/check-charset.js';
```
marcoscaceres: I still wonder if we can import these with “/src/core/...”? If not, we need to figure out why.

package.json
```diff
@@ -80,7 +86,8 @@
     "test:build": "mocha ./tests/test-build.js",
     "test:headless": "node ./tests/headless.js",
     "test:karma": "npm run karma",
-    "prepare": "npm run snyk-protect && npm run build:components"
+    "prepare": "npm run snyk-protect && npm run build:components",
+    "test": "jest"
```
marcoscaceres: Probably want “test:jest” here, until we fully remove Karma.

saschanaz: Or we can just `npx jest`.

saschanaz:
Is Travis dead? It's not running 🤔

tests/spec/core/linter-rules/check-charset-spec.js
CodHeK: The `commonjs` module is needed for such imports , jest doesn't work with `amd` hence the first change...

.babelrc
```diff
@@ -4,8 +4,8 @@
       "@babel/preset-env",
       {
         "targets": "last 1 Firefox version, last 1 Chrome version, last 1 Safari version, last 1 Edge version",
-        "modules": "amd"
+        "modules": "commonjs"
```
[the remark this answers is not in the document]
CodHeK: `amd` was needed with `requireJS` but, since we're using jest now, using `amd` causes all tests to fail ... well, this was the most important change, that resulted in jest working smoothly  😛

.eslintrc.json
CodHeK: Its basically for es6 linting for the import export statements and other es6 syntax.

tests/spec/core/linter-rules/check-charset-spec.js
CodHeK: > I still wonder if we can import these with “/src/core/...”? If not, we need to figure out why.

We need the path to the file, otherwise it just returns an error saying it cannot find the file.

.babelrc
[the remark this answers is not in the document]
saschanaz: But we didn't fully remove requirejs yet and there are several tests still depending on requirejs (via `js/profile-w3c-common.js`).

.eslintrc.json
CodHeK: [eslint-plugin-import](uhttps://www.npmjs.com/package/eslint-plugin-import) It's this package.

.eslintrc.json
saschanaz: Can we do the eslint thing with another PR? I don't strongly feel the need as I'm not seeing any real style difference after adding the plugin.

.babelrc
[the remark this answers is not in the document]
CodHeK: But eventually, we'd have to change it to `commonjs` right ?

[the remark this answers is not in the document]
saschanaz: I think no, we intend to ultimately change it into `"esmodules": true`. #1640 

Currently the blocker is Firefox not supporting dynamic imports yet.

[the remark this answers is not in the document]
CodHeK: But, now we have the support for es6 imports in firefox too, so this should work right?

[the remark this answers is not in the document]
saschanaz: >we have the support for es6 imports in firefox too

For static imports yes, for dynamic imports no.

CodHeK:
We aren’t really using dynamic imports anywhere right , all these are just static ones ...

On Tuesday, February 12, 2019, 11:07 AM, Kagami Sascha Rosylight <notifications@github.com> wrote:


@saschanaz commented on this pull request.

In .babelrc:
> @@ -4,8 +4,8 @@
       "@babel/preset-env",
       {
         "targets": "last 1 Firefox version, last 1 Chrome version, last 1 Safari version, last 1 Edge version",
-        "modules": "amd"
+        "modules": "commonjs"


we have the support for es6 imports in firefox too


For static imports yes, for dynamic imports no.

saschanaz:
>We aren’t really using dynamic imports anywhere

`/js/profile-w3c-common.js` still depends on dynamic `require()` and to migrate to ES module we need support for dynamic imports. Luckily Firefox Nightly have dynamic import, but then we still have to remove the loader-specific syntax (`import {} from "text!filepath"`) because ES module system doesn't support it.

CodHeK:
So, how do we proceed for now ? I’ll change it to ‘commonjs’ locally but then I’ll push changing to ‘amd’ ? 


On Tuesday, February 12, 2019, 11:20 AM, Kagami Sascha Rosylight <notifications@github.com> wrote:



We aren’t really using dynamic imports anywhere


/js/profile-w3c-common.js still depends on dynamic require() and to migrate to ES module we need support for dynamic imports. Luckily Firefox Nightly have dynamic import, but then we still have to remove the loader-specific syntax (import {} from "text!filepath") because ES module system doesn't support it.

tests/spec/core/linter-rules/check-charset-spec.js
marcoscaceres: Could you try this for me:
```
console.log('../../../../src/core/linter-rules/check-charset.js', document.location); 
```

Interested to see what the whole URL is. From that, we should be able to determine the full path.

marcoscaceres: Fixed the above... forgot the URL bit... my brain is not working today.

marcoscaceres:
I think we might only be able to incrementally test via jest the modules that can be imported independently (e.g., linter, maybe "core/utils").  

Other alternative is seeing if Karma has a reported that can spit out the coverage format? That would solve the problem without needing Jest.

marcoscaceres:
Maybe:
https://github.com/caitp/karma-coveralls

marcoscaceres:
And:
https://github.com/karma-runner/karma-coverage

CodHeK:
@marcoscaceres whoa! This'll make the job much easier ! We wouldn't have to change any tests to then right ?

marcoscaceres:
> We wouldn't have to change any tests too then right ?

Correct... but those packages look kinda out of date. I'm being told to maybe use "Istanbul" instead?

CodHeK:
The karma-coverge package that’s you suggested uses Istanbul itself I guess ...


On Tuesday, February 12, 2019, 5:44 PM, Marcos Cáceres <notifications@github.com> wrote:



We wouldn't have to change any tests too then right ?


Correct... but those packages look kinda out of date. I'm being told to maybe use "Istanbul" instead?

CodHeK:
@marcoscaceres I was also thinking if we could get rid of the different frameworks ( mocha, jasmine, karma ) and use only jest instead ? The migration would take a lot of time, but that'll centralise everything to just one framework and less dependencies. just a thought ...

marcoscaceres:
> The migration would take a lot of time, but that'll centralise everything to just one framework and less dependencies

Yes, that would be ideal - but it will take us quite a while to get there. This has been great to see what we actually need to do to get coverage (so I’m super thankful for the work you’ve done here), and also to see the challenges we face with trying to move to jest. My feeling is that it’s probably not worth the time investment until we can get rid of Requirejs. 

If we can’t get coverage directly from Karma, we might have to go without for now. We have plenty of more important things to work on in the meantime.

CodHeK:
Well then, should I try to integrate the karma-coverage package and try it on some module first ?

marcoscaceres:
Yeah, let’s try it. Why not? 😀

CodHeK:
Great! I'll get started then!

CodHeK:
@marcoscaceres what does `mocha` in the `reporters` list do in `karma.conf.js` ?

CodHeK:
@marcoscaceres how do we proceed forward with this ?

marcoscaceres:
@CodHeK, we need to figure out get it working on Travis. Then I think we might be good.

CodHeK:
@marcoscaceres we also need to integrate `coveralls` right ? We might need a account on `coveralls.io` then

marcoscaceres:
> we also need to integrate `coveralls` right ? We might need a account on `coveralls.io` then

Yep, I already have an account. It's just waiting to receive data from TravisCI.

karma.conf.js
```diff
@@ -101,12 +102,17 @@ module.exports = function(config) {
...
+      // do not include tests or libraries
+      // (these files will be instrumented by Istanbul)
+      "src/**/*.js": "coverage",
+    },
 
     // test results reporter to use
     // possible values: "dots", "progress"
     // available reporters: https://npmjs.org/browse/keyword/karma-reporter
-    reporters: ["mocha"],
+    reporters: ["mocha", "progess", "coverage", "coveralls"],
```
saschanaz: `progess` probably a typo?

marcoscaceres:
@CodHeK it seems to be crashing Travis... you able to run this locally, right?

marcoscaceres:
Note, I checked the documentation and we don't need to add any tokens... it "should just work".

marcoscaceres:
Note, we might need: 
https://github.com/WebReflection/hyperHTML/blob/master/.travis.yml#L10

Some additional guidance:
https://medium.com/@WebReflection/js-vanilla-test-code-coverage-7b7ba3740776

CodHeK:
@marcoscaceres I had missed the `coveralls` script in `package.json` should work now I guess...

marcoscaceres:
@CodHeK could you try reverting the changes to package-lock.json?

marcoscaceres:
@CodHeK, did you get this running locally? I've checked out your branch and it's not working for me.

marcoscaceres:
```BASH
$ npm install
$ karma start --single-run karma.conf.js
19 02 2019 16:32:07.039:ERROR [preprocess]: Can not load "coverage", it is not registered!
  Perhaps you are missing some plugin?
19 02 2019 16:32:07.044:ERROR [karma-server]: Server start failed on port 9876: Error: No provider for "framework:jasmine"! (Resolving: framework:jasmine)
```

marcoscaceres:
Managed to get it going via global installs... which is really not great.... and now it's complaining that it doesn't recognize our usage of ES6 code?

marcoscaceres:
This is also what travis is complaining about here:
https://travis-ci.org/w3c/respec/jobs/495274901

CodHeK:
@marcoscaceres testing it locally, `npm run karma` that's what I need to run right ?

marcoscaceres:
yep, `npm run karma`

CodHeK:
[https://www.npmjs.com/package/coveralls#istanbul](https://www.npmjs.com/package/coveralls#istanbul) we'd have to add this to the scripts section I guess

marcoscaceres:
Ok, try and see if you can work it out locally. You'll need to do some investigation on your own. However, if it turns out it's too hard, then we might need to abort. Unfortunately, I've only got limited time to spend on this, and it's turning out to be more of a time drain than I'd originally hoped :(

CodHeK:
@marcoscaceres also, I think what the problem here is, coveralls is not able to fetch the `lcov` reports since the path set here is `cat ./coverage/lcov.info | coveralls` but the coverage folder generated has, 3 other folders with their respective `lcov.info` files

marcoscaceres:
Thanks again @CodHeK for doing this investigation. We learned a ton from you doing this and now have a better idea of what we need to do in the future to get coverage.

CodHeK:
Thanks @marcoscaeres ! It was great experience contributing to `respec`,  got lots to learn ! :D

PR #2088, opened 2019-02-12
fix(tools/respecDocWriter): emit all errors and warnings

### LINKED ISSUE #2062: Emit all errors/warning when with `-e -w`
Currently respec2html with `-e -w` halts when it meets the first error or warning message. It should instead emit all messages and then halt, so that the author can fix all things at once.

EvsChen:
Fix #2062


tools/respecDocWriter.js
```diff
@@ -250,9 +263,6 @@ function makeConsoleMsgHandler(page) {
           console.warn(colors.warn(`🚨 ${output}`));
           break;
       }
-      if (abortOnError || abortOnWarning) {
```
marcoscaceres: We still need to exit the process with the right code based on either a warning or error.

marcoscaceres: Ah sorry, I see you move it to the top. Ignore this comment.

tools/respecDocWriter.js
```diff
@@ -235,9 +244,13 @@ function makeConsoleMsgHandler(page) {
         // https://github.com/GoogleChrome/puppeteer/issues/1939
         return;
       }
-      const abortOnWarning = whenToHalt.haltOnWarn && type === "warning";
-      const abortOnError = whenToHalt.haltOnError && type === "error";
       const output = `ReSpec ${type}: ${colors.debug(text)}`;
+      if (type === "error") {
```
marcoscaceres: Nit: use switch statement here I think.

tools/respecDocWriter.js
```diff
@@ -91,6 +95,11 @@ async function fetchAndWrite(
     }
     await checkIfReSpec(page);
     const html = await generateHTML(page, url);
+    const abortOnWarning = whenToHalt.haltOnWarn && haltFlags.warn;
```
marcoscaceres: I feel like the the `handleConsoleMessages()` should now move into the `generateHTML()` function, and that the `generateHTML()` should return the html AND the if it emitted an error and/or a warning.

tools/respecDocWriter.js
saschanaz: Maybe can be merged with the following switch.

tools/respecDocWriter.js
saschanaz: `handleConsoleMessages()` must be called before `page.goto()`, that was the point of #2057.

marcoscaceres: Ah, keep forgetting. Sorry about that.

tools/respecDocWriter.js
EvsChen: Thanks for the advice! Already fixed.

saschanaz:
I'm yet to locally test this, I'll do soon™️

PR #2092, opened 2019-02-17
bug (core/highlights) : highlighting links in <pre> block

CodHeK:


CodHeK:
@marcoscaceres this simple fix, works!

src/core/highlight.js
```diff
@@ -58,7 +59,6 @@ export async function run(conf) {
         if (id !== msg.id) {
           return; // not for us!
         }
-        element.innerHTML = value;
```
marcoscaceres: Not sure how this works, as the `value` is supposed to be the highlighted HTML? What am I missing?

CodHeK:
What I am thinking is, previously the whole of <pre> tag content is being replaced with the text in all the elements inside it which was the problem originally right ? So, I believe the right change would be to not do that ... 


On Sunday, February 17, 2019, 3:16 PM, Marcos Cáceres <notifications@github.com> wrote:


@marcoscaceres commented on this pull request.

In src/core/highlight.js:
> @@ -58,7 +59,6 @@ export async function run(conf) {
         if (id !== msg.id) {
           return; // not for us!
         }
-        element.innerHTML = value;

Not sure how this works, as the value is supposed to be the highlighted HTML? What am I missing?

marcoscaceres:
> What I am thinking is, previously the whole of <pre> tag content is being replaced with the text in all the elements inside it which was the problem originally right ? So, I believe the right change would be to not do that ...

marcoscaceres:
> What I am thinking is, previously the whole of `<pre>` tag content is being replaced with the text in all the elements inside it which was the problem originally right ? So, I believe the right change would be to not do that ...

The `value` is the highlighted code tho (i.e., it's the HTML the syntax highlighter produced). So, if we don't replace the `innerHTML`, then we don't get any highlights. So, you are correct... it does fix the issue, but it means we get no highlighted code :)

CodHeK:
@marcoscaceres so, do we just proceed with using `class="nohighlight"` for such cases or re-write `respec-worker` to be able to handle such cases ?

marcoscaceres:
We decided not to fix this. Thanks for investigating it @CodHeK.

PR #2099, opened 2019-02-18
refactor(core/github-api): Centralize github fetching

### LINKED ISSUE #1631: Centralize github fetching
Fetching from github can require credentials and can be rate limited. It would be good to centralize fetching of github data through the github module.

devanshbatra04:
closes #1631 
I still need to write a bunch of tests however there were some small nits I wanted to clarify.
After fetching right now, I have stored the github issues as ```window.ghissues``` and the whole issues, comments and contributors as window.repoInfo. 
I don't see any similar usages in window scope for objects being shared across multiple modules. How should I go about it? A new object like conf?


marcoscaceres:
Yes, please avoid adding things to window. If you need to export an object use “export”.

devanshbatra04:
Here's the thing, I need to call a function in ```run(conf)``` and export the result. 
Basically, in ```core/github-api`` I fetch all the issues, comments and contributors and have them available in ```core/contrib``` and ```core/issues-notes```
How do I export this object only after the run method promise has resolved?

marcoscaceres:
> How do I export this object only after the run method promise has resolved?

You can try exporting the promise that eventually resolves with the object.

src/core/contrib.js
```diff
@@ -5,7 +5,7 @@
 // #gh-contributors: people whose PR have been merged.
 // Spec editors get filtered out automatically.
 import { flatten, joinAnd } from "./utils";
-import { fetchIndex } from "./github";
+import { getArgs } from "./github-api";
```
marcoscaceres: The name `getArgs` is overly generic so it's not clear what it does... could you give this a better name?

js/profile-w3c-common.js
```diff
@@ -37,7 +37,8 @@ define([
   "./core/style",
   "./w3c/style",
   "./w3c/l10n",
-  "./core/github",
+  "./core/github-config",
```
marcoscaceres: I wonder if we can avoid one of the renames?

src/core/contrib.js
```diff
@@ -23,8 +23,18 @@ function findUserURLs(...thingsWithUsers) {
   return [...new Set(usersURLs)];
 }
 
+function getHeaders(conf) {
```
marcoscaceres: This extracts the GitHub credential,s and returns them as a header... so `extractGithubCredentials(conf)`

src/core/contrib.js
```diff
@@ -23,8 +23,18 @@ function findUserURLs(...thingsWithUsers) {
...
+  const headers = {};
+  const { githubUser, githubToken } = conf;
+  if (githubUser && githubToken) {
+    const credentials = btoa(`${githubUser}:${githubToken}`);
+    const Authorization = `Basic ${credentials}`;
+    Object.assign(headers, { Authorization });
+  }
```
marcoscaceres: We should add support for token only authorization too, while we are here... so, something like:

```JS
if (!githubUser && githubToken) {
  const credentials = btoa(`${githubToken}`);
  const Authorization = `Token ${credentials}`;
  Object.assign(headers, { Authorization });
}
```

The above could be merged more cleanly with the current code.

src/core/contrib.js
marcoscaceres: This function seems like it should live in the Github API module... not in contrib.

src/core/github-api.js
```diff
@@ -0,0 +1,157 @@
+import { fetchAndCache } from "./utils";
+import { pub } from "./pubsubhub";
+export const name = "core/github-api";
+
+const MAX_GITHUB_REQUESTS = 60;
```
marcoscaceres: We probably don't need this anymore, to be honest... we should be using the "requests left" information from github directly.

src/core/contrib.js
marcoscaceres: Ok yeah... this is all redundant. See:
https://github.com/w3c/respec/pull/2099/files#diff-fd8e14e6c335ae838f06d979287518d8R27

This should be going through that code path... or that code path should be refactored to handle this case too.

src/core/github-api.js
```diff
@@ -0,0 +1,157 @@
...
+export function fetchIndex(url, headers) {
+  // converts URLs of the form:
+  // https://api.github.com/repos/user/repo/comments{/number}
+  // into:
+  // https://api.github.com/repos/user/repo/comments
+  // which is what you need if you want to get the index.
+  return fetchAll(url.replace(/\{[^}]+\}/, ""), headers);
+}
+
+export async function getArgs(urls, editors, elements, headers) {
```
marcoscaceres: This one seems overly specific to be here. You will notice half the arguments are not used.

src/core/github-api.js
```diff
@@ -0,0 +1,157 @@
...
+  return await Promise.all(urls.map(url => fetch(url, { headers })));
+}
+
+export async function run(conf) {
```
marcoscaceres: This should definitely not be here... this is all related to issues and notes.

Swapnilr1:
@marcoscaceres @devanshbatra04 Sorry for the delay - I was busy with academic work. 
I have implemented some changes as suggested, though quite a lot of work is still left. Please review.

PR #2101, opened 2019-02-18
Refactor(core/utils): joinAnd uses Intl.ListFormat if available

### LINKED ISSUE #2096: core/utils joinAnd function should use ListFormat
It would be great if core/utils's [`joinAnd()`](https://github.com/w3c/respec/blob/develop/src/core/utils.js#L382) would make use of the new [ListFormat API](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ListFormat). 

What needs to change:

 1. core/utils should import `lang as docLang` from core/i18n. 
 1. `joinAnd()` should gain a third argument lang that defaults to `docLang`. 
 1. If `intl.ListFormat` is available, use it. 
 1. Otherwise, fallback to existing code. 
 1. Bonus: add support for some localization of `joinAnd()` - we should at least support Dutch, but would be nice to support some other languages.

devanshbatra04:
closes #2096


devanshbatra04:
The build shouldn't really be failing :(

src/core/utils.js
```diff
@@ -379,19 +380,29 @@ export class IDBKeyVal {
...
-      const str = items.join(", ");
-      const lastComma = str.lastIndexOf(",");
-      return `${str.substr(0, lastComma + 1)} and ${str.slice(lastComma + 2)}`;
+  if (Intl.ListFormat && typeof Intl.ListFormat === "function") {
+    const formatter = new Intl.ListFormat(lang, {
+      style: "long",
+      type: "conjunction",
+    });
+    return formatter.format(items);
+  } else {
```
marcoscaceres: Nit: you don’t else here, as the code returns above.

tests/spec/core/utils-spec.js
```diff
@@ -363,7 +363,11 @@ describe("Core - Utils", () => {
   });
 
   describe("joinAnd", () => {
-    it("joins with proper commas and 'and'", () => {
+    it("joins with proper commas and 'and' when Intl.ListFormat is not available", () => {
```
marcoscaceres: I don’t think we need these modifications, as the use of the intl API is an implementation detail.

tests/spec/core/utils-spec.js
```diff
@@ -375,6 +379,22 @@ describe("Core - Utils", () => {
         })
       ).toEqual("X, X, X, and X");
     });
+    it("joins with proper commas and 'and' when Intl.ListFormat is available", () => {
```
marcoscaceres: Same as above, all that matters is that the right output is produced, not that the intl API is involved.

tests/spec/core/utils-spec.js
devanshbatra04: I think this is essential as the tests run in a browsers, and only one of the two code blocks will be executed, depending on whether or not the browser has ```Intl.ListFormat```
More clearly, if we do away with the tests then only our original code will be tested, and once chrome/firefox incorporate ```Intl.ListFormat```, only the newer code will be tested.

src/core/utils.js
devanshbatra04: I must say, I saw that coming :laughing:

tests/spec/core/utils-spec.js
marcoscaceres: I understand the rationale, but then it’s a matter of testing in multiple browsers (what Karma does) of which at least one contains the new API. The tests should primarily test the expected output, otherwise it becomes increasingly difficult to refactor the code. Imagine tomorrow a new intl API is released that was better/faster - it would mean these tests would potentially fail even though they produce the correct output.

devanshbatra04: Okay. Makes sense. Thank you so much for the clarification!

devanshbatra04:
Build seems to be failing at random. Something with Snyk. This happened with my other PR too. Something that can be done?

tests/spec/core/utils-spec.js
marcoscaceres: No problem. Thanks for hearing me out. Appreciate that you went and tested the behaviour nonetheless!

marcoscaceres:
Yeah, Snyk is being bad tonight :( I see them trying to fix stuff. Should hopefully stabilize soon.

PR #2107, opened 2019-02-19
chore(package): don't snyk during prepare

marcoscaceres:
Snyk is completely broken at the moment. It's good to run snyk during build, but running it during prepare is 🤷‍♂️.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2457,7 +2457,7 @@
         },
         "string-width": {
           "version": "1.0.2",
-          "resolved": "https://registry.npmjs.org/string-width/-/string-width-1.0.2.tgz",
+          "resolved": "http://registry.npmjs.org/string-width/-/string-width-1.0.2.tgz",
           "integrity": "sha1-EYvfW4zcUaKn5w0hHgfisLmxB9M=",
           "requires": {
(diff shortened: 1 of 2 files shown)


saschanaz:
I think that line is previously intentionally added by the snyk bot... BTW, #2104 passed the test, what's the problem?

marcoscaceres:
> BTW, #2104 passed the test, what's the problem?

I had to restart it a few times to get it to pass. The problem appear to be that it intermittently fails (like 50% on every stage). I've been restarting all the failing stages manually to get them to pass... but it's getting a little old.

marcoscaceres:
I think what may be happening is that snyk is corrupting the cache on TravisCI when it does the fixes. I think it's doing the patches, then it gets confused because it can't find things that its patched.

saschanaz:
#1413 and #1420 added the hook to run snyk, maybe removing it would trigger a new PR by snyk-bot to restore, but not sure.

marcoscaceres:
I think we can add back to prepare again soon (and it runs at build time, so I'm ok with the current state of things)... but right now it's just too broken. Other option is disabling the cache on TravisCI, but that also slows us down.

Issue #1858, opened 2018-10-21
Generating MDN annotations in spec output


sideshowbarker:
This is a feature proposal for a new feature to automatically generate MDN annotations in spec output.

For any specific feature in a spec, ReSpec would automatically generate an MDN annotation with:

* a link to any MDN articles that document that feature
* the summary description of that feature from the MDN article (available as tooltip text on hovering over the annoation)
* the (caniuse-like) browser-compatibility support matrix for that feature

A similar feature has been implemented for the HTML spec, and output for that is available here:

https://arcane-cove-57093.herokuapp.com/multipage/

Here’s a ~15-second screen recording showing the annotations as implemented for the HTML spec —

![mdnannos720](https://user-images.githubusercontent.com/194984/46925623-c0d20b80-d067-11e8-85d9-94e70e267135.gif)

Data that can be used to build the annotation content is available at https://github.com/w3c/mdn-spec-links and https://w3c.github.io/mdn-spec-links/#user-content-spec-json-files

For example, data for the Payment Request API spec is available here:

https://w3c.github.io/mdn-spec-links/payment-request.json

EvsChen:
Any updates on this issue? would like to pick this up

marcoscaceres:
feel free to take it. You might be able to integrate with whatever the HTML spec is doing/using.

EvsChen:
Thanks! I looked at the way html spec does this. They look for element `id` in the MDN json to determine whether to insert the MDN annotation or not. Are we going to follow this or use some new attributes?

For me, a prefix id syntax like `<code id="mdn:sessionStorage"></code>` may be more effective in dom parsing.

marcoscaceres:
I'd prefer we do whatever HTML is already doing: that way we can leverage and maybe improve whatever @sideshowbarker already built. Then we don't need to maintain as much code and we get benefit across HTML and all ReSpec specs.

EvsChen:
I think their build tool `whatwg/wattsi` is using pascal. It's kind of hard to "integrate" what they are doing. At best we could follow their procedure.

marcoscaceres:
Ok, I meant more just on using the same HTML IDs/hooks and the JavaScript that is driving the widget/loading the JSON form MDN... assuming the widget above is just being generated with JavaScript. If the integration is happening at using Pascal, then yeah... we will roll our own.

sideshowbarker:
[...]

PR #2110, opened 2019-02-20
feat(core/mdn-annotation): allow MDN annotations

EvsChen:
Create a new plugin in core for the feature request in #1858


src/core/mdn-annotation.js
```diff
@@ -0,0 +1,34 @@
...
+export const name = "core/mdn-annoatation";
+
+const SPEC_MAP_URL =
+  "https://raw.githubusercontent.com/w3c/mdn-spec-links/master/SPECMAP.json";
+const JSON_BASE = "https://w3c.github.io/mdn-spec-links/";
+
+async function fetchAndCacheJson(url, maxAge) {
+  if (!url) return {};
+  const request = new Request(url);
+  const response = await fetchAndCache(request, maxAge);
```
marcoscaceres: Maybe just 
```suggestion
  return fetchAndCache(request, maxAge).then(r => r.json());
```

EvsChen: Yes, seems like a separate function may not be necessary here.

EvsChen:
Thanks @marcoscaceres ! I'll just keep working on this PR without opening a new one.

EvsChen:
Hi @marcoscaceres, I've made some updates on this. Now it supports displaying a collapsible annotation box with link. 
image
image

src/core/mdn-annotation.js
```diff
@@ -0,0 +1,83 @@
...
+  );
+  const nodesWithId = document.querySelectorAll("[id]");
+  const unlikelyTagNames = ["STYLE", "SCRIPT", "BODY"];
+  document.head.appendChild(hyperHTML`<style>${[mdnCss]}</style>`);
+  document.head.appendChild(hyperHTML`<script>
+     function toggleStatus(div) {
+       div.parentNode.classList.toggle('wrapped');
+     }
+  </script>`);
+  nodesWithId.forEach(node => {
```
marcoscaceres: We might want to consider for the first two `if` statements, given that they are acting as filters. 

```suggestion
  [...nodesWithId].filter(node => {
```

src/core/mdn-annotation.js
```diff
@@ -0,0 +1,83 @@
...
+    }
+    const mdnSpecArray = mdnSpecJson[node.id];
+    let targetAncestor = node;
+    // Find the furthest ancestor that is a direct child of <body>
+    while (targetAncestor.parentNode.tagName !== "BODY") {
+      targetAncestor = targetAncestor.parentNode;
+    }
+    const targetSibling = targetAncestor.previousElementSibling;
+    let mdnBox;
+    if (targetSibling && targetSibling.class === "mdn before") {
```
marcoscaceres: use `classList.contains()` instead of string comparisons here.

src/core/mdn-annotation.js
```diff
@@ -0,0 +1,83 @@
...
+      // If the target ancestor already has a mdnBox inserted, we just use it
+      mdnBox = targetSibling;
+    } else {
+      mdnBox = hyperHTML`<aside class="mdn before">
+        <input type="button" onclick="toggleStatus(this)" value="⋰" />
```
marcoscaceres:
```suggestion
        <input type="button" onclick="toggleStatus(this)" value="⋰">
```

src/core/mdn-annotation.js
```diff
@@ -0,0 +1,83 @@
...
+      </aside>`;
+      document.body.insertBefore(mdnBox, targetAncestor);
+    }
+    mdnSpecArray.forEach(spec => {
```
marcoscaceres: Nicer to split this up into map, and then add them
```JS
    mdnSpecArray
      .map(spec => {...})
      .forEach(mdnDiv => mdnBox.appendChild(mdnDiv) )
```

src/core/mdn-annotation.js
```diff
@@ -0,0 +1,170 @@
...
+  samsunginternet_android: "Samsung Internet",
+  // uc_android: "UC browser", // not enough data for features in HTML
+  // uc_chinese_android: "Chinese UC Browser", // not enough data for features in HTML
+  webview_android: "WebView Android",
+};
+
+async function fetchAndCacheJson(url, maxAge) {
+  if (!url) return {};
+  const request = new Request(url);
+  return await fetchAndCache(request, maxAge).then(r => r.json());
```
marcoscaceres:
```suggestion
  return fetchAndCache(request, maxAge).then(r => r.json());
```

src/core/mdn-annotation.js
```diff
@@ -0,0 +1,170 @@
...
+  // qq_android: "QQ Browser", // not enough data for features in HTML
+  safari: "Safari",
+  safari_ios: "Safari iOS",
+  samsunginternet_android: "Samsung Internet",
+  // uc_android: "UC browser", // not enough data for features in HTML
+  // uc_chinese_android: "Chinese UC Browser", // not enough data for features in HTML
+  webview_android: "WebView Android",
+};
+
+async function fetchAndCacheJson(url, maxAge) {
```
marcoscaceres:
```suggestion
function fetchAndCacheJson(url, maxAge) {
```

src/core/mdn-annotation.js
```diff
@@ -0,0 +1,170 @@
...
+  while (targetAncestor.parentNode.tagName !== "BODY") {
+    targetAncestor = targetAncestor.parentNode;
+  }
+  const targetSibling = targetAncestor.previousElementSibling;
+  if (targetSibling && targetSibling.classList.contains("mdn")) {
+    // If the target ancestor already has a mdnBox inserted, we just use it
+    return targetSibling;
+  }
+  const mdnBox = hyperHTML`<aside class="mdn before">
+    <input type="button" onclick="toggleStatus(this)" value="⋰" >
```
marcoscaceres: Maybe... 
```suggestion
    <button onclick="toggleStatus(this)"  aria-label="Expand MDN details">⋰</button>
```

src/core/mdn-annotation.js
```diff
@@ -0,0 +1,170 @@
...
+      } else {
+        processBrowserData(browserId, support[browserId]);
+      }
+    }
+  });
+}
+
+export async function run(conf) {
+  const { shortName, mdnAnnotation } = conf;
+  if (!shortName) {
```
marcoscaceres: So, we should only add the mdn box if `mdnAnnotation` is set...
 
```suggestion
  if (!shortName || !mdnAnnotation) {
```

Also, we probably just want to call it `mdn` as "Annotation" is hard to spell.

src/core/mdn-annotation.js
```diff
@@ -0,0 +1,170 @@
...
+  return await fetchAndCache(request, maxAge).then(r => r.json());
+}
+
+function insertMDNBox(node) {
+  let targetAncestor = node;
+  // Find the furthest ancestor that is a direct child of <body>
+  while (targetAncestor.parentNode.tagName !== "BODY") {
+    targetAncestor = targetAncestor.parentNode;
+  }
+  const targetSibling = targetAncestor.previousElementSibling;
```
marcoscaceres:
```suggestion
  const {previousElementSibling: targetSibling} = targetAncestor;
```

src/core/mdn-annotation.js
```diff
@@ -0,0 +1,170 @@
...
+}
+
+function attachMDNDetail(container, mdnSpec) {
+  const { slug, summary } = mdnSpec;
+  container.innerHTML += `<b>MDN </b>`;
+  const mdnSubPath = slug.slice(slug.indexOf("/") + 1);
+  const mdnDetail = document.createElement("details");
+  mdnDetail.innerHTML += `
+    <summary>
+      <a title="${summary}" href="${MDN_URL_BASE}${slug}">${mdnSubPath}</a>
```
marcoscaceres: maybe put ${MDN_URL_BASE}${slug} into a single variable `const href=`?

src/core/mdn-annotation.js
```diff
@@ -0,0 +1,170 @@
...
+  attachMDNBrowserSupport(mdnDetail, mdnSpec);
+  container.appendChild(mdnDetail);
+}
+
+function attachMDNBrowserSupport(container, mdnSpec) {
+  if (!mdnSpec.support) {
+    container.innerHTML += `<p class="nosupportdata">No support data.</p>`;
+    return;
+  }
+  const supportTable = hyperHTML`<p class="mdnsupport"></p>`;
```
marcoscaceres: Was expecting something like:
```JS
const supportTable = hyperHTML`<p class="mdnsupport">${
  buildBrowserSupportTable()
}
</p>`;
```

Also... if it's a table... use a `<table>` instead of a paragraph and with spans 😉

EvsChen:
Thanks @marcoscaceres! I'll update the code soon.

src/core/mdn-annotation.js
EvsChen: That's a good idea. I find it hard to spell too...

src/core/mdn-annotation.js
EvsChen: Yes, that's true. It's more like a table or a list. But that would also cause a lot of changes in the css file. Maybe it's simpler to keep it consistent with the HTML spec?

marcoscaceres: heh, did they seriously use spans?

EvsChen: Yes, the browser support list is a `<p>` containing containing `<span>`s for differents browsers. Also each browser `<span>` contains `<span>` for icon and support version.

marcoscaceres: I guess if it's good enough for the html spec 🤷‍♂️

EvsChen:
Hi @marcoscaceres, any updates on this? Seems I've resolved all the comments.

marcoscaceres:
Apologies @EvsChen! This fell off the radar. I'll have another look now and try it out.

assets/mdn-annotation.css
```diff
@@ -0,0 +1,236 @@
...
+    background: no-repeat center center;
+    background-image: none;
+    background-size: auto auto;
+    background-size: contain;
+    text-align: right;
+    font-size: 0.75em;
+    font-weight: bold;
+}
+
+.mdnsupport>.chrome::before {
```
sidvishnoi: Can we use CSS variables here?

marcoscaceres:
Having a little play in Payment Request... It seems to attach itself correctly to the interface, but not the the actual definitions attribute and method:

Screen Shot 2019-04-16 at 4 00 04 PM

I was expecting, for instance, "PayerErrors" to be on the actual PayerError definition, and so on... is that because the data is bad? or is it something we might be able to overcome?

marcoscaceres:
Be nice to also get rid of the blue highlight:

Screen Shot 2019-04-16 at 4 03 33 PM

And make all the  "MDNs" closed by default... has them all semi-open by default, but doesn't seem great to have them open.

src/core/mdn-annotation.js
```diff
@@ -0,0 +1,174 @@
...
+function fetchAndCacheJson(url, maxAge) {
+  if (!url) return {};
+  const request = new Request(url);
+  return fetchAndCache(request, maxAge).then(r => r.json());
+}
+
+function insertMDNBox(node) {
+  let targetAncestor = node;
+  // Find the furthest ancestor that is a direct child of <body>
+  while (targetAncestor.parentNode.tagName !== "BODY") {
```
marcoscaceres: @sidvishnoi and I were discussing this... we think you should attach to the nearest `.closest("section")`.

marcoscaceres: That would solve the grouping problem, where you end up with too many definitions being attached to an interface.

assets/mdn-annotation.css
```diff
@@ -0,0 +1,236 @@
...
+    margin-left: 8px;
+    padding-bottom: 8px;
+}
+
+.mdnsupport {
+    display: table;
+    margin-top: 4px;
+}
+
+.mdnsupport>span {
```
sidvishnoi: Please run this file through prettier

assets/mdn-annotation.css
```diff
@@ -0,0 +1,236 @@
...
+.mdnsupport>span.unknown::before {
+    opacity: 0.5;
+}
+
+.mdnsupport>span::before {
+    content: "";
+    display: table-cell;
+    min-width: 1.5em;
+    height: 1.5em;
+    background: no-repeat center center;
```
sidvishnoi: With CSS variable (comment below), we can simplify this to:
``` css
body {
  /* ... */
  background: var(--bg) no-repeat center center / contain;
}
```

assets/mdn-annotation.css
sidvishnoi: Like
``` css
.mdnsupport > .chrome {
  --bg: url(https://resources.whatwg.org/browser-logos/chrome.svg);
}
```

src/core/mdn-annotation.js
```diff
@@ -0,0 +1,174 @@
...
+  // qq_android: "QQ Browser", // not enough data for features in HTML
+  safari: "Safari",
+  safari_ios: "Safari iOS",
+  samsunginternet_android: "Samsung Internet",
+  // uc_android: "UC browser", // not enough data for features in HTML
+  // uc_chinese_android: "Chinese UC Browser", // not enough data for features in HTML
+  webview_android: "WebView Android",
+};
+
+function fetchAndCacheJson(url, maxAge) {
```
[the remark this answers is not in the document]
sidvishnoi: Maybe we should remove this function altogether and use `fetchAndCache(...).then(r => r.json())` directly below

src/core/mdn-annotation.js
```diff
@@ -0,0 +1,174 @@
...
+  if (!hasSpecJson) {
+    return;
+  }
+  const mdnSpecJson = await fetchAndCacheJson(
+    `${JSON_BASE}/${shortName}.json`,
+    maxAge
+  );
+  document.head.appendChild(hyperHTML`<style>${[mdnCss]}</style>`);
+  document.head.appendChild(hyperHTML`<script>
+     function toggleStatus(div) {
```
sidvishnoi:
```suggestion
     function toggleMDNStatus(div) {
```

marcoscaceres:
@EvsChen, please also add `"core/mdn-annotation"` to `profile-w3c.js` and `profile-w3c-common.js`.

assets/mdn-annotation.css
EvsChen: Will change. I copied all the css from the original HTMLWG repo and just kept their version.

EvsChen: Checked again. Maybe I'll just group the selector without adding variables, like
```css
.mdnsupport > .chrome,
.mdnsupport>.chrome_android::before {
  --bg: url(https://resources.whatwg.org/browser-logos/chrome.svg);
}
```

marcoscaceres: I wonder if we should link https://resources.whatwg.org/standard-shared-with-dev.css instead? Then we don't need to worry about maintaining the style and it will always match HTML?

src/core/mdn-annotation.js
EvsChen: Is there any case that an element does not have a `<section>` as ancestor?

marcoscaceres: no, `<section>`s are required in ReSpec.

src/core/mdn-annotation.js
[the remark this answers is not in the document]
EvsChen: Since we have other conf like `specMapUrl`, would it be better to group them together within the `run` function? Putting config value and default value together seems more readable to me.

EvsChen:
I tried using the closest section to attach the mdn box. It's much better but still there will be cases that  multiple definitions are grouped. I think it's acceptable.

图片

marcoscaceres:
Agree, the grouping is fine. I guess it will fix itself if the IDs get updated upstream.

src/core/mdn-annotation.js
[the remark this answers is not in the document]
marcoscaceres: Don't have a strong opinion, so ok to keep it in `run()`

marcoscaceres:
Ok, this is looking pretty good now. I think we just need tests and it should be good to go.

marcoscaceres:
Actually, could you please fix this:

Screenshot 2019-04-18 16 32 29


Seems more margin top is needed to get them to MDN box to align with the heading.

EvsChen:
Cool! Thx @marcoscaceres. I'll update the tests soon.

marcoscaceres:
ok, let us know if we can help with the tests. If you do start writing some, be mindful that you need to do:

```
 ./tools/builder.js --profile=w3c --debug; karma start
```

Testing is kinda busted right now until we get the Webpack dev server running (you need to rebuild every time). Hoping to get that fixed soon.

assets/mdn-annotation.css
```diff
@@ -149,6 +149,10 @@
   background-image: url(https://cdnjs.loli.net/ajax/libs/browser-logos/41.0.0/android-webview-beta/android-webview-beta_32x32.png);
 }
 
+.mdn.wrapped div:nth-child(n + 3) {
```
marcoscaceres: What does this actually do?

EvsChen: Otherwise there will be blank space if multiple mdn slugs are associated with one box. It seems a bit ugly to me.
图片

EvsChen:
Add a test for this plugin. PTAL:)

tests/spec/core/mdn-annotation-spec.js
```diff
@@ -0,0 +1,72 @@
...
+    const { previousElementSibling } = paymentInterfaceSection;
+    const mdnSupport = previousElementSibling.querySelector("p.mdnsupport");
+    expect(mdnSupport).toBeTruthy();
+    const browserRows = previousElementSibling.querySelectorAll(
+      "p.mdnsupport > span"
+    );
+    expect(browserRows.length).toBeGreaterThan(0);
+  });
+
+  it("should display correct browser support info", () => {
```
marcoscaceres:
```suggestion
  it("displays correct browser support info", () => {
```

marcoscaceres:
Enabled in Payment Request spec:
https://w3c.github.io/payment-request/

🎉 🔥🎉 🔥🎉 🔥🎉 🔥🎉 🔥🎉 🔥

EvsChen:
Thanks for the help! @marcoscaceres @sidvishnoi

PR #2111, opened 2019-02-20
fix(core/highlight): Highlighted code wrapped in <code>

### LINKED ISSUE #2098: Highlighted code should be wrapped in code element
Code coming back from the syntax highlighter should be wrapped in a `<code>` element, so to semantically conform with HTML spec.

CodHeK:
Fixes #2098


src/core/highlight.js
```diff
@@ -58,7 +58,12 @@ export async function run(conf) {
...
         }
-        element.innerHTML = value;
+        const codeElements = element.querySelectorAll('code');
+        if(codeElements.length > 0)
+          element.innerHTML = value;
+        else {
+          element.innerHTML = `<code>${value}</code>`;
+        }
         if (element.localName === "pre") {
           element.classList.add("hljs");
```
saschanaz: Looks okay if we only want simple wrapping, but I think new classes should also go `<code>` instead of `<pre>`. Not exactly sure about `hljs` but at least the language hint should. Thoughts? @marcoscaceres

CodHeK: new classes as in ?

saschanaz: I mean:

```html
<pre class="js">
...code...
</pre>
```

should become:

```html
<pre><code class="js">
...code...
</code></pre>

Comment (saschanaz) [reply src/core/highlight.js]: (Note that we have `utils.renameElement()`, so we may just "rename" the `<pre>` into a `<code>` and then add it to a new `<pre>`.)

Comment (CodHeK) [reply src/core/highlight.js]: `class="js"` is just an example right ? normally, we won't add a `language-class`  to `<pre>` bcz the highlighter identifies the language by itself ?

Comment (CodHeK) [reply src/core/highlight.js]: > I mean:
> 
> ```
> <pre class="js">
> ...code...
> </pre>
> ```
> 
> should become:
> 
> ```
> <pre><code class="js">
> ...code...
> </code></pre>
> ```

You mean any class that is associated with `<pre>` also needs to added to `<code>` or only the language class, bcz only that'd be relevant to the `<code>` tag right ?

Comment (saschanaz) [reply src/core/highlight.js]: We shouldn't transfer everything, but only language class and `hljs`.

My suggestion:

1. If `<pre>` have no `<code>` then `utils.wrapInner()` by `<code>` but still use the class names from `<pre>` to get language hints for highlighter.
2. The highlighter will return the exact language name later. Add it to `<code>`, and remove it from `<pre>`.

Comment (CodHeK) [reply src/core/highlight.js]: There is a specific format for writing a language class in `<pre>` right ? `class="lang-js"` prefixed with `lang` ?

Comment (CodHeK) [reply src/core/highlight.js]: When the `<pre>` has no language class mentioned, it predicts it being `javascript` almost every time 😛  I tried with c++/python it still returns `language: ["javascript"]` from the worker.

Comment (marcoscaceres) [reply src/core/highlight.js]: The highlight bundle deliberately only knows about html/xml, css, js, http, and a very restricted set of others. It’s be design that it only matches those - c++ or phython won’t ever match because we deliberately don’t support them.

Comment (CodHeK) [reply src/core/highlight.js]: The changes, do they look good ? It's for the case when no `<code>` element is found inside of `<pre>`

**Comment by CodHeK** [comment]
I've 

> @CodHeK, what might help you more is using [test driven development](https://en.wikipedia.org/wiki/Test-driven_development): i.e., write tests as you code to help you answer questions like this (and actually run your code, try stuff out in the browser). Also, go and look at other ReSpec specs to see how people have used `pre` and `code`.
> 
> Although we really do appreciate you helping, you need to be really mindful that neither @saschanaz or I work on this project full time (it's a side project for us). We are also mindful that you are a student and still learning (and want to help you on that journey). But trying things and asking if you are right every step of the way is not the right approach: it eats up too much of our time, and we don't have time to be mentoring or answering questions like this every step of the way.
> 
> I don't want to discourage you from participating, but I feel like a lot of this might be beyond your skill level at the moment. If you want to skill up, try taking the TDD approach I mentioned above (i.e., write code against tests - and come up with your own hard set of tests, based on what users do - or what you assume they will do).

Really sorry sir, it's just that I am new to contributing to open source, I'll make sure to implement my workflow as you've suggested ...

**Comment by marcoscaceres** [comment]
> Really sorry sir, it's just that I am new to contributing to open source, I'll make sure to implement my workflow as you've suggested ...

No need to apologize. I think you will do great and you've shown really charisma by jumping in contributing the way you have: That takes real guts and I command you for doing that. Keep reading and keep learning and you'll get good at this - promise!

**Comment by CodHeK** [comment]
Yes, sure Sir!

File: tests/spec/core/highlight-spec.js [hunk 2]
```diff
@@ -76,4 +76,20 @@ describe("Core — Highlight", () => {
...
+
+  it("checks if <pre> content is warpped in <code>", async () => {
+    const ops = {
+      config: makeBasicConfig(),
+      body: `${makeDefaultBody()}<section>
+          <pre>
+            function foo() {
+              alert('foo');
+            }
+          </pre>
```
marcoscaceres: don't forget to also test:

```JS
<p><dfn>a link</dfn></p>
<pre><!--empty--></pre>
<pre>
   <a>a link</a>
  <code>
    function one(){}
  </code>
  this function(){} is not highlighted. 
  <code class="http">
     Header: Test
  </code>
</pre>

```

examples/basic.html
```diff
@@ -31,6 +31,11 @@
     <section id='sotd'>
       <p>
         This is required.
+        <pre>
```
marcoscaceres: Please remove this. It's ok to use basic.html for your own testing, but no need to modify the basic.html file.

tests/spec/core/highlight-spec.js
CodHeK: Yes, as of now I haven't added the case when there is `<code>` inside `<pre>` so, I'll add this when I handle this case in `highlight.js`, so this test is just when there is no `<code>` inside `<pre>`.

marcoscaceres: Ok, that's fine - so make sure you always check the default condition too (i.e., when there is a `<code>`):

```JS
<pre><code lang="http">test</code></pre>
```

To make sure that it doesn't get double wrapped by code, and that the language remains intact, etc.

saschanaz: I thought `lang` is for spoken languages?

marcoscaceres: sorry, yes... that should be class="http".

marcoscaceres: fixed above.

CodHeK:
I am a little confused as to what should it do in cases like these ... The case when there are no `<code>` tags seems to be done.

```
<pre>
   <a>a link</a>
  <code>
    function one(){}
  </code>
  this function(){} is not highlighted. 
  <code class="http">
     Header: Test
  </code>
</pre>
```
Do I have to preserve the `<code>` tags, which are currently getting converted into `<span>` elements.

saschanaz:
>Do I have to preserve the `<code>` tags

Yes. In that case we shouldn't pass the content of `<pre>`, instead we should pass the content of each `<code>` using their own language hints.

CodHeK:
@saschanaz for the case when there is `<code>` inside `<pre>` ...

```
<pre>
  <code>
    function one(){}
  </code>
  this function(){} is not highlighted. 
  <code class="http">
     Header: Test
  </code>
</pre>
```
This gets converted to : 

```
<pre aria-busy="false">
    <code class="js hljs">
        <span class="hljs-function"><span class="hljs-keyword">function</span> <span class="hljs- 
        title">one</span>(<span class="hljs-params"></span>)</span>{}
    </code>
   this function(){} is not highlighted.
   <code class="http hljs js">
        <span class="hljs-function"><span class="hljs-keyword">function</span> <span class="hljs- 
        title">one</span>(<span class="hljs-params"></span>)</span>{}
   </code>
</pre>
```
It's how it should be right ?

marcoscaceres:
Yes, that looks fine. But make sure you check a second `code`.

src/core/highlight.js
```diff
@@ -45,32 +45,61 @@ export async function run(conf) {
...
-          return; // not for us!
-        }
-        element.innerHTML = value;
-        if (element.localName === "pre") {
-          element.classList.add("hljs");
-        }
-        if (language) {
-          element.classList.add(language);
+      const codeElements = element.querySelectorAll("code");
+      if(!codeElements.length) {
```
marcoscaceres: Note all the code duplication here. To address that have an “elementsToHighlight” array, and to it add either the pre or, if there are some, the code elements.

marcoscaceres: Modified the above

marcoscaceres: so:
```JS
const elementsToHighlight = element.querySelector("code") ? [...element.querySelectorAll("code")] : [element] 
```

saschanaz:
Tip: It's great to run `npm run lint -- --fix` before committing any codes.

src/core/highlight.js
```diff
@@ -45,32 +45,39 @@ export async function run(conf) {
...
+        };
+        worker.addEventListener("message", function listener(ev) {
+          const {
+            data: { id, language, value },
+          } = ev;
+          if (id !== msg.id) {
+            return; // not for us!
+          }
+          let lang = language !== undefined ? language : msg.languages[0];
+          if (each.localName === "pre") {
```
marcoscaceres: for string related tests, we prefer `switch`:
```JS
switch(each.localName) {
   case "pre": 
    // ... pre stuff... 
    break; 
   case "code":
   // ... code stuff... 
   break;
}
```

tests/spec/core/highlight-spec.js
```diff
@@ -76,4 +76,55 @@ describe("Core — Highlight", () => {
...
+        </section>`,
+    };
+    const doc = await makeRSDoc(ops);
+    const pre = doc.querySelectorAll("pre");
+    expect(pre.querySelectorAll("code").length).toBe(1);
+    expect(pre.querySelectorAll("code[class^='js']").length).toBe(1);
+  });
+
+  it("When class defined in <pre> is not correct, returns language: undefined", async () => {
```
marcoscaceres: Make sure the test descriptions read like "it does x when y" or similar (without the "it")...

tests/spec/core/highlight-spec.js
```diff
@@ -76,4 +76,55 @@ describe("Core — Highlight", () => {
...
+          <pre class="js">
+            function foo() {
+              alert('foo');
+            }
+          </pre>
+        </section>`,
+    };
+    const doc = await makeRSDoc(ops);
+    const pre = doc.querySelectorAll("pre");
+    expect(pre.querySelectorAll("code").length).toBe(1);
```
marcoscaceres: Here you really want to be testing if the code element is the `firstElementChild`

src/core/highlight.js
```diff
@@ -45,32 +45,39 @@ export async function run(conf) {
...
-          element.classList.add(language);
-        }
-        clearTimeout(timeoutId);
-        worker.removeEventListener("message", listener);
-        done();
-      });
-      element.setAttribute("aria-busy", "true");
-      worker.postMessage(msg);
+      const elementsToHighlight = element.querySelectorAll("code").length > 0 ? element.querySelectorAll("code") : [element];
+      for(let each of elementsToHighlight) {
```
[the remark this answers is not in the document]
CodHeK: Also here we have a problem we cant just loop since we have a `eventListener` ! The contents of the first `<code>` element gets overwritten into the content of the `<code>` below it due to the delay of response from the worker!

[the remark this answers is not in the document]
marcoscaceres: Right, I think you should do the following... add this function outside this function: 

```JS
function makeListener(id){
  return function listener(ev) {
     // current listener code
  } 
}
```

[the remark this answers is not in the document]
saschanaz: I think the problem is the `<code>`s inside a single `<pre>` share the same `id` for highlight request.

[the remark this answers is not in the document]
marcoscaceres: ah yes, the id should increment per postMessage.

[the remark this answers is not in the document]
CodHeK: Done! Fixed that issue, pushed the code, I'll just add the tests ...

CodHeK:
@marcoscaceres looks go to me now, added the tests too, if you could please review ...

src/core/highlight.js
```diff
@@ -45,32 +45,44 @@ export async function run(conf) {
...
-          element.classList.add(language);
-        }
-        clearTimeout(timeoutId);
-        worker.removeEventListener("message", listener);
-        done();
+      const elementsToHighlight =
+        element.querySelectorAll("code").length > 0
+          ? Array.from(element.querySelectorAll("code"))
+          : [element];
+      elementsToHighlight.map((each, id) => {
```
marcoscaceres: using this id is problematic, because if you have multiple `pre`s with multiple `code`s you are still going get messages to the wrong element. You need to either:

1. combine the outer (`pre`) id with this `code` `id`, 
1. or have a module level message `id` that gets incremented. 
1. or repurpose the `timeoutId` as the message id. Those are also unique.

marcoscaceres: As convention, when we work with elements, we call them "elem" so we know what we are working with.  

```suggestion
      elementsToHighlight.map((elem, id) => {
```

Please change this bellow too.

tests/spec/core/highlight-spec.js
```diff
@@ -76,4 +76,93 @@ describe("Core — Highlight", () => {
...
+    expect(pre.querySelectorAll("code[class^='javascript']").length).toBe(1);
```
marcoscaceres: this is behavior that's part of highlightjs, and not your code, so I think this whole test can be removed.

tests/spec/core/highlight-spec.js
```diff
@@ -76,4 +76,93 @@ describe("Core — Highlight", () => {
...
+            function foo() {
+              alert('foo');
+            }
+          </pre>
+        </section>`,
+    };
+    const doc = await makeRSDoc(ops);
+    const pre = doc.querySelectorAll("pre");
+    expect(pre.querySelectorAll("code").length).toBe(1);
+    expect(pre.querySelectorAll("code[class^='js']").length).toBe(1);
```
marcoscaceres: The expectation here is that it contains "js", but there is no assurance that "js" will be the first item in the class list.

tests/spec/core/highlight-spec.js
```diff
@@ -76,4 +76,93 @@ describe("Core — Highlight", () => {
...
+          <pre class="js">
+            function foo() {
+              alert('foo');
+            }
+          </pre>
+        </section>`,
+    };
+    const doc = await makeRSDoc(ops);
+    const pre = doc.querySelectorAll("pre");
+    expect(pre.firstChild.localName).toBe("code");
```
marcoscaceres: Move this to the test above, then this whole test can go away.

tests/spec/core/highlight-spec.js
```diff
@@ -76,4 +76,93 @@ describe("Core — Highlight", () => {
...
+  });
+
+  it("adds the correct language class to <code> if languge is undefined or not right in <pre>", async () => {
```
marcoscaceres: this test is pretty much exactly the same as the previous test. It can be deleted.

src/core/highlight.js
```diff
@@ -45,32 +45,44 @@ export async function run(conf) {
...
+          } = ev;
+          if (id !== msg.id) {
+            return; // not for us!
+          }
+          const lang = language !== undefined ? language : msg.languages[0];
+          switch (each.localName) {
+            case "pre":
+              each.innerHTML = `<code class="${lang}">${value}</code>`;
+              each.classList.remove(lang);
+              each.firstChild.classList.add("hljs");
```
marcoscaceres: I think this needs to stay on the `pre`, not the firstChild. Note also, `firstChild` will match text nodes, so try to be specific (i.e., `firstElementNode` when you want to target an element).

tests/spec/core/highlight-spec.js
```diff
@@ -76,4 +76,93 @@ describe("Core — Highlight", () => {
...
+    const pre = doc.querySelectorAll("pre");
+    expect(pre.querySelectorAll("code").length).toBe(1);
+    expect(pre.querySelectorAll("code[class^='http']").length).toBe(1);
+  });
+
+  it("checks case when <code> is present inside <pre>", async () => {
+    const ops = {
+      config: makeBasicConfig(),
+      body: `${makeDefaultBody()}<section>
+          <pre>
```
marcoscaceres: To check the message passing is working properly, you want two `pre` blocks and each of those should have multiple `code`s.

marcoscaceres:
@CodHeK, your the test also seem to also be failing. Please take a look.

src/core/highlight.js
CodHeK: Ah, oh yeah! :o

tests/spec/core/highlight-spec.js
CodHeK: But, I am checking if the language detected by highlightjs also gets added into the code element apart from the conventional way of defining it as `<pre class="js">...</pre>` so shouldn't I include this case too ?

marcoscaceres: Ah, ok. Good point. Make sure you use: `[class~='javascript']`.

src/core/highlight.js
CodHeK: instead of using a global `id` , we could pass a pair `(i, id)` won't require much modification .

tests/spec/core/highlight-spec.js
CodHeK: `code[class~='js']` would work for this ...

src/core/highlight.js
marcoscaceres: The combination id is fine, so long as putting the id together is not more computationally expensive than doing i++.

src/core/highlight.js
CodHeK: I guess the firstChild here will only be `<code>` since, I am wrapping the complete `<pre>` into `<code>`.

CodHeK:
> @CodHeK, your the test also seem to also be failing. Please take a look.

It probably because of `class^=` in the tests written above the ones I wrote.

examples/basic.html
marcoscaceres: Don’t forget to remove this from basic.html

src/core/highlight.js
```diff
@@ -45,32 +46,45 @@ export async function run(conf) {
...
+          languages: getLanguageHint(elem.classList),
+        };
+        worker.addEventListener("message", function listener(ev) {
+          const {
+            data: { id, language, value },
+          } = ev;
+          if (id !== msg.id) {
+            return; // not for us!
+          }
+          console.log(ev.data);
```
marcoscaceres: Stray console

examples/basic.html
```diff
@@ -31,6 +31,24 @@
...
```
marcoscaceres: Please address the previous feedback here.

src/core/highlight.js
```diff
@@ -45,32 +46,44 @@ export async function run(conf) {
...
-        done();
+      const elementsToHighlight =
+        element.querySelectorAll("code").length > 0
+          ? Array.from(element.querySelectorAll("code"))
+          : [element];
+      elementsToHighlight.map(elem => {
+        const msg = {
+          action: "highlight",
+          code: elem.innerText,
+          id: `highlight:${id++}`,
```
saschanaz: Increment inside a complex expression is typically a bad thing... Would be better to do this on its own line.

src/core/highlight.js
```diff
@@ -45,32 +46,45 @@ export async function run(conf) {
...
+          clearTimeout(timeoutId);
+          worker.removeEventListener("message", listener);
+          done();
+        });
+        element.setAttribute("aria-busy", "true");
+        worker.postMessage(msg);
       });
-      element.setAttribute("aria-busy", "true");
-      worker.postMessage(msg);
+      id++;
```
marcoscaceres: this is going to be wrong :) imagine you have 2+ `<code>`s, they are going to get the same message id. You should increment the id on line 54 instead (before the `cont msg =`).

CodHeK: Ah! my bad! I had to add on the line above i.e below `worker.postMessage`

marcoscaceres: I think there might be a larger problem with the code tho... the `promisesToHighlight` were intended to only handle one highlight request at time... but `code` blocks now do multiple post messages... but resolve done() after one message 😱. 

I think I need to jump into this code a bit and refactor some of that because we are going to get weird async errors like "aria-busy" will be false before a second `code` update happens.

CodHeK: could we not `element.setAttribute("aria-busy", "false");` set this for `elem` instead and at the end of each loop set it to `true` for the `elem` itself instead of `element` ?

PR #2115, opened 2019-02-21
fix(package): hljs cli now in main branch

### LINKED ISSUE #2112: Difficulty Installing respec dependencies.
* ReSpec version: 24.3.3
* Linux Fedora OS

Difficulty Installing respec dependencies.

Steps to reproduce: `npm install`

I have npm errors while trying to install respec dependencies for the first time so I can run some  tests.  Later I discovered it was due to a dependency *highlight.js*  from _package.json_ `(    "highlight.js": "github:highlightjs/highlight.js#cli" )` which breaks the installation as it is being stuck for a reason I suspect is related to the given error: 

`Command failed: /usr/bin/git checkout cli`
`error: pathspec 'cli' did not match any file(s) known to git.`

Removing this line from _package.json_ and running `npm install` installs all dependencies. But highlight is needed and i do not know how to install this dependency, Installing it independently from npm it has a  file structure which is different from what respec uses for builds/commands so I had to remove it.

Any help would be appreciated as I am so humbled to be able contributes here and I have been stuck trying different things so far.

marcoscaceres:
fixes #2112

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1359,17 +1359,17 @@
       "dev": true
     },
     "acorn-globals": {
-      "version": "3.1.0",
-      "resolved": "https://registry.npmjs.org/acorn-globals/-/acorn-globals-3.1.0.tgz",
-      "integrity": "sha1-/YJw9x+7SZawBPqIDuXUZXOnMb8=",
+      "version": "1.0.9",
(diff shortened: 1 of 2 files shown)


saschanaz:
Can't use the updated npm package?

marcoscaceres:
Soon, the hljs release failed and it’s not something I can fix :(

PR #2116, opened 2019-02-21
fix(core/utils): treat introductory matter as non-normative

### LINKED ISSUE #2114: SOTD section is informative
## Important info

* URL to affected spec: https://w3c.github.io/pointerevents/
  (more precisely, as of
 https://raw.githubusercontent.com/w3c/pointerevents/298f3f4730b7afe18d31b86856672ea71ca0d141/index.html )

* ReSpec version: 24.3.3

## Description of problem

Respec seems to treat the SOTD section as normative for the purpose of external references.

The PointerEvents reference appears as a normative reference with ReSpec. [[PointerEvents]] appear 3 times in the draft:
1. in the SOTD, to mention  that pointerevents2 is an update of pointerevents 1.
2. in a note
3. in an informative section

Changing the occurrence [[PointerEvents]] to [[?PointerEvents]] in the SOTD fixes the issue.

But what I was expecting?:

References in the SOTD are informative references by default instead of normative.

Optional, steps to reproduce:
 1. Use the [pointerevents](https://raw.githubusercontent.com/w3c/pointerevents/298f3f4730b7afe18d31b86856672ea71ca0d141/index.html) draft as of February 11, 2019.
2. Load and see that Pointerevents is a normative reference in the references section
3. Change the occurrence [[PointerEvents]] to [[?PointerEvents]] in the SOTD
4. See that the reference is now informative in the references section

marcoscaceres:
* fixes #2114

### DIFF
--- a/src/core/utils.js
+++ b/src/core/utils.js
@@ -100,7 +100,7 @@ const fetchDestinations = new Set([
 
 // CSS selector for matching elements that are non-normative
 export const nonNormativeSelector =
-  ".informative, .note, .issue, .example, .ednote, .practice";
+  ".informative, .note, .issue, .example, .ednote, .practice, .introductory";
 
 export function calculateLeftPad(text) {
   if (typeof text !== "string") {
--- a/src/core/utils.js
+++ b/src/core/utils.js
@@ -4,7 +4,7 @@
 // As the name implies, this contains a ragtag gang of methods that just don't fit
 // anywhere else.
 import { lang as docLang } from "./l10n";
-import marked from "../deps/marked";
+import marked from "marked";
 import { pub } from "./pubsubhub";
 export const name = "core/utils";
 
--- a/src/core/utils.js
+++ b/src/core/utils.js
@@ -4,7 +4,7 @@
 // As the name implies, this contains a ragtag gang of methods that just don't fit
 // anywhere else.
 import { lang as docLang } from "./l10n";
-import marked from "marked";
+import marked from "deps/marked";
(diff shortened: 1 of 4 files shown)


src/core/utils.js
```diff
@@ -4,7 +4,7 @@
 // As the name implies, this contains a ragtag gang of methods that just don't fit
 // anywhere else.
 import { lang as docLang } from "./l10n";
-import marked from "marked";
+import marked from "deps/marked";
```
marcoscaceres: @saschanaz this was causing issues in the testing environment... but I think it will break in VSCode. Any suggestion for what I should set the path to?

saschanaz: `"marked"` should just work as we set both requirejs and karma mappings. AFAIK this hasn't caused a problem recently...

marcoscaceres: It's giving me grief via http://localhost:5000/tests/SpecRunner.html

marcoscaceres: If "../deps/marked" is not too offensive, it seems to work across everything. Would it be ok to go with that?

marcoscaceres: Actually, I think I know how to fix it.

saschanaz: Anything except `"marked"` will break `@ts-check` (although `core/utils` currently lacks it). Maybe the SpecRunner should get the `paths` field synchronized with `profile-w3c-common.js`?

marcoscaceres:
Happy now 😇

PR #2118, opened 2019-02-23
Part 1 - refactor: use textContent instead of innerText

marcoscaceres:


marcoscaceres:
@saschanaz sending quite a large refactor in parts for you.

src/core/algorithms.js
```diff
@@ -1,5 +1,5 @@
 /**
-Currently used only for adding 'assert' class to algorithm lists  
+Currently used only for adding 'assert' class to algorithm lists
```
saschanaz: Maybe we should add `no-trailing-space` eslint rule.

marcoscaceres: Yeah, would be good.

src/core/utils.js
```diff
@@ -211,7 +211,7 @@ export function normalizePadding(text = "") {
   }
   doc.normalize();
   // use the first space as an indicator of how much to chop off the front
-  const firstSpace = doc.body.innerText
+  const firstSpace = doc.body.textContent
```
saschanaz: I think we should keep this as it is... Or we just replace this bulky function with `reindent()` in `core/reindent`.

marcoscaceres: I'd like to eventually use `reindent()` - that would be great. I need to try it and make sure it handles all the special cases that this one needed to deal with.

marcoscaceres: In any case, I don't think there is any valid reason for this to be using `innerText`

saschanaz: I was concerned about `innerText`s special behavior that converts `<br>` to `\n` and we are using the exact character here. Thinking more, yeah, I think it's okay. This is currently only for markdown and IDL strings that won't (normally) use any `<br>`.

marcoscaceres: Understood. I didn't think of that, but I agree that it should be ok. The markdown is supposed to be processed as given, not as displayed.

PR #2119, opened 2019-02-23
Part 2 - feat(core/biblio): export const biblio

marcoscaceres:


src/core/biblio.js
```diff
@@ -53,7 +54,7 @@ export async function updateFromNetwork(
   const refsToFetch = [...new Set(refs)].filter(ref => ref.trim());
   // Update database if needed, if we are online
   if (!refsToFetch.length || navigator.onLine === false) {
-    return;
+    return null;
```
saschanaz: Do we prefer returning `null`? I have no strong opinion.

marcoscaceres: yeah, the others all return either null or an object.

PR #2121, opened 2019-02-23
Part 3.1 -  refactor: import biblio in modules

marcoscaceres:
(Ignore 3... linting error in that one that I didn't catch)

### DIFF
--- a/src/core/render-biblio.js
+++ b/src/core/render-biblio.js
@@ -107,9 +107,8 @@ export function run(conf) {
 /**
  * returns refcontent and unique key for a reference among its aliases
  * and warns about circular references
- * @param {String} ref
  */
-function getRefContent(conf) {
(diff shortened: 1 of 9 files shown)


src/core/render-biblio.js
```diff
@@ -107,27 +107,37 @@ export function run(conf) {
...
+  return { ref, refcontent };
+}
+
+/**
+ * Render an inline citation
+ *
+ * @param {String} ref the inline reference.
+ * @returns HTMLElement
+ */
+export function renderInlineCitation(ref) {
```
saschanaz: Wait, but what's this?

marcoscaceres: Sorry, should have done this as different hunk. We then use this in the next part to render inline citations.

PR #2122, opened 2019-02-23
Part 4 - fix: spec lacking conformance section are informative

### LINKED ISSUE #2051: If a spec lacks <section id=conformance>, then it reference should be informative
## Important info

* URL to affected spec:
https://w3c.github.io/dxwg/ucr/

## Description of problem
Specs can have either Informative or Normative references. By default, ReSpec treats all documents as if they were Normative. However, some documents, like Use Cases docs, generally are don't use Normative references.  

But what I was expecting?:
The citations in a document that doesn't declare a conformance section (i.e., `<section id="conformance">`) to be treated as Informative. 

To achieve this, all we need to do is add "informative" to the class `attribute` of the `<body>` element  here: 
https://github.com/w3c/respec/blob/develop/src/w3c/conformance.js#L10

cc @andrea-perego

marcoscaceres:
Closes #2051


src/core/inlines.js
```diff
@@ -100,25 +106,21 @@ export function run(conf) {
...
-            ref = ref.replace(/^(!|\?)/, "");
-            df.appendChild(document.createTextNode("["));
-            const refHref = `#bib-${ref.toLowerCase()}`;
-            const cite = hyperHTML`<cite><a class="bibref" href="${refHref}">${ref}</a></cite>`;
-            df.appendChild(cite);
-            df.appendChild(document.createTextNode("]"));
-
-            if (illegal && !conf.normativeReferences.has(ref)) {
+            const cite = renderInlineCitation(ref);
+            const cleanRef = ref.replace(/^(!|\?)/, "");
```
saschanaz: Previously the code removed `!`/`?` first and this now doesn't. Would there be a behavior change?

marcoscaceres: Well spotted. It's now done in by `renderInlineCitation()` - so same behavior.

tests/spec/core/idl-index-spec.js
```diff
@@ -73,10 +74,10 @@ interface Bar {
       body,
     };
     const doc = await makeRSDoc(ops);
-    const idlIndex = doc.getElementById("idl-index");
-    expect(idlIndex.innerText.includes("Foo")).toEqual(false);
-    expect(idlIndex.innerText.includes("Bar")).toEqual(true);
-    expect(idlIndex.innerText.includes("Baz")).toEqual(false);
+    const { textContent } = doc.querySelector("#idl-index pre");
+    expect(textContent).toMatch(/Include/gm);
```
saschanaz: Maybe we can use `toContain()` here?

tests/spec/core/idl-index-spec.js
```diff
@@ -85,44 +86,45 @@ interface Bar {
...
       config: makeBasicConfig(),
       body,
     };
     const doc = await makeRSDoc(ops);
     const idlIndex = doc.getElementById("idl-index");
-    expect(idlIndex.innerText.includes("Foo")).toEqual(false);
-    expect(idlIndex.innerText.includes("Bar")).toEqual(false);
-    expect(idlIndex.innerText.includes("Baz")).toEqual(false);
-    expect(idlIndex.innerText.includes("Qux")).toEqual(true);
+    expect(idlIndex.textContent.includes("Example")).toEqual(false);
```
saschanaz: Also here.

tests/spec/core/idl-index-spec.js
```diff
@@ -131,36 +133,37 @@ interface Bar {
...
     const ops = {
       config: makeBasicConfig(),
       body,
     };
     const doc = await makeRSDoc(ops);
     const idlIndex = doc.getElementById("idl-index");
-    expect(idlIndex.innerText.includes("Foo")).toEqual(true);
-    expect(idlIndex.innerText.includes("Bar")).toEqual(false);
-    expect(idlIndex.innerText.includes("Baz")).toEqual(false);
+    expect(idlIndex.textContent.includes("Pass")).toBe(true);
```
saschanaz: Here too

tests/spec/core/idl-index-spec.js
```diff
@@ -264,6 +267,7 @@ dictionary PromptResponseObject {
         <h2>PASS</h2>
         <p>Custom paragraph.</p>
       </section>
+      <section id="conformance"></section>
```
saschanaz: Multiple elements with the same ID are a weird situation, what's this for?

marcoscaceres:
Codacy and ESLint disagree 😂

marcoscaceres:
Removed the Codacy webhook. It's was nice, but we don't need it anymore.

tests/spec/core/inlines-spec.js
```diff
@@ -19,23 +20,26 @@ describe("Core - Inlines", () => {
     const doc = await makeRSDoc(ops);
 
     const norm = [...doc.querySelectorAll("#normative-references dt")];
-    expect(norm.length).toEqual(3);
+    expect(norm.length).toEqual(4);
```
saschanaz: Also `toBe`?

tests/spec/core/idl-index-spec.js
marcoscaceres: this is ok I think... It's part of a different test, but above it looks like the same test.

marcoscaceres: (note the above is on line 203, and this is line 268)

saschanaz: Oops, my mistake, sorry.

marcoscaceres:
Thanks for the speedy review @saschanaz!

PR #2125, opened 2019-02-26
chore(package): revert to cli branch

saschanaz:
As the feature has been reverted in highlight.js.

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,7 +88,7 @@
     "command-line-usage": "^5.0.5",
     "epipebomb": "^1.0.0",
     "fs-extra": "^7.0.1",
-    "highlight.js": "github:highlightjs/highlight.js",
+    "highlight.js": "github:highlightjs/highlight.js#cli",
     "hyperhtml": "^2.25.4",
     "loading-indicator": "^2.0.0",
     "marked": "^0.6.1",


marcoscaceres:
We might need to change the way the highlighting code is imported... our build scripts are depending on this feature, so we might need to fix a few other things.

marcoscaceres:
See https://github.com/w3c/respec/blob/develop/tools/build-components.js#L11

saschanaz:
What was the reason to revert it BTW? The error message people reported is that `tools/build.js` was lacking but AFAIK it hasn't been removed.

marcoscaceres:
yes, build.js hasn't been removed (reason: [I shouldn't be allowed near computers](https://github.com/highlightjs/highlight.js/issues/1984#issuecomment-467257131))

saschanaz:
>We might need to change the way the highlighting code is imported... our build scripts are depending on this feature, so we might need to fix a few other things.

Can we merge this first and do it later? Currently we can't even do `npm install` 🤣

marcoscaceres:
What I'd like to do is treat hljs like any other library and not do the bundling. So (pseudo code):
```
import hljs from "highlight.js"
import js from "highlight.js/languages/js"
import xml from "highlight.js/languages/js"
... and so on... 
hljs.registerLanguage(js);
hljs.registerLanguage(xml);
... and so on... 
```

marcoscaceres:
ok, I'm off to bed. Let's pick this up again tomorrow.

saschanaz:
Good night!

PR #2126, opened 2019-02-26
Fix obscuring ReSpec button UI

### LINKED ISSUE #948: ReSpec button UI obscures the W3C logo
Great work on the UI enhancements in ReSpec 6.0!

One minor thing I observed: using certain screen width the ReSpec button is on top of the W3C logo, obscuring it (well, one might argue this is a feature, not a bug ;-)).

w3c-logo-respec

yatri1609:
Fixes #948 

I have tried to implement the Intersection Observer in `ui.js` file which should take care of the button being transparent.


src/core/ui.js
```diff
@@ -114,6 +114,58 @@ function createWarnButton(butName, arr, title) {
     ["label", `Document ${title.toLowerCase()}`],
   ]);
   ariaDecorate(button, ariaMap);
+
+
+  var numSteps = 20.0;
+  var boxElement
+  var prevRatio = 0.0;
+  var increasingColor = "rgba(255, 255, 255)";
+  var decreasingColor = "rgba(255, 255, 255, ratio)";
```
sidvishnoi: I like this idea of changing the alpha. CSS variables might be worth looking into here :D

yatri1609: Ohk! I'll have a look at it...

rishabh3112: Using inline arrows seems better IMHO.
```suggestion
  var decreasingColor = (ratio) => `rgba(255, 255, 255, ${ratio})`;
```

src/core/ui.js
```diff
@@ -114,6 +114,58 @@ function createWarnButton(butName, arr, title) {
...
+    }
+
+    thresholds.push(0);
+    return thresholds;
+  }
+
+  function handleIntersect(entries, observer) {
+    entries.forEach(function (entry) {
+      if (entry.intersectionRatio > prevRatio) {
+        entry.target.style.backgroundColor = increasingColor.replace("ratio", entry.intersectionRatio);
```
rishabh3112: `increasingColor` doesn't have `ratio` as a substring to replace. Maybe left there.

src/core/ui.js
```diff
@@ -114,6 +114,58 @@ function createWarnButton(butName, arr, title) {
...
+      } else {
+        entry.target.style.backgroundColor = decreasingColor.replace("ratio", entry.intersectionRatio);
```
rishabh3112:
```suggestion
        entry.target.style.backgroundColor = decreasingColor(entry.intersectionRatio);
```

src/core/ui.js
```diff
@@ -114,6 +114,58 @@ function createWarnButton(butName, arr, title) {
     ["label", `Document ${title.toLowerCase()}`],
   ]);
   ariaDecorate(button, ariaMap);
+
+
+  var numSteps = 20.0;
```
rishabh3112: Using `const` instead of `var` would be better for all the variables here, except `boxElement` there too `let` could be used.
Also this `numSteps` is redundant.

src/core/ui.js
marcoscaceres: Probably can trigger a simple CSS transition to fade to 50% when intesecting. Hovering or otherwise activating the button should make it fade back in, even if over the logo.

yatri1609: @marcoscaceres Yeah that could be done I'll implement it

yatri1609:
> Any progress on this @yatri1609? As noted in the feedback, we should probably be able to do the the actual animation changes using just CSS.

I have worked on this and will be sending a PR by this weekend. It was delayed as I have my exams going on currently. Sorry for this delay @marcoscaceres I'll send PR asap.

marcoscaceres:
> Sorry for this delay @marcoscaceres I'll send PR asap.

No problem! Appreciate the update and best of luck with the exams! 🏆

src/core/ui.js
```diff
@@ -47,7 +47,7 @@ sub("end-all", () => document.body.prepend(respecUI), { once: true });
 
 const respecPill = hyperHTML`<button id='respec-pill' disabled>ReSpec</button>`;
 respecUI.appendChild(respecPill);
-respecPill.addEventListener("click", function(e) {
+respecPill.addEventListener("click", function (e) {
```
```suggestion
respecPill.addEventListener("click", function(e) {
```

src/core/ui.js
```diff
@@ -87,8 +87,55 @@ function errWarn(msg, arr, butName, title) {
 
 function createWarnButton(butName, arr, title) {
   const buttonId = `respec-pill-${butName}`;
+  var numSteps = 20.0;
+  var boxElement
```
marcoscaceres: So, I think all this setup should be referring to `respecUI` as a whole:
https://github.com/w3c/respec/blob/develop/src/core/ui.js#L37

The "warn button" are only yellow buttons that show warnings.

src/core/ui.js
```diff
@@ -87,8 +87,55 @@ function errWarn(msg, arr, butName, title) {
...
+  var prevRatio = 0.0;
+
+  button.addEventListener("load", function (event) {
```
marcoscaceres: As above. Probably don't need this, you have direct access to the `respecUI` div.

src/core/ui.js
```diff
@@ -87,8 +87,55 @@ function errWarn(msg, arr, butName, title) {
...
+    boxElement = document.querySelector("#respec-pill");
+
+    createObserver();
+  }, false);
+
+  function createObserver() {
```
marcoscaceres: you can probably inline this function as it doesn't return anything.

src/core/ui.js
```diff
@@ -87,8 +87,55 @@ function errWarn(msg, arr, butName, title) {
...
+      root: null,
+      rootMargin: "0px",
+      threshold: buildThresholdList()
+    };
+
+    observer = new IntersectionObserver(handleIntersect, options);
+    observer.observe(boxElement);
+  }
+
+  function buildThresholdList() {
```
marcoscaceres: Can you help me understand what this does? In my mind, all we are concerned about is detecting when `respecUI` intersects a `.logo`.

src/core/ui.js
```diff
@@ -87,8 +87,55 @@ function errWarn(msg, arr, butName, title) {
...
+    }
+
+    thresholds.push(0);
+    return thresholds;
+  }
+
+  function handleIntersect(entries, observer) {
+    entries.forEach(function (entry) {
+      if (entry.intersectionRatio > prevRatio) {
+        const button = hyperHTML`<button id='${buttonId}' class='respec-info-button'>`;
```
marcoscaceres: I'm not sure we need to create a new button? Shouldn't a you just set a css class on `respecUI` causing it to fade out?

src/core/ui.js
marcoscaceres: As above... I wouldn't expect any CSS stuff here at all. All CSS transitions or animations should be handled in the CSS of file.

src/core/ui.js
yatri1609: Yeah correct but this help us make a threshold (i.e. he percentage of intersection after which the logo must fade) after which the logo will fade out so unless and until we do not assign a threshold the intersection observer will not do anything

marcoscaceres: ok, but couldn't it just be 1 or 0? (sorry if that's a silly question, I honestly have no idea how it works)

yatri1609: Yeah that might be done.. I need to check it out once doing it locally.

src/core/ui.js
```diff
@@ -34,6 +34,21 @@ function ariaDecorate(elem, ariaMap) {
...
+  entries => {
+    entries.forEach(entry => {
+      if (entry.intersectionRatio > 0) {
+        entry.target.setAttribute("visibility", "hidden");
+      } else {
+        entry.target.setAttribute("visibility", "visible");
+      }
+    });
+  },
+  { root: document.querySelector("#logo") }
```
marcoscaceres: Note that specs can have multiple logos.  Logos have a “.logo” class.

yatri1609: Ohh yes ! That just didn't crossed my mind.

marcoscaceres: I wonder if the root shouldn’t be respecUI? The you observe() each logo?

yatri1609: Yeah the root needs to be the logo with which we need to observe out button as in this case it needs to be the w3c logo... But since I thought it will be the only logo I assigned root to be the logo.. that's my fault here

yatri1609: So here it will be the respec-ui as we will be observing with respect to the whole ui.

src/core/ui.js
```diff
@@ -34,6 +34,21 @@ function ariaDecorate(elem, ariaMap) {
   }, elem);
 }
 
+const observer = new IntersectionObserver(
+  entries => {
+    entries.forEach(entry => {
+      if (entry.intersectionRatio > 0) {
```
sidvishnoi: Can we use [`IntersectionObserverEntry.isIntersecting`](https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserverEntry/isIntersecting)?

marcoscaceres:
My reading is that the IntersectionObserver API cannot detect intersection between two elements. It can detect elements coming into the viewport or it can detect an element being scrolled and a child element coming into view. 

Thus, I think we should close this bug because there doesn't seem to be a sane way to do this in browsers without doing some heavy computation?

yatri1609:
As per what I read it said that we could detect Intersection between root and target. Here the root being respec-ui and target being the respec button

yatri1609:
@marcoscaceres As per what I read it said that we could detect Intersection between root and target. Here the root being respec-ui and target being the respec button

yatri1609:
@marcoscaceres The W3C is the image and it doesn't have any specific id through which it could be used as a root for IntersectionObserver

marcoscaceres:
@yatri1609, you can try getting it to work here: https://jsfiddle.net/vrwh2oyx/1/ - however, I think it has to be in a child parent relationship. Because the `respecUI` and the `.logos` are not in a child/parent relationship, I don't think it's possible. 

The only way I can see this working would be to intersect .logos on the document, and then check the bounding rect of  `respecUI` is overlapping with the logos. However, would be computationally expensive.

yatri1609:
> @yatri1609, you can try getting it to work here: https://jsfiddle.net/vrwh2oyx/1/ - however, I think it has to be in a child parent relationship. Because the `respecUI` and the `.logos` are not in a child/parent relationship, I don't think it's possible.
> 
> The only way I can see this working would be to intersect .logos on the document, and then check the bounding rect of `respecUI` is overlapping with the logos. However, would be computationally expensive.

Yeah it might be computationally expensive.

src/core/ui.js
marcoscaceres: No, as it reports if it’s intercepting inside the viewport.

marcoscaceres:
Closing as wont fix. Thanks for your efforts @yatri1609.

PR #2141, opened 2019-03-06
fix(w3c/style): w3c style must be last when exporting

### LINKED ISSUE #2140: respec-highlight style sheet should appear before the W3C TR stylesheet
## Important info

* URL to affected spec: https://w3c.github.io/performance-timeline/
* ReSpec version: 24.4.0

## Description of problem

The W3C TR style sheet is no longer the last one to appear in html/head. The respec-highlight inline style sheet "github.com style" appears (race condition?) after the TR style sheet.

[[
  👎 Error code “last”.
      ⮡ cf https://www.w3.org/pubrules/doc/rules/?profile=WD#lastStylesheet
]]
https://lists.w3.org/Archives/Public/public-tr-notifications/2019Mar/0014.html

But what I was expecting?:

The W3C TR style sheet appears as the last style in html/head.

Optional, steps to reproduce:
 1. Open https://w3c.github.io/performance-timeline/
2. Go in the DOM inspector and note that there is a style element after the <link rel=stylesheet href='...W3C-ED'

Or use the spec-generator
  https://labs.w3.org/spec-generator/?type=respec&url=https://w3c.github.io/performance-timeline/?specStatus=WD&shortName=performance-timeline

marcoscaceres:
fixes #2140 

@saschanaz I had trouble testing this and I'm bit short on time today 😭 - but need to get this out as it's blocking publications 🔥. I've manually confirmed that it works tho. 

screenshot 2019-03-06 14 44 54

The problem I encountered is that the `pubsubhub` exports seem to be different in the test suite. That is, everything works fine, but when ReSpec tries to `pub("beforesave")` all the subscriptions have vanished. It's very strange. 

CC @sidvishnoi ... maybe encountered the same issue when he wrote this module?

### DIFF
--- a/src/w3c/style.js
+++ b/src/w3c/style.js
@@ -103,6 +103,15 @@ if (!document.head.querySelector("meta[name=viewport]")) {
 
 document.head.prepend(elements);
 
+function styleMover(linkURL) {
+  return exportDoc => {
+    const w3cStyle = exportDoc.querySelector(`head link[href="${linkURL}"]`);
+    exportDoc
+      .querySelector("head")
+      .insertAdjacentElement("beforeend", w3cStyle);
+  };
+}
+
 export function run(conf) {
   if (!conf.specStatus) {
     const warn = "`respecConfig.specStatus` missing. Defaulting to 'base'.";
@@ -156,6 +165,8 @@ export function run(conf) {
   }
   const finalVersionPath = version ? `${version}/` : "";
   const finalStyleURL = `https://www.w3.org/StyleSheets/TR/${finalVersionPath}${styleFile}`;
-
   linkCSS(document, finalStyleURL);
+  // Make sure the W3C stylesheet is the last stylesheet, as required by W3C Pub Rules.
+  const moveStyle = styleMover(finalStyleURL);
+  sub("beforesave", moveStyle);
 }


marcoscaceres:
For references, this would be the test
```JS
  it("moves the W3C style sheet to be last thing in documents head", async () => {
    ops.body = `
      <!-- add WebIDL style -->
      <pre class="idl">
        interface Foo {};
      </pre>
      <!-- add examples and hljs styles -->
      <pre class="example js">
        function Foo(){};
      </pre>`;
    const doc = await getExportedDoc(ops);
    const head = doc.querySelector("head");
    expect(head.lastElementChild.href).toBe(
      "https://www.w3.org/StyleSheets/TR/2016/W3C-ED"
    );
  });
```

saschanaz:
Moving is okay, but how about using `insertBefore(ghCss, doc.head.lastElementChild)` in the highlighter module?

marcoscaceres:
> Moving is okay, but how about using insertBefore(ghCss, doc.head.lastElementChild) in the highlighter module?

marcoscaceres:
> Moving is okay, but how about using insertBefore(ghCss, doc.head.lastElementChild) in the highlighter module?

I'm worried about having to adjust things on a per module basis... I'd rather do the re-organization on cleanup, tbh. 

The other thing is that the ordering requirement from the w3c is a bug. We've been pushing them to get rid of that requirement.

saschanaz:
That's reasonable.

>The problem I encountered is that the pubsubhub exports seem to be different in the test suite. That is, everything works fine, but when ReSpec tries to pub("beforesave") all the subscriptions have vanished. It's very strange.

I'm yet to debug this but how did you confirm that?

marcoscaceres:
I'll send a PR with the test in a new branch and a debug statement.

PR #2142, opened 2019-03-06
Badness

marcoscaceres:
Closes #2141

### DIFF
--- a/src/core/exporter.js
+++ b/src/core/exporter.js
@@ -86,6 +86,8 @@ function cleanup(cloneDoc) {
 
   insertions.appendChild(metaGenerator);
   head.prepend(insertions);
+  console.log("BEFORE SAVE");
+  debugger;
   pub("beforesave", documentElement);
 }
 
--- a/src/core/exporter.js
+++ b/src/core/exporter.js
@@ -87,6 +87,7 @@ function cleanup(cloneDoc) {
   insertions.appendChild(metaGenerator);
   head.prepend(insertions);
   console.log("BEFORE SAVE");
+  // eslint-disable-next-line no-debugger
   debugger;
   pub("beforesave", documentElement);
 }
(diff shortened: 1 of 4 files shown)


src/core/exporter.js
```diff
@@ -86,6 +86,8 @@ function cleanup(cloneDoc) {
 
   insertions.appendChild(metaGenerator);
   head.prepend(insertions);
+  console.log("BEFORE SAVE");
+  debugger;
```
marcoscaceres: Stepping here will jump into pubsubhub... and checking `subscriptions` shows that that all the subscriptions have vanished :(

marcoscaceres: I think for whatever reason, this `pubsubhub` export must be different from the document's one... but I honestly don't know.

saschanaz: image

I can see them alive? 🤔

marcoscaceres: Seems to happen in the test harness. If you run Karma, the test fails (for me at least).

marcoscaceres:
No way???? What was the bug???

saschanaz:
I replaced `require()` with `doc.defaultView.require()`. I thought "What would it even mean when the function is from the parent document and the actual document element is from an iframe?", but I'm still trying to understand why it fixes the problem 🤔

marcoscaceres:
Different realms?

marcoscaceres:
Also, if you agree, I’d like to get this out as it’s going to affect the ability for people to automatically publish at the w3c.

saschanaz:
I think I got it.

1. The modules from the parent document are the raw modules that haven't `run()`, so naturally its `pubhubsub` will get no subscriptions.
2. The previous code was `pub()`ing to the parent document.
3. So nothing happens 😁

marcoscaceres:
Ok, that makes sense. Really appreciate you digging into this!

PR #2148, opened 2019-03-09
fix(w3c/conformance): always emit an end event

### LINKED ISSUE #2147: w3c/conformance end event is not being published in ARIA spec
## Important info

* URL to affected spec: https://w3c.github.io/aria/
* ReSpec version: Click the ReSpec pill (e.g., 20.0.1): 24.4.4
* [] I did a "hard refresh", but it's still busted. 
(not sure what this means)

## Description of problem

ARIA specific scripts are not running. 
For example https://github.com/w3c/aria/blob/master/common/script/aria.js - the events end and end-all do not seem to be called.

But what I was expecting?:
The end and end-all events to fire and the script to run

saschanaz:
Also fixed a problem where `makeRSDoc()` doesn't run inline `<script>` because of timing problem. (ReSpec typically waits until all DOM elements including inline `<script>` load, but currently `makeRSDoc` inserts ReSpec and scripts after the document load.)

Fixes #2147

### DIFF
--- a/tests/spec/core/examples-spec.js
+++ b/tests/spec/core/examples-spec.js
@@ -145,7 +145,7 @@ describe("Core — Examples", () => {
     const ops = makeStandardOps({}, body);
     const doc = await makeRSDoc(ops);
     const mybutton = doc.getElementById("mybutton");
-    expect(mybutton.onclick).toBeDefined();
+    expect(mybutton.onclick).toBeTruthy();
   });
 });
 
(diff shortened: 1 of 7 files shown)


tests/spec/core/examples-spec.js
```diff
@@ -145,7 +145,7 @@ describe("Core — Examples", () => {
     const ops = makeStandardOps({}, body);
     const doc = await makeRSDoc(ops);
     const mybutton = doc.getElementById("mybutton");
-    expect(mybutton.onclick).toBeDefined();
```
saschanaz: This didn't really work as `toBeDefined()` is effectively `not.toBeUndefined()` 😅

PR #2149, opened 2019-03-12
feat(tools/builder): add debug mode

saschanaz:
`npm run build:respec-w3c-common -- --debug` emits an unoptimized result.


marcoscaceres:
We need to figure out a workflow that uses this. We should also consider just shipping the unminified debug version. Running `npm test` should do this for us.

marcoscaceres:
We should also migrate SpecHelper.js to use this. Then RequireJS won’t be needed anymore for the tests.

PR #2151, opened 2019-03-13
refactor(core/inlines.js): split target code into sub functions

himanish-star:
@marcoscaceres as discussed, I have implemented the first stage of code refactor for `core/inlines.js`. Will be looking forward to a review from your side :)

### DIFF
--- a/src/core/inlines.js
+++ b/src/core/inlines.js
@@ -151,7 +151,6 @@ export function run(conf) {
             matched
           )
         ) {
-          // RFC 2119
           inlineRFC2119Matches(matched, df);
         } else if (matched.startsWith("{{{")) {
           // External IDL references (xref)
--- a/src/core/inlines.js
+++ b/src/core/inlines.js
@@ -36,7 +36,7 @@ function inlineRFC2119Matches(matched) {
 /**
  * @param {string} matched
  */
-function inlineXrefMatches(matched, df) {
+function inlineXrefMatches(matched) {
   // slices "{{{" at the beginning and "}}}" at the end
   const ref = matched
     .slice(3, -3)
@@ -46,11 +46,10 @@ function inlineXrefMatches(matched, df) {
 
 /**
  * @param {string} matched
- * @param {DocumentFragment} df
  * @param {Text} txt
  * @param {Object} conf
  */
-function inlineBibrefMatches(matched, df, txt, conf) {
+function inlineBibrefMatches(matched, txt, conf) {
   // slices "[[" at the start and "]]" at the end
   const ref = matched.slice(2, -2);
   if (ref.startsWith("\\")) {
@@ -78,11 +77,10 @@ function inlineBibrefMatches(matched, df, txt, conf) {
 
 /**
  * @param {string} matched
- * @param {DocumentFragment} df
  * @param {Text} txt
  * @param {Map<string, string>} abbrMap
  */
-function inlineAbbrMatches(matched, df, txt, abbrMap) {
+function inlineAbbrMatches(matched, txt, abbrMap) {
   return /** @type {HTMLElement} */ txt.parentElement.tagName === "ABBR" ?
     document.createTextNode(matched) : (hyperHTML`<abbr title="${abbrMap.get(matched)}">${matched}</abbr>`);
 }
@@ -140,16 +138,16 @@ export function run(conf) {
             matched
           )
         ) {
-          const node = inlineRFC2119Matches(matched, df);
+          const node = inlineRFC2119Matches(matched);
           df.appendChild(node);
         } else if (matched.startsWith("{{{")) {
-          const node = inlineXrefMatches(matched, df);
+          const node = inlineXrefMatches(matched);
           df.appendChild(node);
         } else if (matched.startsWith("[[")) {
-          const nodes = inlineBibrefMatches(matched, df, txt, conf);
+          const nodes = inlineBibrefMatches(matched, txt, conf);
           df.append(...nodes);
         } else if (abbrMap.has(matched)) {
-          const node = inlineAbbrMatches(matched, df, txt, abbrMap);
+          const node = inlineAbbrMatches(matched, txt, abbrMap);
           df.appendChild(node);
         } else {
           // FAIL -- not sure that this can really happen
--- a/src/core/inlines.js
+++ b/src/core/inlines.js
@@ -153,7 +153,6 @@ export function run(conf) {
         ) {
           inlineRFC2119Matches(matched, df);
         } else if (matched.startsWith("{{{")) {
-          // External IDL references (xref)
           inlineXrefMatches(matched, df);
         } else if (matched.startsWith("[[")) {
           inlineBibrefMatches(matched, df, txt, conf);
--- a/src/core/inlines.js
+++ b/src/core/inlines.js
@@ -53,7 +53,7 @@ function inlineBibrefMatches(matched, txt, conf) {
   // slices "[[" at the start and "]]" at the end
   const ref = matched.slice(2, -2);
   if (ref.startsWith("\\")) {
-    return /** type {HTMLElement[]} */ [
+    return /** @type {HTMLElement[]} */ [
       document.createTextNode(`[[${ref.slice(1)}]]`),
     ];
   } else {
@@ -73,7 +73,7 @@ function inlineBibrefMatches(matched, txt, conf) {
     } else {
       conf.normativeReferences.add(cleanRef);
     }
-    return /** type {HTMLElement[]} */ cite.childNodes;
+    return /** @type {HTMLElement[]} */ cite.childNodes;
   }
 }
 
--- a/src/core/inlines.js
+++ b/src/core/inlines.js
@@ -22,6 +22,86 @@ import { renderInlineCitation } from "./render-biblio";
 export const name = "core/inlines";
 export const rfc2119Usage = {};
 
+/**
+ * @param {string} matched
+ * @param {DocumentFragment} df
+ */
+function inlineRFC2119Matches(matched, df) {
+  matched = matched.split(/\s+/).join(" ");
+  df.appendChild(
+    hyperHTML`<em class="rfc2119" title="${matched}">${matched}</em>`
+  );
+  // remember which ones were used
+  rfc2119Usage[matched] = true;
+}
+
+/**
+ * @param {string} matched
+ * @param {DocumentFragment} df
+ */
+function inlineXrefMatches(matched, df) {
+  const ref = matched
+    .replace(/^\{{3}/, "")
+    .replace(/\}{3}$/, "")
+    .trim();
+  if (ref.startsWith("\\")) {
+    df.appendChild(
+      document.createTextNode(`{{{${ref.replace(/^\\/, "")}}}}`)
+    );
+  } else {
+    df.appendChild(idlStringToHtml(ref));
+  }
+}
+
+/**
+ * @param {string} matched
+ * @param {DocumentFragment} df
+ * @param {Text} txt
+ * @param {Object} conf
+ */
+function inlineBibrefMatches(matched, df, txt, conf) {
+  let ref = matched;
+  ref = ref.replace(/^\[\[/, "");
+  ref = ref.replace(/\]\]$/, "");
+  if (ref.startsWith("\\")) {
+    df.appendChild(
+      document.createTextNode(`[[${ref.replace(/^\\/, "")}]]`)
+    );
+  } else {
+    const { type, illegal } = refTypeFromContext(ref, txt.parentNode);
+    const cite = renderInlineCitation(ref);
+    const cleanRef = ref.replace(/^(!|\?)/, "");
+    df.append(...cite.childNodes);
+    if (illegal && !conf.normativeReferences.has(cleanRef)) {
+      showInlineWarning(
+        cite.childNodes[1], // cite element
+        "Normative references in informative sections are not allowed. " +
+          `Remove '!' from the start of the reference \`[[!${ref}]]\``
+      );
+    }
+
+    if (type === "informative" && !illegal) {
+      conf.informativeReferences.add(cleanRef);
+    } else {
+      conf.normativeReferences.add(cleanRef);
+    }
+  }
+}
+
+/**
+ * @param {string} matched
+ * @param {DocumentFragment} df
+ * @param {Text} txt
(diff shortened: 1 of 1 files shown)


himanish-star:
Oops! Missed out the tests :tongue:

src/core/inlines.js
```diff
@@ -22,6 +22,82 @@ import { renderInlineCitation } from "./render-biblio";
...
+}
+
+/**
+ * @param {string} matched
+ * @param {DocumentFragment} df
+ * @param {Text} txt
+ * @param {Map<string, string>} abbrMap
+ */
+function inlineAbbrMatches(matched, df, txt, abbrMap) {
+  if (txt.parentElement.tagName === "ABBR")
```
marcoscaceres: nit: can probably use a ternary here.

src/core/inlines.js
```diff
@@ -70,66 +146,22 @@ export function run(conf) {
...
-
-            if (type === "informative" && !illegal) {
-              conf.informativeReferences.add(cleanRef);
-            } else {
-              conf.normativeReferences.add(cleanRef);
-            }
-          }
+          inlineBibrefMatches(matched, df, txt, conf);
         } else if (abbrMap.has(matched)) {
           // ABBR
```
marcoscaceres: Don't need this anymore
```suggestion
```

src/core/inlines.js
```diff
@@ -70,66 +146,22 @@ export function run(conf) {
...
-          if (ref.startsWith("\\")) {
-            df.appendChild(
-              document.createTextNode(`{{{${ref.replace(/^\\/, "")}}}}`)
-            );
-          } else {
-            df.appendChild(idlStringToHtml(ref));
-          }
+          inlineXrefMatches(matched, df);
         } else if (matched.startsWith("[[")) {
           // BIBREF
```
marcoscaceres:
```suggestion
```

src/core/inlines.js
```diff
@@ -70,66 +146,22 @@ export function run(conf) {
...
-          matched = matched.split(/\s+/).join(" ");
-          df.appendChild(
-            hyperHTML`<em class="rfc2119" title="${matched}">${matched}</em>`
-          );
-          // remember which ones were used
-          rfc2119Usage[matched] = true;
+          // RFC 2119
+          inlineRFC2119Matches(matched, df);
         } else if (matched.startsWith("{{{")) {
           // External IDL references (xref)
```
marcoscaceres:
```suggestion
```

src/core/inlines.js
```diff
@@ -70,66 +146,22 @@ export function run(conf) {
...
             matched
           )
         ) {
-          matched = matched.split(/\s+/).join(" ");
-          df.appendChild(
-            hyperHTML`<em class="rfc2119" title="${matched}">${matched}</em>`
-          );
-          // remember which ones were used
-          rfc2119Usage[matched] = true;
+          // RFC 2119
```
marcoscaceres: Function name now makes this clear 👍 
```suggestion
```

src/core/inlines.js
```diff
@@ -22,6 +22,82 @@ import { renderInlineCitation } from "./render-biblio";
...
+}
+
+/**
+ * @param {string} matched
+ * @param {DocumentFragment} df
+ */
+function inlineXrefMatches(matched, df) {
+  const ref = matched
+    .replace(/^\{{3}/, "")
+    .replace(/\}{3}$/, "")
```
saschanaz: Maybe we can just `.slice(3, -3)` here because the assumption here is that the matched string is formed as `{{{ ... }}}`.

src/core/inlines.js
```diff
@@ -22,6 +22,82 @@ import { renderInlineCitation } from "./render-biblio";
...
+    .trim();
+  if (ref.startsWith("\\")) {
+    df.appendChild(document.createTextNode(`{{{${ref.replace(/^\\/, "")}}}}`));
```
saschanaz:
```suggestion
    df.appendChild(document.createTextNode(`{{{${ref.slice(2)}}}}`));
```

src/core/inlines.js
```diff
@@ -22,6 +22,82 @@ import { renderInlineCitation } from "./render-biblio";
...
+ * @param {DocumentFragment} df
+ * @param {Text} txt
+ * @param {Object} conf
+ */
+function inlineBibrefMatches(matched, df, txt, conf) {
+  let ref = matched;
+  ref = ref.replace(/^\[\[/, "");
+  ref = ref.replace(/\]\]$/, "");
+  if (ref.startsWith("\\")) {
+    df.appendChild(document.createTextNode(`[[${ref.replace(/^\\/, "")}]]`));
```
saschanaz:
```suggestion
    df.appendChild(document.createTextNode(`[[${ref.slice(1)}]]`));
```

src/core/inlines.js
```diff
@@ -22,6 +22,82 @@ import { renderInlineCitation } from "./render-biblio";
...
+  } else {
```
saschanaz: I suggest we do an early return:

```js
if (ref.startsWith("\\")) {
  // ...
  return;
}
// ...
```

src/core/inlines.js
marcoscaceres: ah, cute solution @saschanaz!

src/core/inlines.js
```diff
@@ -22,6 +22,82 @@ import { renderInlineCitation } from "./render-biblio";
...
+/**
+ * @param {string} matched
+ * @param {DocumentFragment} df
+ * @param {Text} txt
+ * @param {Object} conf
+ */
+function inlineBibrefMatches(matched, df, txt, conf) {
+  let ref = matched;
+  ref = ref.replace(/^\[\[/, "");
+  ref = ref.replace(/\]\]$/, "");
```
saschanaz:
```js
const ref = matched.slice(2, -2);
```

marcoscaceres: @himanish-star, if you adopt these suggestions, please be sure you document what `.slice()` is chopping off. The current code is really obvious/clear - but `.slice(2, -2)` won't be.

marcoscaceres: i.e., "// chops off "[[" at start, and "]]" at end"

saschanaz: Currently it's somewhat unclear what's the assumptions here (because the monolithic regex) so a proper comment will be great 👍🚀

himanish-star: fixing this and all the above suggestions in a while :)

himanish-star: > proper comment will indeed be great :+1: :rocket: 

will do so

src/core/inlines.js
```diff
@@ -70,66 +146,18 @@ export function run(conf) {
       if (subtxt.length) matched = subtxt.shift();
       df.appendChild(document.createTextNode(t));
       if (matched) {
```
marcoscaceres: Can probably even return early here... saves one level of code nesting.

src/core/inlines.js
himanish-star: done

src/core/inlines.js
himanish-star: missed out on this review. I'll do it in the next set of changes :)

src/core/inlines.js
himanish-star: done

src/core/inlines.js
himanish-star: done

src/core/inlines.js
himanish-star: done

src/core/inlines.js
```diff
@@ -22,6 +22,72 @@ import { renderInlineCitation } from "./render-biblio";
...
+ * @param {Object} conf
+ */
+function inlineBibrefMatches(matched, txt, conf) {
+  // slices "[[" at the start and "]]" at the end
+  const ref = matched.slice(2, -2);
+  if (ref.startsWith("\\")) {
+    return /** @type {HTMLElement[]} */ [
+      document.createTextNode(`[[${ref.slice(1)}]]`),
+    ];
+  } else {
```
marcoscaceres: Tip: return early means you don't need else. 
```suggestion
  }
```

src/core/inlines.js
```diff
@@ -22,6 +22,72 @@ import { renderInlineCitation } from "./render-biblio";
...
+      );
+    }
+
+    if (type === "informative" && !illegal) {
+      conf.informativeReferences.add(cleanRef);
+    } else {
+      conf.normativeReferences.add(cleanRef);
+    }
+    return /** @type {HTMLElement[]} */ cite.childNodes;
+  }
```
marcoscaceres:
```suggestion
  
```

src/core/inlines.js
```diff
@@ -22,6 +22,72 @@ import { renderInlineCitation } from "./render-biblio";
...
 
+/**
+ * @param {string} matched
+ */
+function inlineRFC2119Matches(matched) {
+  const normalize = matched.split(/\s+/).join(" ");
+  const nodeElement = hyperHTML`<em class="rfc2119" title="${normalize}">${normalize}</em>`;
+  // remember which ones were used
+  rfc2119Usage[normalize] = true;
+  return /** @type {HTMLElement} */ nodeElement;
```
saschanaz: Prefer `@return {type}` over `return /** @type {type} */`.

src/core/inlines.js
```diff
@@ -22,6 +22,72 @@ import { renderInlineCitation } from "./render-biblio";
...
+}
+
+/**
+ * @param {string} matched
+ */
+function inlineXrefMatches(matched) {
+  // slices "{{{" at the beginning and "}}}" at the end
+  const ref = matched.slice(3, -3).trim();
+  return /** @type {HTMLElement} */ ref.startsWith("\\")
```
saschanaz:
```suggestion
  return ref.startsWith("\\")
```

Doesn't look needed, it should be automatically inferred as `Node`.

src/core/inlines.js
```diff
@@ -22,6 +22,72 @@ import { renderInlineCitation } from "./render-biblio";
...
+  }
+}
+
+/**
+ * @param {string} matched
+ * @param {Text} txt
+ * @param {Map<string, string>} abbrMap
+ */
+function inlineAbbrMatches(matched, txt, abbrMap) {
+  return /** @type {HTMLElement} */ txt.parentElement.tagName === "ABBR"
```
saschanaz:
```suggestion
  return txt.parentElement.tagName === "ABBR"
```

No need to manually cast here.

src/core/inlines.js
```diff
@@ -22,6 +22,72 @@ import { renderInlineCitation } from "./render-biblio";
...
+          `Remove '!' from the start of the reference \`[[!${ref}]]\``
+      );
+    }
+
+    if (type === "informative" && !illegal) {
+      conf.informativeReferences.add(cleanRef);
+    } else {
+      conf.normativeReferences.add(cleanRef);
+    }
+    return /** @type {HTMLElement[]} */ cite.childNodes;
```
marcoscaceres: I guess here @saschanaz comments about `@return` apply also.

src/core/inlines.js
marcoscaceres: (also that returns a boolean)

saschanaz: @marcoscaceres It returns `Node`. (This is actually a ternary operation 😆)

src/core/inlines.js
```diff
@@ -70,66 +136,22 @@ export function run(conf) {
...
           )
         ) {
-          matched = matched.split(/\s+/).join(" ");
-          df.appendChild(
-            hyperHTML`<em class="rfc2119" title="${matched}">${matched}</em>`
-          );
-          // remember which ones were used
-          rfc2119Usage[matched] = true;
+          const node = inlineRFC2119Matches(matched);
+          df.appendChild(node);
```
marcoscaceres: I wonder if we should instead structure this section like:

```JS
if ( condition ) {
   const node = inlineRFC2119Matches(matched);
   df.appendChild(node);
   continue;
}
if (matched.startsWith("{{{")) {
   const node = inlineXrefMatches(matched);
   df.appendChild(node);
   continue;
}
// And so on...
```

saschanaz: I'd prefer `else if` because it explicitly means the branches are mutually exclusive.

marcoscaceres: ok, yeah... I was kinda abusing "continue;" as meaning that :)

src/core/inlines.js
marcoscaceres: oh lol, oops.

himanish-star:
@saschanaz, I have made the requested changes :)

marcoscaceres:
Great work @himanish-star. I want a few more folks to poke at this before we merge.

src/core/inlines.js
```diff
@@ -22,6 +22,71 @@ import { renderInlineCitation } from "./render-biblio";
...
+  if (type === "informative" && !illegal) {
+    conf.informativeReferences.add(cleanRef);
+  } else {
+    conf.normativeReferences.add(cleanRef);
+  }
+  return cite.childNodes;
+}
+
+/**
+ * @param {string} matched
```
marcoscaceres: nit: missing `@return`

saschanaz:
It’s inferred automatically so not always needed.

BTW I’d prefer @return to be at the bottom, just another nit.

src/core/inlines.js
himanish-star: well @saschanaz had suggested that the type was obvious so I can just leave that out.

himanish-star:
@saschanaz, did the nit changes suggested by you.

src/core/inlines.js
saschanaz: Yes, it can be inferred automatically by the compiler. It can still be manually written but it adds maintenance burden.

marcoscaceres: Hmm... ok yeah. I'm kinda 50/50 on this, as I like the code comments primarily for human usage: They help me outside the context of VS Code - like when doing code review or navigating the code through GitHub's web interface.

himanish-star: so what should I do :sweat_smile: ? shall I add those return types?

saschanaz: >like when doing code review

I think reviewing types should be automatically done by typescript. That's why we use it, right?

marcoscaceres: Agree.

marcoscaceres:
Great work @himanish-star!

PR #2153, opened 2019-03-14
Replace practise desc with advisement class

### LINKED ISSUE #2133: Best practices should use .advicement class
Best practice boxes should use the W3C’s advicement css class. We currently use out own custom style.

isunitha98selvan:
The wiki needs to be modified for best practises from
`<p class="practicedesc">
`to 
`<p class="advisement">`

closes #2133


saschanaz:
This doesn't close #2133. The change itself is good as the generator never uses `.practicedesc`.

I think we want to replace `.practicebox` with `.advisement` and `.practicetitle` with `.marker`.

isunitha98selvan:
@saschanaz changes have been made

assets/bp.css
```diff
@@ -6,16 +6,6 @@ span.practicelab {
...
-  color: #005A9C;
-  text-decoration: underline;
-  margin-bottom: 1em;
-}
-
-.practicebox {
-    background: black;
-}
-
 @media (max-width: 767px) {
```
marcoscaceres: Probably don't need this anymore, as practicetitle won't be used.

isunitha98selvan: Removed

assets/bp.css
```diff
@@ -4,29 +4,4 @@ span.practicelab {
...
-.practicebox {
-    background: black;
-}
-
-@media (max-width: 767px) {
-  span.practicetitle {
-    display: block;
-  }
-}
+}
```
saschanaz: nit: a newline at the end of the file

isunitha98selvan: has been added

saschanaz:
Looks like I was wrong, the `practicebox` should be kept and `.practicelab` should be replaced with `.advisement`. Also, the tr-design document currently defines `.advisement` only valid in `<strong>`/`<p>`/`<div>` while we only supports `<span>`. Maybe we should convert `<span>`s into those valid elements.

marcoscaceres:
Using `strong` for `span`s would be good.

marcoscaceres:
... but honestly, I'm not sure why this isn't just handled in the issues-notes.js module. This is just another kind of box with a different color.

PR #2155, opened 2019-03-15
refactor(core/inlines) : Minor optimizations

Swapnilr1:
I have done some minor optimizations in the inlines.js module.

1. Rather than using the JavaScript `shift()`, I have switched to simple array-indexing. Although JavaScript functions do not have an implementation standard, from what I read, most implementations would cause array re-indexing on each use of `shift()` which is somewhat inefficient.

2. I think that it is better to check for RFC 2119 keywords at the end - it is more efficient to check if a string starts with `{{{` or `[[` first than to run the entire regex expression on all strings. 

These are minor optimizations, because while they probably are efficient, I am not sure if it will make a difference beyond a few milliseconds - this depends how many inline keywords occur in a W3C spec in a practical scenario. 

There are 2 other modifications I was planning on doing, but would like some comments from others first since I do not have much experience with regex:

1. Do we really need the FAIL condition at the end? To me it seems like a bug if that condition ever arises, and so the focus should be on fixing that bug, and not have an extra condition.

2. On line 108, we are sorting the `aKeys` array in descending order of length. But since the regex is actually matches `<space>abbr<space>` do we really to sort it? Even if we have abbreviations like `abbr` and `abbre`, due to the space, regex should not match an instance of `abbre` with `abbr`.

### DIFF
--- a/src/core/inlines.js
+++ b/src/core/inlines.js
@@ -150,7 +150,7 @@ export function run(conf) {
         df.appendChild(node);
       } else {
         // FAIL -- not sure that this can really happen
-        throw `Found token '${t}' but it does not correspond to anything`;
+        throw new Error(`Found token '${t}' but it does not correspond to anything`);
       }
     }
     txt.parentNode.replaceChild(df, txt);
--- a/src/core/inlines.js
+++ b/src/core/inlines.js
@@ -17,7 +17,6 @@
 import { getTextNodes, refTypeFromContext, showInlineWarning } from "./utils";
 import hyperHTML from "hyperhtml";
 import { idlStringToHtml } from "./inline-idl-parser";
-import { pub } from "./pubsubhub";
 import { renderInlineCitation } from "./render-biblio";
 export const name = "core/inlines";
 export const rfc2119Usage = {};
@@ -105,7 +104,6 @@ export function run(conf) {
     abbrMap.set(abbr.textContent, abbr.title);
   }
   const aKeys = [...abbrMap.keys()];
-  aKeys.sort((a, b) => b.length - a.length);
   const abbrRx = aKeys.length ? `(?:\\b${aKeys.join("\\b)|(?:\\b")}\\b)` : null;
 
   // PROCESSING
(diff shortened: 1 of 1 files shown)


Swapnilr1:
@saschanaz @marcoscaceres Please review this.

src/core/inlines.js
```diff
@@ -129,20 +129,16 @@ export function run(conf) {
...
-    while (subtxt.length) {
-      const t = subtxt.shift();
+    for (let i = 0; i < subtxt.length; i++) {
+      const t = subtxt[i];
       let matched = null;
-      if (subtxt.length) matched = subtxt.shift();
+      if (i != subtxt.length - 1) {
+        matched = subtxt[i + 1];
+        i++;
+      }
```
saschanaz: I've heard `shift()` is fast enough in modern engines, and personally found it's more readable here. Thoughts? @marcoscaceres

Swapnilr1: @saschanaz Yes, I was somewhat doubtful about the performance improvements (basically it would only matter if `subtxt` had some ~1000 elements, not really occuring in an actual spec). I will revert back to `shift()`.

src/core/inlines.js
```diff
@@ -129,20 +129,16 @@ export function run(conf) {
...
       df.appendChild(document.createTextNode(t));
       if (matched) {
```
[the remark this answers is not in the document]
Swapnilr1: Yes, I will add that.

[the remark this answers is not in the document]
Swapnilr1: @saschanaz Could you also look at the proposed changes I described (not coded yet)?

src/core/inlines.js
```diff
@@ -129,20 +129,16 @@ export function run(conf) {
     if (subtxt.length === 1) continue;
 
     const df = document.createDocumentFragment();
-    while (subtxt.length) {
-      const t = subtxt.shift();
+    for (let i = 0; i < subtxt.length; i++) {
```
marcoscaceres: Is there some reason we can't do `for (const t of  subtext) {`?

marcoscaceres:
> I think that it is better to check for RFC 2119 keywords at the end - it is more efficient to check if a string starts with {{{ or [[ first than to run the entire regex expression on all strings.

Yes, that could be good. 

> These are minor optimizations, because while they probably are efficient, I am not sure if it will make a difference beyond a few milliseconds - this depends how many inline keywords occur in a W3C spec in a practical scenario.

Payment Request has ~150 `[[` inlines. As we are still working on `{{{`, there are currently none... but if it starts working properly, there could be another ~100 or so.  There are also around ~50-70 RFC2119 inlines. 

> There are 2 other modifications I was planning on doing, but would like some comments from others first since I do not have much experience with regex:

> Do we really need the FAIL condition at the end? To me it seems like a bug if that condition ever arises, and so the focus should be on fixing that bug, and not have an extra condition.

I agree we can probably remove this - or making sure it's actually "exceptional" (i.e., should `throw`, not display a user error).

> On line 108, we are sorting the aKeys array in descending order of length. But since the regex is actually matches <space>abbr<space> do we really to sort it? Even if we have abbreviations like abbr and abbre, due to the space, regex should not match an instance of abbre with abbr.

Sounds worth fixing.

src/core/inlines.js
marcoscaceres: It be nice if we didn't need to `shift()` at all. Haven't looked at the code in detail in a while, so don't know if it's possible... but with thinking about.

src/core/inlines.js
saschanaz: Hmm yeah, we don't even need an exact index number. Maybe just:

```js
let match = true;
for (const t of subtext) {
  match = !match;
  if (match) {
    df.appendChild(document.createTextNode(t));
  } else if (t.startsWith(...)) {
    // ...
  }
}

**Comment by Swapnilr1** [comment]
Thanks for the suggested changes. I am slightly busy at the moment, I will make the modifications by Tuesday.

**Comment by marcoscaceres** [comment]
Thanks for the update @Swapnilr1!

**Comment by Swapnilr1** [comment]
@saschanaz @marcoscaceres Currently I have changed the fail condition to `throw`. Please review the changes.

File: src/core/inlines.js [hunk 4]
```diff
@@ -129,35 +127,30 @@ export function run(conf) {
...
+      } else if (
+        /MUST(?:\s+NOT)?|SHOULD(?:\s+NOT)?|SHALL(?:\s+NOT)?|MAY|(?:NOT\s+)?REQUIRED|(?:NOT\s+)?RECOMMENDED|OPTIONAL/.test(
+          t
+        )
+      ) {
+        const node = inlineRFC2119Matches(t);
+        df.appendChild(node);
+      } else {
+        // FAIL -- not sure that this can really happen
+        throw `Found token '${t}' but it does not correspond to anything`;
```
marcoscaceres: Always throw real errors, otherwise we get not stack! :) 

```suggestion
        throw new Error(`Found token '${t}' but it does not correspond to anything`);
```

Swapnilr1: Thanks, I will read about the difference between simple `throw` and creating a new error.

PR #2157, opened 2019-03-17
feat(core/xref): collect similar errors

sidvishnoi:
image
image


sidvishnoi:
Question: should we show warning or errors?

marcoscaceres:
That's an error, I think. Not linking is bad.

PR #2160, opened 2019-03-17
fix(core/link-to-dfn): don't xref autofills

marcoscaceres:


marcoscaceres:
Don't know how we missed this one... only showed up after release 🤷‍♂️

PR #2162, opened 2019-03-18
fix: use external hljs bundle

marcoscaceres:
We previously forgot a critical thing - the W3C Server mirrors: 

 * js/deps/respec-highlight.js as https://www.w3.org/Tools/respec/respec-highlight.js


.gitignore
```diff
@@ -14,3 +14,4 @@ js/templates.js
 js/ui/*.js
 js/w3c/templates/*
 js/w3c/*.js
+!js/deps/highlight.js
```
saschanaz: Is this a must? Can we publish this from its repo instead?

marcoscaceres: We definitely could, but I need to make a request to the w3c system's team, etc. It's not a big deal, but will probably take a few days to set up on the w3c side.

saschanaz:
Can't see the snyk error message 🤔

marcoscaceres:
> Can't see the snyk error message 🤔

Hljs itself (gear)... moved the dep to be a developer dependency.

.gitignore
marcoscaceres: Can you live with this for now? I can put thing in motion on the W3C side, but releasing is somewhat broken without patch because the hljs build tool is no more (RIP 💀).

saschanaz: I definitely can.

saschanaz:
Is removing package-lock.json intentional?

marcoscaceres:
> Is removing package-lock.json intentional?

oh, whoops! totally not. Well spotted 😳

marcoscaceres:
Oh, package-lock.js is not deleted... just had a lot of deletions 😅

saschanaz:
Oops 😲

PR #2164, opened 2019-03-18
fix(core/link-to-dfn): don't xref autofills (again)

marcoscaceres:
What's better here is:
 1. more clear... 
 1. avoid call to .closest()

### DIFF
--- a/src/core/link-to-dfn.js
+++ b/src/core/link-to-dfn.js
@@ -238,7 +238,7 @@ function findExplicitExternalLinks() {
   return [...links]
     .filter(el => {
       // ignore empties
-      if (el.textContent === "") return false;
+      if (el.textContent.trim() === "") return false;
       /** @type {HTMLElement} */
       const closest = el.closest("[data-cite]");
       return !closest || closest.dataset.cite !== "";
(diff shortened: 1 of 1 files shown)


src/core/link-to-dfn.js
```diff
@@ -237,11 +237,11 @@ function findExplicitExternalLinks() {
   );
   return [...links]
     .filter(el => {
+      // ignore empties
+      if (el.textContent === "") return false;
```
saschanaz: Why removing `.trim()`?

marcoscaceres: The convention is `<a data-cite="thing"></a>` to autofill a thing, but I guess to be safe it should trim.

saschanaz: No strong opinion but a formatter may insert a newline between tag opening and close.

marcoscaceres: Yeah, I was thinking about that too... I don't think they do for `<a>` (at least tidy doesn't) because it would create an empty link with a space in it.

saschanaz: AFAIK the HTML whitespace rule won't render a space if the element follows or followed by whitespaces.

marcoscaceres: ah, ok. I wasn't aware. Browsers are fun 😍

PR #2168, opened 2019-03-19
refactor(core/caniuse): use respec.org API

sidvishnoi:
- remove semverCompare from core/utils
- update test data and tests
- moved addBrowser() out of createTableHTML()
- Minor perf: assigning addBrowser.addLatestVersion.buttonText as a
  string results in a single text node, instead of 3 if we add the same
  in hyperHTML tagged template
- getNormalizedConf is a pure function now
- maxAge is controlled by server response now
- default browser is handled by server now
- lots of nits, cleanup

Don't merge until we've a stable https://respec.org


src/core/caniuse.js
```diff
@@ -47,16 +46,15 @@ export async function run(conf) {
...
+  if (!options.feature) {
     return; // no feature to show
   }
-  const { feature } = caniuse;
-  const featureURL = `https://caniuse.com/#feat=${feature}`;
+  const featureURL = `https://caniuse.com/#feat=${options.feature}`;
   const link = createResourceHint({
     hint: "preconnect",
-    href: "https://raw.githubusercontent.com",
+    href: "https://respec.org",
```
marcoscaceres: Ah, we should do this for xref too!

src/core/caniuse.js
```diff
@@ -88,36 +87,29 @@ export async function run(conf) {
...
-    return;
+    return { feature: conf.caniuse, ...DEFAULTS };
   }
-  if (Array.isArray(conf.caniuse.browsers)) {
-    conf.caniuse.browsers = conf.caniuse.browsers
+  const caniuseConf = { ...DEFAULTS, ...conf.caniuse };
+  if (Array.isArray(caniuseConf.browsers)) {
+    caniuseConf.browsers = caniuseConf.browsers
       .map(b => b.toLowerCase())
       .filter(isValidBrowser);
```
saschanaz: Could we only warn rather than filtering unknown browsers? Then users won't have to wait for ReSpec to add a new browser.

marcoscaceres:
I’ve reached out to Dreamhost about stability and auto restarts.

src/core/caniuse.js
sidvishnoi: Valid point. We iterate over the browsers in server response anyway, so this filter isn't needed.

src/core/caniuse.js
```diff
@@ -105,22 +105,16 @@ function getNormalizedConf(conf) {
   }
   const caniuseConf = { ...DEFAULTS, ...conf.caniuse };
   if (Array.isArray(caniuseConf.browsers)) {
-    caniuseConf.browsers = caniuseConf.browsers
-      .map(b => b.toLowerCase())
-      .filter(isValidBrowser);
+    for (const browser of caniuseConf.browsers)
+      if (!BROWSERS.has(browser)) {
+        pub(
```
marcoscaceres: Collect the offending browsers and publish a single warning? 🙏

sidvishnoi: image
Fixed

marcoscaceres:
Good news: server is now stable. It gracefully handles server restarts ⭐️

Issue #2084, opened 2019-02-11
Modules should contain their own localization information


marcoscaceres:
We currently have localization strings in core/l10n.js that are used throughout ReSpec. However, it's rare that a localized string is used in more than one place at a time. What would be better is for where localization happens, the strings be included in the appropriate module itself. 

Then core/l10n should only export the `lang` of the document, which should be configurable. Then each module that requires localization should just key off that value.  

This model is already used by the linter modules:
https://github.com/w3c/respec/blob/develop/src/core/linter-rules/check-charset.js#L6-L18

So, we just need to copy that. 

The following remain:

- [x] src/core/figures.js
- [x] src/core/github.js
- [x] src/core/issues-notes.js
- [x] src/core/structure.js
- [x] src/w3c/headers.js
- [x] src/w3c/templates/cgbg-headers.js
- [x] src/w3c/templates/cgbg-sotd.js
- [x] src/w3c/templates/headers.js
- [x] src/w3c/templates/sotd.js

ThobyV:
Can I take on this issue please?

marcoscaceres:
Sure @ThobyV. Go for it. Are you clear on what you need to do?

ThobyV:
I would need a bit more explanation if you could, but can you please point me to an example module that uses the *core/i10n*  localization strings but not the newly proposed, required model used in https://github.com/w3c/respec/blob/develop/src/core/linter-rules/check-charset.js#L6-L18

I think I would be able to find out more from there.

marcoscaceres:
You should be able to find lots of instances by searching the code for "conf.l10n". 

Try:
```BASH
 git grep -C5 "conf.l10n" src
```

to see where they are used, and in what context.

ThobyV:
Thanks @marcoscaceres for your time. I am about to Implement your instructions and hopefully send a PR for review. However There are some things I wish to be clear about. 

Firstly, Is it ok to remove  the `run` function from *core/I10n* and where it's called since I think we wouldn't need or export it anymore as we are no longer setting the `config.I10n` object to localization strings  (If I am correct). 

Secondly would this mean what I would need to Include the localization strings in the appropriate  modules  with the available language options as used here:
https://github.com/w3c/respec/blob/c115b0438832b4f555933297793356a8add0260d/src/w3c/l10n.js#L4-L27

ThobyV:
[...]

PR #2169, opened 2019-03-19
refactor(core/structure): localize module 

ThobyV:
P.R to add localization in respect to issue https://github.com/w3c/respec/issues/2084 for the module `structures.js`.  I think this is last module in /core that needs localization. Thanks @marcoscaceres  and everyone for review! Will make sure to complete the rest asap.


ThobyV:
Thanks @marcoscaceres. Do you think I would be able to send a pre-proposal for early review or  wait until gsoc official time?

marcoscaceres:
Send me an email and let's discuss (marcos@marcosc.com)!

marcoscaceres:
Quick `git grep --name-only conf.l10n src` shows the following modules still need localization (@ThobyV, not asking you to do these... just so I have an idea of where we are at!):  

src/core/figures.js
src/core/github.js
src/core/issues-notes.js

And: 
src/w3c/headers.js
src/w3c/templates/cgbg-headers.js
src/w3c/templates/cgbg-sotd.js
src/w3c/templates/headers.js
src/w3c/templates/sotd.js

ThobyV:
@marcoscaceres I think those modules left in /core have localization which are not used in the interface, would they still be refactorable?

marcoscaceres:
Not sure I understand what you mean @ThobyV. Could you please clarify? 

Looking at figures.js, we could definitely fold `conf.l10n.fig` into the figures.js file.

PR #2171, opened 2019-03-20
Add <cite> when data-cite='spec' is used

### LINKED ISSUE #2021: Add <cite> when data-cite="spec" is used.
Given the following `data-cite` (with no path or fragment): 

```HTML
 <a data-cite="somespec">Some spec</a>
```

It would be nice if that was outputted as:

```HTML
<cite><a href="https://foo.com/whatever-spec">Some spec</a></cite>
```

I think the place where this needs is around here:
https://github.com/w3c/respec/blob/develop/src/core/data-cite.js#L53

After the link is created, if there is no fragment or path, then the `<cite>` can wrap `<a>`.

pradeepgangwar:
fixes #2021


pradeepgangwar:
@marcoscaceres I understand that this is a very raw PR but I was stuck when I tried to wrap `<a>` in `<cite>`. While doing so `<a href="...">spec</a>` stopped rendering. It would be great if you can help me a bit to get some initial start.

src/core/data-cite.js
```diff
@@ -52,12 +52,17 @@ function requestLookup(conf) {
...
       href = new URL(frag, href).href;
     }
+    const cite = document.createElement("cite");
     switch (elem.localName) {
       case "a": {
         if (elem.textContent === "") {
           elem.textContent = title;
         }
         elem.href = href;
+        // ToDo (It creates problem because HTML does not know ehere to place newly created node that is <cite> <a..> </a> </cite>)
```
pradeepgangwar: @marcoscaceres I was able to add `<cite>` for dfn successfully but it creates problem with `<a>` probably because HTML does not know where to place newly created node <cite>. 
In dfn we are able to tackle this by appending `<cite>` to dfn element. I tried looking for solution but could not figure out any concrete solution. Do you have any suggestions?

pradeepgangwar:
I have already added tests for `dfn` case. Seems to work fine for now.

src/core/data-cite.js
marcoscaceres: > because HTML does not know where to place newly created node 

Not sure what you mean here? I'd suggest stepping into  wrapInner(cite, elem); and seeing exactly what's going on.

saschanaz: The below line just appends `elem` into `cite` and nothing more. You should instead do `elem.replaceWith(cite); cite.append(elem);`.

pradeepgangwar: @saschanaz That worked. Thanks. 😄 
I ran into some issues recently while I was trying an alternative fix. But that was quick. 😬

tests/spec/core/data-cite-spec.js
```diff
@@ -372,4 +372,26 @@ describe("Core — data-cite attribute", () => {
     dfnA = doc.querySelector("#t2 > dfn > cite > a");
     expect(dfnA.href).toBe("https://fetch.spec.whatwg.org/");
   });
+
+  it("Adds <cite> around <a> when frag and path are missing from <a data-cite='some-spec'>", async () => {
+    const ops = {
```
sidvishnoi: We also have a `makeStandardOps(body, respecConfigOverride)` function which gives more abstraction as:
``` js
const body = `
  <section>
    <p id="t1"><a data-cite="HTML"></a></p>
    <p id="t2"><a data-cite="Fetch"></a></p>
  </section>`;
const ops = makeStandardOps(body, null);
```

pradeepgangwar: Yeah I saw this structure in other test files but since most of the tests in data-cite were following the method that is there in this PR so I went with that. 😬 
Should I change it?

saschanaz: Didn't know that! A refactoring for other tests would be good, especially when even active maintainers are frequently just copy-pasting existing tests.

sidvishnoi: >  Should I change it?

Yes please.

pradeepgangwar: Should I change it in this PR itself or open a new one? I am okay with either of them.

sidvishnoi: Lets keep the the fix for this PR here only, then we can do a refactor for other tests later in some other PR.

pradeepgangwar: Sure. Will do it. 🙂

pradeepgangwar:
Did the changes. Please review. :)

src/core/data-cite.js
```diff
@@ -68,6 +73,10 @@ function requestLookup(conf) {
         } else {
           wrapInner(elem, anchor);
         }
+        if (!path && !frag) {
+          wrapInner(cite, anchor);
```
saschanaz: `wrapInner` is for elements that already have their children. You can just `cite.append(anchor)` in this case.

pradeepgangwar:
@saschanaz Fixed that. Let me know if it requires further changes. :)

src/core/data-cite.js
```diff
@@ -52,12 +52,17 @@ function requestLookup(conf) {
     if (frag) {
       href = new URL(frag, href).href;
     }
+    const cite = document.createElement("cite");
```
sidvishnoi: Maybe keep this `createElement()` local to `case` where it is used?

pradeepgangwar: Sorry, I didn't get you. Actually I have used it inside `case: "a" and case "dfn"` block so if I define it there then I will have to define it twice.

sidvishnoi: I was thinking we should define it as close to as it is used. Declaring twice shouldn't be an issue I think. WDYT @marcoscaceres

marcoscaceres: Agree... would be fine to do:
 
```JS
        if (!path && !frag) {
           const cite = document.createElement("cite");
           elem.replaceWith(cite);
           cite.append(elem);
         }
```

pradeepgangwar:
Did the changes. Please review. :D

tests/spec/core/data-cite-spec.js
```diff
@@ -350,4 +350,46 @@ describe("Core — data-cite attribute", () => {
     ).toBeTruthy();
     expect(dahut).toBe(null);
   });
+
+  it("Adds <cite> around <a> when frag and path are missing from <dfn data-cite='some-spec'>", async () => {
+    const body = `
+      <section>
+        <p id="t1"><dfn data-cite="HTML"></dfn></p>
+        <p id="t2"><dfn data-cite="Fetch"></dfn></p>
```
marcoscaceres: t2 seems identical to t1. I think we can drop t2.

tests/spec/core/data-cite-spec.js
```diff
@@ -350,4 +350,46 @@ describe("Core — data-cite attribute", () => {
...
+    expect(cite).toBeTruthy();
+    dfnA = doc.querySelector("#t2 > dfn > cite > a");
+    expect(dfnA.href).toBe("https://fetch.spec.whatwg.org/");
+  });
+
+  it("Adds <cite> around <a> when frag and path are missing from <a data-cite='some-spec'>", async () => {
+    const body = `
+      <section>
+        <p id="t1"><a data-cite="HTML"></a></p>
+        <p id="t2"><a data-cite="Fetch"></a></p>
```
marcoscaceres: As above. t2 seems identical to t1.

pradeepgangwar:
Working on it.

tests/spec/core/data-cite-spec.js
```diff
@@ -324,4 +324,47 @@ describe("Core — data-cite attribute", () => {
...
+    const ops = makeStandardOps(null, body);
+    const doc = await makeRSDoc(ops);
+
+    const cite = doc.querySelector("#t1 > cite");
+    expect(cite).toBeTruthy();
+    const a = doc.querySelector("#t1 > cite > a");
+    expect(a.href).toBe("https://html.spec.whatwg.org/multipage/");
+  });
+
+  it("<cite> does not appear around tags other than a/dfn when frag and path are missing with data-cite='some-spec' attribute", async () => {
```
sidvishnoi:
```suggestion
  it("does not wrap tags other than a/dfn with <cite> even with [data-cite='spec']", async () => {
```

tests/spec/core/data-cite-spec.js
```diff
@@ -324,4 +324,47 @@ describe("Core — data-cite attribute", () => {
     ).toBeTruthy();
     expect(dahut).toBe(null);
   });
+
+  it("Adds <cite> around <a> when frag and path are missing from <dfn data-cite='some-spec'>", async () => {
```
sidvishnoi: Can we merge this and the test below into a single test? Test boilerplate is larger than test itself 😅

pradeepgangwar:
@saschanaz Added the tests for the case mentioned by you.

tests/spec/core/data-cite-spec.js
sidvishnoi: @marcoscaceres @pradeepgangwar

marcoscaceres: agree.

pradeepgangwar: Okay, So should I remove this? @sidvishnoi

pradeepgangwar: oh okay so i should merge it into below test.

pradeepgangwar:
@marcoscaceres @sidvishnoi Did the changes and merged the tests.

marcoscaceres:
Awesome, thanks @pradeepgangwar! will merge as soon as travis is happy.

marcoscaceres:
@pradeepgangwar sorry for the delay... TravisCI is having a meltdown :( Hopefully they will get things sorted soon.

PR #2172, opened 2019-03-20
(core/structure): Fixed checking for empty text anchors

### LINKED ISSUE #2170: Autofilling an anchor not working in deeply nested algorithm?
## Important info

* URL to affected spec: https://github.com/w3c/performance-timeline/pull/115
* ReSpec version: Click the ReSpec pill (e.g., 20.0.1).

## Description of problem

What happened (e.g., it crashed)?: In https://github.com/w3c/performance-timeline/pull/115 something is preventing the section's title from getting filled. 

But what I was expecting?:

for the section's title to appear in the `<a>`. 

Optional, steps to reproduce:
 1. See https://github.com/w3c/performance-timeline/pull/115

Swapnilr1:
fixes #2170 

The trimmed `textContent` should be compared to empty string.

### DIFF
--- a/src/core/structure.js
+++ b/src/core/structure.js
@@ -242,7 +242,7 @@ function updateEmptyAnchors(secMap) {
   [...document.querySelectorAll("a[href^='#']:not(.tocxref)")]
     .filter(
       anchor =>
-        anchor.textContent === "" &&
+        anchor.textContent.trim() === "" &&
         anchor.getAttribute("href").slice(1) in secMap
     )
     .forEach(anchor => {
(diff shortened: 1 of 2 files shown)


Swapnilr1:
@marcoscaceres I have added the test. Also, I found that the localization test was outside the `describe()` block, so I added it inside the block. 

Also I was wondering why the localization test gives TypeError: "e is undefined" in the browser console?

marcoscaceres:
@Swapnilr1 does the error have a stack? Might help us figure out where the other issue is.

PR #2173, opened 2019-03-20
Add <cite> when data-cite="spec" is used

### LINKED ISSUE #2021: Add <cite> when data-cite="spec" is used.

yatri1609:
Fixes #2021 

`<cite><a href="https://foo.com/whatever-spec">Some spec</a></cite>`

This type of link is created when there is no fragment or path.

### DIFF
--- a/src/core/data-cite.js
+++ b/src/core/data-cite.js
@@ -52,6 +52,10 @@ function requestLookup(conf) {
     if (frag) {
       href = new URL(frag, href).href;
     }
+    if (frag == null && path == null) {
+      const cite = hyperHTML`<cite>`;
+      wrapInner(cite, elem);
+    }
     switch (elem.localName) {
       case "a": {
         if (elem.textContent === "") {
@@ -62,6 +66,8 @@ function requestLookup(conf) {
       }
       case "dfn": {
         const anchor = hyperHTML`<a href="${href}">`;
+        const cite = hyperHTML`<cite>`;
+        wrapInner(cite, anchor);
         if (!elem.textContent) {
           anchor.textContent = title;
           elem.append(anchor);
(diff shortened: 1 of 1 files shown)


yatri1609:
@marcoscaceres @sidvishnoi If you could have a look at this

src/core/data-cite.js
```diff
@@ -52,6 +52,10 @@ function requestLookup(conf) {
     if (frag) {
       href = new URL(frag, href).href;
     }
+    if (frag == null && path == null) {
+      const cite = hyperHTML`<cite>`;
```
marcoscaceres:
```suggestion
      const cite = document.createElement("cite");
```

Same below.

yatri1609:
@marcoscaceres I'll add the test by tomorrow

src/core/data-cite.js
```diff
@@ -52,6 +52,10 @@ function requestLookup(conf) {
     if (frag) {
       href = new URL(frag, href).href;
     }
+    if (frag == null && path == null) {
```
sidvishnoi: I'm not sure if `== null` is the right check here (`null` is generally used to tell that's it's explicitly set to null).
```suggestion
    else if (!path) {
```
Though `if (!frag && !path)` might be preferable as it's more explicit.

marcoscaceres:
Closing as duplicate of #2171

PR #2175, opened 2019-03-21
feat(core/inlines): add new syntax for inline <var>

sidvishnoi:
``` html
|varName: type|     <=>    <var data-type="type">varName</var>
|varName|           <=>    <var>varName</var>
```


tests/spec/core/inlines-spec.js
```diff
@@ -72,4 +72,26 @@ describe("Core - Inlines", () => {
     expect(rfc2119[0].textContent).toBe("MUST");
     expect(rfc2119[1].textContent).toBe("NOT RECOMMENDED");
   });
+
+  it("processes inline variable syntax", async () => {
+    const body = `
+      <section>
+        <p id="a">TEXT |Interface:variable| TEXT</p>
+        <p id="b">TEXT |variable| TEXT</p>
+        <p id="c">TEXT | ignored | TEXT</p>
```
marcoscaceres: I think we need one more ignore case:

```
<p id="a">TEXT|Ignore:ignore|TEXT</p>
```

src/core/inlines.js
```diff
@@ -118,6 +133,7 @@ export function run(conf) {
       "\\b(?:NOT\\s+)?RECOMMENDED\\b",
       "\\bOPTIONAL\\b",
       "(?:{{3}\\s*.*\\s*}{3})", // inline IDL references,
+      "\\|\\w+(?:\\:\\w+)?\\|", // inline variable regex
```
marcoscaceres: @saschanaz we plan to match on Interface names and primitives (e.g., string, object, promise) etc. Does this look ok?

saschanaz: I'm not very familiar with this kind of things... Especially haven't seen scoped variables. Do you have some existing examples?

marcoscaceres: We are inventing something new :) anyway, I think I should trial it in one or two specs. BikeShed uses unscoped ones. Scoped ones offer all sorts of neat opportunities.

saschanaz: I want to understand what would it even mean, will there be same-named variables with different scopes? How would it be represented visually?

src/core/inlines.js
```diff
@@ -86,6 +86,21 @@ function inlineAbbrMatches(matched, txt, abbrMap) {
...
+ * @example |Interface:varName| => <var data-type="Interface">varName</var>
+ * @example |varName| => <var>varName</var>
+ * @param {string} matched
+ */
+function inlineVariableMatches(matched) {
+  // remove "|" at the beginning and at the end, then split at an optional `:`
+  const matches = matched.slice(1, -1).split(":", 2);
+  if (matches.length === 2) {
+    const [interfaceName, varName] = matches;
+    return hyperHTML`<var data-type="${interfaceName}">${varName}</var>`;
```
saschanaz: AFAIK currently `data-type` is used to represent the type of IDL definitions, so this looks somewhat confusing. Maybe `data-parent`, `data-scope`, or `data-context`?

src/core/inlines.js
marcoscaceres: Visually, I’m imagining something similar (but simpler) to what VSCode shows when you hover over a variable. Once the type is declared, all other matching vars inherent the type. This really helpful when reading algorithms, as you can imagine.

src/core/inlines.js
saschanaz: Oh wait, it's not really a scope?

marcoscaceres: Yeah, we might have a conflict. I need to check how we use data-type in idl.

marcoscaceres: I guess it depends on you definition of scope.

saschanaz: I thought I misunderstood and it literally represented the variable's type. And now I'm confused 😵

marcoscaceres: Just to be clear, the use case is:

`Let |DOMString:name| be “hello”.`

saschanaz: Okay then, `data-type` looks good as-is. But then the syntax looks too much like `DOMString.name` for me...

marcoscaceres: Sorry for the confusion. I realize now we didn’t give enough context/explanation of what we were trying to achieve. I’ll add a proper bug description tomorrow so it’s more clear.

marcoscaceres: > looks too much like DOMString.name

We were also thinking of using c++ style `::` namespaces instead of just `:`. I don’t feel too strongly about it.

saschanaz: At first sight I thought it's something like C++ namespaces 🤣

marcoscaceres: Heh, I actually took inspiration from c++... as all DOM interfaces end up being c++ anyway.

saschanaz: >We were also thinking of using c++ style :: namespaces

Based on what I read, this is not about scoping something but about typing a variable, right? Maybe TypeScript-like `|name: DOMString|` would be good enough then.

marcoscaceres: > typing a variable, right? 

Correct. 

> |name: DOMString|

I guess because I’ve not used too much typescript, I still find the above odd to look at... I kinda want the type first, then the variable name. Maybe we don’t need the “:” at all... we can just use a space:

Let |DOMString name| be “hello”.

saschanaz: I think I've seen some variables that have spaces in their name.

src/core/inlines.js
```diff
@@ -86,6 +86,21 @@ function inlineAbbrMatches(matched, txt, abbrMap) {
...
+/**
+ * @example |Interface:varName| => <var data-type="Interface">varName</var>
+ * @example |varName| => <var>varName</var>
+ * @param {string} matched
+ */
+function inlineVariableMatches(matched) {
+  // remove "|" at the beginning and at the end, then split at an optional `:`
+  const matches = matched.slice(1, -1).split(":", 2);
+  if (matches.length === 2) {
+    const [interfaceName, varName] = matches;
```
sidvishnoi: Maybe we should've called it `type`, to avoid confusion.

src/core/inlines.js
marcoscaceres: Blasphemy! But yes, I’ve seen it too. Could be an opportunity to discourage that... or in that case, the typescript approach could work rather well.

marcoscaceres: Note the example above is all sorts of confusing... is it a concept used as a variable? That should really be an internal slot.

marcoscaceres: Ok, so in ReSpec’s tradition of not being overly opinionated, we should go with the typescript syntax.

sidvishnoi:
Improved regex to match following cases (suggestions welcome):
image

src/core/inlines.js
```diff
@@ -118,6 +130,7 @@ export function run(conf) {
       "\\b(?:NOT\\s+)?RECOMMENDED\\b",
       "\\bOPTIONAL\\b",
       "(?:{{3}\\s*.*\\s*}{3})", // inline IDL references,
+      "\\B\\|\\w+(?:\\s*\\w+)*(?:\\:\\s*\\w+)?\\|\\B", // inline variable regex
```
saschanaz: How about `\w[\s\w]*\w\s*(?:\:\s*\w+)?`?

marcoscaceres: good suggestion too.

```suggestion
      "\\B\\|\\w[\\s\\w]*\\w(?:\\s*\\:\\s*\\w+)?\\|\\B", // inline variable regex
```

tests/spec/core/inlines-spec.js
```diff
@@ -72,4 +72,37 @@ describe("Core - Inlines", () => {
     expect(rfc2119[0].textContent).toBe("MUST");
     expect(rfc2119[1].textContent).toBe("NOT RECOMMENDED");
   });
+
+  it("processes inline variable syntax", async () => {
+    const body = `
+      <section>
+        <p id="a1">TEXT |variable: Type| TEXT</p>
+        <p id="a2">TEXT |variable with spaces:Type| TEXT</p>
+        <p id="a3">TEXT |with spaces: Type| TEXT</p>
```
sidvishnoi:
```suggestion
        <p id="a3">TEXT |with spaces :  Type| TEXT</p>
```

src/core/inlines.js
sidvishnoi: I was being lazy and use commit suggestions to edit two files from github.com UI

tests/spec/core/inlines-spec.js
```diff
@@ -72,4 +72,37 @@ describe("Core - Inlines", () => {
...
+    const doc = await makeRSDoc(makeStandardOps(null, body));
+
+    const a1 = doc.querySelector("#a1 var");
+    expect(a1.textContent).toEqual("variable");
+    expect(a1.dataset.type).toEqual("Type");
+
+    const a2 = doc.querySelector("#a2 var");
+    expect(a2.textContent).toEqual("variable with spaces");
+    expect(a2.dataset.type).toEqual("Type");
+
```
sidvishnoi:
```suggestion

```

marcoscaceres:
@saschanaz, you happy for us to go with this?

src/core/inlines.js
```diff
@@ -138,6 +151,9 @@ export function run(conf) {
       } else if (t.startsWith("[[")) {
         const nodes = inlineBibrefMatches(t, txt, conf);
         df.append(...nodes);
+      } else if (t.startsWith("|") && t.endsWith("|")) {
```
saschanaz: nit: Do we need both checks? The first one should be sufficient.

marcoscaceres: Good catch.

PR #2181, opened 2019-03-22
feat(core/github-api): Centralised Github Fetching And Rate Limiting using API v3

### LINKED ISSUE #1631: Centralize github fetching

devanshbatra04:
closes #1631

I closed the earlier PR (#2099) as merge conflicts had popped up and there were just too many changes to be made. 
Still to write the tests.


src/core/github-api.js
```diff
@@ -0,0 +1,126 @@
...
+  } catch (err) {
+    issue.message = `Error JSON parsing issue #${issueNumber} from GitHub.`;
+  }
+  if (!response.ok || issue.message) {
+    const msg = `Error fetching issue #${issueNumber} from GitHub. ${
+      issue.message
+    } (HTTP Status ${response.status}).`;
+    pub("error", msg);
+  }
+  return /** @type {[number, GitHubIssue]} */ ([issueNumber, issue]);
```
sidvishnoi: Nit: We should move the in JSDoc comment above as `@returns`, instead of casting here.

src/core/github-api.js
```diff
@@ -0,0 +1,126 @@
...
+async function processResponse(response, issueNumber) {
+  // "message" is always error message from GitHub
+  const issue = { title: "", number: issueNumber, state: "", message: "" };
+  try {
+    const json = await response.json();
+    Object.assign(issue, json);
+  } catch (err) {
+    issue.message = `Error JSON parsing issue #${issueNumber} from GitHub.`;
+  }
+  if (!response.ok || issue.message) {
```
sidvishnoi: This should probably be split into two different error messages

src/core/github-api.js
```diff
@@ -0,0 +1,126 @@
...
+  return /** @type {[number, GitHubIssue]} */ ([issueNumber, issue]);
+}
+
+function findNext(header) {
+  // Finds the next URL of paginated resources which
+  // is available in the Link header. Link headers look like this:
+  // Link: <url1>; rel="next", <url2>; rel="foo"; bar="baz"
+  // More info here: https://developer.github.com/v3/#link-header
+  const m = (header || "").match(/<([^>]+)>\s*;\s*rel="next"/);
+  return (m && m[1]) || null;
```
sidvishnoi:
```suggestion
  return m ? m[1] : null;
```

src/core/github-api.js
```diff
@@ -0,0 +1,126 @@
...
+}
+
+export async function fetchAll(url, headers, output = []) {
+  const urlObj = new URL(url);
+  if (urlObj.searchParams && !urlObj.searchParams.has("per_page")) {
+    urlObj.searchParams.append("per_page", "100");
+  }
+  const request = new Request(urlObj, {
```
sidvishnoi: Nit: Prevent splitting into multiple lines maybe
``` js
const request = new Request(urlObj, { headers });
```

src/core/github-api.js
```diff
@@ -0,0 +1,126 @@
...
+    headers,
+  });
+  request.headers.set("Accept", "application/vnd.github.v3+json");
+  const response = await window.fetch(request);
```
sidvishnoi:
```suggestion
  const response = await fetch(request);
```

src/core/github-api.js
devanshbatra04: Is that right? Isn't the earlier version supposed to release a boolean. But this would result in m[1] whatever it turns out to be. 
(I didn't write it just moved around)

devanshbatra04: I finally understood what the code meant. Thanks!

src/core/github-api.js
devanshbatra04: As in the first error message being Error fetching issue #33 from GitHub. And the second being the issue.message?

sidvishnoi: `!response.ok` good be any http error.
I'm not sure where `issue.megsage` comes

sidvishnoi: ah. just saw the comment above. 
`// "message" is always error message from GitHub`
never mind. this is fine.

src/core/github-api.js
```diff
@@ -0,0 +1,124 @@
...
+    referrerPolicy: "no-referrer",
+    headers,
+  });
+  const responseJSON = await fetch(request).then(r => r.json());
+  return responseJSON.rate.remaining;
+}
+
+export async function fetchAndStoreGithubIssues(conf) {
+  const { githubAPI } = conf;
+  /** @type {NodeListOf<HTMLElement>} */
```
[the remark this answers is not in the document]
devanshbatra04: We don't apparently.

marcoscaceres:
@sidvishnoi could please give this one last look over?

src/core/contrib.js
```diff
@@ -24,9 +24,12 @@ function findUserURLs(...thingsWithUsers) {
 }
 
 async function toHTML(urls, editors, element, headers) {
-  const args = await Promise.all(urls.map(url => fetch(url, { headers })));
-  const names = args
-    .map(([user]) => user.name || user.login)
+  const args = await Promise.all(
+    urls.map(url => fetch(new Request(url, { headers })))
+  );
+  const argsResolved = await Promise.all(args.map(arg => arg.json()));
```
sidvishnoi: Can we do this `.json()` in the `args` above?

marcoscaceres: Good point.

devanshbatra04: Right. Done. :-)

PR #2183, opened 2019-03-22
fix(core/xref): duplicate references

### LINKED ISSUE #2180: Regression: normative biblio things end up in informative section
## Important info

* URL to affected spec: https://w3c.github.io/payment-request/#references

## Description of problem

Duplicate bibliographical entries:
https://w3c.github.io/payment-request/#references

DOM and dom, weidl and WEBIDL

But what I was expecting?:

For there to be no duplicates. 

Optional, steps to reproduce:
 1. See https://w3c.github.io/payment-request/#references

marcoscaceres:
Fixes #2180


src/core/xref.js
```diff
@@ -241,12 +241,24 @@ function addDataCiteToTerms(results, xrefMap, conf) {
 
       // add specs for citation (references section)
       const closestInform = elem.closest(nonNormativeSelector);
-      if (
```
sidvishnoi: we better split this into it's own function

src/core/xref.js
```diff
@@ -241,12 +241,24 @@ function addDataCiteToTerms(results, xrefMap, conf) {
...
+      const closestNormative = elem.closest(".normative");
+      let isNormative = false;
+      if (!closestInform) {
+        isNormative = true;
+      } else if (elem === closestNormative) {
```
sidvishnoi: maybe merge this in `if` statement above?
``` js
if (!closestInform || elem === closestNormative) {
  isNormative = true;
} else if (// ...
```

marcoscaceres: Good idea - will add. Could we treat this a nit (i.e., pre-approve this PR)? I’ve got a slight emergency with as this is blocking publications 😓

PR #2185, opened 2019-03-23
Shifted localization tests into specification test suites

Swapnilr1:
The localization tests recently added were outside the `describe()` block, so I have shifted them inside. I think I have shifted all of them.


marcoscaceres:
Thanks @Swapnilr1!

PR #2187, opened 2019-03-23
feat: executable-less API

### LINKED ISSUE #631: Add code coverage support
We should add a code quality tool like code climate to get an idea of gaps in our tests.

### LINKED ISSUE #1469: Use/support jsdom to preprocess
ReSpec shouldn't require installing real browsers to use `respec2html.js` as jsdom supports browser-like DOM manipulation in Node.js. Dropping Nightmare (and so Electron) dependency will allow easier build environment settings on GPU-less CLI environments.

saschanaz:
Just an initial draft...

Closes #1469, closes #631


src/core/reindent.js
```diff
@@ -20,8 +24,22 @@ function reindent(text) {
...
 }
+
+export function run() {
+  preprocess(document);
+}
+
+/**
+ * @param {import("../respec-document").RespecDocument} respecDoc
+ */
+export default function({ document }) {
```
devanshbatra04: why not directly export the function ```preprocess``` instead?

devanshbatra04: Never mind, just saw its a WIP. Sorry.

src/core/base-runner.js
```diff
@@ -31,6 +31,7 @@ function toRunnable(plug) {
       try {
         if (typeof plug.default === "function") {
           await plug.default(respecDoc);
+          resolve();
```
marcoscaceres: oops? Was this a long standing bug?

saschanaz: Nope, I added `plug.default` thing and just forgot calling this.

marcoscaceres: oh, ok 😁

marcoscaceres:
Wonder if we should start cherrypicking some of this? There are some good changes here, so worried about bit-rot.

saschanaz:
Trying some merge and being blocked by ES module problem. Will continue tomorrow with potential cherrypicks.

marcoscaceres:
Any goodies you want to start bringing over?

saschanaz:
I think I'll start bringing parts of this after I add everything in src/index.js and the new test passes. Not sure how I can split this into parts, as it seems every test depends on every module 🤔

marcoscaceres:
Yeah, did a quick scroll through and see what you mean.

src/core/utils.js
```diff
@@ -303,12 +296,21 @@ export function norm(str) {
...
-  return ISODate.format(date).replace(dashes, sep);
+  return toShortIsoDate(date).replace(dashes, sep);
 }
 
-// formats a date to "yyyy-mm-dd"
+/**
+ * formats a date to "yyyy-mm-dd"
+ * @param {Date} date
+ */
 export function toShortIsoDate(date) {
```
sidvishnoi: How about `date.toISOString().slice(0, 10)`

saschanaz: Much simpler 🎉

saschanaz:
Tests partially fail, but now so what 🤪 (Done to reduce changes)

marcoscaceres:
@saschanaz, is it ok to close this? we can keep the branch in case we want to bring anything back from here.

saschanaz:
Looking back, I think the core issue that lead me to do this is that ReSpec runs as a big singleton and there is no easy way to initialize all the internal states. (The biblio dictionary for example.) Adding a way to initialize could vastly simplify this.

Does adding initialization function interest you? @sidvishnoi

saschanaz:
That said, I don't think I have time to find any good thing here. Let's close this and perhaps thing about the initialization I mentioned.

sidvishnoi:
@saschanaz something like #2981?

saschanaz:
That's extending the initialization process, I mean the initialization after everything so that it can run again for another document.

saschanaz:
Oh wait, I guess it can be done in #2981 👀

sidvishnoi:
Not sure what (short term?) benefits it can add over using puppeteer/respecDocWriter. Can make an attempt later some time in future though.

saschanaz:
> If I understand correctly, something like baserunner.runAll, but callable from node?

Ultimately yes. I guess #2981 can be a good step before that.

PR #2188, opened 2019-03-23
(core/xref) : xref defaults

### LINKED ISSUE #2167: xref defaults
Been playing around with @sidvishnoi xref implementation, and it's simply amazeballs. Some things I thought of as I was playing around - we should extend the configuration:

• `xref: true` //enable it, what we have today. 
• or, `xref: ["list", "of", "specs"]` (get automatically added to `<body data-cite="">` )
• or `xref: "w3c"` (defaults for w3c..., which would include "html, infra, url, promises-guide, webidl, dom, fetch" to <body data-cite="">)

We could come up with other cool little profiles based on different communities. 

Eventually, all w3c specs would get `xref: w3c` by default. 😍😍😍😍

Swapnilr1:
fixes #2167 

Well, I think I have written this in a very messy way with too many conditionals, please suggest changes. More importantly, please check if this is the desired implementation.
I will add tests later once I get to know this is what is to be implemented.


src/core/xref.js
```diff
@@ -37,7 +41,35 @@ export async function run(conf, elems) {
     upgradeDB.createObjectStore("xrefs");
   });
   const cache = new IDBKeyVal(idb, "xrefs");
+  const isObject =
```
marcoscaceres: Yeah, the logic is right, but way too convoluted :)  A better approach is to use object spread... see an example here where we do something similar:

https://github.com/w3c/respec/blob/develop/src/core/github.js#L117

So, in this case, start with a default object, and then go through each case:

```JS

const xrefDefaults = {
   url: API_URL,
   profile: "",
   specs: null,
}

// in run()
if (!conf.xref) return;

const normalizedConfig = {}; 
// get the type.. 
const type = Array.isArray(conf.xref) ? "array" : typeof conf.xref;
switch(type) {
   case "boolean": 
      // ... true
      Object.assign(normalizedConfig,  xrefDefaults);
      break;
   case "string":
      // do we have this profile? no: fallback to true, but show an error to the the user. 
      break;
   case "array":
       Object.assign(normalizedConfig, xrefDefaults, {specs: [].concat(conf.xref)});
      break;
   case "object":
         Object.assign(normalizedConfig, xrefDefaults, conf.xref);
         break;
    default: 
       // show user error - bad configuration 
       Object.assign(normalizedConfig, xrefDefaults);
}
// now apply the `normalizedConfig`... 

```

Swapnilr1: Thanks, I will make the changes by tomorrow.

Swapnilr1:
Just realized that I still need to add a check for invalid configuration within the object. On it.

src/core/xref.js
```diff
@@ -72,6 +87,50 @@ export async function run(conf, elems) {
...
+      Object.assign(normalizedConfig, xrefDefaults);
+      break;
+    case "string":
+      if (profiles[conf.xref]) {
+        // We have the profile
+        Object.assign(normalizedConfig, xrefDefaults, {
+          specs: profiles[conf.xref],
+        });
+      } else {
+        pub("error", "Invalid cross-references profile");
```
marcoscaceres: Make sure you make this more descriptive. It should tell the user what profile is wrong, what the right options are, and how to fix the problem.

src/core/xref.js
```diff
@@ -72,6 +87,50 @@ export async function run(conf, elems) {
...
+        specs: [].concat(conf.xref),
+      });
+      break;
+    case "object":
+      Object.assign(normalizedConfig, xrefDefaults, conf.xref);
+      break;
+    default:
+      // show user error - bad configuration
+      Object.assign(normalizedConfig, xrefDefaults);
+      pub("error", "Invalid cross-references configuration");
```
marcoscaceres: Same as above here.

src/core/xref.js
```diff
@@ -39,7 +45,16 @@ export async function run(conf, elems) {
     },
   });
   const cache = new IDBKeyVal(idb, "xrefs");
+
+  normalizeConfig(conf);
   const { xref } = conf;
+  if (xref.specs)
```
sidvishnoi: Maybe we can make this a bit simpler:
``` js
if (xref.specs) {
  const bodyCite = document.body.dataset.cite ? document.body.dataset.cite.split(/\s+/) || [];
  document.body.dataset.cite = bodyCite.concat(xref.specs).join(" ");
}
```
(I'm not liking the `getAttribute()` inside a `setAttribute()`)

marcoscaceres: Agree. Good suggestion.

src/core/xref.js
```diff
@@ -72,6 +87,50 @@ export async function run(conf, elems) {
...
 }
 
+/**
+ * converts conf.xref to object with url and spec properties
+ */
+function normalizeConfig(conf) {
+  const xrefDefaults = {
+    url: API_URL,
+    profile: "",
+    specs: null,
```
sidvishnoi: I'm not sure if we want both `profile` and `specs`. @marcoscaceres Do we plan to add something more based on profile? If not, I think we can have only `specs: null | string | string[]` and if `typeof specs === "string"`, then treat it as profile and convert it to array of specs based on profile.

marcoscaceres: > @marcoscaceres Do we plan to add something more based on profile?

Yes, I was thinking "WHATWG" profile for instance, which doesn't include any w3c specs. And there are a few others, like Web Perf working group would have w3c + their specs.

marcoscaceres: (we would eventually need to figure out some extensible way of creating profiles... but there might be profiles that are popular enough to always include.... specially for w3c specs)

sidvishnoi: I meant, do we plan to use `profile` for something other than a list of specs?

marcoscaceres: oh, apologies, I see what you mean. No, we don't plan to use it for anything else.

Swapnilr1: OK, I was thinking it sounds more natural for the user to write `xref: { profile: "w3c"}` than `xref: { specs: "w3c"}`. But internally, yes we can have only `xref: { url: , specs: }`.

marcoscaceres: I agree with @Swapnilr1 ... there is also less type swapping. I'd prefer to only have multiple types at the `xref` level.

src/core/xref.js
Swapnilr1: Yeah, sorry about that. This is much cleaner, thanks.

src/core/xref.js
```diff
@@ -72,6 +87,50 @@ export async function run(conf, elems) {
...
+      break;
+    case "object":
+      Object.assign(normalizedConfig, xrefDefaults, conf.xref);
+      break;
+    default:
+      // show user error - bad configuration
+      Object.assign(normalizedConfig, xrefDefaults);
+      pub("error", "Invalid cross-references configuration");
+  }
+  conf.xref = normalizedConfig;
```
marcoscaceres: Don't reassign it, as it's now an internal representation. 

```suggestion
  return normalizedConfig;
```

Then use normalizedConfig in `run()`

src/core/xref.js
Swapnilr1: I agree that I need to make this more descriptive, but how do I do that? The list of available profiles will expand when this feature is implemented, so is this a better message:
`Invalid cross-references profile {nameOfProfile}. Please use one of the supported profile configurations listed <a href=linkToWiki>here</a>.

Perhaps, we could maintain a list of supported profiles on the Wiki?

src/core/xref.js
```diff
@@ -72,6 +84,71 @@ export async function run(conf, elems) {
...
+    case "string":
+      if (profiles[xref]) {
+        // We have the profile
+        Object.assign(normalizedConfig, xrefDefaults, {
+          specs: profiles[xref],
+        });
+      } else {
+        pub(
+          "error",
+          `Invalid cross-references profile "${xref}". Please use one of the supported specification profiles.`
```
marcoscaceres:
```suggestion
          `Invalid cross-references profile "${xref}". Please use one of the supported specification profiles: ${Object.keys(profiles).join(", ")}.`
```

marcoscaceres: Otherwise, the user has no way of knowing what the right option is without going to the wiki or looking at the source code.

src/core/xref.js
```diff
@@ -72,6 +84,71 @@ export async function run(conf, elems) {
...
+  switch (type) {
+    case "boolean":
+      // ... true
+      Object.assign(normalizedConfig, xrefDefaults);
+      break;
+    case "string":
+      if (profiles[xref]) {
+        // We have the profile
+        Object.assign(normalizedConfig, xrefDefaults, {
+          specs: profiles[xref],
```
marcoscaceres: Users don't care for case :) 

```suggestion
          specs: profiles[xref.toLowerCase()],
```

src/core/xref.js
```diff
@@ -72,6 +84,71 @@ export async function run(conf, elems) {
...
+        specs: [].concat(xref),
+      });
+      break;
+    case "object":
+      if (xref.profile && !profiles[xref.profile]) {
+        pub(
+          "error",
+          `Invalid cross-references profile "${
+            xref.profile
+          }". Please use one of the supported specification profiles.`
```
marcoscaceres: As above... you might want to make this a function to avoid code duplication.

Swapnilr1: Yes, I was thinking of that. Will do.

src/core/xref.js
```diff
@@ -72,6 +84,71 @@ export async function run(conf, elems) {
...
+            specs: xref.specs
+              ? xref.specs.concat(profiles[xref.profile])
+              : profiles[xref.profile],
+          });
+      }
+      break;
+    default:
+      // show user error - bad configuration
+      Object.assign(normalizedConfig, xrefDefaults);
+      pub("error", `Invalid syntax for cross-references configuration.`);
```
marcoscaceres:
```suggestion
      pub("error", `Invalid value for `xref` configuration option. Got passed: ${xref}`);
```

src/core/xref.js
```diff
@@ -72,6 +84,72 @@ export async function run(conf, elems) {
   addDataCiteToTerms(results, xrefMap, conf);
 }
 
+function invalidProfileError(profile) {
+  pub(
+    "error",
+    `Invalid cross-references profile "${profile}". Please use one of the supported specification profiles: ${Object.keys(
```
marcoscaceres: Note, users generally won't know that "cross-references" means. So always link it back to things they can find and fix.

```suggestion
    `Invalid `xref` profile option passed in the configuration: "${profile}". Please use one of the supported specification profiles: ${Object.keys(
```

src/core/xref.js
```diff
@@ -72,6 +84,72 @@ export async function run(conf, elems) {
...
+              : profiles[xref.profile.toLowerCase()],
+          });
+      }
+      break;
+    default:
+      // show user error - bad configuration
+      Object.assign(normalizedConfig, xrefDefaults);
+      pub(
+        "error",
+        `Invalid value for 'xref' configuration option. Got passed: ${xref}.`
```
marcoscaceres: These are parsed as markdown... sorry I forgot to escape them before: 

```suggestion
        `Invalid value for \`xref\` configuration option. Got passed: ${xref}.`
```

marcoscaceres: updated above.

Swapnilr1:
I am thinking of writing tests for:
1. `xref: ["a", "b", "c"]`
2. `xref: "W3C"`
3. `xref: "InvalidProfile"`
4. `xref: { specs:["a", "b", "c"], profile:"W3C" }`
5. `xref: { profile: "InvalidProfile" }`

These are not even all possible configurations, and yet it feels like too many tests for a small feature. Should I integrate some of these configurations into the existing tests?

marcoscaceres:
> Should I integrate some of these configurations into the existing tests?

Yes, please add them to existing xref-spec.js file... just make a new `description()` for the configuration.

You should also test for:

```
6. xref: 123
```

tests/spec/core/xref-spec.js
```diff
@@ -721,6 +721,53 @@ describe("Core — xref", () => {
...
+      const body = `
+      <section id='abstract'>
+        <p>This is required.</p>
+      </section>
+      <section id='sotd'>
+        <p>This is required.</p>
+      </section>`;
+      const ops = makeStandardOps(config, body);
+      const doc = await makeRSDoc(ops);
+      expect(doc.body.dataset.cite).toContain("HTML INFRA URL PROMISES-GUIDE WEBIDL DOM FETCH");
```
marcoscaceres: Maybe something like: 

```suggestion
      expect(doc.body.dataset.cite.split(" ")).toContain(["HTML", "INFRA", "URL", "PROMISES-GUIDE", "WEBIDL", "DOM", "FETCH");
```

src/core/xref.js
```diff
@@ -11,6 +11,11 @@ import {
   showInlineWarning,
 } from "./utils";
 import { openDB } from "idb";
+import { pub } from "./pubsubhub";
+
+const profiles = {
+  w3c: ["HTML", "INFRA", "URL", "PROMISES-GUIDE", "WEBIDL", "DOM", "FETCH"],
```
marcoscaceres: Promise Guide is being kinda crappy... and I think it's being folded into the WebIDL spec... so let's not include it. 
```suggestion
  w3c: ["HTML", "INFRA", "URL", "WEBIDL", "DOM", "FETCH"],
```

Swapnilr1:
Sorry, for the commit history, will clean it up.
Also, I am not sure how to write tests for wrong configurations. What do I test for?

marcoscaceres:
> Also, I am not sure how to write tests for wrong configurations. What do I test for?

So, for example:

```
xref: 123
```

You would check that is still `body data-cite=""` ... implicit is also that ReSpec doesn't crash.

marcoscaceres:
I.e., think about.... if someone gives me bad data (`123`, `{lols: []}`), how do I recover? And how do I know I've recovered?

marcoscaceres:
Also... what happens is:
```
{xref: "w3c"}
```

But?:

```
 <body data-cite="SVG">
```

(make sure they get merged!)

tests/spec/core/xref-spec.js
marcoscaceres: (you will need to check Jasmine docs for how to do the comparison between arrays..)

Swapnilr1:
> I.e., think about.... if someone gives me bad data (`123`, `{lols: []}`), how do I recover? And how do I know I've recovered?

We are supposed to display a user error and ignore bad data. xref with bad data runs as if it was `xref: true` (BTW, I hope this is intended behavior)
So I could add some `data-cite` to body and see if xref ran successfully?

Also I hope the squashed commits are not a problem - rebased on the wrong branch :sob:

Swapnilr1:
@marcoscaceres @sidvishnoi Sorry if I overlooked this :worried: , but I am not able to find a way to add `data-cite="some specs"` to the body tag inside tests. I checked all functions in `SpecHelper.js` and could not find something that would allow me to add attributes to `body`.

tests/spec/core/xref-spec.js
Swapnilr1: Yeah, I found out how to check an array for a subarray. I will add it along with the wrong profile tests.

devanshbatra04:
what about ```doc.body.dataset.cite = "some specs"```?

sidvishnoi:
Any operation on `doc.body` will be executed after xref and all other plugins have executed, so updating `doc.body.dataset.cite` won't help. We might need to make some changes to `SpecHelper.js`

src/core/xref.js
```diff
@@ -72,6 +84,72 @@ export async function run(conf, elems) {
   addDataCiteToTerms(results, xrefMap, conf);
 }
 
+function invalidProfileError(profile) {
+  pub(
+    "error",
+    `Invalid \`xref\` profile option passed in the configuration: "${profile}". Please use one of the supported specification profiles: ${Object.keys(
```
sidvishnoi: Nit: split this message into multiple lines.

Swapnilr1:
@devanshbatra04 Thanks for your suggestion, but as @sidvishnoi said, we need to do that before `makeRSDoc()` is run. 

@sidvishnoi Can we add `bodyAttrs` analogous to https://github.com/w3c/respec/blob/88a0d537d5759c2f253d30b3e6f095ef34d34f0f/tests/spec/SpecHelper.js#L103

src/core/xref.js
Swapnilr1: `\n` before "Please use one of the..." would work?

sidvishnoi:
>  Can we add bodyAttrs analogous to htmlAttrs

Good idea. Let's go with this.

src/core/xref.js
sidvishnoi: Sorry I wasn't clear. I mean, the line is too long (more than 80 characters), so split it and join multiple lines using `+`.

sidvishnoi:
Do it in `decorateBody` maybe

src/core/xref.js
Swapnilr1: Ah, my bad, you meant too long in the source code, right? Will do.

Swapnilr1:
I will complete adding tests once #2206 gets merged.

Swapnilr1:
I think tests are mostly complete. Checking whether body still has it's original `data-cite` attribute should be enough to ensure that ReSpec did not crash.
I think I don't need to check whether the `data-cite` specs for `body` are actually used by xref for linking?

marcoscaceres:
> I think I don't need to check whether the data-cite specs for body are actually used by xref for linking?

No need.

src/core/xref.js
```diff
@@ -11,6 +11,11 @@ import {
   showInlineWarning,
 } from "./utils";
 import { openDB } from "idb";
+import { pub } from "./pubsubhub";
+
+const profiles = {
+  w3c: ["HTML", "INFRA", "URL", "WEBIDL", "DOM", "FETCH"],
```
marcoscaceres: I decided "w3c" is probably not the right name... "web-platform" is more accurate.

Swapnilr1: OK, I will make the changes. 

Although "web-platform" is a bit vague to me (almost every W3C and WHATWG profile could be included in web platform - so what would other profiles names be in future?).

marcoscaceres: No stress, spec editors have a pretty good idea what constitutes “the platform”. 

There could be: “performance” for performance specs. And “a11y” for Accessibility specs... “css” for css specs and so on.

PR #2189, opened 2019-03-23
feat(core/dfn-abbr): Support definitions which abbreviate

### LINKED ISSUE #2174: Definitions that abbreviate
Given, for instance, 

```HTML
<dfn>primary account number</dfn>
```

It would be nice to able to add `data-abbr`:

```HTML
<dfn data-abbr="PAN">primary account number</dfn>
```

which would then render as:

"primary account number (PAN)"

And PAN would be added to the inline.js abbreviations map. So, then, one could just write:

```HTML
Check the PAN.
``` 

And ReSpec would automatically wrap that in `<abbr title="primary account number">PAN</abbr>`. 

What would be nice also, is for ReSpec to automatically derive the abbreviation if just given:

```HTML
<dfn data-abbr>primary account number</dfn>
``` 

(i.e., the first letter of each word is combined and upper-cased... so "PAN")

Would be equivalent to:

```HTML
<dfn data-abbr="PAN">primary account number</dfn>
```

devanshbatra04:
fixes #2174


devanshbatra04:
After some debugging, I realise this seems to have created a cyclic dependency of sort. 
```core/inline``` needs to run before ```core/dfn``` to handle dfns with non alphanumeric characters. And as per our needs we need to add the dfn abbreviations in ```core/inline``` abbrMap before it is run. I am currently doing that in ```core/dfn```. So that conflicts each other, leading to one module misbehaving in either case.

File: src/core/dfn.js [hunk 1]
```diff
@@ -1,12 +1,28 @@
...
 // - Finds all <dfn> elements and populates conf.definitionMap to identify them.
+// - Also populates abbrMap from core/inlines for definitions that have an abbreviation
 
 import { getDfnTitles } from "./utils";
 import { registerDefinition } from "./dfn-map";
 
 export const name = "core/dfn";
 
+export function addDfnAbbrtoMap(abbrMap) {
+  document.querySelectorAll("dfn").forEach(dfn => {
```
marcoscaceres: This seems backwards, because we are doing this query on line 27. 

How about create a new module called data-abbr.js instead that runs before inlines.js... that module only processes `querySelectorAll("dfn[data-abbr]")` and does the appropriate text replacements. It can own and export `abbrMap`, which inlines.js can then import.

devanshbatra04: Thats exactly what I wanted to do in the first place. But was hesistant in adding a new module. Thanks!

devanshbatra04:
Screenshot from 2019-03-24 16-37-49
I get weird type errors at https://github.com/w3c/respec/pull/2189/files#diff-71b168b6cf6e83b4136dbabab317387eR12

I just can't get to shake it. Some help would be nice. What would be the right casting?

src/core/dfn-abbr.js
```diff
@@ -0,0 +1,38 @@
...
+
+import { registerDefinition } from "./dfn-map";
+
+export const name = "core/dfn-abbr";
+export const abbrMap = new Map();
+
+export function run() {
+  document.querySelectorAll("dfn[data-abbr]").forEach(dfn => {
+    if (dfn.dataset.abbr === "")
+      dfn.dataset.abbr = dfn.textContent
```
sidvishnoi: This should be a function in itself, like `createAbbreviationFromText(textContent)`. Makes clearer what we're doing

devanshbatra04: Done. Thanks!

src/core/dfn-abbr.js
```diff
@@ -0,0 +1,48 @@
...
+ */
+function parseAbbreviatedDefinition(dfn) {
+  // checks if text content is already in the form Permanent Account Number (PAN)
+  const matched = /\((.*?)\)/.exec(dfn.textContent);
+  if (!matched || !(matched.pop() === dfn.dataset.abbr))
+    dfn.textContent = dfn.textContent.concat(` (${dfn.dataset.abbr})`);
+}
+
+export function run() {
+  document.querySelectorAll("dfn").forEach(dfn => {
```
devanshbatra04: If I use ```document.querySelectorAll("dfn[data-abbr]")```, I run into jsDoc issues. Apparently unlike other ```querySelectorAll``` functions,  ```document.querySelectorAll("tag[attr]")``` returns an Element, and not an HTMLElement, which does not have several properties used (like dataset)

devanshbatra04: @marcoscaceres is it fine if we leave it this way?

src/core/utils.js
```diff
@@ -914,3 +914,19 @@ export function msgIdGenerator(namespace, counter = 0) {
...
 }
+
+/**
+ * Generates simple Abbreviations from text passed.
+ *
+ * @param {String} text A string like "Permanent Account Number".
+ * @returns {String} abbr Abbreviation like "PAN"
+ */
+
+export function getAbbreviationFromText(text) {
```
sidvishnoi: How about this?
``` js
return text.match(/\b([a-z])/ig).join('').toUpperCase();
```

src/core/dfn-abbr.js
marcoscaceres: document.querySelectorAll returns a NodeList. What you are describing doesn't sound correct (or is a bug in TS... which would also be odd, given that document.querySelectorAll is used all the time, even in our code base)

marcoscaceres: cc @saschanaz - might know what's going on from a typescript perspective

src/core/utils.js
marcoscaceres: this should probably live in core/dfn-abbr.js, no? or is it used here?

src/core/dfn-abbr.js
saschanaz: You'll want to do this way:

```js
const dfns = document.querySelectorAll("dfn[data-abbr]");
dfns.forEach(dfn => { ... });
```

TS knows `dfn` is an `HTMLElement` but starts to be confused when there is any more selector thing.

devanshbatra04: Interesting. Breaking it up that way works. Thanks :smile:

src/core/utils.js
devanshbatra04: yup. Works. Isn't that smart. :smile:

devanshbatra04: I wasn't sure, just felt like a good Util function. Shifting it back.

src/core/dfn-abbr.js
```diff
@@ -0,0 +1,63 @@
...
+  const matched = /\((.*?)\)/.exec(dfn.textContent);
+  if (!matched || !(matched.pop() === dfn.dataset.abbr))
+    dfn.textContent = dfn.textContent.concat(` (${dfn.dataset.abbr})`);
+}
+
+export function run() {
+  /** @type {NodeListOf<HTMLElement>} */
+  const dfns = document.querySelectorAll("dfn[data-abbr]");
+  dfns.forEach(dfn => {
+    if (!dfn.hasAttribute("data-abbr")) return;
```
marcoscaceres:
```suggestion
```

src/core/dfn-abbr.js
```diff
@@ -0,0 +1,63 @@
...
+
+    if (dfn.dataset.abbr === "")
```
marcoscaceres: Nit: add { } as multiple lines

src/core/dfn-abbr.js
```diff
@@ -0,0 +1,63 @@
...
+    .join("")
+    .toUpperCase();
+}
+
+/**
+ * Parses Elements of form <dfn data-abbr="PAN">Permanent Account Number</dfn>
+ * to <dfn data-abbr="PAN">Permanent Account Number (PAN)</dfn>
+ * @param {HTMLElement} dfn Element to be parsed.
+ */
+function parseAbbreviatedDefinition(dfn) {
```
marcoscaceres:
```suggestion
function renderAbbreviatedDefinition(dfn) {
```

src/core/dfn-abbr.js
```diff
@@ -0,0 +1,63 @@
...
+  // checks if text content is already in the form Permanent Account Number (PAN)
+  const matched = /\((.*?)\)/.exec(dfn.textContent);
```
marcoscaceres: I don’t think we need this. The author ops into this behavior by explicitly using data-abbr.

src/core/dfn-abbr.js
```diff
@@ -0,0 +1,63 @@
...
+  if (!matched || !(matched.pop() === dfn.dataset.abbr))
+    dfn.textContent = dfn.textContent.concat(` (${dfn.dataset.abbr})`);
```
marcoscaceres: Does the abbreviation inside () end up getting marked up in inlines.js with `<abbr title=“full text”>`? It’s unclear from the tests - but it should be.

src/core/dfn-abbr.js
devanshbatra04: The author may do something like 
```<dfn data-abbr>Permanent Account Number (PAN)</dfn>``` 
or something like ```<dfn data-abbr>Statement of Purpose (SoP)</dfn>```  our another where our logic could get it wrong. He would still want that to be ```<abbr>```

File: src/core/dfn-abbr.js (hunk 9 above)
Comment (devanshbatra04) [reply src/core/dfn-abbr.js]: yes. I have added the tests. :-)

File: src/core/dfn-abbr.js (hunk 8 above)
Comment (marcoscaceres) [reply src/core/dfn-abbr.js]: Sure, but she could have just done:

> `<dfn>Statement of Purpose (<abbr title="Statement of Purpose">SoP</abbr>)</dfn>` 

As this is a new feature, we should refrain from dealing with a case that literally doesn’t exist.

Comment (devanshbatra04) [reply src/core/dfn-abbr.js]: guess what. ```<dfn data-abbr="PoR">Position of Responsibility</dfn>``` works just as good. Correcting.

Comment (devanshbatra04) [reply src/core/dfn-abbr.js]: Corrected. Should be good to go now.

File: tests/spec/core/dfn-abbr-spec.js [hunk 10]
```diff
@@ -0,0 +1,89 @@
...
+    expect(dfnUS.dataset.abbr).toEqual("US");
+    expect(dfnPoR.dataset.abbr).toEqual("PoR");
+    expect(dfnUI.dataset.abbr).toEqual("UI");
+    expect(dfnUS.textContent.trim()).toEqual("United States (US)");
+    expect(dfnPoR.textContent.trim()).toEqual(
+      "Position of Responsibility (PoR)"
+    );
+    expect(dfnUI.textContent.trim()).toEqual("User Interface (UI)");
+
+    for (const x of Array(3).keys()) {
```
marcoscaceres: Maybe do: 
```
    divs.forEach((div, i) => {
    
    }
```

src/core/dfn-abbr.js
```diff
@@ -0,0 +1,60 @@
...
+}
+
+/**
+ * Parses Elements of form <dfn data-abbr="PAN">Permanent Account Number</dfn>
+ * to <dfn data-abbr="PAN">Permanent Account Number (PAN)</dfn>
+ * @param {HTMLElement} dfn Element to be parsed.
+ */
+function renderAbbreviatedDefinition(dfn) {
+  // checks if text content is already in the form Permanent Account Number (PAN)
+  dfn.textContent = dfn.textContent.concat(` (${dfn.dataset.abbr})`);
```
marcoscaceres: This might be overly destructive. Consider: 

```HTML
<dfn data-abbr>The <span>Funky</span> Bar</dfn>
```

Replacing the `textContent` could trash the `span`. The ` (${dfn.dataset.abbr})` should probably just be an appended TextNode

tests/spec/core/dfn-abbr-spec.js
```diff
@@ -0,0 +1,89 @@
...
+      expect(dfns[x].dataset.abbr).toEqual(
+        divs[x].getElementsByTagName("abbr")[0].textContent
+      );
+      expect(
+        dfns[x].textContent
```
marcoscaceres: please put this into a variable. It will make it a bit easier to debug in the future if there is a regression.

tests/spec/core/dfn-abbr-spec.js
```diff
@@ -0,0 +1,89 @@
...
+        <dfn data-abbr>United States</dfn>
+        <dfn data-abbr="PoR">Position of Responsibility</dfn>
+        <dfn data-abbr="UI">User Interface</dfn>
+        <div>US</div>
+        <div>PoR</div>
+        <div>UI</div>
+      </section>`,
+    };
+    const doc = await makeRSDoc(ops);
+    const divs = doc.querySelectorAll("#section div");
```
marcoscaceres:
```suggestion
    const divs = [...doc.querySelectorAll("#section div")];
```

tests/spec/core/dfn-abbr-spec.js
```diff
@@ -0,0 +1,89 @@
...
+    const dfns = doc.querySelectorAll("#section dfn");
```
marcoscaceres:
```suggestion
    const dfns = [...doc.querySelectorAll("#section dfn")];
```

tests/spec/core/dfn-abbr-spec.js
```diff
@@ -0,0 +1,89 @@
...
+
+    const dfnUS = dfns[0];
```
marcoscaceres: Cleaner as: 
```
 const [dfnUS, dfnPoR, dfnUI]  = dfns;
```

tests/spec/core/dfn-abbr-spec.js
saschanaz: Why? We are not using any Array-specific functions here.

marcoscaceres: Oh true. I forgot NodeList has .forEach(). I guess only for destructoring below.

src/core/dfn-abbr.js
```diff
@@ -0,0 +1,59 @@
...
+export const name = "core/dfn-abbr";
+export const abbrMap = new Map();
+
+/**
+ * Generates simple Abbreviations from text passed.
+ *
+ * @param {String} text A string like "Permanent Account Number".
+ * @returns {String} abbr Abbreviation like "PAN"
+ */
+
```
saschanaz: A redundant newline.

```suggestion
```

src/core/dfn-abbr.js
saschanaz:
```suggestion
  dfn.append(` (${dfn.dataset.abbr})`);
```

saschanaz: I think we can just inline this anyway.

marcoscaceres: Agree.

tests/spec/core/dfn-abbr-spec.js
saschanaz: NodeList already supports destructuring so I think `dfns` neither have to be an array.

tests/spec/core/dfn-abbr-spec.js
```diff
@@ -0,0 +1,87 @@
...
+        <div>PoR</div>
+        <div>UI</div>
+      </section>`,
+    };
+    const doc = await makeRSDoc(ops);
+    const divs = doc.querySelectorAll("#section div");
+    const dfns = [...doc.querySelectorAll("#section dfn")];
+    const [dfnUS, dfnPoR, dfnUI] = dfns;
+
+    expect(dfnUS.dataset.abbr).toEqual("US");
```
saschanaz: nit: Prefer `toBe()` for string checks.

tests/spec/core/dfn-abbr-spec.js
```diff
@@ -0,0 +1,87 @@
...
+        <dfn data-abbr="PoR">Position of Responsibility</dfn>
+        <dfn data-abbr="UI">User Interface</dfn>
+        <div>US</div>
+        <div>PoR</div>
+        <div>UI</div>
+      </section>`,
+    };
+    const doc = await makeRSDoc(ops);
+    const divs = doc.querySelectorAll("#section div");
+    const dfns = [...doc.querySelectorAll("#section dfn")];
```
marcoscaceres:
```suggestion
    const dfns = doc.querySelectorAll("#section dfn");
```

tests/spec/core/dfn-abbr-spec.js
marcoscaceres: TIL! Thanks @saschanaz.

src/core/dfn-abbr.js
```diff
@@ -0,0 +1,49 @@
...
+export const name = "core/dfn-abbr";
+export const abbrMap = new Map();
+
+/**
+ * Generates simple Abbreviations from text passed.
+ *
+ * @param {String} text A string like "Permanent Account Number".
+ * @returns {String} abbr Abbreviation like "PAN"
+ */
+function getAbbreviationFromText(text) {
```
marcoscaceres: This can be inlined too.

devanshbatra04: @sidvishnoi suggests otherwise. 
https://github.com/w3c/respec/pull/2189#discussion_r268429359

marcoscaceres: I don't mind... but I think we can communicate that just as effectively with a code comment there.

devanshbatra04: Inlined. :smiley:

marcoscaceres:
@saschanaz, you want to give this one last look over?

saschanaz:
Looking over...

src/core/dfn-abbr.js
```diff
@@ -0,0 +1,41 @@
...
+      // A string like "Permanent Account Number" returns an abbreviation like "PAN"
+      dfn.dataset.abbr = dfn.textContent
+        .match(/\b([a-z])/gi)
+        .join("")
+        .toUpperCase();
+    }
+
+    dfn.append(` (${dfn.dataset.abbr})`);
+
+    const fullForm = dfn.textContent
```
saschanaz: Maybe:

```js
const fullForm = dfn.textContent.trim();
dfn.append(` (${dfn.dataset.abbr})`);
```

and then we won't have to substr.

tests/spec/core/dfn-abbr-spec.js
```diff
@@ -0,0 +1,85 @@
...
+        <dfn data-abbr>foo bar</dfn> can be referenced as
+        <span id="foobarLink">
+          <a id="dfnlink">FB</a>
+        </span>
+      </section>`,
+    };
+    const doc = await makeRSDoc(ops);
+
+    const dfnFooBar = doc.querySelector("#section dfn");
+    expect(dfnFooBar.id).toEqual("dfn-fb");
```
saschanaz: nit: toBe for string checks

src/core/dfn-abbr.js
devanshbatra04: Done. Hope it's good now. :smiley_cat:

devanshbatra04:
Hope it's good to merge now. :)

marcoscaceres:
Looking forward to using it. Before I forget, we need to document it in the wiki.

marcoscaceres:
Added [wiki page](https://github.com/w3c/respec/wiki/data-abbr-(dfn-element-only)). @devanshbatra04 could you please check it.

devanshbatra04:
Looks good to me!
Sorry for delayed reply.

devanshbatra04:
The page seems to have disappeared.
Also it would help if we added that ```<dfn data-abbr>User Interface</dfn>``` can be referenced as ```<a>UI</a>```, ```<a>User Interface</a>``` as well as ```<a>User Interface (UI)</a>```

marcoscaceres:
Ah, I renamed the page.

marcoscaceres:
Now here https://github.com/w3c/respec/wiki/data-abbr)

PR #2191, opened 2019-03-25
chore: Bumps idb To V4

devanshbatra04:
WIP
should allow #2154 to be merged


marcoscaceres:
@devanshbatra04, be mindful that even time you push it generates an email to all maintainers. Make sure you are testing things locally before and only push things up as needed.

devanshbatra04:
Didn't know that. Thanks.

saschanaz:
>generates an email to all maintainers

@marcoscaceres Are you watching the full repo here? That... would be a fair amount of emails.

marcoscaceres:
> @marcoscaceres Are you watching the full repo here? That... would be a fair amount of emails.

📧 👀📧 👀📧 👀📧 👀

saschanaz:
BTW, try [drafts](https://github.blog/2019-02-14-introducing-draft-pull-requests/) to open WIP things.

devanshbatra04:
Apparently, idb module now requires a global scope. I saw that there was a similar case with ```shortcut.js```.
I will try following the webpack config path and see how it works too. 
https://github.com/w3c/respec/blob/develop/tools/builder.js#L139
Referencing the window definitely isn't right.

saschanaz:
>Apparently, idb module now requires a global scope.

I don't see such change on [the changelog](https://github.com/jakearchibald/idb/blob/master/changes.md#breaking-changes). Requiring global scope looks like a bug.

devanshbatra04:
```Import "idb"``` works, it makes idb object available in Window scope. 
Anything like ```Import idb from "idb"``` doesn't. It says its undefined. 
The file being included now looks like 
```javascript
var idb = function(e) {
...
} ({ })
```
It was previously a normal iife function.
Pretty sure its global scope now

saschanaz:
Oh, it's because you are using iife build. And... I found it doesn't support AMD module anymore. 😱

devanshbatra04:
The ES module build wouldn't work with requireJS.
I think https://webpack.js.org/loaders/exports-loader/ might just be the way to do it. Hacking on it.

saschanaz:
Webpack specific thing shouldn’t be used as we are not going to ditch requirejs just yet…

devanshbatra04:
How about shims? 
https://stackoverflow.com/questions/15471088/requirejs-why-and-when-to-use-shim-config

devanshbatra04:
My bad. sorry.

tests/spec/core/xref-spec.js
```diff
@@ -4,14 +4,15 @@ describe("Core — xref", () => {
 
   let cache;
   beforeAll(async () => {
-    const { openDb } = await new Promise(resolve =>
-      require(["deps/idb"], resolve)
-    );
+    await new Promise(resolve => require(["deps/idb"], resolve));
+    const { openDB } = window.idb;
```
devanshbatra04: I want to use some sort of shim here, so that I don't need to reference the window object. Just couldn't find the right documentation.

marcoscaceres: you are only doing this in the tests, right?

devanshbatra04: yes. Only in tests. I tried adding the shim to ```test-main.js```. doesn't help

PR #2135, opened 2019-03-03
refactor(core/figures): localize module

ThobyV:
This P.R aims to apply proper changes requested to resolve issue https://github.com/w3c/respec/issues/2084 for the module `figures.js`. Thanks to all who review and @marcoscaceres for assistance and useful git commands.

PR #2192, opened 2019-03-25
chore: migrate Geonovum profile

thijsbrentjens:
Initial setup for migration of the Geonovum profile to the W3C repo. @marcoscaceres could you please let me know if this workflow is okay to you? Or what we could do better.


src/geonovum/informative.js
```diff
@@ -0,0 +1,13 @@
+// Module w3c/informative
```
marcoscaceres: Let's fix this on the W3C side... we can move this module to be "core" and localize it appropriately.

js/profile-geonovum.js
```diff
@@ -0,0 +1,101 @@
...
+  "./core/inlines",
+  "./w3c/conformance",
+  "./core/dfn",
+  "./core/pluralize",
+  "./core/examples",
+  "./core/issues-notes",
+  "./core/requirements",
+  "./core/best-practices",
+  "./core/figures",
+  "./core/webidl",
```
marcoscaceres: Not sure if you really want this one at geonovum? It's for browser APIs so you probably don't need it.

js/profile-geonovum.js
```diff
@@ -0,0 +1,101 @@
...
+});
+
+define([
+  // order is significant
+  "./core/base-runner",
+  "./core/ui",
+  "./core/reindent",
+  "./core/location-hash",
+  "./core/l10n",
+  "./w3c/defaults",
```
marcoscaceres: We probably don't want this one... we should make a geonovum/defaults.js instead.

js/profile-geonovum.js
```diff
@@ -0,0 +1,101 @@
...
+  "./core/style",
+  "./w3c/style",
```
marcoscaceres: We probably don't want this one either.

js/profile-geonovum.js
```diff
@@ -0,0 +1,101 @@
...
+  "./w3c/l10n",
```
marcoscaceres: Same here... we should check what's usable from this one...

js/profile-geonovum.js
```diff
@@ -0,0 +1,101 @@
...
+  "./core/github",
+  "./core/data-include",
+  "./core/markdown",
+  "./w3c/headers",
```
marcoscaceres: You might not need these two.

js/profile-geonovum.js
```diff
@@ -0,0 +1,101 @@
...
+  "./w3c/abstract",
+  "./core/data-transform",
+  "./core/inlines",
+  "./w3c/conformance",
```
marcoscaceres: Same with this one... or other W3C ones below...

src/geonovum/informative.js
```diff
@@ -0,0 +1,13 @@
+// Module geonovum/informative
```
marcoscaceres: See my previous comment about us making this a core module instead. We can then add the localization for English and Dutch.

js/profile-geonovum.js
```diff
@@ -0,0 +1,101 @@
+"use strict";
+// In case everything else fails, we want the error
+window.addEventListener("error", ev => {
+  console.error(ev.error, ev.message, ev);
+});
+
+// this is only set in a build, not at all in the dev environment
+require.config({
```
marcoscaceres: Take a look at the current w3c profile about this configuration... we've managed to reduce it quite a bit in the w3c profile, so you might not need most of this.

src/geonovum/informative.js
thijsbrentjens: That would be perfect. I choose this one because of simplicity.

js/profile-geonovum.js
thijsbrentjens: Ah, yes, I just took the W3C profile, but didn't do a check on the modules yet. We don't need WebIDL indeed. Will remove it.

js/profile-geonovum.js
thijsbrentjens: We have a geonovum/defaults.js already. For now, to make each PR small, should I just remove the ./w3c/defaults? I could also try to replace it directly with the geonovum defaults, but a following PR will be a bit more complicated if we do this. But it makes sense to set it up correctly first. Also with the styles etc

js/profile-geonovum.js
marcoscaceres: Ooops! sorry, I was wrong.

js/profile-geonovum.js
thijsbrentjens: Agreed

js/profile-geonovum.js
```diff
@@ -0,0 +1,101 @@
...
+  paths: {
+    clipboard: "deps/clipboard",
+    hyperhtml: "deps/hyperhtml",
+    idb: "deps/idb",
+    jquery: "deps/jquery",
+    marked: "deps/marked",
+    pluralize: "deps/pluralize",
+    text: "deps/text",
+    webidl2: "deps/webidl2",
```
marcoscaceres: WebIDL can go from here too.

js/profile-geonovum.js
thijsbrentjens: Which two modules do you mean? We have a headers module, so we should use that instead of w3c's. Markdown is used for several of our docs though.

js/profile-geonovum.js
thijsbrentjens: Agreed

src/geonovum/informative.js
thijsbrentjens: Agreed.

js/profile-geonovum.js
thijsbrentjens: I took the w3c profile (from develop) as an example. I could check what we need from it.

js/profile-geonovum.js
thijsbrentjens: Agreed

thijsbrentjens:
My conclusion: the workflow is fine. But to get a good start (a clean build) I think I need to at least remove w3c modules and the core modules we don't need. I'll try to do that first.

js/profile-geonovum.js
marcoscaceres: Yeah, let's remove it for now. 
```suggestion

```

js/profile-geonovum.js
marcoscaceres: Sorry, I meant  "./w3c/headers", and "./w3c/abstract",

thijsbrentjens: Ah, makes sense :). We've got geonovum versions for both.

src/geonovum/informative.js
thijsbrentjens: Regarding this module: how would you like to proceed? Should I for now keep this geonovum module and as soon as we have a core module, change it?

marcoscaceres: ok, as part of this PR, let's update `core/informative.js`. Could you delete this file. Then we just need to copy the structure and test from:

https://github.com/w3c/respec/pull/2135/files

marcoscaceres: Could you do the the changes to core/informative.js ? I'll do the test, as the test file doesn't exist yet.

marcoscaceres: >  Should I for now keep this geonovum module and as soon as we have a core module, change it?

Let's fix things as we go. That way we eliminate duplication and reduce technical debt.

marcoscaceres:
> My conclusion: the workflow is fine. But to get a good start (a clean build) I think I need to at least remove w3c modules and the core modules we don't need. I'll try to do that first.

sounds like a plan.

src/geonovum/informative.js
thijsbrentjens: Okay. And I'll do the changes to core/informative.js (better: give it a try )

js/profile-geonovum.js
```diff
@@ -0,0 +1,101 @@
...
+  "./core/webidl",
+  "./core/data-cite",
+  "./core/biblio",
+  "./core/webidl-index",
+  "./core/link-to-dfn",
+  "./core/render-biblio",
+  "./core/contrib",
+  "./core/fix-headers",
+  "./core/structure",
+  "./geonovum/informative",
```
marcoscaceres:
```suggestion
  "./core/informative",
```

thijsbrentjens:
@marcoscaceres because we have quite some changes, I'd prefer to create a new PR (it's hard for me to keep the overview now). Is that okay for you?

src/geonovum/linter-rules/privsec-section.js
```diff
@@ -0,0 +1,47 @@
+// @ts-check
+/**
```
marcoscaceres: This file can probably be removed. It looks the same as the w3c one.

src/geonovum/defaults.js
```diff
@@ -0,0 +1,85 @@
+/**
+ * Sets the defaults for Geonovum documents
+ */
+export const name = "geonovum/defaults";
+import { coreDefaults } from "../core/defaults";
+import linter from "../core/linter";
+import { rule as privsecSectionRule } from "./linter-rules/privsec-section";
```
marcoscaceres: Filed https://github.com/w3c/respec/issues/2223

marcoscaceres: now fixed.

marcoscaceres:
Move this here: https://github.com/w3c/respec/pull/2247

PR #2194, opened 2019-03-25
refactor(core/informative): move W3C informative module to core incl l10n

thijsbrentjens:
As a result of discussion in https://github.com/w3c/respec/pull/2192, move the W3C informative module to core and add localization. For now only nl added to en.
 
I tried to follow your conventions, but please let me know if things need improvement.


marcoscaceres:
Looks great. I'll set up test now.

marcoscaceres:
Actually, will have to be tomorrow for the test. But looking good.

thijsbrentjens:
Thanks. And no problem of course, you've already responded very fast.

saschanaz:
That GitHub thing again fails? 🤔

marcoscaceres:
Yep, we need #2197.

saschanaz:
Last time we didn't change any tests. I think this may mean we are creating a more frequent API calls...

marcoscaceres:
> Last time we didn't change any tests. I think this may mean we are creating more frequent API calls...

Because Travis' IP are all shared, a single call to GitHub can push it over the 60 calls limit (as someone else on the server might have consumed all the available calls)... in theory, once  #2197 goes in, it should not fail anymore... but if it does, then the original patch is doing bad things by making redundant network calls to github.

PR #2197, opened 2019-03-26
Remove Hardcoded GitHub URL

### LINKED ISSUE #2195: Remove hardcoded github url
The Github module contains a hard-coded URL, meaning that it’s pinging GitHub during testing. 
https://api.github.com/rate_limit

This should be dynamically set via the appropriate configuration option instead so that it works with our testing server.

devanshbatra04:
closes #2195

### DIFF
--- a/src/core/github-api.js
+++ b/src/core/github-api.js
@@ -21,8 +21,11 @@ export function githubRequestHeaders(conf) {
 }
 
 export async function getRateLimit(conf) {
+  const githubRateLimitAPI = conf.githubRateLimitAPI
+    ? conf.githubRateLimitAPI
+    : "https://api.github.com/rate_limit";
   const headers = githubRequestHeaders(conf);
-  const request = new Request(`https://api.github.com/rate_limit`, {
+  const request = new Request(githubRateLimitAPI, {
     mode: "cors",
     referrerPolicy: "no-referrer",
     headers,
--- a/tests/data/ratelimit.json
+++ b/tests/data/ratelimit.json
@@ -1,0 +1,29 @@
+{
+  "resources": {
+    "core": {
+      "limit": 60,
+      "remaining": 60,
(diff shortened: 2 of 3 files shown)


src/core/github-api.js
```diff
@@ -21,8 +21,11 @@ export function githubRequestHeaders(conf) {
 }
 
 export async function getRateLimit(conf) {
+  const githubRateLimitAPI = conf.githubRateLimitAPI
+    ? conf.githubRateLimitAPI
```
marcoscaceres: Please use `conf.githubAPI` instead. The URL should always be:

```JS
new URL("./rate_limit", githubAPI).href
```

So you will need to rename `ratelimit.json` to `rate_limit`.

devanshbatra04: The githubAPI currently defaults to https://api.github.com/repos/${org}/${repo}
I will make convert it to https://api.github.com/repos/ and make the required changes in all other files.

marcoscaceres: Ok, as long as it works that same, then go for it.

marcoscaceres: @devanshbatra04 ... could you check something for me? when you make a `Request` to Github API, I think the Response's headers actually say how many remaining requests there are (`X-Rate-Limit:` or something)... if that's true, maybe we don't need getRateLimit() at all?

devanshbatra04: Thats right. I wanted to take that path earlier. There's a blocker though. I believe we would need the number of requests before we make the multiple requests required. The rate limit api doesnt count to a decrement in requests.

marcoscaceres: Ok, let's fix this first, as it's blocking us from merging other things as travis keeps falling over. Then we can work out a better strategy...

marcoscaceres:
Closed by  #2200

PR #2199, opened 2019-03-26
fix(w3c/headers): noRecTrack text not showing up

marcoscaceres:
Moved some of the stuff around in src/w3c/templates/sotd.js because otherwise it's quite difficult to debug. 

We should eventually move all of them logic into smaller render functions, otherwise the template is too hard to use or maintain.


saschanaz:
Maybe a test for this?

marcoscaceres:
> Looks 👍 but seems we need a test for this.

Will add one.

tests/spec/w3c/headers-spec.js
```diff
@@ -1283,6 +1283,22 @@ describe("W3C — Headers", () => {
...
+      status: "ED",
+    };
+    Object.assign(ops.config, newProps);
+    const doc = await makeRSDoc(ops);
+    const sotdText = doc
+      .getElementById("sotd")
+      .textContent.replace(/\s+/gm, " ");
+    const expectedString =
+      "does not expect this document to become a W3C Recommendation";
+    expect(sotdText).toMatch(expectedString);
```
saschanaz: `toContain()`?

marcoscaceres: yeah... copy/🍝:)

PR #2200, opened 2019-03-26
Use Remaining Requests from GitHub Response Headers

devanshbatra04:
I came up with this strategy to use the request headers instead of a new endpoint https://github.com/w3c/respec/pull/2197#discussion_r268945469

We don't need to merge that PR if this is good.

### DIFF
--- a/src/core/contrib.js
+++ b/src/core/contrib.js
@@ -32,7 +32,7 @@ async function toHTML(urls, editors, element, headers) {
     urls
       .map(url =>
         fetch(new Request(url, { headers })).then(r => {
-          if (checkLimitReached(r)) return;
+          if (checkLimitReached(r)) return null;
           return r.json();
         })
       )
--- a/src/core/contrib.js
+++ b/src/core/contrib.js
@@ -4,7 +4,7 @@
 // #gh-commenters: people having contributed comments to issues.
 // #gh-contributors: people whose PR have been merged.
 // Spec editors get filtered out automatically.
-import { fetchIndex, getRateLimit, githubRequestHeaders } from "./github-api";
+import { fetchIndex, checkLimitReached, githubRequestHeaders } from "./github-api";
 import { flatten, joinAnd } from "./utils";
 import { pub } from "./pubsubhub";
 export const name = "core/contrib";
@@ -25,7 +25,10 @@ function findUserURLs(...thingsWithUsers) {
 
 async function toHTML(urls, editors, element, headers) {
   const args = await Promise.all(
-    urls.map(url => fetch(new Request(url, { headers })).then(r => r.json()))
+    urls.map(url => fetch(new Request(url, { headers })).then(r => {
+      checkLimitReached(r);
+      return r.json()
+    }))
   );
   const names = args
     .map(user => user.name || user.login)
@@ -69,14 +72,14 @@ export async function run(conf) {
   const editors = conf.editors.map(nameProp);
   const commenterUrls = ghCommenters ? findUserURLs(issues, comments) : [];
   const contributorUrls = ghContributors ? contributors.map(urlProp) : [];
-  const remainingRequests = await getRateLimit(conf);
-  if (commenterUrls.length + contributorUrls.length > remainingRequests) {
-    const msg =
-      `Your GitHub Repository contains ${commenterUrls.length +
-        contributorUrls.length} contributors and commenters` +
-      `but your current GitHub quota only allows ${remainingRequests} more requests. Some contributors will not show up`;
-    pub("warning", msg);
-  }
+  // const remainingRequests = await getRateLimit(conf);
+  // if (commenterUrls.length + contributorUrls.length > remainingRequests) {
+  //   const msg =
+  //     `Your GitHub Repository contains ${commenterUrls.length +
+  //       contributorUrls.length} contributors and commenters` +
+  //     `but your current GitHub quota only allows ${remainingRequests} more requests. Some contributors will not show up`;
+  //   pub("warning", msg);
+  // }
   try {
     await Promise.all(
       toHTML(commenterUrls, editors, ghCommenters, headers),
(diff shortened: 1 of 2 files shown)


devanshbatra04:
@marcoscaceres hope this helps

src/core/github-api.js
```diff
@@ -20,32 +20,28 @@ export function githubRequestHeaders(conf) {
...
-  const headers = githubRequestHeaders(conf);
-  const request = new Request(`https://api.github.com/rate_limit`, {
-    mode: "cors",
-    referrerPolicy: "no-referrer",
-    headers,
-  });
-  const responseJSON = await fetch(request).then(r => r.json());
-  return responseJSON.rate.remaining;
+export function checkLimitReached(response) {
+  if (response.ok) return false;
```
devanshbatra04: Our current stubs would definitely satisfy this, so no new stubs required :D

src/core/github-api.js
```diff
@@ -20,32 +20,28 @@ export function githubRequestHeaders(conf) {
...
+  if (status === 403 && headers.get("X-RateLimit-Remaining") === "0") {
+    if (typeof checkLimitReached.warned === "undefined") {
+      checkLimitReached.warned = true;
+      const msg = `You have run out of github requests. Some github assets will not show up.`;
+      pub("warning", msg);
+      window.pub = pub;
+    }
+    return true;
+  }
+  // Not sure we can reach here.
```
marcoscaceres:
```suggestion
  return true;
```

src/core/github-api.js
```diff
@@ -20,32 +20,28 @@ export function githubRequestHeaders(conf) {
...
+  if (response.ok) return false;
+  const { headers, status } = response;
+  if (status === 403 && headers.get("X-RateLimit-Remaining") === "0") {
+    if (typeof checkLimitReached.warned === "undefined") {
+      checkLimitReached.warned = true;
+      const msg = `You have run out of github requests. Some github assets will not show up.`;
+      pub("warning", msg);
+      window.pub = pub;
+    }
+    return true;
```
marcoscaceres:
```suggestion

```

src/core/contrib.js
```diff
@@ -25,7 +29,14 @@ function findUserURLs(...thingsWithUsers) {
 
 async function toHTML(urls, editors, element, headers) {
   const args = await Promise.all(
-    urls.map(url => fetch(new Request(url, { headers })).then(r => r.json()))
+    urls
+      .map(url =>
+        fetch(new Request(url, { headers })).then(r => {
+          if (checkLimitReached(r)) return;
```
marcoscaceres: Worried that returning undefined here will break things. Make sure you check and return `null` or `{}` so whatever is expecting a result doesn't crash.

marcoscaceres:
Remember to prioritize https://github.com/w3c/respec/pull/2197 or I'll have to back everything out :( It's blocking all other merges.

devanshbatra04:
```
Core — xref
    ✔ uses inline references to provide context
    ✖ uses data-cite to disambiguate (server side)
    ✔ shows error if external term doesn't exist
    ✔ uses data-cite to disambiguate (client side)
    ✔ shows error if cannot resolve by data-cite
    ✔ ignores terms if local dfn exists
    ✔ treats terms as local if empty data-cite on parent
    inline IDL references
      ✔ ignores inlines starting with backslash
      ✔ ignores malformed syntax
      ✔ links inline IDL references
      ✔ links dictionary members
      ✔ links methods
      ✔ links local definitions first
      ✔ links attributes
      ✔ links internalSlots
    ✖ caches results and uses cached results when available
    ✔ does nothing if xref is not enabled
    ✔ takes dfn pluralization into account
    ✔ doesn't link auto-filled anchors
    ✔ adds normative and informative references
    ✔ adds link to unique external terms
    ✔ takes data-lt into account
```
These failing tests are related to #2191 and not this PR

marcoscaceres:
Ah, server is currently down... doing some maintenance

marcoscaceres:
Back up... should be ok now.

devanshbatra04:
We can only have one of #2197 or this. What would you prefer?

marcoscaceres:
Happy to go with this one 👍 Let's see what travis says.

src/core/github-api.js
devanshbatra04: if the response code is neither 200 nor 403. 
(Means if we neither get the correct response, nor are we out of requests) the function would in this case return true. 
Is this okay?

devanshbatra04: I think I should add a warning for such cases (when !response.ok) so the viewer knows something is not right.

devanshbatra04:
```javascript
if (response.ok) return false;
else pub("warning", "unexpected response received from github");
  const { headers, status } = response;
  if (status === 403 && headers.get("X-RateLimit-Remaining") === "0") {
    if (typeof checkLimitReached.warned === "undefined") {
      checkLimitReached.warned = true;
      const msg = `You have run out of github requests. Some github assets will not show up.`;
      pub("warning", msg);
      window.pub = pub;
    }
   return true;
  }
return false;
```
This feels like the correct thing to do

marcoscaceres: Ah, right, ok... yeah, would be good to warn and show the HTTP error code.

marcoscaceres: > pub("warning", "unexpected response received from github");

Yep, see other HTTP errors we emit. Can just copy those.

devanshbatra04: Don't need to warn. It's being done by both core/contrib (for contributors) and core/github-api for issues and comments.

marcoscaceres: Excellent, thanks for checking!

src/core/github-api.js
devanshbatra04: Slightly changed the code so it stays here.

devanshbatra04:
I ran the failing tests on my local without internet. Works like a charm. :)

marcoscaceres:
Excellent, thank for confirming that!

marcoscaceres:
I think we will have to code name this release "baby groot" :)

marcoscaceres:
@devanshbatra04 all good to go?

devanshbatra04:
Yup!

PR #2203, opened 2019-03-27
refactor(core/override-configuration): clean comments

marcoscaceres:
Fixed up/added some comments

### DIFF
--- a/src/core/override-configuration.js
+++ b/src/core/override-configuration.js
@@ -9,7 +9,7 @@ import { pub, sub } from "./pubsubhub";
 export const name = "core/override-configuration";
 
 function overrideConfig(config) {
-  // For legacy reasons, we still support ";" instead of "&"
+  // For legacy reasons, we still support both ";" and "&"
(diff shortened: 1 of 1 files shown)


src/core/override-configuration.js
```diff
@@ -4,14 +4,12 @@
...
 //      ?specStatus=LC&publishDate=2012-03-15
-// Note that fields are separated by semicolons and not ampersands.
-// TODO
-//  There could probably be a UI for this to make it even simpler.
 import { pub, sub } from "./pubsubhub";
 
 export const name = "core/override-configuration";
 
 function overrideConfig(config) {
+  // For legacy reasons, we still support both ";" and "&"
```
saschanaz: Do those reasons still exist?

marcoscaceres: yes 😭 There are a ton of these little files in various repos:
https://github.com/w3c/payment-request/blob/gh-pages/echidna-manifest.txt#L1

notice the ";" there. They are passed to ReSpec.

PR #2204, opened 2019-03-27
fix(core/data-abbr): generalize implementation

marcoscaceres:
Started using this today and realized we'd missed a couple of things. Also, this should be generalized to just `data-abbr`.


src/core/data-abbr.js
```diff
@@ -0,0 +1,43 @@
+// @ts-check
+// Module core/dfn
```
sidvishnoi:
```suggestion
// Module core/data-abbr
```

src/core/data-abbr.js
```diff
@@ -0,0 +1,43 @@
...
+  });
+}
+/**
+ * @param {HTMLElement} dfn
+ */
+function processDfnElement(dfn) {
+  if (dfn.dataset.abbr === "") {
+    // Generates abbreviation from textContent
+    // e.g., "Permanent Account Number" -> "PAN"
+    dfn.dataset.abbr = dfn.textContent
```
sidvishnoi: Do we need this assignment to `dfn.dataset.abbr`? Maybe we can do
``` js
const abbr = dfn.dataset.abbr || createAbbrFromText(fullForm);
```

src/core/data-abbr.js
```diff
@@ -0,0 +1,43 @@
...
+// - Finds all elements with data-abbr attribute and processes them.
+import { showInlineWarning } from "./utils";
+export const name = "core/dfn-abbr";
+
+export function run() {
+  document.querySelectorAll("[data-abbr]").forEach(elem => {
+    const { localName } = elem;
+    switch (localName) {
+      case "dfn":
+        // @ts-ignore
```
sidvishnoi: Why `@ts-ignore` is needed?

src/core/data-abbr.js
```diff
@@ -0,0 +1,43 @@
...
+      .join("")
+      .toUpperCase();
+  }
+  const fullForm = dfn.textContent.trim();
+  const { abbr } = dfn.dataset;
+  dfn.insertAdjacentHTML(
+    "afterend",
+    ` (<abbr title="${fullForm}">${abbr}</abbr>)`
+  );
+  dfn.dataset.lt = dfn.dataset.lt ? `${dfn.dataset.lt}|${abbr}` : abbr;
```
devanshbatra04: This means we cannot link to the definition using ```<a class="dfnlink">foo bar (FB)</a>```.
Although not a huge deal, I think it might be counter intuitive as visually the definition after rendering reads foo bar (FB). The user may want to refer it as such.

tests/testFiles.json
```diff
@@ -4,16 +4,20 @@
   "spec/core/biblio-db-spec.js",
   "spec/core/biblio-spec.js",
   "spec/core/caniuse-spec.js",
+  "spec/core/data-abbr-spec.js",
```
devanshbatra04: Oops. Didn't know this.

tests/spec/core/data-abbr-spec.js
```diff
@@ -1,57 +1,48 @@
...
+      .querySelectorAll(".dfnlink")
+      .forEach(link => expect(link.getAttribute("href")).toBe("#dfn-fb"));
   });
   it("correctly parses and abbreviates different Abbreviation syntax", async () => {
     const ops = {
       config: makeBasicConfig(),
       body: `
       <section id="section">
-        <dfn data-abbr>United States</dfn>
+        <dfn data-abbr="">United States</dfn>
```
devanshbatra04: Is there a difference? The issue originally asked for the the former way. (```<dfn data-abbr>```)

src/core/data-abbr.js
devanshbatra04: Is it something related to ```HtmlElement``` and ```Element```? Breaking up the querySelectorAll and forEach loop could help.

tests/spec/core/data-abbr-spec.js
sidvishnoi: We're testing `<dfn data-abbr>` above already :)

devanshbatra04: oh right.

src/core/data-abbr.js
marcoscaceres: Understood. However, I've never seen that kind linking in a spec in practice. It's always:

 * `<a>FB</a>` or `<a>foo bar</a>`, but never `<a class="dfnlink">foo bar (FB)</a>`.

tests/testFiles.json
marcoscaceres: It's ok, it's only for local testing :) I keep forgetting to update this file too.... hence all the changes below.

src/core/data-abbr.js
marcoscaceres: Ah yeah, let me do what you did originally @devanshbatra04.

src/core/data-abbr.js
devanshbatra04: Alright. I meant ```<a>foo bar (FB)</a>``` and not the class too. 
Entirely your call of course.

marcoscaceres: yeah, still the same. We can always add it easily if it comes up in practice or someone asks for it.

src/core/data-abbr.js
marcoscaceres: good suggestion

src/core/data-abbr.js
```diff
@@ -0,0 +1,44 @@
+// @ts-check
+// Module core/data-abbr
+// - Finds all elements with data-abbr attribute and processes them.
+import { showInlineWarning } from "./utils";
+export const name = "core/dfn-abbr";
+
+export function run() {
+  /** @type {NodeListOf<HTMLElement>} */
+  const elements = document.querySelectorAll("[data-abbr]");
+  for (const elem of elements) {
```
devanshbatra04: ```elements.forEach``` should work well now, if it's on a different line.
Don't need the type casting above as well.

marcoscaceres: Decided `for` was nicer here... no benefit of using forEach.

src/core/data-abbr.js
```diff
@@ -26,19 +26,24 @@ export function run() {
...
+
   dfn.insertAdjacentHTML(
     "afterend",
     ` (<abbr title="${fullForm}">${abbr}</abbr>)`
   );
   dfn.dataset.lt = dfn.dataset.lt ? `${dfn.dataset.lt}|${abbr}` : abbr;
 }
+
+function generateAbbreviation(elem) {
+  if (elem.dataset.abbr) return elem.dataset.abbr;
```
sidvishnoi: umm.. maybe don't do this and do the call to `generateAbbreviation` conditionally.

src/core/data-abbr.js
```diff
@@ -26,19 +26,24 @@ export function run() {
...
-  if (dfn.dataset.abbr === "") {
-    // Generates abbreviation from textContent
-    // e.g., "Permanent Account Number" -> "PAN"
-    dfn.dataset.abbr = dfn.textContent
-      .match(/\b([a-z])/gi)
-      .join("")
-      .toUpperCase();
-  }
-  const fullForm = dfn.textContent.trim();
+  // Generates abbreviation from textContent
```
sidvishnoi: nit: this comment should be on the function definition instead

src/core/data-abbr.js
marcoscaceres: I prefer to keep the error handling here, tbh.

marcoscaceres:
@devanshbatra04, in case you are interested... real word usage of this feature:  
https://github.com/w3c/payment-method-basic-card/pull/74/files

devanshbatra04:
Thats Nice and quick. :")

PR #2205, opened 2019-03-27
tests: custom respec2html tests

### LINKED ISSUE #2061: Add unit test for respec2html
I would like to see tests for issues like #2057.

devanshbatra04:
closes #2061 

This is an early draft and totally WIP.
I came up with this method to add unit tests to respec2html. 
The respec2htmlTests array contains some blocks. These are analogous to describe blocks. 
The tests are analogous to unit tests themselves.

The logic needs to be written into the evalFunction each time a new feature is added to r2h. 

Comments would be really appreciated.
Also it would be great if you could suggest the unit tests to be written for this initial PR.

It currently looks like this
Screenshot from 2019-03-27 19-47-50


devanshbatra04:
Also I opened a draft as @saschanaz suggested. Hope you don't get a ton of emails.

sidvishnoi:
> Hope you don't get a ton of emails.

But you mentioned the name in a comment. Now there will be tons of emails 😆

devanshbatra04:
Here's a video demo of how this fails testing for multiple errors and warnings. I reverted the respecDocWriter.js to before #2088

failing

And this is it passing with the latest source code.
passing

What do you think @marcoscaceres @saschanaz @sidvishnoi ?

saschanaz:
Behavior looks good, make sure you mark this as reviewable when it's done.

tests/headless.js
```diff
@@ -3,10 +3,115 @@
...
+          } catch (error) {
+            throw error;
+          }
+        },
+        message: `Processes example spec employing default respec-w3c-common.js profile`,
+      },
+      {
+        URL: `http://localhost:${port}/examples/basic.html`,
+        evalFunction: async exec => {
+          try {
```
marcoscaceres: Don't need these try/catch blocks I think..

devanshbatra04: Agreed. Only needed when the tests need to play around with errors and warnings.

tests/respec2htmlTests/utils.js
```diff
@@ -0,0 +1,71 @@
...
+    run() {
+      return new Promise((resolve, reject) => {
+        const childProcess = exec(cmd, (err, data) => {
+          if (err) {
+            return reject(err);
+          }
+          resolve(data);
+        });
+        childProcess.stdout.pipe(process.stdout);
+        childProcess.stderr.pipe(process.stderr);
```
devanshbatra04: We could possibly get rid of this by directly working on the process.stderr stream. And avoid the try catch blocks completely in eval function. 
I haven't done it yet as it seemed unneeded complexity. Don't mind going either way personally.

devanshbatra04:
@marcoscaceres @saschanaz can you please review the PR once you get the time? Thanks :)

tests/respec2htmlTests/utils.js
```diff
@@ -0,0 +1,71 @@
...
+        case "Fatal error":
+          return {
+            type: "Fatal error",
+            text: errorStr,
+          };
+      }
+    });
+  },
+
+  URLTorespec2htmlExecutable(url) {
```
saschanaz: This is not using camelCase... `urlToRespec2htmlExecutable` or just `urlToExecutable`?

tests/respec2htmlTests/utils.js
```diff
@@ -0,0 +1,71 @@
...
+  },
+  debug(msg) {
+    // eslint-disable-next-line no-console
+    console.log(colors.grey(moment().format("LTS")) + colors.cyan(` ${msg}`));
+  },
+};
+
+const colors = require("colors");
+const { exec } = require("child_process");
+const moment = require("moment");
```
saschanaz: Imports typically go to the top of the file... I would move the `module.exports` thing to the bottom.

tests/respec2htmlTests/r2h.js
```diff
@@ -0,0 +1,124 @@
...
+            await exec.run();
+          } catch (error) {
+            const expectedErrors = new Set([
+              "A custom SotD paragraph is required for your type of document.",
+              'Document must have one element with `id="abstract"',
+            ]);
+            const expectedWarnings = new Set([
+              "Can't find Table of Contents. Please use <nav id='toc'> around the ToC.",
+            ]);
+            const RecordedTerminalErrors = parseErrorsAndWarnings(
```
saschanaz: camelCase, so `recordedTerminalErrors`

tests/respec2htmlTests/r2h.js
```diff
@@ -0,0 +1,124 @@
...
+              error.toString().split("\n")
+            ).filter(err => err);
+            RecordedTerminalErrors.forEach(({ type, text }) => {
+              switch (type) {
+                case "ReSpec error":
+                  if (!expectedErrors.has(text))
+                    throw `Unexpected ReSpec error ${text}`;
```
saschanaz: Should throw only `Error` instance rather than a raw text.

devanshbatra04:
@saschanaz I am thinking about including #1419 into this PR as well. It requires testing ```save-html.js```. 
It should be helpful in deciding upon the directory structure and seeing how convenient it is to write headless tests for other modules this way.

tests/respec2htmlTests/multiple-warn.html
```diff
@@ -0,0 +1,104 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <meta charset='utf-8'>
+  <title>
+    Replace me with a real title
+  </title>
+  <script src='../../js/deps/require.js' data-main='../../js/profile-w3c-common'
```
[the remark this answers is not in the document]
devanshbatra04: I tried this. But this removes the ToC warning (both from r2h and the actual html render)

devanshbatra04:
I have made the required changes, please have a look. I think we need a well defined directory structure to store the headless tests. 

Also should I incorporate #1419 in the same PR. It involves writing headless tests for spec writer. This should help making the structure more scalable.

tests/respec2htmlTests/multiple-warn.html
```diff
@@ -0,0 +1,104 @@
+<!DOCTYPE html>
```
marcoscaceres: This document seems overly complicated for something that should emit a couple of warnings. I'd suggest stripping this right down so it just emits two warnings. These could be a thing missing in the respecConfig and something missing in the document (e.g., missing heading).

marcoscaceres:
Closing, as I think we want to take a different direction with this. @devanshbatra04 appreciate the work you did on this, nonetheless.

PR #2209, opened 2019-03-28
feat(assets/respec2): make inline warnings visually pop

marcoscaceres:
When clicking navigating to inline errors, or certain areas of interest, this makes the things "pop"... making them easier to find amongst other text.

### DIFF
--- a/assets/respec2.css
+++ b/assets/respec2.css
@@ -3,6 +3,19 @@
  * Robin Berjon - http://berjon.com/
  *****************************************************************/
 
+@keyframes pop {
+  0% {
+    transform: scale(1, 1);
+  }
+  25% {
+    transform: scale(1.25, 1.25);
+    opacity: 0.75;
+  }
+  100% {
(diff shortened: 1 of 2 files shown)


sidvishnoi:
Squiggly line is rendered twice, once as background image, then due to "text-decoration"
image

sidvishnoi:
Also, can we make it a animation bit more visible? use a background maybe?

marcoscaceres:
Will play around. Might make the background a bit yellow then fade to whatever it originally was

sidvishnoi:
Looks like `ui.css` loaded before `respec2.css` giving us both underline and squiggly background.

marcoscaceres:
> Will play around. Might make the background a bit yellow then fade to whatever it originally was

Played around with different background... none of them really looked good, tbh. I decided to stick with this for now. Can improve it over time.

paulirish:
What a subtle and beautiful lil detail. ❤️

PR #2215, opened 2019-03-29
Test Errors and Warnings

### LINKED ISSUE #1084: Make it possible to test errors and warning

devanshbatra04:
closes #1084 

I have been stuck on this for a two days now. Just can't figure what where have I gone wrong. 
Trying to implement [this](https://github.com/w3c/respec/issues/1084#issuecomment-476949095). 

However, the promise in the tests never gets resolved. It works great on trying locally (outside the test environment). 
Help appreciated @marcoscaceres @saschanaz


Swapnilr1:
I just had a brief look at the changes, so just dismiss this if I am wrong (most probably am).

Since you are just running `makeRSDoc()` with the basic configuration, won't `conf.onRespecError` and `conf.onRespecWarn` evaluate to `false` in `pubsubhub.js` ? So your variables declared in `pubsubhub.js` would remain null probably.

devanshbatra04:
That's definitely not the reason. The Promise declaration sets the right config. I made sure of that. Thanks for the help though :)

Swapnilr1:
Ah, my bad, yep you did handle the config - rest of the code seemed correct so I assumed the fault must be here.

Turns out I wasn't entirely wrong. The problem is that https://github.com/w3c/respec/blob/1121be0cdb5244b12a6462993b64cc6f59ba77b8/tests/spec/SpecHelper.js#L87

does not preserve your error function you defined in your promise. So when `pubsubhub.js` gets the `conf` parameter, `conf.onRespecError` is undefined. I don't know how to fix this, probably [StackOverflow can help](https://stackoverflow.com/questions/7759200/is-there-any-possibility-to-have-json-stringify-preserve-functions).

devanshbatra04:
Yes, That seems like it. I had been suspecting this LoC for some time now. I will start cracking again. Thanks a ton!

src/core/pubsubhub.js
```diff
@@ -74,10 +77,40 @@ export function unsub({ topic, cb }) {
...
+  if (conf.collectErrors && !conf.hasOwnProperty("onRespecError")) {
+    // Automatically assign a callback function to collect errors.
+    conf.onRespecError = err => {
+      if (!conf.hasOwnProperty("collectedErrors"))
+        conf.collectedErrors = new Set();
+      const { collectedErrors } = conf;
+      collectedErrors.add(err);
+    };
+  }
+  if (conf.collectWarnings && !conf.hasOwnProperty("onRespecWarn")) {
```
devanshbatra04: Thanks to Swapnil's observation, I came up with this slightly dirty trick to collect Errors and Warnings. 
We can't pass the promises as we planned earlier. I think this is similar to why we need [pre-process-spec file](https://github.com/w3c/respec/blob/develop/tests/spec/core/pre-process-spec.html)

devanshbatra04: #886 needs something similar. It should be a good way to integrate that as well.

devanshbatra04:
@marcoscaceres do you mind reviewing this PR once? Just want to be sure that the approach is right and I will go ahead and test the linter.

saschanaz:
This is adding a new feature rather than testing existing one. `tests/specs/SpecHelper.js` can collect messages so we could use that instead?

devanshbatra04:
I see the point. 
The issue is to be able to test Errors and Warnings themselves.
I will see if the addEventListener idea can help out.

devanshbatra04:
Doesn't ```addEventListener("message")``` work on an element? Wouldn't it be wrong to associate errors and warnings with an element?

saschanaz:
>Doesn't addEventListener("message") work on an element? Wouldn't it be wrong to associate errors and warnings with an element?

It works on a window global object to catch messages from iframes, on which we runs tests.

devanshbatra04:
Got it. I will try it out. Thanks!

marcoscaceres:
I don't think this is the right approach to take here, as it means adding an unnecessary config option that's only relevant for testing. 

There should be a means to use `conf.preProcess` directly from tests to hook into `sub("warm")` and `sub("error")`.

marcoscaceres:
Same as the other bug... want to try a simpler approach. Thanks again @devanshbatra04.

PR #2217, opened 2019-03-30
feat(w3c/headers): support `orcid` member on people.

### LINKED ISSUE #2113: Issue with "person" extras field
## Important info

* URL to affected spec: currently not online
* ReSpec version: 24.3.3

## Description of problem

It seems that, if the `name` field in `extras` has HTML code as value, this is returned escaped in the document.

E.g., re-using the example in https://github.com/w3c/respec/wiki/person :

````
{              
	name:       "Ben De Meester",
	mailto:     "ben.demeester@ugent.be",
	company:    "Ghent University - iMinds - Data Science Lab",
	companyURL: "http://www.iminds.be/",
	url:        "http://users.ugent.be/~bjdmeest/",
	extras: [{
		name:  "<img src='figures/orcid_logo.png' alt='orcid logo'/>",
		href:  "http://orcid.org/0000-0003-0248-0987",
		class: "orcid"
	}],
	w3cid:      "73403"
}
````

This is what will be shown in the spec:

<blockquote>
<dd class="p-author h-card vcard" data-editor-id="73403"><a class="ed_mailto u-email email p-name" href="mailto:ben.demeester@ugent.be">Ben De Meester</a>
            (<a class="p-org org h-org h-card" href="http://www.iminds.be/">Ghent University - iMinds - Data Science Lab</a>)
          , <span class="orcid"><a href="http://orcid.org/0000-0003-0248-0987">&lt;img src='figures/orcid_logo.png' alt='orcid logo'/&gt;</a></span></dd>
</blockquote>

Here's the underlying HTML generated by ReSpec:

````html
<dd class="p-author h-card vcard" data-editor-id="73403">
  <a class="ed_mailto u-email email p-name" href="mailto:ben.demeester@ugent.be">Ben De Meester<!---0.248711%--></a>
  (<a class="p-org org h-org h-card" href="http://www.iminds.be/">Ghent University - iMinds - Data Science Lab<!---0.248711%--></a>) , 
  <span class="orcid"><a href="http://orcid.org/0000-0003-0248-0987">&lt;img src='figures/orcid_logo.png' alt='orcid logo'/&gt;</a></span><!---0.248711%-->
</dd>
````

### PR DESCRIPTION
**Pull request by markasoftware**
Closes  #2113. 

This was split out of my first PR. Since then, I have:

* Inlined the ORCID logo into CSS (although I personally think an inline SVG in HTML, as before, makes more sense).
* Allowed ORCIDs without the origin portion.
* Enhanced the error messages when an ORCID is invalid.

Let me know if anything else should be changed!

### CODE REVIEW

File: assets/respec2.css [hunk 1]
```diff
@@ -55,6 +55,16 @@ th code {
...
...zdGFuY2VJRD0ieG1wLmlpZDo2REM1MzRDNEZGNTcxMUUxODdBOEVCODg2RjdCQ0QwOSIgeG1wOkNyZWF0b3JUb29sPSJBZG9iZSBQaG90b3Nob3AgQ1M1IE1hY2ludG9zaCI+IDx4bXBNTTpEZXJpdmVkRnJvbSBzdFJlZjppbnN0YW5jZUlEPSJ4bXAuaWlkOjlFNTI0NkNBMDgyMDY4MTE5NUZFRDc5MUM2MUUwNEREIiBzdFJlZjpkb2N1bWVudElEPSJ4bXAuZGlkOjA3ODAxMTc0MDcyMDY4MTE5OTRDOTM1MTNGNkRBODU3Ii8+IDwvcmRmOkRlc2NyaXB0aW9uPiA8L3JkZjpSREY+IDwveDp4bXBtZXRhPiA8P3hwYWNrZXQgZW5kPSJyIj8+SwKtIQAAAjZJREFUeNpiZCASLDtnaQCkAoDYHogdcCg7AMQHgXhDlNHxC8SYy0iExQlAKh+IDRhIAyAHTAQ6ZAFZDgBarACk5uPxLbEAFCqJQIc8INoBQMsDoJYLMFAHfIA6YgNBB0CDfD4DbUAiepQwYvH5emQxMR4jBnFeQzD71MNpDJxs7JQ6IhA5JJiwxDkcfP/1E2y5jkQyGFMJzIfaheoAKsc5PiCA7FFGfPEOCgFyARFRBU4PLFBOPi5V8kKWKGkAXQwZ/PrzheH999sMr76cAzuegCNAdi5ggZZwOAsZWBpAdgCyGDbw6+8XhvvvtjKcfjiVgZn5Dy5lBiC7WaDFK1ng66/nQIu2ISKXU5VBEIi52SQZ1EXDwTno5MMWcKjgAAEs0LKdTAe8wJo1ZfjtGMzla8COAdE7bsTjMsKeiQpFLQZ48vEQw77b2WA2yBGg0MABHJholddAwQ6LHkVhL5zqmGiZ4Z98OAQPhQFxACg3EAI0dQAbM8/AOkBGwA6eHvA54AAtLAfFu6IQJPHdf7sNZ2OFCdqGo67PgeWAk+pUuO9vvl6JS+lBUEEEqpvrybGIm02CwUw+C2tJCLMcVBLiARtYQK1XYJl8gYxGJ9gibHUCkXXBBZDdsNpwIq5m2MvP54HkXIJiZNSGE1GaZMBQ2I9eLNOwPXAA6HtHEIMFuYEAxOeRW0VUaP/hbCFjlAPQdnsiHZpkKH0ElIII2lpNpLHlGwZ3x2RQdM0GReeUXt1zgAADALON7V7SXPS8AAAAAElFTkSuQmCC');
```
saschanaz: I think the original SVG was fine, do you have a reason to do PNG instead?

src/w3c/headers.js
```diff
@@ -665,6 +674,43 @@ function collectSotdContent(sotd, { isTagFinding = false }) {
...
+  // trailing slash would mess up checksum
+  const orcidId = orcidUrl.pathname.slice(1).replace(/\/$/, "");
+  if (!/^\d{4}-\d{4}-\d{4}-\d{3}(\d|X)$/.test(orcidId)) {
+    throw new Error(
+      `ORCIDs have the format "1234-1234-1234-1234", not "${orcidId}"`
+    );
+  }
+
+  // calculate checksum as per https://support.orcid.org/hc/en-us/articles/360006897674-Structure-of-the-ORCID-Identifier
+  const lastDigit = orcidId[orcidId.length - 1];
```
saschanaz: Could this checksum part be a separate function?

saschanaz: Ah, actually, never mind.

tests/spec/w3c/headers-spec.js
```diff
@@ -73,6 +73,7 @@ describe("W3C — Headers", () => {
             mailto: "EMAIL",
             note: "NOTE",
             w3cid: "1234",
+            orcid: "https://orcid.org/0000-0002-1694-233X",
```
saschanaz: I'd prefer to just do `0000-0002-1694-233X` rather than the full URL, like `w3cid`... What do you think?

marcoscaceres: Do you mean in this particular test or in general? 

If you mean in general, the problem is that organization that defines ORCID mandates that the id be a URL. However the code supports both.

assets/respec2.css
marcoscaceres: Agree. We should use the svg here.

marcoscaceres: Ah wait... I see why @markasoftware was against putting this in the CSS. @saschanaz, if you see below, Mark added a div where the SVG would normally go. I was confused and thought the actual ID was being shown to the reader - when in fact, it's only the little logo next to their name. 

So, Mark was correct: the SVG should be in the document itself.

markasoftware: For what it's worth, I tried to put the `background-image` on the link directly. However, this requires setting a width and height on the link, but width and height have no effect on inline elements. Setting the `display` to `inline-block` completely messes up the alignment, hence the separate div.

I'll add the SVG back into the HTML. Should I add an eslint exception so I can stuff the whole SVG on one line, or format it as before?

marcoscaceres: Formatted it as before. Sorry for that additional work.

tests/spec/w3c/headers-spec.js
saschanaz: I think the host part is a kind of namespace thing. It looks redundant because the field name is already `orcid` and requires double check for the host.

marcoscaceres: So I agree it's redundant, thought their [docs state](https://support.orcid.org/hc/en-us/articles/360006897674) that:
 
> The ORCID iD is an https URI with a 16-digit number that is compatible with the ISO Standard (ISO 27729), also known as the International Standard Name Identifier (ISNI), e.g. https://orcid.org/0000-0001-2345-6789

My worry is that if we drop don't do the host part, people who use ORCID will complain that just the numeric components don't constitute a valid ORCID (technically correct)... I feel most people will be copy/pasting these URIs from the ORCID website into their ReSpec config. And then we will have other parts of the community using just the "16-digit number" part. The current design allows for both: error correction and we always end up with a conforming ORCID URL.

saschanaz: Hmm... how about `orcidPath`, as the same documentation talks about the host and path?

saschanaz: I feel this PR might be an overimplementation, doing full url and checksum, but if you feel good about it then it may be just okay.

marcoscaceres: heh, yeah, I thought the same :) The checksum is a bit of overkill but it's tiny. The community (really a tiny set of editors in the Web Publishing WG) are the only ones wanting this and they are already using full URLs. For example, to original bug, and:
https://github.com/w3c/dxwg/blob/b32452e9fc003cfd2c1ddeebaa9355826313f62a/profilesont/config.js#L128

So I don't think we need to spend too many cycles on this. It's not something 99% of other folks will use.

assets/respec2.css
markasoftware: Done!

markasoftware:
I've added a test or two.

PR #2219, opened 2019-03-30
fix(core/issue-notes): Make issue-summary section more modular

### LINKED ISSUE #2207: Handling of <section id="issue-summary"> is poor
## Important info

* URL to affected spec: (not online yet) 
* ReSpec version: 24.7.0

## Description of problem

ReSpec doesn't work well if you try to customize the Issue Summary.

According to [the wiki on `issue-summary`](https://github.com/w3c/respec/wiki/issue-summary),
if you create a section with that `id`, it will be used for the auto-generated issue summary.  Like similar auto-generated sections, I assumed that meant I could add my own heading and intro paragraph.

**What happened?:** 

The auto generated issue summary generates and adds it's own heading even if I've included one, so I end up with two `<h2>` elements in the section.

If I _don't_ include my own heading in the section, I get the "no-headingless-sections" linter error!

**Minimal test case (for the body of a generic ReSpec document):**

```html
<p class="issue">Write the rest of this spec!</p>

<section class="appendix" id="issue-summary">
<h2>Open Issues</h2><!-- or remove this to see the linter error -->
<p>
We are tracking issues on
<a href="https://github.com/w3c/respec/issues">Github</a>.
</p>
</section>
```

pradeepgangwar:
fixes #2207

Works as follows:
- If heading and paragraph is provided it uses both of them and appends issues list at the end.
- If only paragraph is given it uses default title and appends issues list after adding paragraph.
- If nothing is give then it chooses default title and appends issues list.


pradeepgangwar:
@marcoscaceres Working as expected. Added few tests to check functionality. Please review.

src/core/issues-notes.js
```diff
@@ -133,7 +133,29 @@ function handleIssues(ins, ghIssues, conf) {
     }
     pub(report.type, report);
   });
-  createIssueSummarySection(l10n.issue_summary, issueList);
+  const issueSummaryElement = document.getElementById("issue-summary");
```
marcoscaceres: Can you help me understand this change a bit? I'm not sure why inline this is different/better than the original code?

marcoscaceres: Oh, sorry, I got confused... I was looking at a subset of changes. However, having this in a function makes sense. This code block is getting rather long.

pradeepgangwar:
Did the changes. Please review. :)

src/core/issues-notes.js
```diff
@@ -203,6 +193,36 @@ function createIssueSummaryEntry(l10nIssue, report, id) {
...
 }
 
+/**
+ *
+ * @param {*} issueList
+ */
+function makeIssueSectionSummary(issueList) {
+  const issueSummaryElement = document.getElementById("issue-summary");
+
+  if (issueSummaryElement) {
```
marcoscaceres: Avoid a level of nesting: 
```suggestion
  if (!issueSummaryElement) return;
```

src/core/issues-notes.js
```diff
@@ -203,6 +193,36 @@ function createIssueSummaryEntry(l10nIssue, report, id) {
...
+        issueSummaryElement.removeChild(paragraph);
+        const issueSummary = hyperHTML`
+          <div><h2>${l10n.issue_summary}</h2>${paragraph}${issueList}</div>`;
+        issueSummaryElement.append(...issueSummary.childNodes);
+      } else {
+        const issueSummary = hyperHTML`
+          <div><h2>${l10n.issue_summary}</h2>${issueList}</div>`;
+        issueSummaryElement.append(...issueSummary.childNodes);
+      }
+    } else {
```
marcoscaceres: I'm thinking that this is no longer an error. If the `issueList` doesn't have any children, then it should say: `<p>There are no issues listed in this specification.</p>`. The string should be localizable, but we can get to that.

src/core/issues-notes.js
```diff
@@ -203,6 +193,36 @@ function createIssueSummaryEntry(l10nIssue, report, id) {
...
 
+/**
+ *
+ * @param {*} issueList
+ */
+function makeIssueSectionSummary(issueList) {
+  const issueSummaryElement = document.getElementById("issue-summary");
+
+  if (issueSummaryElement) {
+    if (document.querySelectorAll(".issue").length) {
```
marcoscaceres: Should really use `issueList` here... not query the document.

src/core/issues-notes.js
```diff
@@ -203,6 +193,36 @@ function createIssueSummaryEntry(l10nIssue, report, id) {
...
+      const heading = issueSummaryElement.querySelector("h2");
```
marcoscaceres: Note that  `issue-summary` can appear anywhere, and hence there is no guarantee that the first heading will be a h2. It's perfectly valid for it to be any heading level 2-6.

src/core/issues-notes.js
```diff
@@ -203,6 +193,36 @@ function createIssueSummaryEntry(l10nIssue, report, id) {
...
+      const paragraph = issueSummaryElement.querySelector("p");
```
marcoscaceres: Likewise, there is no restriction on what content there can be inside the summary:

```HTML
<section>
   <h3>Appendix</h3>
   <section id="issue-summary">
     <h4>OMG so many issues!</h4>
     <aside class="issue">Do we have too many issues?</aside>
     <figure><figcaption>A diagram</figcaption></figure>
   </section>
</section> 
```

src/core/issues-notes.js
```diff
@@ -203,6 +193,36 @@ function createIssueSummaryEntry(l10nIssue, report, id) {
...
+      if (heading && paragraph) {
+        issueSummaryElement.append(issueList);
+      } else if (!heading && paragraph) {
+        issueSummaryElement.removeChild(paragraph);
+        const issueSummary = hyperHTML`
+          <div><h2>${l10n.issue_summary}</h2>${paragraph}${issueList}</div>`;
```
marcoscaceres: We should ditch the `<div>`

src/core/issues-notes.js
pradeepgangwar: I figured out that function `handleIssues` is not called when there are no issues. Hence the function that i created is never called thus never providing me opportunity to show the error `<p>There are no issues listed in this specification.</p>`.

Check:

pradeepgangwar: If you want a different behaviour we will probably have to change this. I am not sure how to change this to meet ReSpec requirements.

src/core/issues-notes.js
```diff
@@ -203,6 +193,30 @@ function createIssueSummaryEntry(l10nIssue, report, id) {
...
+  const issueSummaryElement = document.getElementById("issue-summary");
+
+  if (!issueSummaryElement) return;
+  const heading = issueSummaryElement.querySelectorAll("h2, h3, h4, h5, h6");
+  const defaultHeading = hyperHTML`<h2>${l10n.issue_summary}</h2>`;
+
+  issueList.hasChildNodes()
+    ? issueSummaryElement.append(issueList)
+    : issueSummaryElement.append(
+        hyperHTML`<p>There are no issues listed in this specification.</p>`
```
marcoscaceres: We should localize this string. Please add it to `localizationStrings` above as "no_issues_in_spec". 

```suggestion
        hyperHTML`<p>${l10n.no_issues_in_spec}</p>`
```

For Spanish, please use: 

"No hay problemas enumerados en esta especificación." 

@thijsbrentjens, could you suggest a Dutch translation for: 

"There are no issues listed in this specification."

tests/spec/core/issues-notes-spec.js
```diff
@@ -378,4 +378,47 @@ describe("Core — Issues and Notes", () => {
...
+  it("shows issue-summary section with paragraph and default heading when only <p> is defined", async () => {
+    const ops = {
+      config: makeBasicConfig(),
+      body: `
+      <section>
+        <h2>Test Issues</h2>
+        <p class="issue" data-number=123></p>
+      </section>
+      <section id="issue-summary">
+        <p>Here you will find all issues summary</p>
```
marcoscaceres: I think we should add one more random element, like a `<div class="note">`, just to make sure it's also being added.

src/core/issues-notes.js
```diff
@@ -203,6 +193,30 @@ function createIssueSummaryEntry(l10nIssue, report, id) {
...
 
+/**
+ *
+ * @param {*} issueList
+ */
+function makeIssueSectionSummary(issueList) {
+  const issueSummaryElement = document.getElementById("issue-summary");
+
+  if (!issueSummaryElement) return;
+  const heading = issueSummaryElement.querySelectorAll("h2, h3, h4, h5, h6");
```
marcoscaceres: I think here you just want whatever the first one is. The assumption being that there is a heading in the section, which is fine. 
 
```suggestion
  const heading = issueSummaryElement.querySelector("h2, h3, h4, h5, h6");
```

src/core/issues-notes.js
```diff
@@ -203,6 +193,30 @@ function createIssueSummaryEntry(l10nIssue, report, id) {
...
+  const defaultHeading = hyperHTML`<h2>${l10n.issue_summary}</h2>`;
+
+  issueList.hasChildNodes()
+    ? issueSummaryElement.append(issueList)
+    : issueSummaryElement.append(
+        hyperHTML`<p>There are no issues listed in this specification.</p>`
+      );
+  heading.length
```
marcoscaceres:
```suggestion
  heading
```

src/core/issues-notes.js
```diff
@@ -203,6 +193,30 @@ function createIssueSummaryEntry(l10nIssue, report, id) {
...
+ */
+function makeIssueSectionSummary(issueList) {
+  const issueSummaryElement = document.getElementById("issue-summary");
+
+  if (!issueSummaryElement) return;
+  const heading = issueSummaryElement.querySelectorAll("h2, h3, h4, h5, h6");
+  const defaultHeading = hyperHTML`<h2>${l10n.issue_summary}</h2>`;
+
+  issueList.hasChildNodes()
+    ? issueSummaryElement.append(issueList)
```
marcoscaceres: I'm a bit worried that this `issueSummaryElement.append(issueList)` appears twice (here and above)? I was only expecting to see that once.

src/core/issues-notes.js
pradeepgangwar: I did a google translate for this. Let me know if that needs some changes.

src/core/issues-notes.js
pradeepgangwar: I noticed that and tested it for various cases and it works fine. It does not duplicate the content. Should I do it another way?

pradeepgangwar:
I have made the changes. I have also added some inline comments. Let me know if we need more changes. :)

src/core/issues-notes.js
```diff
@@ -22,12 +22,15 @@ export const name = "core/issues-notes";
 const localizationStrings = {
   en: {
     issue_summary: "Issue Summary",
+    no_issues_in_spec: "There are no issues listed in this specification.",
   },
   nl: {
     issue_summary: "Lijst met issues",
+    no_issues_in_spec: "No hay problemas enumerados en esta especificación.",
```
marcoscaceres: This is spanish ;)

src/core/issues-notes.js
```diff
@@ -22,12 +22,15 @@ export const name = "core/issues-notes";
...
   },
   es: {
     issue_summary: "Resumen de la cuestión",
+    no_issues_in_spec: "Er zijn geen problemen vermeld in deze specificatie.",
```
marcoscaceres: Please be sure @thijsbrentjens verifies this translation. Using auto-translations is always a big risk.

src/core/issues-notes.js
marcoscaceres: yeah, please see if there is a way that `issueSummaryElement.append(issueList)` is only ever called once.

src/core/issues-notes.js
```diff
@@ -203,6 +196,26 @@ function createIssueSummaryEntry(l10nIssue, report, id) {
...
+  const issueSummaryElement = document.getElementById("issue-summary");
+
+  if (!issueSummaryElement) return;
+  const heading = issueSummaryElement.querySelector("h2, h3, h4, h5, h6");
+
+  issueList.hasChildNodes()
+    ? issueSummaryElement.append(issueList)
+    : issueSummaryElement.append(hyperHTML`<p>${l10n.no_issues_in_spec}</p>`);
+  heading ||
+    issueSummaryElement.insertBefore(
```
marcoscaceres: Just using `if (!heading) {` would be cleaner. 

I still think there is a bug tho... you could have:

```HTML
<section id="issue-summary">
  <p>Issues are amazing</p>
  <section>
     <h3>Amazing</h3>
  </section>
</section>
```

(Make sure you add the above to a test)

So there is check needed: that the `heading !== issueSummaryElement.firstElementChild`.

marcoscaceres: I wonder if `insertAdjacentHTML("afterbegin”, ...)` couldn’t be used here instead 🧐 if yes, then it’s good as it avoids having to use hyperHTML).

pradeepgangwar:
@marcoscaceres Replaced `insertAfter` with `inserAdjecantHTML` and added tests for below case: 😃 

```
<section id="issue-summary">
  <p>Issues are amazing</p>
  <section>
     <h3>Amazing</h3>
  </section>
</section>
```

src/core/issues-notes.js
```diff
@@ -203,6 +196,30 @@ function createIssueSummaryEntry(l10nIssue, report, id) {
   `;
 }
 
+/**
+ *
+ * @param {*} issueList
```
marcoscaceres:
```suggestion
 * @param {HTMLUListElement} issueList
```

pradeepgangwar:
@marcoscaceres Undoing commit https://github.com/w3c/respec/pull/2219/commits/5840790024e57a108dfaebf15132d39a3284fcae makes all the tests pass. 😕 
I don't know why that happens.

marcoscaceres:
Oops, I broke it. Will fix.

marcoscaceres:
Sorry about that @pradeepgangwar ... was distracted in meeting and "fixing" things in the GitHub editor... that always ends badly :)

PR #2221, opened 2019-04-01
fix(core/inlines): Single letter variables don't get converted to var

### LINKED ISSUE #2208: Single letter variables don't get converted to var

If you have: `Let |p| be new Promise`... 

But what I was expecting?:

It should convert `|p|` to `<var>p</var>`.

Similarly, if you have: `Let |p:Promise| be newly created Promise`.

Swapnilr1:
Fixes #2208 

The new RegEx is:
```js
/\B\|\w[\w\s]*(?:\s*\:[\w\s&;<>]+)?\|\B/
```

Thanks to @sidvishnoi for the RegEx. This matches single letter variables, and also types with spaces. I have tested it on some possible variable types, and it does not seem to create any new issues.

In the variable type, the RegEx matches word characters (A-Z, a-z, 0-9, _) and whitespace characterss and the characters `&;<>`. 

I want to confirm if no other character occurs in a variable type. @saschanaz Could you take a look at this?

### DIFF
--- a/src/core/inlines.js
+++ b/src/core/inlines.js
@@ -137,7 +137,7 @@ export function run(conf) {
       "\\b(?:NOT\\s+)?RECOMMENDED\\b",
       "\\bOPTIONAL\\b",
       "(?:{{3}\\s*.*\\s*}{3})", // inline IDL references,
-      "\\B\\|\\w[\\s\\w]*\\w(?:\\s*\\:\\s*\\w+)?\\|\\B", // inline variable regex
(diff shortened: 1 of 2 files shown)


Swapnilr1:
Just occurred to me that I should also write tests with a single letter variable. Will add them.

marcoscaceres:
This is looking really good, but it would be good to include the test cases that @sidvishnoi  included in the bug:
https://github.com/w3c/respec/issues/2208#issuecomment-478316862

In particular, the Generics case, and the case where it was originally overmatching.

Swapnilr1:
Tests should be enough now.

PR #2222, opened 2019-04-02
feat(core/inlines): Auto-expand inline references

### LINKED ISSUE #2211: Auto-expanding inline references
In ReSpec, we currently support the ability to, for example, write:

```HTML
<a data-cite="payment-request"></a>
```

That gets expanded to:

```HTML
<a href="https://...">Payment Request API</a>
```

However, `<a data-cite="payment-request"></a>` is kinda ugly... What would be nicer would be to write inline (3 square brackets):

```
[[[payment-request]]]
```

And that then gets transformed to:

```HTML
<a data-cite="payment-request"></a>
```

And ReSpec then handles the rest.

Swapnilr1:
Fixes #2211 

Please suggest a better function name than `inlineRefMatches()`. 

Need feedback regarding a nit: I have added the test to `inlines-spec.js` and I am only testing whether `[[[xyz]]]` got converted to an anchor tag or not. Now the end result is that `[[[xyz]]]` gets converted to `<a href=...>XYZ API</a>` but `inlines.js` only converts it to `<a data-cite=...></a>`. 
So converting `<a data-cite=...></a>` to `<a href=...>XYZ API</a>` is outside the scope of `inlines.js` module and I should not test for it in `inlines-spec.js`. But it's not clear to me if I should check whether the inline reference got expanded, and if so where should I put the test for it.


marcoscaceres:
> But it's not clear to me if I should check whether the inline reference got expanded, and if so where should I put the test for it.

The tests should be implementation agnostic, meaning that they check the expected output (so yes, test the innerText too). What matters is that they hit your code path and that you get the right thing at the end. Also, make sure you test with a `localBiblio` - don’t query a real spec otherwise it will break if the network is down.

Swapnilr1:
> The tests should be implementation agnostic, meaning that they check the expected output (so yes, test the innerText too). 

Yes, I get that. Just want to know if `inlines-spec.js` is the appropriate place for that. Because let's say `data-cite.js` breaks, then if I test for the expanded reference, the test in `inlines-spec.js` would fail. To me that is acceptable, but just wanted to confirm it.

>Also, make sure you test with a localBiblio - don’t query a real spec otherwise it will break if the network is down.

Yes, will change the test.

marcoscaceres:
> Just want to know if inlines-spec.js is the appropriate place for that. 

Yep 👍

marcoscaceres:
Make sure you also test that the reference itself ends up in the References section ... you will need a `<section id="conformance">` to trigger normative references. Make sure you also test [[[?foo]]] ends up in the Informative References section. Likewise, test that:

`<p class=example>[[[bar]]]</p>` also ends up in the Informative References section.

tests/spec/core/inlines-spec.js
```diff
@@ -134,4 +134,15 @@ describe("Core - Inlines", () => {
     expect(h[1].textContent).toEqual("var2");
     expect(h[1].dataset.type).toEqual("Generic<unsigned short int>");
   });
+
+  it("expands inline references", async () => {
```
marcoscaceres:
```suggestion
  it("expands inline references and they get classified as normative/informative correctly", async () => {
```

Swapnilr1:
About the normative and informative references, currently enforcing is not working. So `[[[!xyz]]]` ends up as an informative reference right now.
To my understanding, I am only allowing a shorthand here. So `[[[ref]]]` converts to `<a data-cite="ref"></a>`. So what should `[[[!ref]]]` or `[[[?ref]]]` convert to? 
Currently, if I want a `<a data-cite="ref">` to be enforced normative or informative, how do I do that?

marcoscaceres:
> So what should [[[!ref]]] or [[[?ref]]] convert to?

```
<a data-cite="!ref"></a>
<a data-cite="?ref"></a>
```

> how do I do that?

You should only need to add:

```
<section id="conformance"></section>
```

So, the whole thing would be:

```HTML
<section id="conformance"></section>
<section>
  <p>[[[normative]]] and [[[!normative-2]]]</p>
  <p>[[[?informative]]]</p>
</section>
```

Swapnilr1:
I tried
```html
<section class="informative">
    <p><a data-cite="!payment-request"></a></p>
</section>
```
and it simply added Payment Request to Informative references, which is not what is intended.
So just confirming, I also have to make changes to `data-cite.js` to handle `data-cites` with enforced reference type?

marcoscaceres:
Did you try?:

```
<section id="conformance"></section>
<section class="informative">
    <p><a data-cite="!payment-request"></a></p>
</section>
```

marcoscaceres:
OK, just tried it... the behavior is correct: the "!" is ignored in the informative section with no warning. That's fine tho.

Swapnilr1:
> OK, just tried it... the behavior is correct: the "!" is ignored in the informative section with no warning. That's fine tho.

Sorry, didn't get this. Does this mean I can skip adding tests for `[[[!foo]]]` and `[[[?foo]]]`?

marcoscaceres:
Ignore `[[[!foo]]]`, but `[[[?foo]]]` is important.

marcoscaceres:
If possible, could you please document this feature here: https://github.com/w3c/respec/wiki/ReSpec-Editor's-Guide#references

Let me know if you are busy and I can do it otherwise.

Swapnilr1:
I will do that today.

Swapnilr1:
@marcoscaceres I have documented this feature, but I wasn't able to write much, especially since referencing works very similar to `[[Foo]]`. I hope it is still clear enough. 

Also I just realized, this implementation does not support `[[[Spec#Section]]]`, but only `[[[Spec]]]`.
I can change the RegEx to support `[[[Spec#Section]]]` but the `textContent` for the resulting anchor tag would still be `SpecTitle` not `SectionHeading`. Do we intend to support `[[[Spec#Section]]]`?

marcoscaceres:
>Do we intend to support [[[Spec#Section]]]?

No. We don’t intend to support that at this point.

marcoscaceres:
The updated documentation reads well. Thanks for doing that. I’ll send out an email to the W3C editors list to let folks know it’s available.

marcoscaceres:
Working really nicely @Swapnilr1 :) Just updated Payment Request: https://github.com/w3c/payment-request/commit/f12390196d13f2bd0b547675689319086fec4a53

Issue #1975, opened 2018-12-23
Migrate to webpack/rollup and ditch requirejs


saschanaz:
I'm not sure what we should do for existing `text!` thing yet.

saschanaz:
Hmm. @marcoscaceres Thoughts?

saschanaz:
Tried this a bit and found we should first remove the `deps/text` things.

marcoscaceres:
We should be able to use one of the CSS importer things for text, I think. Or are there actual text we import? 

Agree about the UI. We should be able to ditch dynamic imports.

saschanaz:
A question: Do we want to keep our babel transpiler after webpack/rollup migration? Maybe for debugging, as sourcemaps are not perfect yet?

sidvishnoi:
We need babel until Safari supports object spread assignment though.

saschanaz:
webpack/rollup internally have babel so I think it will be okay.

saschanaz:
#1981 made the webpack part but not really ditched requirejs.

saschanaz:
Currently I have two options:

1. Migrate raw text imports to `fetch()` with a custom webpack plugin that converts `fetch()` into text importer call.
2. Implement a service worker that understands text import request and serves proper JavaScript module instead.

Option 1 needs some study because I don't know details to implement such plugins. Option 2 should be easier but 1) needs a way to detect whether a request is for text import or not 2) the worker should be injected separately to a document because the webpack-built version won't need it.

marcoscaceres:
> Implement a service worker that understands text import request and serves proper JavaScript module instead.

Good idea - but that's not going to work, unfortunately :( The same origin restriction would mean that each spec would also need to locally host a copy of the service worker. 

If the text calls are mostly CSS... we could consider CSS-in-JS?

saschanaz:
>The same origin restriction would mean that each spec would also need to locally host a copy of the service worker.

We can keep using webpack to bundle CSS. The suggested service worker is only for non-bundled development version.

marcoscaceres:
Ah, ok. It still doesn't feel right that we would need two versions... from experience, the bundled version should still be debugguble in WebPack (just going to be a big fat file)... so I'm still hopeful we can do this without resorting to 2. 

I need to explore exactly what stuff is coming in with !text.

saschanaz:
>V

saschanaz:
[...]

PR #2228, opened 2019-04-04
refactor(core/biblio-db): Move from indexedDB to idb

### LINKED ISSUE #1955: Refactor biblio-db with idb dependency

Swapnilr1:
Fixes #1955 

I started refactoring few days ago but ran into issues with tests and have not been able to resolve it.
When running the tests in `biblio-db-spec.js`, I get `Error: Script error for "idb", needed by: core/biblio-db` and the tests fail.
I even tested after removing almost all code and keeping just 
```js
import { openDB } from "idb";
export const name = "core/biblio-db";
export const biblioDB = {};
```
in `biblio-db.js`, and I still got the same error `Error: Script error for "idb", needed by: core/biblio-db`. 
I even tried to use the IDB code from `xref.js` and I still got the same error. :cry: 

I noticed that the tests for xref include the database creation code:
```js
await new Promise(resolve => require(["deps/idb"], resolve));
const { openDB } = window.idb;
const { IDBKeyVal } = await new Promise(resolve =>
  require(["core/utils"], resolve)
);
const idb = await openDB("xref", 1, {
  upgrade(db) {
    db.createObjectStore("xrefs");
  },
});
```

So I think I have to make changes to the test files to get idb to be "initialized".

Any guidance is much appreciated.


Swapnilr1:
@sidvishnoi Could you have a look and tell me how to resolve Script error in tests (details in description).

saschanaz:
Just tried fixing your error, it should now at least import the package properly.

Swapnilr1:
@saschanaz Thanks for your help, though the error still persist. But your changes pointed me to towards the problem: The test is trying to fetch `js/idb.js` and getting a 404 instead of `js/deps/idb.js`. 
I tried importing `idb` in some other modules and found that tests fail with the Script error when the test spec includes a `require` statement like https://github.com/w3c/respec/blob/24bf82b40e2e2ccb9efccc04390264b25252db77/tests/spec/core/biblio-db-spec.js#L42 and work otherwise (though I think you already know this). 

So I need to check how to set the correct path for tests. Working on it. Thanks a lot.

Swapnilr1:
Finally fixed. Thank you @marcoscaceres and @saschanaz for help with importing `idb` correctly. 
I have almost completed the refactor, please review it - I think I have wrongly removed the `DOMException` rejections.  
I am looking into why the tests are failing, because they ran well locally in SpecRunner. :confused:

saschanaz:
I suspect you are using `SpecRunner.html`? Maybe that's why I was seeing different result...

Swapnilr1:
@saschanaz Yes I am using `SpecRunner.html`. Thanks for the fix.

src/core/biblio-db.js
```diff
@@ -54,13 +42,8 @@ const readyPromise = new Promise((resolve, reject) => {
...
       }),
     ];
-    try {
-      await Promise.all(promisesToCreateSchema);
-      resolve();
-    } catch (err) {
-      reject(err);
-    }
-  };
+    Promise.all(promisesToCreateSchema);
```
saschanaz:
```suggestion
```

Because this line doesn't really do anything.

src/core/biblio-db.js
```diff
@@ -143,22 +110,13 @@ export const biblioDB = {
...
-      request.onerror = () => {
-        reject(new DOMException(request.error.message, request.error.name));
-      };
-    });
+
+    const objectStore = db
+      .transaction("alias", "readonly")
+      .objectStore("alias");
+    const range = IDBKeyRange.only(id);
+    const request = await objectStore.openCursor(range);
```
saschanaz: How about `result`, because it's not really a request object anymore?

src/core/biblio-db.js
```diff
@@ -175,20 +133,10 @@ export const biblioDB = {
...
-        }
-        resolve(request.result.value);
-      };
-      request.onerror = () => {
-        reject(new DOMException(request.error.message, request.error.name));
-      };
-    });
+    const objectStore = db.transaction([type], "readonly").objectStore(type);
+    const range = IDBKeyRange.only(id);
+    const request = await objectStore.openCursor(range);
```
saschanaz: Same here

src/core/biblio-db.js
```diff
@@ -277,20 +220,9 @@ export const biblioDB = {
...
-    });
+    const stores = await db.transaction(storeNames, "readwrite");
     const clearStorePromises = storeNames.map(name => {
-      return new Promise(resolve => {
-        const request = stores.objectStore(name).clear();
-        request.onsuccess = resolve;
-      });
+      return stores.objectStore(name).clear();
     });
     Promise.all(clearStorePromises);
```
saschanaz: Shouldn't we await this? Looks like a bug.

src/core/biblio-db.js
Swapnilr1: My bad, leaving constant names intact wasn't a very good decision. Will make these changes

src/core/biblio-db.js
Swapnilr1: Yes, missed this. I have checked more thoroughly and fixed this at a few more places. Thanks for the catch.

src/core/biblio-db.js
```diff
@@ -19,29 +19,27 @@ const readyPromise = openDB("respec-biblio2", 12, {
...
+      try {
+        const transaction = db.createObjectStore("reference", {
+          keyPath: "id",
+        }).transaction;
+        transaction.oncomplete = resolve;
+        transaction.onerror = reject;
+      } catch (err) {
+        reject(err);
+      }
+    });
```
saschanaz: Can we just skip the promise part here? The [idb examples](https://github.com/jakearchibald/idb#article-store) or even the raw [IndexedDB examples](https://developer.mozilla.org/en-US/docs/Web/API/IDBObjectStore/createIndex#Example) never uses `transaction.oncomplete` so maybe?

Swapnilr1: I will look into it and other parts of the module. Actually while refactoring I only changed parts for which `idb` returns a promise. Since you wanted this issue fixed for your work, I made a PR as soon as I got the tests resolved. I will make these changes.

marcoscaceres: I was thinking the same. Hopefully idb takes care of everything and the above goes away.

src/core/biblio-db.js
```diff
@@ -173,8 +171,8 @@ export const biblioDB = {
       }, aliasesAndRefs);
     const promisesToAdd = Object.keys(aliasesAndRefs)
       .map(type => {
-        return Array.from(aliasesAndRefs[type]).map(details =>
-          await this.add(type, details)
+        return Array.from(aliasesAndRefs[type]).map(
+          async details => await this.add(type, details)
```
saschanaz: This change is redundant because the return type is still Promise, and it's expected.

src/core/biblio-db.js
```diff
@@ -222,7 +221,7 @@ export const biblioDB = {
     const storeNames = [...ALLOWED_TYPES];
     const stores = await db.transaction(storeNames, "readwrite");
     const clearStorePromises = storeNames.map(name => {
-      return stores.objectStore(name).clear();
+      return await stores.objectStore(name).clear();
```
saschanaz: This is redundant too. It still returns a Promise anyway.

src/core/biblio-db.js
```diff
@@ -78,7 +40,7 @@ export const biblioDB = {
     if (await this.isAlias(id)) {
       id = await this.resolveAlias(id);
     }
-    return this.get("reference", id);
+    return await this.get("reference", id);
```
marcoscaceres:
```suggestion
    return this.get("reference", id);
```

src/core/biblio-db.js
```diff
@@ -95,17 +57,9 @@ export const biblioDB = {
...
-      request.onsuccess = () => {
-        resolve(!!request.result);
-      };
-      request.onerror = () => {
-        reject(new DOMException(request.error.message, request.error.name));
-      };
-    });
+    const objectStore = db.transaction([type], "readonly").objectStore(type);
+    const range = IDBKeyRange.only(id);
+    return !!(await objectStore.openCursor(range));
```
marcoscaceres: I know this was in the original, but it's still a bit ugly... maybe we can do? 

```suggestion
    const result = await objectStore.openCursor(range); 
    return result ? true : false;
```

Same with other one below...

src/core/biblio-db.js
```diff
@@ -118,19 +72,11 @@ export const biblioDB = {
...
-      };
-      request.onerror = () => {
-        reject(new DOMException(request.error.message, request.error.name));
-      };
-    });
+    const objectStore = db
+      .transaction(["alias"], "readonly")
+      .objectStore("alias");
+    const range = IDBKeyRange.only(id);
+    return !!(await objectStore.openCursor(range));
```
marcoscaceres: As above...

src/core/biblio-db.js
```diff
@@ -251,15 +178,10 @@ export const biblioDB = {
...
-      // update or add, depending of already having it in db
-      const request = isInDB ? store.put(details) : store.add(details);
-      request.onsuccess = resolve;
-      request.onerror = () => {
-        reject(new DOMException(request.error.message, request.error.name));
-      };
-    });
+    const store = db.transaction([type], "readwrite").objectStore(type);
+    // update or add, depending of already having it in db
+    const request = isInDB ? store.put(details) : store.add(details);
```
marcoscaceres: Maybe we should do:
Maybe just: 

```suggestion
    return isInDB ? await store.put(details) : await store.add(details);
```

src/core/biblio-db.js
saschanaz: `? true : false`? It doesn't look JavaScript-y for me... Maybe `!!result`?

src/core/biblio-db.js
saschanaz: I'd rather say "if you await something in a function, rather await all", because returning a raw promise sometimes causes a tricky situation:

```js
try {
  const result = await myPromise;
  return result.doAsync(); // A rejected promise here won't call the catch block!
} catch {
  console.error("Ouch!");
}
```

src/core/biblio-db.js
marcoscaceres: !!result wfm

src/core/biblio-db.js
marcoscaceres: In general,  I agree. My general preference is usually to either return the raw promise in cases where it's a simple return type (e.g., boolean here)... or always put the result into a variable, as it makes debugging much easier, as one can check the `result` before returning it. 

@Swapnilr1, we leave it to you to decide what is best.

src/core/biblio-db.js
marcoscaceres: Oops... I've seen this come up too I think. Glad you spotted this @saschanaz.

Swapnilr1:
I have implemented the changes, but I am not able to run tests either by `SpecRunner.html` or by karma. SpecRunner gives `SyntaxError: import declarations may only appear at top level of a module`.

saschanaz:
Investigation incoming...

saschanaz:
The problem is that currently the ES module system does not support resolving by package name. After #2241, karma wants to load `idb` as a dependency of `core/biblio-db.js`, which fails because the module system does not know what `idb` is.

We can replace `idb` with `../../node_modules/idb/buildesm/index.js` as a workaround. It's not pretty, but anyway...

marcoscaceres:
Yeah, this is why copydeps.js exists :( and what’s nice about webpack is that it handles this. I’m still not sure how we overcome copydeps.js in a modules only world. We probably can’t as those files have to live somewhere.

saschanaz:
@marcoscaceres This is where [import-maps](https://github.com/WICG/import-maps) comes handy, but only Chrome 74 supports it as of now...

marcoscaceres:
Must prioritize Gecko implementation :)

saschanaz:
Couldn't even find an relevant issue from bugzilla 🤔

marcoscaceres:
WICG start their Mozilla life here: https://github.com/mozilla/standards-positions/issues/146

Swapnilr1:
> We can replace idb with ../../node_modules/idb/build/esm/index.js as a workaround. It's not pretty, but anyway...

OK, this can allow me to pass karma tests and get this PR merged. BTW, even Firefox 66 does not yet support dynamic imports (still a flag, enabled by default in Firefox 67) so karma tests fail to execute in Firefox at the moment, but that's OK I guess - only need to wait for Firefox 67.

 But `SpecRunner.html` still does not work, be it Firefox 66 or Chrome 73. 

specrunner

saschanaz:
Maybe we should do something separately for SpecRunner...

What situation requires SpecRunner, BTW? I do `npm run karma start` + click `DEBUG` and it works perfectly well for debugger-requiring tasks.

marcoscaceres:
> What situation requires SpecRunner, BTW? 

Just me being lazy and old :( I like the little in-browser reporting. But yeah... if it's becoming a pain, it might be time for me to give it up.

marcoscaceres:
Filed https://github.com/w3c/respec/issues/2244 about killing SpecRunner.

Swapnilr1:
> We can replace idb with ../../node_modules/idb/build/esm/index.js as a workaround.

@saschanaz This breaks everything else except the karma tests. This is because `index.js` is served directly to the browser and browsers do not support `import`. I think reverting to `import { openDB } from "idb";` and adding a shim would solve the issue, just that I am not sure where to add the shim.

saschanaz:
Browsers do support imports, but yes, it will break requirejs compatibility... 🤔

Swapnilr1:
> Browsers do support imports, ...

Oh, I thought browsers not supporting imports was the reason `SpecRunner.html` was failing, but I believe browsers require a `type="module"` attribute which RequireJS does not provide. I need help with configuring shims.

saschanaz:
Seems we are kind of locked. Maybe we should do #1975 first?

marcoscaceres:
Now that SpecRunner is gone, @Swapnilr1 what’s blocking this?

Swapnilr1:
@marcoscaceres SpecRunner wasn't really blocking this (it doesn't run on Travis anyway) - SpecRunner would not work after #2241 so I had just made a comment regarding that. 
The blocker is karma tests. Using `import { openDB } from "idb";` (which loads the IIFE build) leads to `TypeError: Failed to resolve module specifier "idb".` when running karma tests. (Most likely because we require a shim - though I am not sure if it can be done).

Switching to `import { openDB } from "../../node_modules/idb/build/esm/index.js";` as suggested by @saschanaz fixes the issue with karma tests but will break everything else. This is because `index.js` has import/export declarations, and unlike other modules won't be compiled by Babel before being served to the browser. The browser then complains of import declarations in a non-module because RequireJS does not serve it as an ES Module.

[Extremely sorry for any technical inaccuracy; I am not very well versed with ES6 modules].

marcoscaceres:
Hhmm... should we complain upstream? This doesn’t sound good and no other module exhibits this kind of problem?

Swapnilr1:
@marcoscaceres Yes, perhaps we can ask for RequireJS compatibility without needing shims - though I am not completely certain if this will fix the problem.

saschanaz:
This is because karma now consumes ES modules, and this PR tries consuming a non-ES module in an ES module. Webpack and requirejs somehow can deal with it but ES modules need import-maps support.

marcoscaceres:
> this PR tries consuming a non-ES module (or ES module without an actual path) in an ES module

marcoscaceres:
> this PR tries consuming a non-ES module (or ES module without an actual path) in an ES module

Ok, but that seems like a bug upstream, no? The maintainer has shipped something non-conforming. We could ask them to ship an actual ES module, instead of whatever this is.

saschanaz:
No, `idb` itself is good, what happens here is we cannot just `import * from "idb"` because the native ES module system does not support non-path import. It's okay for `marked` and `hyperhtml` because they are not directly imported from karma (as generally tests just uses the compiled full ReSpec), but here we want to directly import a module and write a unit test for it.

marcoscaceres:
> No, idb itself is good, what happens here is we cannot just import * from "idb" because the native ES module system does not support non-path import.

That makes sense :)  

> It's okay for marked and hyperhtml because they are not directly imported from karma (as generally tests just uses the compiled full ReSpec), but here we want to directly import a module and write a unit test for it.

This doesn't make sense to me (and probably where I'm getting confused). We should be testing the idb that is compiled into ReSpec. Additionally, shouldn't we just use use copydeps.js to copy idb over to deps? 

copydeps.js can be removed once we fully transition to some packager.

saschanaz:
>We should be testing the idb that is compiled into ReSpec.

We have [tests that directly imports biblio-db](https://github.com/w3c/respec/blob/48411a34a369c60f6796c7d0cbec2f38e0ae94a8/tests/spec/core/biblio-db-spec.js).

>shouldn't we just use use copydeps.js to copy idb over to deps?

Currently copydeps is only used for requirejs mapping, otherwise TypeScript type system breaks (as they only knows package name to type).

It doesn't help here anyway, if we copy cjs then ES module breaks, or we copy esm then requirejs breaks, so 🤷‍♀️

>

marcoscaceres:
Yay! module systems 💃. Ok, we hold I guess until we kill RequiresJS.

saschanaz:
>until we kill RequiresJS.

Or until we get import maps on Firefox. I hope Mozillians get some push for this new feature.

marcoscaceres:
I believe this is appropriate?: 

Screenshot 2019-04-25 13 02 43

😂

sidvishnoi:
What's blocking this one? Why can't we use `idb` the way we used in [xref.js](https://github.com/w3c/respec/blob/51282455ffb29e42c4c04c3fc807a05e8438e9d9/src/core/xref.js#L20) and [xref-spec.js](https://github.com/w3c/respec/blob/51282455ffb29e42c4c04c3fc807a05e8438e9d9/tests/spec/core/xref-spec.js#L11)?

saschanaz:
> 
> 
> What's blocking this one? Why can't we use `idb` the way we used in [xref.js](https://github.com/w3c/respec/blob/51282455ffb29e42c4c04c3fc807a05e8438e9d9/src/core/xref.js#L20) and [xref-spec.js](https://github.com/w3c/respec/blob/51282455ffb29e42c4c04c3fc807a05e8438e9d9/tests/spec/core/xref-spec.js#L11)?

Because Karma is using ES module, some tests there want to import `core/biblio-db`, and then the ES module system doesn't know where `idb` maps to. `core/xref` is not directly imported within karma, however.

Using the actual path for ES module version of `idb` will solve the problem and then will break RequireJS because it doesn't support ES module.

saschanaz:
Okay, the tests now pass. Now time to review again! @marcoscaceres @sidvishnoi

src/core/biblio-db.js
```diff
@@ -81,7 +45,8 @@ export const biblioDB = {
     if (await this.isAlias(id)) {
       id = await this.resolveAlias(id);
     }
-    return this.get("reference", id);
+    const result = this.get("reference", id);
```
marcoscaceres: We probably want to await this.get, then return the result.

marcoscaceres:
@saschanaz, it shows you as having requested changes? I guess that’s from an old review?

PR #2231, opened 2019-04-07
Refactor GitHub contributor & commenter display; add tests

### LINKED ISSUE #2182: Tests needed for core/contrib
There are a few things wrong with ```core/contrib```.
https://github.com/w3c/respec/blob/develop/src/core/contrib.js#L44 does not allow for token-only Authentication, which is incorporated in ```core/github```
https://github.com/w3c/respec/blob/develop/src/core/contrib.js#L29 too fails as it requires another asynchronous resolution of the response to get a JSON.
I have solved both of these in #2181 
However, I think we should write tests for this module once that gets merged.

markasoftware:
Resolves #2182. This was a bit trickier to test than I anticipated because some API endpoints reference others; See #2182 for relevant discussion with @devanshbatra04. In the end, I added a `githubAPIBase` parameter to the config. It isn't used in production, but rather just allows the test mocks to have origin-less URLs.

The tests itself are pretty small, but I believe they implicitly cover everything contrib.js has to offer: Making sure it gets data from all sources, works with only either contributors or commenters, sorts the usernames, and prefers a user's "name" over their "login".


src/core/contrib.js
```diff
@@ -52,14 +55,15 @@ export async function run(conf) {
   if (!ghCommenters && !ghContributors) {
     return;
   }
-  const { githubAPI } = conf;
+  const { githubAPI, githubAPIBase } = conf;
```
marcoscaceres: If `githubAPIBase` is always the origin of `conf.githubAPI`, then I don't think we should add it to `conf`. Just do:

```JS
const githubAPIBase = new URL(conf.githubAPI, window.location).origin;
```

And maybe rename `githubAPIBase` to `githubAPIOrigin` so it's more clear.

src/core/github.js
```diff
@@ -51,6 +51,7 @@ export async function run(conf) {
     githubToken: undefined,
     githubUser: undefined,
     githubAPI: `https://api.github.com/repos/${org}/${repo}`,
+    githubAPIBase: `https://api.github.com`,
```
marcoscaceres: See above.

src/core/contrib.js
```diff
@@ -71,22 +74,49 @@ export async function run(conf) {
...
+
+  const [
+    issues,
+    issueComments,
+    otherComments,
+    contributors,
+  ] = await Promise.all(
+    [issues_url, issue_comment_url, comments_url, contributors_url].map(url =>
+      fetchAll(
+        new URL(url.replace(/\{[^}]+\}/, ""), window.location.origin).href,
```
marcoscaceres: Do the `url.replace(/\{[^}]+\}/, "")` into a variable. Then use that variable. when building the URL.

src/core/contrib.js
```diff
@@ -17,13 +17,16 @@ function prop(prop) {
   return o => o[prop];
 }
 const nameProp = prop("name");
-const urlProp = prop("url");
+
+function URLByUser(user) {
```
marcoscaceres: See my previous comments about just inlining this.

src/core/contrib.js
```diff
@@ -71,22 +74,49 @@ export async function run(conf) {
...
+    ? findUserURLs(issues, issueComments, otherComments)
+    : [];
+  const contributorUrls = ghContributors ? contributors.map(URLByUser) : [];
   try {
-    await Promise.all(
-      toHTML(commenterUrls, editors, ghCommenters, headers),
-      toHTML(contributorUrls, editors, ghContributors, headers)
-    );
+    const toHTMLPromises = [];
+    if (ghCommenters) {
```
marcoscaceres: There seems to be a lot of duplication in the following lines... and it looks like it could be `.filter(/* whatever the if is doing */).map( /* to html */)`?

markasoftware:
I'm going out of town for a bit, will fix this when i get back.

src/core/contrib.js
markasoftware: Ooh, good idea! I'm a bit concerned that my implementation is overengineered; You be the judge.

src/core/contrib.js
```diff
@@ -17,13 +17,17 @@ function prop(prop) {
...
+  const { githubAPIBase } = conf;
+  return new URL(user.url, githubAPIBase).href;
+}
+
+function findUserURLs(conf, ...thingsWithUsers) {
   const usersURLs = thingsWithUsers
     .reduce(flatten, [])
     .filter(thing => thing && thing.user)
-    .map(({ user }) => user.url);
+    .map(({ user }) => URLByUser(conf, user));
```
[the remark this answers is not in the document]
markasoftware: I have copied the logic directly into the two spots where it's used. I originally decided to split this into its own function because I saw that `nameProp` was, and it is trivial too.

markasoftware:
I believe I've addressed all of your concerns. Sorry for the delay!

I also merged w3c/develop back into my branch to resolve merge conflicts to conform with 90db5b02e0a0fc1c7823bd486f19efe8ba160f5d, 753f0803ae81d8370cb9dc557af4126d968db44b, and 745be334e702fa712945d726dea4f0b64fccd9db.

src/core/contrib.js
marcoscaceres: No, I think it's good. Easy to follow what's going on.

PR #2232, opened 2019-04-07
fix(core/ui) : Make ReSpec menu navigation smoother with keyboard

### LINKED ISSUE #1064: ReSpec menu not very usable with keyboard
It's possible to tab to, open, and navigate the first level of the ReSpec pill, but once you choose an option, the focus gets lost.

pradeepgangwar:
List of changes:
- Pressing "i" on keyboard now opens the menu (acts as shortcut key).
- Can navigate within menu via Tab key
- When you close a menu overlay, you get back to menu items list to navigate.

Fixes #1064

### DIFF
--- a/src/core/ui.js
+++ b/src/core/ui.js
@@ -59,14 +59,29 @@ respecPill.addEventListener("click", function(e) {
   this.setAttribute("aria-expanded", String(menu.hidden));
   menu.hidden = !menu.hidden;
 });
+document.documentElement.addEventListener("keyup", e => {
+  e.stopPropagation();
+  if (e.keyCode === 73) {
+    if (menu.hidden) {
+      menu.classList.remove("respec-hidden");
+      menu.classList.add("respec-visible");
+      respecPill.focus();
+    } else {
+      menu.classList.add("respec-hidden");
+      menu.classList.remove("respec-visible");
+    }
+    respecPill.setAttribute("aria-expanded", String(menu.hidden));
+    menu.hidden = !menu.hidden;
+  }
+});
+respecUI.appendChild(menu);
 document.documentElement.addEventListener("click", () => {
-  if (!menu.hidden) {
+  if (!menu.hidden && !overlay) {
     menu.classList.remove("respec-visible");
     menu.classList.add("respec-hidden");
     menu.hidden = true;
   }
 });
-respecUI.appendChild(menu);
 
 const ariaMap = new Map([
   ["controls", "respec-menu"],
@@ -161,6 +176,10 @@ export const ui = {
     if (owner) {
       owner.setAttribute("aria-expanded", "false");
     }
+    respecPill.setAttribute("aria-expanded", String(false));
+    respecPill.focus();
+    menu.hidden = false;
+    menu.classList.add("respec-visible");
     if (!modal) return;
     modal.remove();
     modal = null;


src/core/ui.js
```diff
@@ -161,6 +176,10 @@ export const ui = {
     if (owner) {
       owner.setAttribute("aria-expanded", "false");
     }
+    respecPill.setAttribute("aria-expanded", String(false));
```
marcoscaceres: The conversion here is redundant: 

```suggestion
    respecPill.setAttribute("aria-expanded", "false");
```

src/core/ui.js
```diff
@@ -59,14 +59,29 @@ respecPill.addEventListener("click", function(e) {
   this.setAttribute("aria-expanded", String(menu.hidden));
   menu.hidden = !menu.hidden;
 });
+document.documentElement.addEventListener("keyup", e => {
+  e.stopPropagation();
+  if (e.keyCode === 73) {
```
marcoscaceres: I think `keyCode` is deprecated - use [`key`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key) instead.

src/core/ui.js
```diff
@@ -59,14 +59,29 @@ respecPill.addEventListener("click", function(e) {
   this.setAttribute("aria-expanded", String(menu.hidden));
   menu.hidden = !menu.hidden;
 });
+document.documentElement.addEventListener("keyup", e => {
```
```suggestion
document.addEventListener("keyup", e => {
```

I wonder if this should be more specific tho... maybe `body`?

src/core/ui.js
```diff
@@ -59,14 +59,29 @@ respecPill.addEventListener("click", function(e) {
...
+  e.stopPropagation();
```
marcoscaceres: We might need to be careful with this...  should only stop propagation when related to the particular target?

src/core/ui.js
```diff
@@ -59,14 +59,29 @@ respecPill.addEventListener("click", function(e) {
...
+  if (e.keyCode === 73) {
+    if (menu.hidden) {
```
marcoscaceres: The following lines or code are basically identical, so why not:

```JS
const classes = ["visible", "hidden"];
const [toAdd, toRemove] =  menu.hidden ? classes.reverse() : classes;  
menu.classList.remove(`respec-${toRemove}`);
menu.classList.add(`respec-${toAdd}`);
if (menu.hidden) respecPill.focus();
```

Or, if "respec-hidden" is set at construction time, maybe even just:

```JS  
menu.classList.toggle("respec-visible");
menu.classList.toggle("respec-hidden", menu.hidden);
if (menu.hidden) respecPill.focus();
```

src/core/ui.js
```diff
@@ -59,14 +59,29 @@ respecPill.addEventListener("click", function(e) {
...
+      menu.classList.remove("respec-hidden");
+      menu.classList.add("respec-visible");
+      respecPill.focus();
+    } else {
+      menu.classList.add("respec-hidden");
+      menu.classList.remove("respec-visible");
+    }
+    respecPill.setAttribute("aria-expanded", String(menu.hidden));
```
marcoscaceres: This automatically coerces to a string, I believe (or is TypeScript causing an error here?). 
```suggestion
    respecPill.setAttribute("aria-expanded", menu.hidden);
```

marcoscaceres:
I think this is a good start towards fixing  #1064, but need to check. I'd like to be able to tab/shift-tab to each option, and make sure tabbing doesn't escape back out to the document once the ReSpec pill has control. Hitting "enter" should execute a selected option. Hitting esc should close the pill.

src/core/ui.js
saschanaz: AFAIK `window.addEventListener("keyup")`, `document.addEventListener("keyup")`, `document.body.addEventListener("keyup")` etc. works all the same way.

src/core/ui.js
saschanaz: Why do we have both `respec-visible` and `respec-hidden`?

src/core/ui.js
saschanaz: TS disallows automatic coercing so probably.

src/core/ui.js
marcoscaceres: Yeah, can probably roll visible of into the base element’s css definition (or the other way around).

src/core/ui.js
marcoscaceres: I was thinking more from a logical/debugging perspective... make the listener easier to find, reduced potential conflicts with other things higher in the tree.

marcoscaceres: In other words, we should attach this directly on the target... maybe even remove the listener when the menu is hidden... re-attach when shown.

saschanaz: >make the listener easier to find

I should admit this is reasonable because `document.addEventListener` doesn't appear in the DOM tree view.

saschanaz: >we should attach this directly on the target

Isn't the purpose here opening ReSpec menu when user presses `i` wherever the user is?

pradeepgangwar:
Will do the changes.

src/core/ui.js
marcoscaceres: Well, that's something @pradeepgangwar is proposing... not something I'm 100% we should go with yet. What's more important is the up and down keys and tabbing from one option to the next.

pradeepgangwar:
The main problem will be with up/down arrow keys and restricting user's focus to respec pill once that is open. I have an alternate solution in mind. Will try that one and update the PR.

PR #2233, opened 2019-04-07
refactor: dynamic import for assets

saschanaz:
~~Currently blocked by `marked` dependency in `core/utils`:~~

* ~~ES modules cannot reference a package via its package name~~
* ~~`marked` does not have a ES module distribution~~

~~The first problem can be fixed by [import maps](https://github.com/WICG/import-maps/) but currently only [Chrome beta 74](https://www.chromestatus.com/feature/5315286962012160) implements it. The second one is [being worked on](https://github.com/markedjs/marked/pull/1349) but it seems not to be active now.~~

Resolved by #2240


marcoscaceres:
Amazing :) If you want to send this in pieces, it would be awesome.

saschanaz:
Will soon send some smaller pieces from this.

saschanaz:
@marcoscaceres This contains asset loading part and karma part... The former is okay, but the latter requires Chrome because it's the only browser that supports dynamic import now. Should we wait until Firefox stable enables dynamic import?

marcoscaceres:
> Should we wait until Firefox stable enables dynamic import?

No, I don't think we should wait... If we want to test against Firefox, we could figure out how to enable them via setting `javascript.options.dynamicImport`.

marcoscaceres:
Bingo: https://github.com/karma-runner/karma-firefox-launcher/blob/master/README.md#custom-preferences

saschanaz:
Closing for now, because this should really be a part of #2187 or a fix for #1640, and both are not ready yet.

PR #2236, opened 2019-04-08
fix(core/webidl): EventHandler URL was wrong

marcoscaceres:


marcoscaceres:
Found via https://github.com/w3c/wake-lock/pull/176

Issue #2161, opened 2019-03-18
Migrate to webpack built version on tests


saschanaz:
Tests based on html files e.g. pubsubhub-spec.html still depend on requirejs. We may want to use webpack-built ReSpec instead to do #1975.

saschanaz:
I'm not adding `good first issue` because I'm seeing a blocker: currently `makeRSDoc()` with webpack-built ReSpec finishes processing even before inserting our custom test configuration.

devanshbatra04:
Just wanted some clarification. Are you referring to the requireJS usage [here](https://github.com/w3c/respec/blob/develop/tests/spec/core/pubsubhub-spec.html#L21)? Because pubsubhub-spec.html uses ```builds/respec-w3c-common.js``` (which as far as I can tell is the webpack build).
However this Require function comes from [core/expose-modules](https://github.com/w3c/respec/blob/develop/src/core/expose-modules.js) which isn't RequireJS (at least in case of pubsubhub)
 
Other HTML files ```webidl.html, highlight.html, includer.html and pre-process.html``` do use the RequireJS profile, I ran into some issue on [webidl.html](https://github.com/w3c/respec/blob/develop/tests/spec/core/webidl.html) but the others do work well by simply changing the script src.

saschanaz:
I think you understand correctly. I think some other tests also had problems but I might be wrong.

devanshbatra04:
WebIDL spec does have this problem. trying to come up with a solution.

devanshbatra04:
I worked on this for several hours today.The issue encountered is only in ```core/webidl.html``` as of now.
I think I know exactly what is happening. 
When we fetch the dfn element form [definiton map here](https://github.com/w3c/respec/blob/develop/src/core/link-to-dfn.js#L103), In test environment with webpack build, we get an object that is identical to the dfn element originally pushed here. 

However this ```dfn``` object is not a DOM Element, and any operations on it do not show up in the DOM. 
This causes addId function [here](https://github.com/w3c/respec/blob/develop/src/core/link-to-dfn.js#L139) useless. As it attempts to set an ID on this static object copy of the required dfn element. Making it look like the tests executed before ```awaitRSDoc()``` was complete and the ID remained Unset. It causes the tests to fail which require dynamically setting of id attribute on the dfn elements.

devanshbatra04:
This doesn't happen on running the HTML file in browser where the Map stores the actual HTMLElements. This is probably a bug.

[...]

PR #2241, opened 2019-04-09
refactor: use ES modules for karma tests

saschanaz:
Turns out that the asset loader part of #2233 wasn't strictly needed. (But still helpful for #1640.)

I recommend to review this PR commit-by-commit, as this contains many files but the changes are rather simple.


tests/spec/core/highlight.html
```diff
@@ -3,8 +3,7 @@
   <head>
     <meta charset='utf-8'>
     <title>Replace me with a real title</title>
-    <script class='remove' src='/js/deps/require.js' data-main='/js/profile-w3c-common'>
-    </script>
+    <script class='remove' src='/builds/respec-w3c-common.js'></script>
```
marcoscaceres: I'm worried about this one, because it means we need to pre-build before testing, which we don't normally do. We might need to come up with an alternative solution.

marcoscaceres: Actually, ignore me... this only applies to Karma, which on Travis is built so we are fine.

saschanaz: >it means we need to pre-build before testing

I think you already are required to pre-build, otherwise even the requirejs build won't get the latest code.

That said, this change wasn't to be included here... I'll roll this back.

saschanaz: >Actually, ignore me... this only applies to Karma, which on Travis is built so we are fine.

Oh, okay then 😄

saschanaz:
>I guess we will need to do this for all the tests, because they currently all rely on require... hmmm....

#2161 👍

PR #2245, opened 2019-04-10
refactor(tests): Deprecate SpecRunner.html and move to html reporter

### LINKED ISSUE #2244: Deprecate SpecRunner.js
We have two ways of running tests... SpecRunner and Karma... we should only have one: Karma. 

We should look at removing SpecRunner.html and make sure any documentation that lists it gets updated too.

Swapnilr1:
Fixes #2244 

I will update the Developers Guide also.


Swapnilr1:
Yup, I also find `SpecRunner` more convenient than running karma tests, and once ReSpec moves to webpack I would definitely want to see if this can be patched and added back.

marcoscaceres:
We should check https://www.npmjs.com/package/karma-jasmine-html-reporter

Swapnilr1:
> We should check https://www.npmjs.com/package/karma-jasmine-html-reporter

Great! Will look into it once my existing PRs are merged.

Swapnilr1:
OK, this was easy to add.

marcoscaceres:
Tested locally with BROWSERS set... works great!

PR #2249, opened 2019-04-11
feat(w3c): make a jQuery-less profile (w3c)

marcoscaceres:
How about we make a new profile without jQuery? It shaves 30kb and we can move new specs that don't need jQuery off w3c-common.


examples/basic.built.html
```diff
@@ -3,7 +3,7 @@
   <head>
     <meta charset='utf-8'>
     <title>Replace me with a real title</title>
-    <script src="../builds/respec-w3c-common.js" class='remove'></script>
+    <script src="../builds/respec-w3c.js" class='remove'></script>
```
saschanaz: We have several tests that uses respec-w3c-common, those should be fixed as well.

marcoscaceres: I'll change those too.

marcoscaceres:
Sorry for all the noise... I think this is good now.

PR #2251, opened 2019-04-11
fix(w3c/headers): Add Preview of PR to title

### LINKED ISSUE #2030: Show "Preview of #" in title
See the following for details:
https://github.com/tobie/pr-preview/issues/23

Basically, the spec's title should come "Preview of PR xxx: " when the spec `isPreview` is true. 

This would build on @yatri1609's work: 
https://github.com/w3c/respec/pull/1843

Swapnilr1:
Fixes #2030 

This is how it looks: 
pr

Just confirm whether this is the intended fix and I will add a test for it.


tests/spec/w3c/headers-spec.js
```diff
@@ -1408,4 +1408,13 @@ describe("W3C — Headers", () => {
...
     });
   });
+
+  it("adds Preview status to title", async () => {
+    const ops = makeStandardOps();
+    const doc = await makeRSDoc(
+      ops,
+      "spec/core/simple.html?isPreview=true&prNumber=123"
+    );
+    expect(doc.title).toBe("Preview of PR #123: Simple Spec");
```
marcoscaceres: And make sure the H1 is the same :)

src/w3c/headers.js
```diff
@@ -273,7 +273,7 @@ export function run(conf) {
     }
   }
   const title = document.title || "No Title";
-  conf.title =
+  document.title = conf.title =
```
marcoscaceres: We generally avoid the double assignments.

marcoscaceres:
I would be really nice to also link the issue number in the H1... that will take the user back to the correct pull request on Github.

Swapnilr1:
I need a small clarification here: Should we add "Preview of PR #" only if `isPreview`, `prNumber` and `prURL` all are configured or even when only `isPreview` and `prNumber` (so here no linking to the PR) are configured?

Swapnilr1:
@marcoscaceres Is there a simpler way to prepend the text than what I have done?

src/w3c/headers.js
```diff
@@ -273,6 +273,9 @@ export function run(conf) {
     }
   }
   conf.title = document.title || "No Title";
+  if (document.title && conf.isPreview && conf.prNumber) {
```
marcoscaceres: I'm unsure about the first check for "document.title"?

src/w3c/templates/headers.js
```diff
@@ -22,6 +22,20 @@ function getSpecTitleElem(conf) {
     specTitleElem.textContent = conf.title;
     specTitleElem.id = "title";
   }
+  if (conf.isPreview && conf.prNumber) {
+    if (conf.prUrl)
```
marcoscaceres: as a general rule, please always use "{" and "}" for multiline if statements.

src/w3c/templates/headers.js
```diff
@@ -22,6 +22,20 @@ function getSpecTitleElem(conf) {
...
+      specTitleElem.prepend(
```
marcoscaceres: you could do:
```JS
 const {childNodes}  = html`Preview of PR <a href=${conf.prUrl}>#${conf.prNumber}</a>:`;
```
And that gets rid of the span.

marcoscaceres:
> I need a small clarification here: Should we add "Preview of PR #" only if isPreview, prNumber and prURL all are configured or even when only isPreview and prNumber (so here no linking to the PR) are configured?

We can ask that we always get the `prURL`, then we don't need to special case it.

src/w3c/templates/headers.js
```diff
@@ -22,6 +22,20 @@ function getSpecTitleElem(conf) {
...
+        html`
+          <span
+            >Preview of PR <a href=${conf.prUrl}>#${conf.prNumber}</a>:
```
marcoscaceres:
```suggestion
            >Preview of PR <a href="${conf.prUrl}">#${conf.prNumber}</a>:
```

marcoscaceres: As convention of the project, we like using the quotes for attributes (even if optional).

Swapnilr1: My bad, how did this even work at all when I was testing? :fearful:

src/w3c/headers.js
Swapnilr1: If I don't check this, and document.title is undefined, I would be setting document.title to be equal to "Preview of PR #XYZ: undefined". But if I leave it undefined when it is undefined, it would be set by  https://github.com/w3c/respec/blob/3c0dda705c7de88f6ec68a913d4b34d8126964d0/src/w3c/templates/headers.js#L27

src/w3c/templates/headers.js
Swapnilr1: Can I use innerHTML instead? ESLint causes the the string to be split across lines, adding whitespace, which I cannot trim.

src/w3c/headers.js
marcoscaceres: ok, thanks for clarifying. I was unsure.

src/w3c/templates/headers.js
marcoscaceres: we prefer not to use `innerHTML` because hyperHTML gives us some security assurances. If you need to disable ESLint for a particular line, then that's acceptable. Just put a comment.

Swapnilr1: A comment asking not to remove the comment disabling eslint prettier?

tests/spec/w3c/headers-spec.js
```diff
@@ -1408,4 +1408,42 @@ describe("W3C — Headers", () => {
...
+        },
+        body: `<h1 id='title'>Simple Spec</h1>${makeDefaultBody()}`,
+      };
+      const doc = await makeRSDoc(ops);
+      expect(doc.title).toBe("Preview of PR #123: Simple Spec");
+      const headingtitle = doc.querySelector("h1#title");
+      expect(headingtitle.textContent).toEqual(
+        "Preview of PR #123: Simple Spec"
+      );
+      expect(headingtitle.querySelector("a").href).toEqual(
```
marcoscaceres:
```suggestion
      expect(headingtitle.querySelector("a").href).toBe(
```

tests/spec/w3c/headers-spec.js
```diff
@@ -1408,4 +1408,42 @@ describe("W3C — Headers", () => {
...
+  describe("Add Preview Status in title", () => {
+    it("when document title is present", async () => {
+      const ops = makeStandardOps();
+      const doc = await makeRSDoc(
+        ops,
+        "spec/core/simple.html?isPreview=true&prNumber=123&prUrl=%22http%3A//w3c.github.io/respec/%22"
+      );
+      expect(doc.title).toBe("Preview of PR #123: Simple Spec");
+      const headingtitle = doc.querySelector("h1#title");
+      expect(headingtitle.textContent).toEqual(
```
marcoscaceres:
```suggestion
      expect(headingtitle.textContent).toBe(
```

tests/spec/w3c/headers-spec.js
```diff
@@ -1408,4 +1408,42 @@ describe("W3C — Headers", () => {
...
+          isPreview: true,
+          prNumber: 123,
+          prUrl: "http://w3c.github.io/respec/",
+        },
+        body: `<h1 id='title'>Simple Spec</h1>${makeDefaultBody()}`,
+      };
+      const doc = await makeRSDoc(ops);
+      expect(doc.title).toBe("Preview of PR #123: Simple Spec");
+      const headingtitle = doc.querySelector("h1#title");
+      expect(headingtitle.textContent).toEqual(
```
marcoscaceres:
```suggestion
      expect(headingtitle.textContent).toBe(
```

PR #2255, opened 2019-04-12
refactor(core/xref): use xref api v4, update tests

sidvishnoi:
Yes. Huge PR in single commit 🙈 
- core/xref is almost a re-write. Would advice side-by-side diff.
- Added two more functions to IDlKeyVal. They do multiple operations in a single transaction.
- Moved tests data to separate files. Now there is no external API request on these tests.
- Added a `types.d.ts`, which might be a good option for module specific types (interfaces in JSDoc get really ugly)


src/types.d.ts
```diff
@@ -0,0 +1,25 @@
+module "core/xref" {
```
marcoscaceres: 👋👶 "types-escript" noob here ... we have ./src/type-helper.d.ts ... could this go in there?

src/core/xref.js
```diff
@@ -164,150 +110,117 @@ function normalizeConfig(xref) {
...
 
-// adds data to cache
-async function cacheResults(data, cache) {
-  const promisesToSet = Object.entries(data).map(([term, results]) =>
-    cache.set(term, results)
-  );
-  await cache.set("__CACHE_TIME__", Date.now());
-  await Promise.all(promisesToSet);
+  // merge above two Maps. faster than following equivalent:
+  // return new Map([...resultsFromCache, ...fetchedResults]);
```
marcoscaceres: I think you should do the: 

```JS
 return new Map([...resultsFromCache, ...fetchedResults]);
```

Let engines catch up on perf. Don't work around it. It's also going to be a tiny per hit.

src/core/xref.js
```diff
@@ -330,176 +244,155 @@ function isNormative(elem) {
...
-        const existingKey = conf.informativeReferences.has(cite)
-          ? conf.informativeReferences.getCanonicalKey(cite)
-          : cite;
-        conf.normativeReferences.add(existingKey);
-        conf.informativeReferences.delete(existingKey);
-        continue;
+      default: {
+        const collector =
+          errorsAmbiguous.get(term) ||
+          errorsAmbiguous.set(term, { specs: new Set(), elems: [] }).get(term);
```
marcoscaceres: cute

sidvishnoi: cute means don't do it? ❓

marcoscaceres: no, means it's a clever solution :)

PR #2257, opened 2019-04-13
feat(core/inlines): Mouseover for variable data type

Swapnilr1:
Added mouseover for data type. Yet to add propagating variable type to other occurences.
Tests will be added when implementation is approved.


Swapnilr1:
Yeah sure, will do.

marcoscaceres:
Note, don’t use balloon tho... it’s full of bloat.

Swapnilr1:
Is [this](https://www.w3schools.com/css/tryit.asp?filename=trycss_tooltip_arrow_top) acceptable? Looks similar to me. If not, please suggest something. I use CSS very minimally.

sidvishnoi:
@Swapnilr1 Ideally, we would like to make it work with following markup (and CSS) only:
``` html
<var data-type="type">identifier</a>
```
So, the one above at w3schools won't do well. This might be handy: https://codepen.io/pure-css/pen/bddggP

Swapnilr1:
@sidvishnoi Yeah, also I just saw W3Schools is copyrighted content, so probably can't use that. What you linked to looks good. 
I have no opinion on what to use. Maybe you and @marcoscaceres can discuss this.

marcoscaceres:
I think we should just strip down balloon (I prefer the way it looks and animates) and only use “above”.

Swapnilr1:
> I think we should just strip down balloon (I prefer the way it looks and animates) and only use “above”.

So to be clear, we won't add `balloon-css` as a dependency since that gives the entire `balloon.css` file? So do we just copy a part of `balloon.css` into `assets`? Or is there a way to selectively import CSS rules?

sidvishnoi:
> strip down balloon

Take the good parts that we need, and write it in, say `assets/var.css`

saschanaz:
Tried some tweaking: https://codepen.io/SaschaNaz/pen/KYXBPg

marcoscaceres:
I think we just need a sans-serif font on that and a speedy fade in and we have a winner 🏆

marcoscaceres:
Tried with a slightly larger interface name and it kinda goes wonky: https://codepen.io/anon/pen/YMrdyw

Balloon seems to have figured out how to keep things scalable.

sidvishnoi:
Here's my take on this: https://codepen.io/sidvishnoi/pen/oOGmOy?editors=0100

marcoscaceres:
Works nicely... added a little more padding and a little roundness:

https://codepen.io/anon/pen/bJoZRp

Swapnilr1:
@marcoscaceres Looks great, apart from a little problem:

tooltip

Basically resize the window just right to get the text over which tool-tip is to appear at the very starting of a line - happens in both Firefox 66, Chrome 73.

marcoscaceres:
I wonder if we need to then display vars as inline blocks?

Swapnilr1:
Yes, displaying as inline-block fixes that, and doesn't seem to cause any new problems.

Swapnilr1:
A few questions:
1. Let's say I get `First instance of |p|, second instance of |p:int| and third instance of |p:long long|` in a section. I should show an error that data types are inconsistent, but should I apply a data type at all to the first instance and other instances without data types? Or should I show the error but over-write all instances with the last data type encountered?

2. Should I rename `highlight-vars.js` to `process-vars.js` and add the data type propagation? Or create a separate module?

marcoscaceres:
> Let's say I get First instance of |p|, second instance of |p:int| and third instance of |p:long long| in a section. I should show an error that data types are inconsistent, but should I apply a data type at all to the first instance and other instances without data types? Or should I show the error but over-write all instances with the last data type encountered?

Good question. If for whatever reason the Editor chooses to switch types, that's ok. The new type will override the old type.

> Should I rename highlight-vars.js to process-vars.js and add the data type propagation? Or create a separate module?

Depends... if you are going to piggyback the implementation on the highlight vars code, then: we can probably keep the current name, just happens to have this additional functionality. It's still part of "highlighting vars" (part of the highlight is that you get type information).  

If it looks like it's different enough, we would call the module data-var.js, as we've done with others... that reflects clearly what it's being applied to. I'd be inclined to just do a new module called data-var.js... the processing of var elements will be rather quick, so we might as well make a new module (though feel free to steal ideas from highlight-vars).

Swapnilr1:
> If for whatever reason the Editor chooses to switch types, that's ok. The new type will override the old type.

I was thinking about an unintentional type change. Let's say in a fairly large section, the author accidentally re-uses the same variable but different data type - not realizing it's part of the same section. 
But I guess that's unlikely to happen.

marcoscaceres:
That could definitely happen.. but it's complicated because we don't have real scopes. like:

```
the algorithm of (things, moreThings):
  for each (|thing:Int| of things) do:
     for each (|thing:string| of moreThings) do: 
    
```

Swapnilr1:
OK, overriding data type it is then. No warnings, or errors to be shown?

marcoscaceres:
> OK, overriding data type it is then. No warnings, or errors to be shown?

For now no... we can then test it out in the field and see what Editors say they would like.

marcoscaceres:
(note: we can add option type checking as a pref later if we want)

Swapnilr1:
Right now the tooltip is appearing when hovering over the place where the tooltip is supposed to appear, rather than hovering over the text itself. Looking into it.
Also I realize the implementation is a bit messy, feedback appreciated while I am also thinking of ways to simplify it.

src/core/data-type.js
```diff
@@ -0,0 +1,41 @@
...
+export function run(conf) {
+  if (!conf.highlightVars) {
+    return;
+  }
+
+  const style = document.createElement("style");
+  style.textContent = tooltipStyle;
+  document.head.appendChild(style);
+
+  document.querySelectorAll("section").forEach(section => {
```
marcoscaceres: Alternatively, query “section var”, then you can check if you are still in the same section by using something like varElement.closest(“section”) === lastSeenSection. 

You could also keep a map per section, to quickly look up a var’s type by name.

Swapnilr1: I don't know, querying `var` separately for each section seems more intuitive to me than querying `section var`. Any benefits of querying “section var” I am missing? 

The map idea is great, will do.

assets/datatype.css
```diff
@@ -0,0 +1,49 @@
...
+    transform: translateX(-50%) translateY(-100%);
+    background: var(--tooltip-background);
+    text-align: center;
+    /* additional styling */
+    color: var(--tooltip-color);
+    font-family: "Dank Mono", "Fira Code", monospace;
+    font-style: normal;
+    padding: 6px;
+    border-radius: 3px;
+    color: rgb(218,202,136);
```
sidvishnoi: For consistency how we're using colors everywhere else

```suggestion
    color: #daca88;
```

Swapnilr1: Will make the change.

assets/datatype.css
```diff
@@ -0,0 +1,49 @@
+var {
```
sidvishnoi: Please run this file through prettier once

assets/datatype.css
```diff
@@ -0,0 +1,49 @@
...
+    border-top-color: var(--tooltip-background);
+  }
+  /* actual text */
+  var[data-type]::after {
+    content: attr(data-type);
+    transform: translateX(-50%) translateY(-100%);
+    background: var(--tooltip-background);
+    text-align: center;
+    /* additional styling */
+    color: var(--tooltip-color);
```
sidvishnoi: This is redundant (overwritten just below)
```suggestion
```

assets/datatype.css
```diff
@@ -0,0 +1,49 @@
+var {
+    position: relative;
+    cursor: pointer;
+    display: inline-block;
+  }
+  
+  var[data-type] {
```
sidvishnoi: Lets remove this entirely and hardcode these values below.

Swapnilr1:
Thanks @sidvishnoi - I didn't really look at the CSS when adding it. I will examine it and make the necessary changes.

src/core/data-type.js
marcoscaceres: > Any benefits of querying “section var” I am missing?

if flattens the algorithm, but the trade off is then needing to run `.closest()`

So, something like:
 
```JS
let section = null;
const varMap = new Map();
for (const varElem of document.querySelectorAll("section var")) {
  const currentSection = varElem.closest("section");
  if (lastSeenSection !== currentSection) {
     section = currentSection;
     varMap.clear();
  }
  if (varElement.data.type) {
     varMap.add(varElement.textContent.trim(), varElement.data.type);
     continue;
  }
  const type = varMap.get(varElement.textContent.trim());
  if (type) var.dataset.type = type;
}
```

marcoscaceres: fixed some typos above.

Swapnilr1:
> Right now the tooltip is appearing when hovering over the place where the tooltip is supposed to appear, rather than hovering over the text itself.
Was easier to fix than expected.

@marcoscaceres I am not selecting "section var" currently. Since the current behavior is to override all the instances with the type of the last instance of the variable in the section, even using `document.querySelectorAll("section var")` would have required another loop inside to process the vars in a section again. Using map made the code much more concise, and when I tried using `document.querySelectorAll("section var")` the code was getting messy, since you need to check when a section ends to run through that and override the types. 

Will add tests also.

marcoscaceres:
> @marcoscaceres I am not selecting "section var" currently. Since the current behavior is to override all the instances with the type of the last instance of the variable in the section, even using document.querySelectorAll("section var") would have required another loop inside to process the vars in a section again. 

I think you might misinterpret the requirement: it’s to *incrementally* override the type as you iterate. Right now, you would assign the last seen type, which wouldn’t be correct. 

Please also note, you are querying all sections even though the vast number of sections in a spec won’t contain any vars. Using “section var” assures you never process section that have no vars in them.

saschanaz:
>you are querying all sections

`section var` also queries those sections while they won't be in the resulting collection.

saschanaz:
>Please also note, you are querying all sections even though the vast number of sections in a spec won’t contain any vars.

I have to constantly remind for myself that [the last part of the selector will be first evaluated](https://csswizardry.com/2011/09/writing-efficient-css-selectors/).  I have to admit I'm not a CSS expert 😅

marcoscaceres:
Just to be clear, my comments are less about performance and more around code complexity and maintenance. Performance changes will be negligible irrespective of doing two queries or one. However, if anything goes wrong as we continue to develop this out, the unnecessary sections will add to the complexity of debugging.

Swapnilr1:
> I think you might misinterpret the requirement: it’s to incrementally override the type as you iterate. Right now, you would assign the last seen type, which wouldn’t be correct.

My bad, I thought that the intended behavior was that a variable can have only one data type within a section, so I propagated this data type overriding all previous and subsequent instances.
So now to confirm, we only propagate data type to subsequent instances, somewhat like JavaScript does, right?

marcoscaceres:
> So now to confirm, we only propagate data type to subsequent instances, somewhat like JavaScript does, right?

Yep, correct.

Swapnilr1:
I think this is mostly ready, except maybe more tests can be added but I tried to cover all cases. Please review this. A minor problem I noticed while testing was that for types with spaces, the tooltip does not increase in width but rather takes more lines.  Though I guess that is acceptable.

tests/spec/core/data-type-spec.js
```diff
@@ -133,4 +133,58 @@ describe("Core — data-type attribute", () => {
...
+    const body = `
+    <section>
+      <p id="a1">First instance of |varA|.</p>
+      <p id="a2">Second instance of |varA: Type1|.</p>
+      <p id="a3">Third instance of |varA|.</p>
+      <p id="a4">Fourth instance of |varA: Type2|.</p>
+      <p id="b1">First instance of |varB:Type1|.</p>
+      <p id="a5">Fifth instance of |varA|.</p>
+    </section>
+    <section>
```
marcoscaceres: Let’s nest this section inside the previous section, so it confirms that nested sections work as expected.

Swapnilr1: Done. :)

PR #2261, opened 2019-04-14
feat(core/inlines): handle [= inline links =]

marcoscaceres:


marcoscaceres:
@sidvishnoi, this is currently testing against the ReSpec.org... but would like to test against local xref dummy data. Could you please guide me to creating the dummy data for these?

sidvishnoi:
Currently, the method I used for local xref data is: Send the request with `xref: true`, copy the response from devtools network panel, write it locally in a file and use it. 

Also, if you want the query also (so that we know what we're sending - makes test files easier to reason about), replace this with:
https://github.com/w3c/respec/blob/bba846d94db63a7f8fb40fc0e747b5b269192601/src/core/xref.js#L219

``` js
const query = { keys, options: { query: true } }; 
```

The reason we're doing above is dynamically generated `id` (which is SHA-1 of request key, which is used for caching and referencing on server and client). Would it be a good idea to specify a custom `data-test-id` on link's parent for testing so we don't rely on that hashing and create local test json files using those ids? Or, we can use an auto-increment number for each xref element to specify the `id`.

src/core/inlines.js
```diff
@@ -115,6 +115,15 @@ function inlineVariableMatches(matched) {
   return hyperHTML`<var data-type="${type}">${varName}</var>`;
 }
 
+function inlineLinkMatches(matched) {
+  const parts = matched
+    .slice(2, -2) // Chop [= =]
+    .split(/\/(?:\s?)/) // Only at first "/"
+    .map(s => s.trim());
```
sidvishnoi: If we just want to split at only the first `/`, we can do`.split("/", 2)`. Are there any pitfalls?

src/core/inlines.js
```diff
@@ -137,7 +146,11 @@ export function run(conf) {
   const abbrRx = aKeys.length ? `(?:\\b${aKeys.join("\\b)|(?:\\b")}\\b)` : null;
 
   // PROCESSING
-  const txts = getTextNodes(document.body, ["pre"]);
+  // Don't gather text nodes for these:
+  const exclusions = ["#respec-ui", ".head", "pre"];
+  const txts = getTextNodes(document.body, exclusions, {
+    wsNodes: false, // we don't nodes with just whitespace
```
```suggestion
    wsNodes: false, // we don't want nodes with just whitespace
```

src/core/inlines.js
marcoscaceres: no pitfalls, I'd just forgotten about that second argument

tests/spec/core/xref-spec.js
```diff
@@ -5,6 +5,7 @@ import {
   makeDefaultBody,
   makeRSDoc,
   makeStandardOps,
+  xrefTestUrl,
```
sidvishnoi: I think this function should be here only. Can't think of a use case to keep it in SpecHelper

marcoscaceres: thought it would be nice because it's now used in two files

marcoscaceres: (also, there is inline tests that should probably be in the inline-spec file... like the inline idl tests)

sidvishnoi: Which other file?

marcoscaceres: inlines-spec.js

sidvishnoi: Yes. need to refactor these tests. xref-spec is too large

sidvishnoi: In `inlines-spec.js`, we won't be making xref requests, right?

marcoscaceres: https://github.com/w3c/respec/pull/2261/files#diff-e708af9e164f9fa08c467555607eee1cR7

marcoscaceres: yes, for the  `[= For / thing  =]` tests, for example.

Issue #2259, opened 2019-04-13
Incorrect rendering of inline enum types


marcoscaceres:
Given: 

```HTML
<pre class="idl">
    enum WakeLockType { "screen", "system" };
</pre>

<p>{{ WakeLockType["system"] }}</p>
```

It's getting rendered as `WakeLockType["system]"`

![Screen Shot 2019-04-14 at 7 56 54 AM](https://user-images.githubusercontent.com/870154/56085700-e612cb80-5e8a-11e9-8411-27d41cf5396e.png)

PR #2262, opened 2019-04-14
Inline idl overmatch

### LINKED ISSUE #2258: Overmatching on enum values when two are on the same line
## Description of problem
Given: 
```HTML
<pre class="idl">
   enum WakeLockType { "screen", "system" };
</pre>
<p>{{ WakeLockType["screen"] }} and {{ WakeLockType["system"] }}</p>
```

It overmatches and gets a syntax error.

marcoscaceres:
fixes #2258 

Kinda ugly... can anyone think of a better solution?

### DIFF
--- a/src/core/inlines.js
+++ b/src/core/inlines.js
@@ -147,7 +147,7 @@ export function run(conf) {
       "\\b(?:NOT\\s+)?REQUIRED\\b",
       "\\b(?:NOT\\s+)?RECOMMENDED\\b",
       "\\bOPTIONAL\\b",
-      '(?:{{[^}}]*}})', // inline IDL references,
+      "(?:{{[^}}]*}})", // inline IDL references,
       "\\B\\|\\w[\\w\\s]*(?:\\s*\\:[\\w\\s&;<>]+)?\\|\\B", // inline variable regex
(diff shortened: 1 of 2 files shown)


sidvishnoi:
``` js
const re = /{{[^}}]*}}/
```

image

marcoscaceres:
reading https://mariusschulz.com/blog/why-using-in-regular-expressions-is-almost-never-what-you-actually-want#lazy-quantifiers-to-the-rescue

src/core/inlines.js
```diff
@@ -147,7 +147,7 @@ export function run(conf) {
       "\\b(?:NOT\\s+)?REQUIRED\\b",
       "\\b(?:NOT\\s+)?RECOMMENDED\\b",
       "\\bOPTIONAL\\b",
-      "(?:{{2}\\s*.*\\s*}{2})", // inline IDL references,
+      "(?:{{[^}}]+}})", // inline IDL references,
```
saschanaz: As @sidvishnoi said this should really be `[^}]`, because exclude `}` twice is redundant.

marcoscaceres: oh whoops. Merged too quick.

sidvishnoi: I'll send fix with https://github.com/w3c/respec/issues/2259

PR #2267, opened 2019-04-14
fix(w3c/headers): mismatched title warning

Swapnilr1:
This is a continuation to #2251 fixing a small bug with the implementation, when both `h1` and `document.title` are present and `isPreview`, `prUrl`, and `prNumber` are set, then ReSpec emits a warning `The document's title and the "<title>" element differ.` due to whitespace being present.

Sorry for not testing this initially. 

I don't think I can have a test for this until #1084 is fixed.

### DIFF
--- a/src/w3c/templates/headers.js
+++ b/src/w3c/templates/headers.js
@@ -27,7 +27,7 @@ function getSpecTitleElem(conf) {
     `;
     specTitleElem.prepend(...childNodes);
   }
-  conf.title = specTitleElem.textContent.trim();
+  conf.title = specTitleElem.textContent.replace(/\s{2,}/g, " ").trim();
   specTitleElem.classList.add("title", "p-name");
   if (document.querySelector("title") === null) {
     document.title = conf.title;
--- a/src/w3c/templates/headers.js
+++ b/src/w3c/templates/headers.js
@@ -1,5 +1,6 @@
 // @ts-check
 import html from "hyperhtml";
+import { norm } from "../../core/utils.js";
 import { pub } from "../../core/pubsubhub.js";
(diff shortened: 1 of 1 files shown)


src/w3c/templates/headers.js
```diff
@@ -27,7 +27,7 @@ function getSpecTitleElem(conf) {
     `;
     specTitleElem.prepend(...childNodes);
   }
-  conf.title = specTitleElem.textContent.trim();
+  conf.title = specTitleElem.textContent.replace(/\s{2,}/g, " ").trim();
```
marcoscaceres: Probably want `/\s\s+/g` here, no? Just normalizing whitespace?  

```suggestion
  conf.title = specTitleElem.textContent.replace(/\s{2,}/g, " ").trim();
```

sidvishnoi: We have a `norm` in utils.js

marcoscaceres: even better!

Swapnilr1: @marcoscaceres Sorry, what's the difference between `/\s{2,}` and `/\s\s+`? 
@sidvishnoi Yes, we can use that. Thanks.

Swapnilr1:
I hope there aren't any other issues with #2251.

marcoscaceres:
"Only where there is no code there is no bugs" - ancient proverb

src/w3c/templates/headers.js
marcoscaceres: > Sorry, what's the difference between /\s{2,} and /\s\s+?
I guess they are the same... just more used to seeing the other.

PR #2273, opened 2019-04-15
feat(core/inlines): support inline `code`

marcoscaceres:


marcoscaceres:
Apologies for force push... commit message got screwed up.

tests/spec/core/inlines-spec.js
```diff
@@ -188,6 +188,35 @@ describe("Core - Inlines", () => {
...
+    expect(simple).toBeTruthy();
+    expect(simple.textContent).toBe("null");
+
+    // multi per line
+    const multi = doc.querySelectorAll("#multi code");
+    expect(multi.length).toBe(3);
+    const [firstCode, secondCode, thirdCode] = multi;
+    expect(firstCode.textContent).toBe("123");
+    expect(secondCode.textContent).toBe("undefined");
+    expect(thirdCode.textContent.endsWith("string")).toBeTruthy();
```
saschanaz: Wouldn't `multi[i]` be more readable here?

src/core/inlines.js
```diff
@@ -165,6 +170,7 @@ export function run(conf) {
       "(?:\\[\\[(?:!|\\\\|\\?)?[A-Za-z0-9\\.-]+\\]\\])",
       "(?:\\[\\[\\[(?:!|\\\\|\\?)?[A-Za-z0-9\\.-]+\\]\\]\\])",
       "(?:\\[=[^=]+=\\])", // Inline [= For/link =]
+      "(?<!`)(?:`[^`]+`)", // Inline `code`
```
saschanaz: Should we get a way to escape `` ` ``?

tests/spec/core/inlines-spec.js
marcoscaceres: yeah, you are right. Will fix.

src/core/inlines.js
marcoscaceres: Thought about that too... I've not run into that in practice in specs, so didn't bother. Another one is "[= link to `thing` in other spec =]". At least this gives us the 90% use case of `true` `false` `null` `void`... etc.

saschanaz: We can just fall back to `<code>` if we get one, so maybe okay for now.

marcoscaceres: yeah, exactly.

PR #2275, opened 2019-04-15
fix(core/inlines): RegExp syntax error 🔥

marcoscaceres:
Broke Safari and Firefox, yay! 🥳


saschanaz:
Might be great if we have a tracking issue on Firefox side.

marcoscaceres:
Yeah, agree... I honestly don't even know what to search for.

marcoscaceres:
maybe https://bugzilla.mozilla.org/show_bug.cgi?id=1362154

saschanaz:
"Regex lookahead" seems the right keyword. AFAIK named group is another separate feature.

marcoscaceres:
Might need also try enabling Firefox again on Travis... so flaky tho... but maybe better since last time, specially now that we can do headless mode.

saschanaz:
Actually "lookbehind", the relevant issue: https://bugzilla.mozilla.org/show_bug.cgi?id=1225665

marcoscaceres:
Thank for finding it.

PR #2277, opened 2019-04-15
test: add firefox to travis

### LINKED ISSUE #2276: Try enabling Firefox on travis again..
A about a year has passed since we last tried this... might be fun to try Firefox again, now that it has a headless mode. Would have caught #2275

saschanaz:
Closes #2276

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,6 @@ node_js:
 cache:
   directories:
     - node_modules # NPM packages
-env:
-  - BROWSERS=ChromeHeadless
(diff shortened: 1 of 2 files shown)


saschanaz:
We could theoretically even use `puppeteer-firefox` (which is experimental yet) to run headless tests.

saschanaz:
@marcoscaceres [The document says `latest` means `product=firefox-latest`](https://docs.travis-ci.com/user/firefox/), so it should be version 66 now.

PR #2278, opened 2019-04-15
fix(core/inline-idl-parser): enumValueRegex too restrictive

marcoscaceres:
Now uses the regex directly from WebIDL https://heycam.github.io/webidl/#prod-string


marcoscaceres:
Hit this while converting on screen orientation spec.

marcoscaceres:
Yeah, I’ll add “foo-bar” one and “”.

PR #2282, opened 2019-04-16
refactor(tools): use fs.promises API

saschanaz:
Uses `fs.promises` which is [now marked as stable](https://nodejs.org/api/fs.html#fs_fs_promises_api).


tools/builder.js
```diff
@@ -2,7 +2,7 @@
 
 "use strict";
 const colors = require("colors");
-const fsp = require("fs-extra");
```
sidvishnoi: We can remove `fs-extra` from package.json now, right?

sidvishnoi:
Should we update this too?
https://github.com/w3c/respec/blob/2c073773d343e3a93196ad20307d5508a3df83d0/package.json#L7

tools/builder.js
saschanaz: `copydeps` still uses `fs-extra` for its own utility functions, but maybe we could cook our own ones.

saschanaz:
>Should we update this too?

Ah yes... but should we defer this until version 8 losses its support? Not sure.

sidvishnoi:
`fs.promises` isn't available in node 8, right? It'll break if use node 11 features there

tools/builder.js
sidvishnoi: I checked. `cp` is being used only for files (not directories), so we can use the native one easily. And we have `fs.rmdir` and `fs.unlink` for `fsp.remove` 🎉

saschanaz:
The API is introduced in Node.js 10, but yes, it will break 8.

saschanaz:
>This is in node lts, right?

Sure 😃

marcoscaceres:
We should target active lts only. Ok to delete 8.

saschanaz:
One notable thing is that https://github.com/nodejs/node/pull/26581 is not backported into 10 LTS yet, so using ReSpec there may emit some warnings.

saschanaz:
Like this... Looks like we should wait a bit more 😅

image

sidvishnoi:
Lets pipe it through `grep -v "ExperimentalWarning: The fs.promises API is experimental"` 😛

marcoscaceres:
The warnings are ok IMO. Reading the history, the API did not change from when it was experimental to when it became stable... so technically, it was stable. Also, the warning don’t affect processing, so I think it’s fine to ship.

tools/builder.js
sidvishnoi: @saschanaz Do you plan to fix above here or separate PR?

saschanaz: Hmm, maybe a separate one.

PR #2285, opened 2019-04-18
fix(core/utils): getTextNodes() returning ws nodes

marcoscaceres:


src/core/utils.js
```diff
@@ -680,10 +680,10 @@ export function addId(elem, pfx = "", txt = "", noLC = false) {
 export function getTextNodes(el, exclusions = [], options = { wsNodes: true }) {
   const exclusionQuery = exclusions.join(", ");
   const acceptNode = (/** @type {Text} */ node) => {
-    if (node.parentElement.closest(exclusionQuery)) {
+    if (!options.wsNodes && /^\s*$/.test(node.data)) {
```
saschanaz: From a regex objector: How about `!node.data.trim()`?

tests/spec/core/utils-spec.js
```diff
@@ -492,13 +492,23 @@ describe("Core - Utils", () => {
...
         expect(str).toEqual("aabbccdd");
       });
+      it("Excludes white space nodes", () => {
+        const node = document.createElement("div");
+        node.innerHTML =
+          " <exclude> </exclude> <exclude>\t \n</exclude>include me";
+
+        const textNodes = utils.getTextNodes(node, [], "");
+        expect(textNodes.length).toEqual(1);
+        const str = textNodes.map(tn => tn.nodeValue).join("");
```
saschanaz: `.length === 1` here so maybe just `textNodes[0].nodeValue`?

src/core/utils.js
marcoscaceres: yeah, that's clearer.  However, semantically, I kinda liked the regex. I'll change it tho.

PR #2286, opened 2019-04-18
feat(core/inlines): process code inside inline links

marcoscaceres:
Can now handle, for example, [= `link` element =]

### DIFF
--- a/src/core/inlines.js
+++ b/src/core/inlines.js
@@ -27,6 +27,11 @@ import { renderInlineCitation } from "./render-biblio.js";
 export const name = "core/inlines";
 export const rfc2119Usage = {};
 
+// Inline `code`
+// TODO: Replace (?!`) at the end with (?:<!`) at the start when Firefox + Safari
+// add support.
+const inlineCodeRegExp = new RegExp("(?:`[^`]+`)(?!`)");
+
 /**
  * @param {string} matched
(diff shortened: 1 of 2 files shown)


src/core/inlines.js
```diff
@@ -121,14 +126,27 @@ function inlineLinkMatches(matched) {
...
+function processInlineContent(text) {
+  if (inlineCodeRegExp.test(text)) {
+    // We use a capture group to split, so we can process all the parts.
+    return text.split(/(`[^`]+`)(?!`)/).map(part => {
+      return part.startsWith("`")
+        ? inlineCodeMatches(part)
+        : processInlineContent(part);
+    });
+  }
+  return hyperHTML`${text}`;
```
saschanaz: Maybe just `text`?

marcoscaceres: Just text confuses hyper, in that it then coerces all other items in the array to text.

saschanaz: Uhm, what array do you mean?

saschanaz: Ah okay, this is a recursive function.

saschanaz: But maybe `document.createTextNode(text)`? I think we prefered it on other parts of the code.

marcoscaceres: good suggestion.

Issue #2270, opened 2019-04-14
Better error message for ambiguous terms


marcoscaceres:
The current error message for in-spec ambiguous term is somewhat misleading:

> The term "is empty" is defined in 1 spec(s) in multiple ways, so it's ambiguous. To disambiguate, you need to add a data-cite attribute. The specs where it's defined are: infra. at: 1.

It should be:

> The term "is empty" is defined in "Infra" in multiple ways, so it's ambiguous. To disambiguate, you need to add `data-xref-for` and one of the following: `x`, `y`.

PR #2293, opened 2019-04-19
feat(core/xref): provide better hints if term is ambiguous

sidvishnoi:
Closes https://github.com/w3c/respec/issues/2270
image

### DIFF
--- a/src/core/xref.js
+++ b/src/core/xref.js
@@ -281,7 +281,7 @@ function isNormative(elem) {
  * @param {any} conf
  */
 function addDataCiteToTerms(elems, queryKeys, data, conf) {
-  /** @type {Map<string, { elems: HTMLElement[], specs: Set<string> }>} */
+  /** @type {Map<string, { elems: HTMLElement[], results: SearchResultEntry[], term: string }>} */
   const errorsAmbiguous = new Map();
   /** @type {Map<string, HTMLElement[]>} */
   const errorsTermNotFound = new Map();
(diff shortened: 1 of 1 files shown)


src/core/xref.js
```diff
@@ -391,16 +390,22 @@ function showErrors({ errorsAmbiguous, errorsTermNotFound }) {
...
+    const msg = `The term "**${term}**" is defined in ${specs} in multiple ways, so it's ambiguous.`;
+    let hint = "";
+    if (definedInSpecs.size === 1) {
+      // defined only in one spec but in multiple ways
+      const xrefFor = new Set([].concat(...results.map(entry => entry.for)));
+      const forContext = [...xrefFor].map(s => `**${s}**`).join(", ");
+      hint = `add \`data-xref-for\` attribute with value equal to one of the following: ${forContext}.`;
+    } else {
+      // defined in multiple specs
+      hint = `add ${dataCiteLink} attribute with value equal to one of the following: ${specs}.`;
```
marcoscaceres: I wonder if we should mention that they should use the "[= for / term=]" syntax?

sidvishnoi: Maybe add wiki link to `data-xref-for` and mention shorthand syntax there?

marcoscaceres: Yeah, probably better

marcoscaceres: We should move people off using data-*

Issue #2212, opened 2019-03-28
Explore about replacing `normalizePadding()` with `reindent()`


saschanaz:
Turns out that a simple replacement doesn't work, so explore about why.

saschanaz:
Looks like indentation processing is weird on `normalizePadding()`. For example:

```markdown
<section id='sotd'><p>foo</p></section><section id='toc'></section>

    # Title
    Some text
```

Per the markdown spec, the text should be interpreted as `<pre>` because of its indentation. Instead, the function tends to interpret it as normal unindented markdown by trimming the text and all its indentations.

marcoscaceres:
Correct... so, this is for legacy reasons. People were writing markdown inside the body indented like that:(

saschanaz:
Is there a live example? I saw this on our tests that appends indented template string.

marcoscaceres:
I’ll dig some up.

marcoscaceres:
https://github.com/coremob/coremob-2012/blob/gh-pages/index.html

marcoscaceres:
Hello from 2012!

saschanaz:
🤦‍♀️

marcoscaceres:
Heh, yep. Kinda stuck. But we could make a case to break things for sanity’s sake.

marcoscaceres:
@saschanaz do you think we can do anything here? Or “abandon all hope”?

saschanaz:
@marcoscaceres I think we should learn exactly what behavior differences exist, add some tests for it, and then we can revisit this. For now it's rather a black box...

marcoscaceres:
Closing. Markdown should be written as markdown and ReSpec shouldn’t try to correct authoring mistakes for it.

PR #2302, opened 2019-04-22
refactor(core/utils): transfer normalizePadding() to core/markdown

saschanaz:
Part of #2212, this would at least discourage further use of the function.

### DIFF
--- a/src/core/webidl.js
+++ b/src/core/webidl.js
@@ -5,9 +5,9 @@
 //  - It could be useful to report parsed IDL items as events
 //  - don't use generated content in the CSS!
 import * as webidl2 from "webidl2";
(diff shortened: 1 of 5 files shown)


src/core/webidl.js
```diff
@@ -296,7 +296,7 @@ export function run() {
         "error",
         `Failed to parse WebIDL: ${e.message}.
         <details>
-        <pre>${normalizePadding(idlElement.textContent)}\n ${e}</pre>
+        <pre>${idlElement.textContent}\n ${e}</pre>
```
saschanaz: (No need to reindent here because `core/reindent` already takes care of them.)

PR #2303, opened 2019-04-22
feat(core/examples): substitutes empty example anchors

### LINKED ISSUE #1124: Ability to crossref examples
The idea is to have something similar to what is supported for sections, so that the following markup

````html
<a href="#ex-some-example" class="exampleRef"></a>
````

will be rendered as

<a href="#ex-some-example" class="exampleRef">Example 123</a>

This would ensure that examples are referred to consistently, without the need of checking whether the number of the referred example changed (because examples were dropped/added/moved elsewhere).

### PR DESCRIPTION
**Pull request by marcoscaceres**
closes #1124 

Allows for: 

```HTML
<div id="example" class="example" title="foo"><div> 
<!-- adds the example's title below: Example 1: foo --> 
<a href="#example"></a>
```


src/core/examples.js
```diff
@@ -58,7 +59,20 @@ function makeTitle(conf, elem, num, report) {
   `;
 }
 
-export function run(conf) {
+/**
+ * Link <a href="#example">
+ */
+function replaceEmptyAnchors() {
+  for (const id of examplesMap.keys()) {
```
saschanaz: `for (const [id, text] of examplesMap.entries()) {` could be better, which can replace `examplesMap.get(id)` with `text`.

marcoscaceres: good suggestion!

sidvishnoi: A little late. We don't need to do `.entries()` 😅

marcoscaceres: Oh, I didn't know about this! Thanks!

PR #2305, opened 2019-04-22
refactor: apply stricter no-unused-vars

saschanaz:


src/core/contrib.js
```diff
@@ -86,7 +86,7 @@ export async function run(conf) {
       toHTML(commenterUrls, editors, ghCommenters, headers),
       toHTML(contributorUrls, editors, ghContributors, headers)
     );
-  } catch (error) {
+  } catch {
```
marcoscaceres: I feel like we should really `console.error()` all these... it would be useful when these errors happen to be able to see why, for instance, and URL was invalid and so on....

saschanaz: Hmm, maybe we could make a helper function which do console and pub things together.

marcoscaceres: Problem is then that we get the line number helper function, making it somewhat annoying to unwind the stack.

saschanaz: The stack will be stored in the error object, passing it around won't change it.

marcoscaceres: Sorry, yes. I meant more the console.error() line will be wrong and then you need to go to the error to get the stack. I’d prefer to keep the console.error as close as possible to where the exception was throw.

saschanaz: >then you need to go to the error to get the stack

Not sure I follow, what does it mean by "go to the error"?

Both Firefox and Chrome show the error stack and optionally console stack together so I don't think things would be too complicated...

saschanaz: Currently some code do `pub("error", message)` and `console.error(err)` separately while the former also internally does `console.error`, resulting two separate error logs from a single error.

saschanaz: Hmm... maybe we can revisit this later. I thought a merged function may help #2187 but... maybe not?

marcoscaceres: I don't mind leaving it to later. We can always add the console.errors them as needed.

PR #2306, opened 2019-04-22
refactor(tests): remove reliance on RequireJS

marcoscaceres:
Removes MOST of the reliance on RequireJS. However, the worker script is still a problem because !text :( Will need to figure something out for that, but this gets us most of the way there.


tests/spec/w3c/conformance-spec.js
```diff
@@ -62,22 +62,4 @@ describe("W3C — Conformance", () => {
     const doc = await makeRSDoc(ops);
     expect(doc.querySelectorAll("#conformance .rfc2119").length).toBe(0);
   });
-
-  it("emits end event", async () => {
```
marcoscaceres: Don't really care for this as external events are deprecated.

saschanaz:
I have figured out that `!text` thing in #2233, maybe we could use that pattern.

marcoscaceres:
Ok, let’s get this in now and then do the magic from #2233

PR #2307, opened 2019-04-23
refactor: remove redudant .entries() where not needed

marcoscaceres:
So fancy... so javer scripts


marcoscaceres:
Fails are unrelated to the code changes. Feature Policy got published at the W3C.

PR #2309, opened 2019-04-24
 feat(core/inlines): allow [[[#...]]] to expand ids in doc

### LINKED ISSUE #2131: Linking to sections & figures
This has become so annoying that i'm finally making time to raise an issue. Currently automatic completion of link text for links pointing to sections and figures is prompted by creating a `a` element without link text, eg. `<a href="#myfragid"></a>`. This produces no visible rendered text if the link fails, and also in Dreamweaver's design view (which i use all the time). You get something such as:

> See .

This is annoying because i can't see at a glance where it's pointing to. If the rendering fails, it's often not very noticeable, and you have to jump through some hoops to locate the appropriate bit of source text in order to know which fragid is failing.

Could we create an alternative (to allow backwards compatibility) where the souce code would look like this:

```
<a class="secref">myfragid</a>
```

This would then show the following in design view, or in failed renderings:

> See myfragid

For figure references we could use `<a class="figref">myfragid</a>`.

I use this approach for my own document templates, and it works well.  The essential logic shouldn't change much, just get the cue from the class name and the target location from the link text.

marcoscaceres:
closes #2131 

Allow for `[[[#example-1]]]` and `[[[#figure-whatever]]]` and so on... 

The actual expansion is handled by individual modules.

### DIFF
--- a/src/core/inlines.js
+++ b/src/core/inlines.js
@@ -51,7 +51,9 @@ function inlineRFC2119Matches(matched) {
 function inlineRefMatches(matched) {
   // slices "[[[" at the beginning and "]]]" at the end
   const ref = matched.slice(3, -3).trim();
-  const nodeElement = hyperHTML`<a data-cite="${ref}"></a>`;
+  const nodeElement = ref.startsWith("#")
+    ? hyperHTML`<a href="${ref}"></a>`
+    : hyperHTML`<a data-cite="${ref}"></a>`;
   return nodeElement;
 }
 
@@ -191,7 +193,7 @@ export function run(conf) {
       "(?:{{[^}]+}})", // inline IDL references,
       "\\B\\|\\w[\\w\\s]*(?:\\s*\\:[\\w\\s&;<>]+)?\\|\\B", // inline variable regex
       "(?:\\[\\[(?:!|\\\\|\\?)?[A-Za-z0-9\\.-]+\\]\\])",
-      "(?:\\[\\[\\[(?:!|\\\\|\\?)?[A-Za-z0-9\\.-]+\\]\\]\\])",
+      "(?:\\[\\[\\[(?:!|\\\\|\\?)?[#A-Za-z0-9\\.-]+\\]\\]\\])",
       "(?:\\[=[^=]+=\\])", // Inline [= For/link =]
       inlineCodeRegExp.source,
       ...(abbrRx ? [abbrRx] : []),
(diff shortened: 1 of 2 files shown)


src/core/inlines.js
```diff
@@ -191,7 +205,7 @@ export function run(conf) {
       "(?:{{[^}]+}})", // inline IDL references,
       "\\B\\|\\w[\\w\\s]*(?:\\s*\\:[\\w\\s&;<>]+)?\\|\\B", // inline variable regex
       "(?:\\[\\[(?:!|\\\\|\\?)?[A-Za-z0-9\\.-]+\\]\\])",
-      "(?:\\[\\[\\[(?:!|\\\\|\\?)?[A-Za-z0-9\\.-]+\\]\\]\\])",
+      "(?:\\[\\[\\[(?:!|\\\\|\\?)?[#A-Za-z0-9\\.-]+\\]\\]\\])",
```
sidvishnoi: I think we don't want to support `foo#bar` here, so it should be:
```suggestion
      "(?:\\[\\[\\[(?:!|\\\\|\\?)?#[A-Za-z0-9\\.-]+\\]\\]\\])",
```

src/core/inlines.js
```diff
@@ -51,8 +52,21 @@ function inlineRFC2119Matches(matched) {
 function inlineRefMatches(matched) {
   // slices "[[[" at the beginning and "]]]" at the end
   const ref = matched.slice(3, -3).trim();
-  const nodeElement = hyperHTML`<a data-cite="${ref}"></a>`;
-  return nodeElement;
+  if (ref.startsWith("#")) {
+    if (!document.querySelector(ref)) {
```
sidvishnoi: Maybe return early `if (document.querySelector(ref))` and avoid larger nested chunk?

src/core/inlines.js
```diff
@@ -51,8 +52,21 @@ function inlineRFC2119Matches(matched) {
...
+      const badReference = hyperHTML`
```
sidvishnoi: Nit: Can we do this in one line?

saschanaz: `showInlineError()` without actual DOM element is no useful, because the function is to give a link to the document.

src/core/inlines.js
saschanaz: That may cause too many queries even for non-id links. But then how about:

```
if (!ref.startsWith("#")) {
  // return link with data-cite
}
// more codes to return link with href

marcoscaceres: good suggestion

src/core/inlines.js
saschanaz: Somehow commented here because its same line 😁

PR #2319, opened 2019-04-30
chore(package): add build:w3c-debug script

marcoscaceres:


package.json
```diff
@@ -73,10 +73,12 @@
   "scripts": {
     "babel:build": "babel src -d js --source-maps -q",
     "babel:watch": "babel src -d js --watch --source-maps",
-    "build:w3c": "node ./tools/builder.js --profile=w3c",
-    "build:geonovum": "node ./tools/builder.js --profile=geonovum",
     "build:components": "npm run copydeps && npm run babel:build",
+    "build:geonovum-debug": "node ./tools/builder.js --profile=geonovum --debug",
```
saschanaz: How about this to reduce code duplication?

```suggestion
    "build:geonovum-debug": "npm run build:geonovum -- --debug",
```

marcoscaceres: TIL :)

marcoscaceres:
Did the same for the w3c ones.

sidvishnoi:
Why did we add these scripts? Why not just pass the `--debug` when running these scripts?

marcoscaceres:
--debug It's not obvious, it's a hidden "gem" 💎

saschanaz:
To save 4 characters for each time? 😄

PR #2323, opened 2019-04-30
feat(core/webidl): enable xref by default for WebIDL

### LINKED ISSUE #999: Make "maplike" and "setlike" linkable in WebIDL
Maplike and setlike declarations currently don't link.... but they should.

### LINKED ISSUE #1372: Use defined terms list for autolinks in WebIDL when available

### LINKED ISSUE #1907: Link or unlink iterable<>?
Currently we don't prevent links for `iterable<>`, causing `<dfn> for "iterable" does not exist` warning.

https://github.com/w3c/respec/blob/c23d764aa7f6f31f7f87b916c6fa8ba858b7f0f6/src/core/dfn-finder.js#L14-L17

But I haven't seen any spec that has a spec definition for `iterable`. What should be done?

cc: @marcoscaceres

### LINKED ISSUE #2004: Point the IDL block that misses <dfn>
## Important info

* URL to affected spec: https://w3c.github.io/touch-events/
* ReSpec version: 24.0.0

## Description of problem

Currently it's hard to find what elements are causing warnings.

image

But what I was expecting?:

Links to the faulty blocks.

### LINKED ISSUE #2317: Should inheritance take same path as name
Given:

```
 interface Foo : EventTarget {};
```

Right now, the WebIDL module has `inheritance()` handler path, but EventTarget really just another interface (name). @saschanaz, should that just take the "name()" template path instead? Right now, the code doesn't complain when the super class is not linked.

### LINKED ISSUE #2321: should maplike be passed to `reference()`?
@saschanaz, not sure this is right, but I think `maplike` should be passed to the `reference()` template handler (same way that sequence, record, and Promise are passed). I can then link "maplike" to the WebIDL spec via the template.

marcoscaceres:
Closes #2321
Closes #2317
Closes #2004
Closes #1907
Closes #1372
Closes #999

Gets rid of `suppressWarnings` too and "no-link-warnings"


src/core/webidl.js
```diff
@@ -5,181 +5,112 @@
...
+      .map(x => (typeof x === "string" ? new Text(x) : x));
+  },
+  trivia(t) {
+    if (!t.trim()) {
+      return t;
+    }
+    return hyperHTML`<span class='idlSectionComment'>${t}</span>`;
+  },
+  generic(keyword) {
+    // Interfaces start with capital letters, like Promise, FrozenArray
```
saschanaz: Promises and FrozenArrays are not exactly "interface"s per the definition, maybe we could just describe what Shepherd does.

src/core/webidl.js
```diff
@@ -5,181 +5,112 @@
...
+    return /^[A-Z]/.test(keyword)
+      ? hyperHTML`<a data-xref-type="_IDL_">${keyword}</a>`
+      : // Other keywords like sequence, maplike, etc...
+        hyperHTML`<a data-xref-type="dfn">${keyword}</a>`;
+  },
+  reference(wrapped) {
+    return hyperHTML`<a data-xref-type="_IDL_">${wrapped}</a>`;
```
saschanaz: Just curious, does shepherd assumes primitive types as `_IDL_`, as `reference()` should also cover them?

src/core/webidl.js
marcoscaceres: I agree... Shepherd marks them as "interface", so you are right that we should make that clear.

src/core/webidl.js
marcoscaceres: @sidvishnoi can confirm, but I think: "_CONCEPT_" and "_IDL_" are an abstraction we created. 

Shepherd classifies:
attribute, dfn, dict-member, dictionary, element, enum, enum-value, event, exception, extended-attribute, interface, method, typedef

src/core/webidl.js
```diff
@@ -5,181 +5,112 @@
...
-    },
-  };
-  const result = webidl2.write(parse, { templates });
-  return hyperHTML`<pre class="def idl">${result}</pre>`;
-}
+const templates = {
+  wrap(items) {
+    return items
+      .reduce(flatten, [])
+      .filter(x => x !== "")
```
marcoscaceres: @saschanaz, wrap gets quite a few empty "", maybe a small bug/nit in Webidl2.js or are they somehow meaningful?

src/core/webidl.js
marcoscaceres: @saschanaz, to be more clear, if `reference()` was passed the actual types above, then  that's potentially a nice optimization for xref... as it knows exactly what to look for. @sidvishnoi, wdty?

src/core/webidl.js
saschanaz: webidl2 currently doesn't check whether a token has an empty whitespace or not and emits it unconditionally. Maybe we could fix it.

src/core/webidl.js
sidvishnoi: We send `_IDL_` when we are not sure of exactly what IDL type it is. If we're sure it's a dict-memmber, enum etc, we should use that instead. `_IDL_` is for fallback only. The types are defined [sidvishnoi/respec-xref-route](https://github.com/sidvishnoi/respec-xref-route/blob/5538b9e022f9073e1a55124f01a25489ecb02176/utils.js#L48).

src/w3c/defaults.js
```diff
@@ -25,6 +25,7 @@ const w3cDefaults = {
       url: "https://www.w3.org/",
     },
   ],
+  xref: true,
```
sidvishnoi: 🤞

src/core/webidl.js
saschanaz: > if reference() was passed the actual types above, then that's potentially a nice optimization for xref... as it knows exactly what to look for.

Primitive types are primitive, they do not belong to any of those groups in [sidvishnoi/respec-xref-route](https://github.com/sidvishnoi/respec-xref-route/blob/5538b9e022f9073e1a55124f01a25489ecb02176/utils.js#L48), so for now, no.

src/core/webidl.js
```diff
@@ -5,181 +5,113 @@
...
+    if (!t.trim()) {
+      return t;
+    }
+    return hyperHTML`<span class='idlSectionComment'>${t}</span>`;
+  },
+  generic(keyword) {
+    // Shepherd classifies "interfaces" as starting with capital letters,
+    // like Promise, FrozenArray, etc.
+    return /^[A-Z]/.test(keyword)
+      ? hyperHTML`<a data-xref-type="_IDL_">${keyword}</a>`
```
sidvishnoi: If we are certain here that it's an interface, we can pass `data-xref-type="interface"` maybe?

marcoscaceres: I think this holds true for all generics - because of how IDL limits them. @saschanaz, that that sounds right? Like, you can't extends `Promise` or `FrozenArray` or anything that's a generic type etc.

saschanaz: I don't follow how `"interface"` relates to non-extendability?

I think we follow Shepherd behavior here that I don't fully understand, so if it thinks they are interfaces then I agree with @sidvishnoi.

sidvishnoi:
Tests are failing locally, but if I paste the test body in basic.html, it works right 😕

marcoscaceres:
i have the same result as on TravisCI. Did you build the debug build for Karma to run against?

sidvishnoi:
> Did you build the debug build for Karma to run against

Yes. Still same results. Will start with removing some recent changes (locally) to find where it breaks.

src/core/link-to-dfn.js
```diff
@@ -43,11 +43,6 @@ export async function run(conf) {
       return findLinkTarget(target, ant, titleToDfns, possibleExternalLinks);
     });
     if (!foundDfn && linkTargets.length !== 0) {
-      // ignore WebIDL
-      if (ant.closest("pre.idl")) {
```
sidvishnoi: Adding this back fixes tests. 😕

marcoscaceres: heh, ok... delving deeper.

src/core/webidl.js
marcoscaceres: @saschanaz sorry this whole thing is a bit confusing... once we have the xref UI all this will make a lot more sense, as we will easily be able to see how things are classified by Shepherd.

marcoscaceres:
Marking as ready, but still need to figure out why that test is failing.

src/core/webidl.js
marcoscaceres: I don't mind handling them here for now.

src/core/webidl.js
```diff
@@ -5,181 +5,113 @@
...
+    if (!t.trim()) {
+      return t;
+    }
+    return hyperHTML`<span class='idlSectionComment'>${t}</span>`;
+  },
+  generic(keyword) {
+    // Shepherd classifies "interfaces" as starting with capital letters,
+    // like Promise, FrozenArray, etc.
+    return /^[A-Z]/.test(keyword)
+      ? hyperHTML`<a data-xref-type="interface">${keyword}</a>`
```
sidvishnoi: Should we add `data-cite="webidl"`on these if we're sure they are in WebIDL spec?

saschanaz: All generics are guaranteed to be in WebIDL spec so 👍

src/core/webidl.js
```diff
@@ -5,181 +5,113 @@
...
+    }
+    const parentName = parent ? parent.name : "";
+    const { name, idlId } = getNameAndId(data, parentName);
+    return hyperHTML`<span class='${className}' id='${idlId}' data-idl data-title='${name}'>${contents}</span>`;
+  },
+  extendedAttribute(contents) {
+    const result = hyperHTML`<span class="extAttr">${contents}</span>`;
+    return result;
+  },
+  extendedAttributeReference(name) {
```
marcoscaceres: @saschanaz, I've got a bit of a problem here... given:

```
[Exposed=Window, Worker]
```

It turns out `Worker` is not actually defined by any specification. And `Window` is also a bit dubious (probably why BikeShed doesn't link these either). I wonder if we can skip passing those as referenced in WebIDL2.js?

marcoscaceres: Actually, Window is fine... it has [Global]. Let me see if I can track down Worker.

marcoscaceres: It's unclear to me if `Worker` maps to `WorkerGlobalScope`... it feels like it should, but I can't seem to put all the pieces together looks at the various specs.

marcoscaceres: Argh... hmmm... maybe we should avoid linking anything after "=" for all WebIDL things.

saschanaz: Typically adding `GlobalScope` postfix should be enough. `Window` is an exception.

marcoscaceres: ah, ok. We can try that.

marcoscaceres: Linking the workers ended up being ugly and causing issues elsewhere :( Will need to think about it more. 

I'd be happy with not linking for now tho, as it addresses the main problems.

saschanaz:
Is this PR waiting something?

marcoscaceres:
So, was hoping that we would get a GSOC intern to help us deal with any fallout... but we didn't get a slot... so it's mostly just fear. I've been testing enabling xref on various specs and it's not had any significant side effects. So, I guess we could just merge and deal with it.

saschanaz:
Might be able to extract some refactoring to separate PR(s) before landing this.

marcoscaceres:
It's all pretty tightly scoped... let's just do this.

PR #2326, opened 2019-05-03
fix(core/anchor-expander): clone nodes when expanding an anchor

### LINKED ISSUE #2131: Linking to sections & figures

marcoscaceres:
closes #2131

Refactors expanding anchors into its own module. This now supports safely cloning a referenced node (by removing ids and converting `dfn`s to spans), and applies localization heuristics, so the correct language and directionality are applied to an expanded reference.


src/core/figures.js
```diff
@@ -48,23 +46,20 @@ export function run(conf) {
...
       decorateFigure(fig, caption, i, conf);
-      figMap[fig.id] = caption.childNodes;
     } else {
       showInlineWarning(fig, "Found a `<figure>` without a `<figcaption>`");
     }
 
     tof.push(getTableOfFiguresListItem(fig.id, caption));
   });
-  return { figMap, tof };
+  return { tof };
```
saschanaz:
```suggestion
  return tof;
```

src/core/inlines.js
```diff
@@ -50,10 +50,10 @@ function inlineRefMatches(matched) {
   // slices "[[[" at the beginning and "]]]" at the end
   const ref = matched.slice(3, -3).trim();
   if (!ref.startsWith("#")) {
-    return hyperHTML`<a data-cite="${ref}"></a>`;
+    return hyperHTML`<a class="respec-inline-expansion" data-cite="${ref}"></a>`;
```
saschanaz: Is this only for testing?

src/core/structure.js
```diff
@@ -88,13 +85,12 @@ function scanSections(sections, maxTocLevel, { prefix = "" } = {}) {
...
       });
       if (sub) {
-        Object.assign(secMap, sub.secMap);
         item.append(sub.ol);
       }
       ol.append(item);
     }
   }
-  return { ol, secMap };
+  return { ol };
```
saschanaz:
```suggestion
  return ol;
```

src/core/anchor-expander.js
```diff
@@ -0,0 +1,141 @@
...
+function makeSafeCopy(node) {
+  const clone = node.cloneNode(true);
+  clone.querySelectorAll("[id]").forEach(elem => elem.removeAttribute("id"));
+  clone.querySelectorAll("dfn").forEach(dfn => {
+    const span = document.createElement("span");
+    span.append(...dfn.childNodes);
+    [...dfn.attributes].forEach(({ name, value }) => {
+      span.setAttribute(name, value);
+    });
+    dfn.parentNode.replaceChild(span, dfn);
```
saschanaz: The whole thing in this loop could be replaced with `renameElement()` from `core/utils.js`.

src/core/anchor-expander.js
```diff
@@ -0,0 +1,141 @@
...
+
+function localize(matchingElement, newElement) {
+  for (const attrName of ["dir", "lang"]) {
+    // Already set on element, don't override.
+    if (newElement.hasAttribute(attrName)) continue;
+    const matchingClosest = matchingElement.closest(`[${attrName}]`);
+    const newClosest = newElement.closest(`[${attrName}]`);
+
+    // It's the same, so already inherited from closet (probably HTML element or body).
+    if (matchingClosest === newClosest) continue;
```
saschanaz: Shouldn't this be value comparison? Could get the same lang value even with different parents.

src/core/anchor-expander.js
marcoscaceres: oh! Forgot about that! thanks!!!

src/core/inlines.js
marcoscaceres: I was thinking of styling these... I was going to add the "[[[" "]]]" back in using CSS generated content.

marcoscaceres: (styled when they fail, that is)

src/core/anchor-expander.js
marcoscaceres: yes, you are right. Well spotted.

js/profile-w3c-common.js
```diff
@@ -79,6 +79,7 @@ define([
   "./core/highlight-vars",
   "./core/data-type",
   "./core/algorithms",
+  "./core/anchor-expander.js",
```
saschanaz: Does this work? AFAIK requirejs refuses to work with `.js` file extension for relative paths.

marcoscaceres: no, this will break :) I was testing only against w3c-respec profile. Fixing.

src/core/figures.js
```diff
@@ -12,33 +12,46 @@ export const name = "core/figures";
 
 const localizationStrings = {
   en: {
-    table_of_fig: "Table of Figures",
+    list_of_figures: "List of Figures",
+    fig: "Figure ",
+  },
+  ja: {
+    fig: "図",
+    list_of_figures: "図のリスト ",
```
saschanaz:
```suggestion
    list_of_figures: "図のリスト",
```

marcoscaceres:
Thanks for fixing those @saschanaz!

src/core/inlines.js
marcoscaceres: @saschanaz, decided to ditch the "respec-inline-expansion" class.

marcoscaceres:
Ok, I think this might be good to go. I'll keep working with @r12a to improve the l10n aspects, but it covers the use cases that @r12a had.

PR #2327, opened 2019-05-04
fix(core/mdn-annotation): Improve user experience

### LINKED ISSUE #2313: MDN pill hard to use on mobile
The MDN boxes are a little difficult to use on mobile. We should increase hit area to the whole box, and we should open the whole thing up by default (i.e., let's get rid of details). 

We should also improve the markup. When we shipped, we copied what the HTML Spec was doing... but the markup is less than ideal.

Swapnilr1:
Closes #2313 

I have increased the hit area and also the removed the `<details>` section so now browser support table opens up without an additional click. Tested on Firefox and Chrome. Please try it out.

> We should also improve the markup. When we shipped, we copied what the HTML Spec was doing... but the markup is less than ideal.

Does this mean refactoring of the implementation? Like using `addEventListener("click",..)` instead of the current implementation where the event handler is in the HTML itself?

### DIFF
--- a/assets/mdn-annotation.css
+++ b/assets/mdn-annotation.css
@@ -13,16 +13,10 @@
 
 .mdn button {
   cursor: pointer;
-  position: absolute;
-  left: 0;
-  top: -2px;
-  height: 1em;
   border: none;
   color: #000;
   background: transparent;
-  margin-left: -8px;
-  font-size: 10px;
-  line-height: 1em;
+  margin: -8px;
   outline: none;
 }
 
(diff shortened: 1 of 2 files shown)


Swapnilr1:
I haven't improved the markup yet!

marcoscaceres:
Ok ok to do it in multiple PRs :) It's already an improvement.

marcoscaceres:
> Does this mean refactoring of the implementation? Like using addEventListener("click",..) instead of the current implementation where the event handler is in the HTML itself?

The changes might need to be more significant... but we would need to look at structurally what we are doing. Like, we should consider using an actual `<table>`.

PR #2340, opened 2019-05-07
fix(core/best-practices): align style with other boxes

### LINKED ISSUE #2008: best practices boxes look weird

### LINKED ISSUE #2341: Best Practices module needs to be localized.
It's currently only in English.

marcoscaceres:
Fixes #2008 
Closes #2341

Screenshot 2019-05-07 17 44 45

This whole module is still less than ideal and a bit fragile... we need to come up with something better and deprecate "practicelab" (why "lab?"). Anyway...


marcoscaceres:
Just noting, this is designed to more closely match:  https://w3c.github.io/tr-design/src/readme.html#advisement

src/core/utils.js
```diff
@@ -822,3 +822,13 @@ export class InsensitiveStringSet extends Set {
         );
   }
 }
+
+export function makeSafeCopy(node) {
+  const clone = node.cloneNode(true);
+  clone.querySelectorAll("[id]").forEach(elem => elem.removeAttribute("id"));
+  clone.querySelectorAll("dfn").forEach(dfn => renameElement(dfn, "span"));
+  // TODO: make a tree walker and remove all the comment nodes, not just first level.
+  return [...clone.childNodes].filter(
```
saschanaz: Filtering is done later by exporter, why should be done here?

src/core/best-practices.js
```diff
@@ -17,20 +16,22 @@ export function run() {
...
-    ul.classList.add("practicebox");
-    const title = hyperHTML`<span class="practicetitle">Best Practice ${num}</span>: `;
-    bp.prepend(...title.childNodes);
+    const container = bp.closest("div");
+    if (!container) {
+      bp.classList.add("advisement");
+      continue;
+    }
+    container.classList.add("advisement");
+    const marker = hyperHTML`<a href="${`#${id}`}"
```
saschanaz: This is not really a `marker`, it's `contents` of the container that includes a marker right?

src/core/utils.js
marcoscaceres: I was worried that we might pick up these nodes again and pass them to HyperHTML in a different/future module. I don’t know exactly what HyperHTML does with these nodes, and if it might get confused.

src/core/best-practices.js
marcoscaceres: Yeah, it’s a weird one. It’s as close as I could get it to what the TR design guide recommends.

saschanaz: I mean the variable name 😁, the markup includes a marker and the text, so IMO not really appropriate to name it just `marker`...

src/core/utils.js
saschanaz: Hmm. Maybe we could get a general wrapper to filter comments out.

marcoscaceres: Yeah, i looked at pulling the one out you created from the exporter. Should be easy to do.

src/core/utils.js
```diff
@@ -822,3 +822,37 @@ export class InsensitiveStringSet extends Set {
         );
   }
 }
+
+export function makeSafeCopy(node) {
+  const clone = node.cloneNode(true);
+  clone.querySelectorAll("[id]").forEach(elem => elem.removeAttribute("id"));
+  clone.querySelectorAll("dfn").forEach(dfn => renameElement(dfn, "span"));
+  cleanupHyperComments(clone);
```
saschanaz: Should we ultimately prevent comments everywhere? I'm not strongly convinced that only this position needs cleanup...

marcoscaceres: Yeah, I’ve struggled with this too... I guess it would be ok to remove them all.

saschanaz:
I feel the comment filter makes this PR more complex... Can we split that thing to another PR? I'm considering `nanohtml` that works similar with `hyperhtml` but returns `DocumentFragment` rather than hyperhtml-specific JS object (and especially no `<!-- hyper -->`).

marcoscaceres:
Sure, can split it out.

marcoscaceres:
The spec that will benefit immediately form this: https://www.w3.org/TR/fingerprinting-guidance/

PR #2342, opened 2019-05-08
fix: remove all comment nodes when exporting

marcoscaceres:
All comments are removed when exporting. 

`makeSafeCopy()` now returns the copied node, not just the children. 

core/utils now exports `removeCommentNodes()`

### DIFF
--- a/src/core/utils.js
+++ b/src/core/utils.js
@@ -834,7 +834,7 @@ export function makeSafeCopy(node) {
 
 export function removeCommentNodes(node) {
   const walker = document.createTreeWalker(node, NodeFilter.SHOW_COMMENT);
-  for (const comment of [...walkTree(walker)]) {
+  for (const comment of walkTree(walker)) {
     comment.remove();
   }
 }
--- a/src/core/utils.js
+++ b/src/core/utils.js
@@ -834,7 +834,7 @@ export function makeSafeCopy(node) {
 
 export function removeCommentNodes(node) {
   const walker = document.createTreeWalker(node, NodeFilter.SHOW_COMMENT);
-  for (const comment of walkTree(walker)) {
+  for (const comment of [...walkTree(walker)]) {
     comment.remove();
   }
 }
--- a/src/core/utils.js
+++ b/src/core/utils.js
@@ -822,3 +822,30 @@ export class InsensitiveStringSet extends Set {
         );
   }
 }
+
+export function makeSafeCopy(node) {
+  const clone = node.cloneNode(true);
+  clone.querySelectorAll("[id]").forEach(elem => elem.removeAttribute("id"));
+  clone.querySelectorAll("dfn").forEach(dfn => renameElement(dfn, "span"));
+  if (clone.hasAttribute("id")) clone.removeAttribute("id");
+  removeCommentNodes(clone);
+  return clone;
+}
+
+export function removeCommentNodes(node) {
+  const walker = document.createTreeWalker(node, NodeFilter.SHOW_COMMENT);
+  for (const comment of [...walkTree(walker)]) {
+    comment.remove();
+  }
+}
+
+/**
+ * @template {Node} T
+ * @param {TreeWalker<T>} walker
+ * @return {IterableIterator<T>}
+ */
+function* walkTree(walker) {
+  while (walker.nextNode()) {
+    yield /** @type {T} */ (walker.currentNode);
+  }
+}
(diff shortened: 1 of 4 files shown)


tests/spec/core/exporter-spec.js
```diff
@@ -28,25 +28,17 @@ describe("Core - exporter", () => {
     expect(doc.querySelectorAll(".removeOnSave").length).toBe(0);
   });
 
-  it("cleans up hyperHTML comments", async () => {
+  it("removes all comments", async () => {
     const ops = makeStandardOps();
-    ops.body = `<div><!---LEAVE%-->PASS<!-- STAY --></div>`;
+    ops.body = `<div><!-- remove -->PASS <span><!-- remove --></span></div>`;
```
saschanaz: Previously [we decided not to remove everything](https://github.com/w3c/respec/pull/1514#discussion_r168681074):

>Sometimes Editors add random comments that they don't mind having on TR.

Is it okay now?

marcoscaceres: That's what I was struggling with... I think we just remove them all and see if anyone complains.

src/core/utils.js
```diff
@@ -822,3 +822,30 @@ export class InsensitiveStringSet extends Set {
...
+  clone.querySelectorAll("[id]").forEach(elem => elem.removeAttribute("id"));
+  clone.querySelectorAll("dfn").forEach(dfn => renameElement(dfn, "span"));
+  if (clone.hasAttribute("id")) clone.removeAttribute("id");
+  removeCommentNodes(clone);
+  return clone;
+}
+
+export function removeCommentNodes(node) {
+  const walker = document.createTreeWalker(node, NodeFilter.SHOW_COMMENT);
+  for (const comment of [...walkTree(walker)]) {
```
sidvishnoi: Can we do this?
```suggestion
  for (const comment of walkTree(walker)) {
```

marcoscaceres: I think I tried the previously and it breaks.

marcoscaceres: let's try it and see what happens :)

saschanaz: It won't work as it will cut the tree in the middle.

marcoscaceres: yeah, I remember it did bad things... changing it back.

marcoscaceres:
@saschanaz, I'm getting a feeling you are not loving this pull request. Happy to change things.

saschanaz:
I feel it's a chance to explore alternatives of hyperhtml, as the library is the cause of those comments. I checked `nanohtml` but it has some behavior difference which does not live well with prettier formatting.

marcoscaceres:
> I feel it's a chance to explore alternatives of hyperhtml, as the library is the cause of those comments. I checked nanohtml but it has some behavior difference which does not live well with prettier formatting. (It does not trim whitespaces whereas hyperhtml does.)

understood. The rationale behind my choosing HyperHTML in this project was that I'd reviewed the original design and source code when the maintainer of that project created it. That gave me quite a few assurances about how it works, etc. I also trust the maintainer to keep the project going, and know we can call on him for help if we ever get stuck or need anything.

marcoscaceres:
(having said that... we could ask him if he can think about some means that hyperHTML could clean up after itself?)

marcoscaceres:
I see you and others already asked :) I tend to agree that the comments are ugly, but they are not a big deal or a deal breaker.

saschanaz:
I also feel good about `nanohtml` because it always returns a `Node` whereas `hyperhtml` sometimes returns it's own JS structure (when an input string has multiple elements). One big problem is that it only supports CJS so will break requirejs.

AFAIK hyperhtml's comments are placeholders to efficiently respond to dynamic value changes, so I'm not sure the maintainer will want to remove them...

marcoscaceres:
>I also feel good about nanohtml because it always returns a Node whereas hyperhtml sometimes returns it's own JS structure (when an input string has multiple elements). One big problem is that it only supports CJS so will break requirejs.

Understood. Why don't we come back to this once we kill requirejs. I'm open to exploring alternatives... or better, removing dependencies on libraries entirely. 

>AFAIK hyperhtml's comments are placeholders to efficiently respond to dynamic value changes, so I'm not sure the maintainer will want to remove them...

Yeah, that's my reading too.

saschanaz:
>removing dependencies on libraries entirely.

That will be super great, but seeing the sizes of existing ones, I doubt it's doable 😥

saschanaz:
>Why don't we come back to this once we kill requirejs

This needs import-maps implementation, so...

image

marcoscaceres:
"Slow and steady wins the race" -🐢

marcoscaceres:
Filed https://github.com/w3c/respec/issues/2343

PR #2354, opened 2019-05-20
fix(assets/datatype): use CSS defaults for vars

marcoscaceres:
https://github.com/w3c/webdriver/issues/1424

### DIFF
--- a/assets/datatype.css
+++ b/assets/datatype.css
@@ -35,6 +35,8 @@ var[data-type]::after {
   padding: 6px;
   border-radius: 3px;
   color: #daca88;
+  text-indent: 0;
+  font-weight: normal;
 }
 
 var[data-type]:hover::after,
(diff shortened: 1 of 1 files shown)


saschanaz:
I think those properties are for typed variables. Are we sure they won't be broken with this change?

marcoscaceres:
Ah, let me try... I only tried clicking on them but not the hover.

fantasai:
So Firefox definitely has a bug here where it leaves behind an empty piece of inline box on the first line if the var gets pushed to the second. But Chrome seems to render OK as long as you keep the 'position: relative'.

fantasai:
(I will note that the tooltip goes offscreen when it's long and the var is short and you're near the start or end of the window. But fixing that is a different problem.)

marcoscaceres:
Ok, I think this addresses the reported bug at least... @fantasai is right that there is definitely a lot of work to do on this.

marcoscaceres:
> So Firefox definitely has a bug here where it leaves behind an empty piece of inline box on the first line if the var gets pushed to the second. 

Filed #2367 to track that... still seeing some weirdness in Chrome.

PR #2360, opened 2019-05-24
chore(package): pin puppeteer to 1.16.0

saschanaz:
As I said in https://github.com/w3c/respec/issues/2359#issuecomment-495554972 I think this is an intermittent Chrome Canary bug (on which puppeteer depends).


saschanaz:
Will merge after I confirm this is still an issue.

PR #2365, opened 2019-05-27
refactor(core/caniuse): use path, not frag

marcoscaceres:
As the feature id is not actually included in caniuse.com document, link validation fails. Using a path gets around this issue.


marcoscaceres:
Related https://github.com/Fyrd/caniuse/issues/4925

saschanaz:
Didn't know this 😮

marcoscaceres:
neither me... just learned about it from that bug.

PR #2368, opened 2019-05-28
fix(core/webidl): autolink link Window and object

marcoscaceres:


src/core/webidl.js
```diff
@@ -38,8 +38,20 @@ const templates = {
...
   },
-  reference(wrapped) {
-    return hyperHTML`<a data-xref-type="_IDL_">${wrapped}</a>`;
+  reference(wrapped, unescaped) {
+    let type = "_IDL_";
+    let cite = null;
+    switch (unescaped) {
+      case "Window":
+        type = "interface";
+        cite = "HTML";
```
saschanaz: https://respec.org/xref/ only gives a singnle item for `Window`, so not sure why we need this...

marcoscaceres: Ok, let’s remove 👍

saschanaz: Removing it reintroduces the bug. Hard to understand...

marcoscaceres: Magic 👏

marcoscaceres: But seriously, let’s run with this and then deal with xref not matching separately as most specs are  currently in a broken state.

saschanaz: Agreed.

saschanaz: Will you publish this or should I do it instead?

marcoscaceres: Sorry, just saw this... published. Still have the issue with Exposed=Worker... but it only affects a few specs. We can deal with it tomorrow.

PR #2369, opened 2019-05-28
fix(core/webidl): autolink worker types

marcoscaceres:


src/core/webidl.js
```diff
@@ -38,9 +38,11 @@ const templates = {
       : // Other keywords like sequence, maplike, etc...
         hyperHTML`<a data-xref-type="dfn" data-cite="WebIDL">${keyword}</a>`;
   },
-  reference(wrapped, unescaped) {
+  reference(wrapped, unescaped, context = {}) {
```
saschanaz: Oh, do we get undefined context sometimes? That's unexpected.

src/core/webidl.js
```diff
@@ -49,9 +51,22 @@ const templates = {
       case "object":
         type = "interface";
         cite = "WebIDL";
+        break;
+      default: {
+        const isWorkerType = unescaped.includes("Worker");
```
saschanaz: We also have worklets: https://webaudio.github.io/web-audio-api/#audioworkletglobalscope

src/core/webidl.js
marcoscaceres: No, that was my silliness from before I realized the WebIDL2.js hadn’t updated.

src/core/webidl.js
```diff
@@ -49,9 +51,22 @@ const templates = {
...
+        if (isWorkerType && tokens && tokens.name.value === "Exposed") {
```
saschanaz:
```suggestion
        if (isWorkerType && context.type === "extended-attribute" && context.name === "Exposed") {
```

src/core/webidl.js
```diff
@@ -49,9 +51,22 @@ const templates = {
...
+          lt = `${unescaped}GlobalScope`;
+          type = "interface";
+          cite = ["Worker", "DedicatedWorker", "SharedWorker"].includes(
```
saschanaz: Is this some optimization thing?

marcoscaceres: No, cite needs to be explicit for now or it breaks.

src/core/webidl.js
marcoscaceres: Let’s add these in the next round.

marcoscaceres:
Travis got stuck...

saschanaz:
Looks like we have both Travis app and service, maybe we can remove service now.

marcoscaceres:
Done!

PR #2376, opened 2019-05-31
refactor: use jsdoc type augmenting format

saschanaz:
No need to write horizontally looooong types anymore 😁


saschanaz:
@sidvishnoi That might be good for a type that does not belong to a single file but is used across multiple various modules. In general, I'd like to keep it where it's used.

Issue #2377, opened 2019-06-01
DOMHighResTimeStamp fails to link


saschanaz:
## Important info

* URL to affected spec: https://w3c.github.io/performance-timeline/
* ReSpec version: 24.22.3
* [x] I did a "hard refresh", but it's still busted. 

## Description of problem

What happened (e.g., it crashed)?: 

![image](https://user-images.githubusercontent.com/3396686/58742908-45409580-8463-11e9-8776-4ff78172d7dd.png)

The type is from: https://www.w3.org/TR/hr-time-2/#dom-domhighrestimestamp

saschanaz:
https://respec.org/xref/ gives no result for DOMHighResTimeStamp, any idea? @sidvishnoi

marcoscaceres:
It’s not part of the dataset yet. We need to implement #685 first.

sidvishnoi:
The terms that are available to xref can be seen here: https://respec.org/xref/meta?fields=terms (Will add them to the UI as auto-predict today)

sidvishnoi:
https://github.com/tabatkins/bikeshed-data/blob/046a5d7638be7e629265c7a7e6720747b8810c76/data/anchors/anchors-do.data#L2110 shows it's not exported (see L2127), so it's not part of dataset.

saschanaz:
@sidvishnoi I'm not familiar with the file, what do we expect at L2127?

sidvishnoi:
@saschanaz L2127 denotes "export" (1 means true, empty means false) 
https://github.com/tabatkins/bikeshed/blob/0da7328/bikeshed/update/updateCrossRefs.py#L313-L328

saschanaz:
This has been fixed. xref is great 🎉 🎉 🎉

PR #2380, opened 2019-06-03
feat(core/dfn): add data-export by default

### LINKED ISSUE #685: Implement bikeshed's dfn contract
The contract is as follows: 

- [X] The defining element MUST be a dfn or h2...h6. No other element is recognized as defining a term.
- [X] The element MUST have an id attribute.
- [x] The linking text defaults to the text content of the dfn/heading. If the desired text content isn’t suitable for linking text, or you wish to provide multiple linking texts, a `lt` attribute` containing one or more pipe-separated linking texts will override the text content. See also §6.3 Defining Extra-Short "Local" Linking Texts

Note: we don't support `local-lt`. We also "dirty" the `data-lt` with various things, like plurals... we might need to deal with that.  

- [x] `data-dfn-type` MUST be provided, and set one of the [accepted values](https://tabatkins.github.io/bikeshed/#dfn-types). (The dfn element, specifically, actually allows this to be omitted, and defaults to the "dfn" type. But headings require a type, and dfns are clearer with it specified.)

Note: it's still unclear what the heading types should be. We need to. look that up. 

- [x] Either `data-export` or `data-noexport` MAY be provided (both boolean attributes). If neither is provided, "dfn" type definitions default to noexport, while all others default to export. Unexported definitions aren’t linkable by default.

- [x] Several types of definitions are namespaced to another construct; for example, attribute names are namespaced to an interface. These definitions MUST contain a data-dfn-for attribute, containing a comma-separated list of one or more definitions they’re namespaced to.

Note:  we. can support for on IDL definitions easily enough. We need to support them for internal slots too, and possibly other other things. However, we generally don't want to export internal slots, because they are private.

saschanaz:
Closes #685

### DIFF
--- a/src/core/dfn.js
+++ b/src/core/dfn.js
@@ -1,6 +1,9 @@
 // @ts-check
 // Module core/dfn
-// - Finds all <dfn> elements and populates conf.definitionMap to identify them.
+// Finds all <dfn> elements and:
+// - populates definitionMap to identify them.
+// - normalize `data-dfn-for`
+// - add `data-export` unless there is `data-noexport`
 
 import { getDfnTitles } from "./utils.js";
 import { registerDefinition } from "./dfn-map.js";
@@ -17,6 +20,10 @@ export function run() {
     if (dfn.dataset.dfnFor) {
       dfn.dataset.dfnFor = dfn.dataset.dfnFor.toLowerCase();
     }
+    if (!("noexport" in dfn.dataset)) {
+      dfn.dataset.export = "";
+    }
+
     // TODO: we should probably use weakmaps and weaksets here to avoid leaks.
     const titles = getDfnTitles(dfn, { isDefinition: true });
     registerDefinition(dfn, titles);
--- a/tests/spec/core/dfn-spec.js
+++ b/tests/spec/core/dfn-spec.js
@@ -81,4 +81,20 @@ describe("Core — Definitions", () => {
     expect(dfn.dataset.lt).toBe("text|text 1|text 2|text 3");
     expect(dfn.dataset.dfnType).toBe("dfn");
   });
+
+  it("adds data-export", async () => {
+    const ops = {
+      config: makeBasicConfig(),
+      body: `${makeDefaultBody()}<section id='dfn'>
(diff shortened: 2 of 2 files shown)


src/core/dfn.js
```diff
@@ -17,6 +20,10 @@ export function run() {
     if (dfn.dataset.dfnFor) {
       dfn.dataset.dfnFor = dfn.dataset.dfnFor.toLowerCase();
     }
+    if (!("noexport" in dfn.dataset)) {
+      dfn.dataset.export = "";
```
sidvishnoi: Should we do `dfn.removeAttribute("data-export")` instead?

sidvishnoi: Oh. I misunderstood.

saschanaz: This is not removing an attribute, this is adding one.

sidvishnoi:
I'm not clear why this is needed. Can you help me understand @saschanaz ?

saschanaz:
@sidvishnoi To mark L2127 with `1`: https://github.com/w3c/respec/issues/2377#issuecomment-497917390

This follows export-by-default behavior, because otherwise we'll have to modify all spec documents to export something.

marcoscaceres:
The BikeShed/Shepherd contract requires no export by default. The reason is that Editors need to think really carefully about what definitions they export and why. The only exception to this is IDL definitions, which have both a “type” and “for”.

marcoscaceres:
I think we should do the contract incrementally: let’s start by exporting the IDL things. That would be a massive win already.

saschanaz:
>The BikeShed/Shepherd contract requires no export by default

[You liar](https://github.com/w3c/respec/issues/685#issuecomment-487366690) 😭🤣

> let’s start by exporting the IDL things

I'll do that happily.

marcoscaceres:
> You liar 😭🤣

heh, sorry about that... "trust, but verify".

saschanaz:
@marcoscaceres Safe to be closed, right?

marcoscaceres:
yes, I think so.

PR #2381, opened 2019-06-03
feat(core/webidl): add data-export

saschanaz:
Also adding proper `data-dfn-type`.

### DIFF
--- a/src/core/webidl.js
+++ b/src/core/webidl.js
@@ -115,6 +115,8 @@ function createIdlAnchor(escaped, data, parent) {
     parent: parentName,
   });
   if (dfn) {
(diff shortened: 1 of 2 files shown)


src/core/webidl.js
```diff
@@ -143,6 +145,17 @@ function createIdlAnchor(escaped, data, parent) {
   return unlinkedAnchor;
 }
 
+function getDfnType(idlType) {
```
marcoscaceres:
```suggestion
/** 
* Map to Shepherd types, for export. 
* @see https://tabatkins.github.io/bikeshed/#dfn-types
*/
function getDfnType(idlType) {
```

PR #2393, opened 2019-06-17
l10n: update zh and ja translations

xfq:
This PR upstreams the changes in https://github.com/w3c/jlreq/pull/61 and updates some other zh and ja translations.

/cc @r12a @himorin


marcoscaceres:
Assuming these are ok :)

r12a:
Thanks @marcoscaceres. We checked the Japanese ones with the jlreq folks, and Fuqiao is Chinese, so should be no worries.

marcoscaceres:
Appreciate that. I'll do a release first thing tomorrow so they show up in your specs.

marcoscaceres:
Pushed out a new version of ReSpec, which includes these.

r12a:
Thanks, Marcos.

PR #2404, opened 2019-06-27
fix(core/data-cite): prevent re-export of exisiting dfns

marcoscaceres:
Prevents specs stomping on each other's definitions by accident.

### DIFF
--- a/tests/spec/core/data-cite-spec.js
+++ b/tests/spec/core/data-cite-spec.js
@@ -380,6 +380,6 @@ describe("Core — data-cite attribute", () => {
     const dfn2 = doc.querySelector("#t2");
     expect(dfn2.classList).toContain("respec-offending-element");
     expect("export" in dfn2.dataset).toBe(false);
-    expect("no-export" in dfn2.dataset).toBe(true);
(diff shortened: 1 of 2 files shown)


tests/spec/core/data-cite-spec.js
```diff
@@ -361,4 +361,25 @@ describe("Core — data-cite attribute", () => {
...
+    const doc = await makeRSDoc(ops);
+
+    const dfn1 = doc.querySelector("#t1");
+    expect("export" in dfn1.dataset).toBe(false);
+    expect("noExport" in dfn1.dataset).toBe(true);
+
+    const dfn2 = doc.querySelector("#t2");
+    expect(dfn2.classList).toContain("respec-offending-element");
+    expect("export" in dfn2.dataset).toBe(false);
+    expect("no-export" in dfn2.dataset).toBe(true);
```
saschanaz: Probably:

```suggestion
    expect("noExport" in dfn2.dataset).toBe(true);
```

marcoscaceres: good catch.

PR #2408, opened 2019-07-01
feat(core/webidl): allow self-defining IDL

### LINKED ISSUE #2316: Allow IDL types to be self-defined
From https://github.com/w3c/respec/issues/685#issuecomment-487366561

>... like allowing IDL things to be “self defining” (in the IDL itself) instead of requiring things to be defined separately.

saschanaz:
No more definition errors! Fixes #2316 

`<dfn>` by default uses bold font, not sure we want it in IDL text.

### DIFF
--- a/src/core/webidl.js
+++ b/src/core/webidl.js
@@ -75,12 +75,12 @@ const templates = {
     if (data.idlType && data.idlType.type === "argument-type") {
       return hyperHTML`<span class="idlParamName">${escaped}</span>`;
     }
-    const idlAnchor = createIdlAnchor(escaped, data, parent);
-    const className = parent ? "idlName" : "idlID";
-    if (data.type === "enum-value") {
-      return idlAnchor;
+    const idlLink = defineIdlName(escaped, data, parent);
+    if (data.type !== "enum-value") {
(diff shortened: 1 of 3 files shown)


src/core/webidl.js
```diff
@@ -108,7 +108,10 @@ const templates = {
   },
 };
 
-function createIdlAnchor(escaped, data, parent) {
+/**
+ * Returns a link to existing <dfn> or creates one if not exists.
```
marcoscaceres:
```suggestion
 * Returns a link to existing <dfn> or creates one if doesn’t exists.
```

saschanaz:
Sometimes emojis are confusing... I'll consider it as a go sign 😆

marcoscaceres:
sorry for the confusing emojis.... I've been reluctant on supporting this because I worry about Editors under-specifying things. In particular, not defining the steps to take when a method is called would not be great. Similarly with attributes, it makes it easy for developers to forget that "getting" an attribute may have side effects (not a good thing, but the reality is that implementation code runs... same with with setting... like setting an IDL attribute might be reflected in HMTL content).

Anyway, let's see how we go with this... if things start getting messy, we can always make things more strict.

PR #2409, opened 2019-07-02
feat(core/xref): data-no-xref short-circuits xref lookup

marcoscaceres:


marcoscaceres:
I think this is the cheapest option.

src/core/xref.js
```diff
@@ -292,6 +292,7 @@ function addDataCiteToTerms(elems, queryKeys, data, conf) {
 
   for (let i = 0, l = elems.length; i < l; i++) {
     const elem = elems[i];
+    if (elem.closest("[data-no-xref]")) continue;
```
sidvishnoi: Why not add this check here:
https://github.com/w3c/respec/blob/fc967beb562f59335877ed1190005080f5500d45/src/core/link-to-dfn.js#L45-L51

sidvishnoi: I think this should work.
``` js
if (!foundDfn && linkTargets.length !== 0 && !("noXref" in ant.dataset)) { 
```

sidvishnoi:
It's breaking due to:
https://github.com/w3c/respec/blob/8fe9735da0bca1515b2335e62df68898066aa4d7/src/core/link-to-dfn.js#L233-L236

Revert back to first commit, or fix here?

PR #2416, opened 2019-07-11
feat(core/webidl): enable validation

saschanaz:
Initial basic validation without xref...

### DIFF
--- a/examples/basic.built.html
+++ b/examples/basic.built.html
@@ -29,7 +29,7 @@
     <section data-dfn-for="Foo">
       <h2><dfn>Foo</dfn> interface</h2>
       <pre class="idl">
-      [Constructor]
+      [Exposed=Window, Constructor]
       interface Foo {
         attribute DOMString bar;
         void doTheFoo();
--- a/examples/basic.html
+++ b/examples/basic.html
@@ -58,7 +58,7 @@
         <dfn>Foo</dfn> interface
       </h2>
       <pre class="idl">
-      [Constructor]
+      [Exposed=Window, Constructor]
       interface Foo {
         attribute DOMString bar;
         void doTheFoo();
--- a/package-lock.json
+++ b/package-lock.json
@@ -1280,7 +1280,7 @@
     },
     "arg": {
       "version": "2.0.0",
-      "resolved": "http://registry.npmjs.org/arg/-/arg-2.0.0.tgz",
+      "resolved": "https://registry.npmjs.org/arg/-/arg-2.0.0.tgz",
       "integrity": "sha512-XxNTUzKnz1ctK3ZIcI2XUPlD96wbHP2nGqkPKpvk/HNRlPveYrXIVSTk9m3LcqOgDPg3B1nMvdV/K8wZd7PG4w==",
       "dev": true
     },
@@ -1361,7 +1361,7 @@
         },
         "util": {
           "version": "0.10.3",
-          "resolved": "http://registry.npmjs.org/util/-/util-0.10.3.tgz",
+          "resolved": "https://registry.npmjs.org/util/-/util-0.10.3.tgz",
           "integrity": "sha1-evsa/lCAUkZInj23/g7TeTNqwPk=",
           "dev": true,
           "requires": {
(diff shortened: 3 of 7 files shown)


saschanaz:
image

The markdown form is somehow broken here, not sure why 🤔

saschanaz:
Adding respec-offending-element to the box breaks its style with `inline-block`:

image

Hmm...

saschanaz:
@marcoscaceres Do you recall why `.respec-offending-element` has `display: inline-block` and `position: relative`?

marcoscaceres:
Hmm, I don’t recall. Maybe we should add an override for some block level elements (or remove that entirely).

sidvishnoi:
> why `.respec-offending-element` has `display: inline-block` and `position: relative`?

`text-decoration-style` isn't supported everywhere (https://caniuse.com/#search=text-decoration-style), so we use a wavy background image. The background position doesn't work well on inline elements, so `inline-block` .
and `position: relative`: 🤷‍♂️

saschanaz:
>text-decoration-style isn't supported everywhere

I think we are okay to ignore Edge 18 here, and then it seems every actively maintained browser supports it.

sidvishnoi:
> we are okay to ignore Edge 18 here

Safari isn't also supporting `text-decoration-style` (@marcoscaceres can confirm)

The difference that comes from `inline-block` and `inline` isn't very noticable. I think we can remove it.

marcoscaceres:
I’m ok with removing it.

marcoscaceres:
About the broken markdown, you might need to send the validation message into the markdown parser. Standard markdown only treats top-level markdown as markdown: anything inside a HTML tag (I.e. the stuff inside the pre) will just be seen as HTML.

saschanaz:
>you might need to send the validation message into the markdown parser. 

The validation message itself is okay, what broken is the `at:` part which the `showInlineError()` attaches after the message.

marcoscaceres:
Oh, I see. Hmmm indeed. Is that going into the parser as part of showInlineError() - sorry for not checking, responding on my phone ☎️ .

saschanaz:
That part has worked well, not sure what conflicts with it.

marcoscaceres:
Maybe we should change generateMarkdownLink() to generate a real HTML link. That will still work with markdown and HTML and hopefully not break anything?

sidvishnoi:
@saschanaz Adding `\n\n` after the `</details>` fixed the markdown issue for me { [marked demo](https://marked.js.org/demo/?text=WebIDL%20validation%20error%20in%20IDL%20block%201%20at%20line%203%3A%0A%3Cdetails%3E%3Cpre%3EValidation%20error%20at%20line%203%20in%201%2C%20inside%20%60operation%20eat%60%3A%0A%20%20%20%20%20%20void%20eat(optional%20Bread%20bread)%3B%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%5E%20Optional%20dictionary%20arguments%20must%20have%20a%20default%20value%20of%20%60%7B%7D%60.%3C%2Fpre%3E%0A%3C%2Fdetails%3E%20%0A%0Aat%3A%20%5B1%5D(%23respec-off)&options=%7B%0A%20%22baseUrl%22%3A%20null%2C%0A%20%22breaks%22%3A%20false%2C%0A%20%22gfm%22%3A%20true%2C%0A%20%22headerIds%22%3A%20true%2C%0A%20%22headerPrefix%22%3A%20%22%22%2C%0A%20%22highlight%22%3A%20null%2C%0A%20%22langPrefix%22%3A%20%22language-%22%2C%0A%20%22mangle%22%3A%20true%2C%0A%20%22pedantic%22%3A%20false%2C%0A%20%22sanitize%22%3A%20false%2C%0A%20%22sanitizer%22%3A%20null%2C%0A%20%22silent%22%3A%20false%2C%0A%20%22smartLists%22%3A%20false%2C%0A%20%22smartypants%22%3A%20false%2C%0A%20%22xhtml%22%3A%20false%0A%7D&version=master) }

saschanaz:
Oops, that was tricky 😂 Thanks!

But then it looks weird that it creates a line for just a few characters. Also I don't like that every validation errors look similar before the Details part is opened. Maybe we can modify webidl2.js to expose message after `^` so that we can show it on `title` attribute and the non-detail description in the Errors popup.

saschanaz:
Maybe we can merge the Details feature into showInlineError so that it can attach the detail after the `at:` thing, not before.

marcoscaceres:
Whatever you think is best. Just really excited to see this land! 🤩

saschanaz:
Merging this, thank you all for your help!

PR #2418, opened 2019-07-12
refactor(assets/respec2): wavy by default

saschanaz:
Solving problem raised in https://github.com/w3c/respec/pull/2416#issuecomment-510440998.

### DIFF
--- a/assets/respec2.css
+++ b/assets/respec2.css
@@ -56,24 +56,17 @@ a.bibref {
   animation: pop 0.25s ease-in-out 0s 1;
 }
 
-/* TODO: Remove once https://github.com/w3c/tr-design/pull/176 is merged. */
-.respec-offending-element a[href] {
-  margin: 0;
-}
-
 .respec-offending-element {
-  display: inline-block;
-  position: relative;
-  /* Red squiggly line */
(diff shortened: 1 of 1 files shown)


assets/respec2.css
```diff
@@ -56,24 +56,17 @@ a.bibref {
...
-  display: inline-block;
-  position: relative;
-  /* Red squiggly line */
-  background: url(data:image/gif;base64,R0lGODdhBAADAPEAANv///8AAP///wAAACwAAAAABAADAEACBZQjmIAFADs=)
-    bottom repeat-x;
+  text-decoration: red wavy underline;
 }
-@supports (text-decoration-style: wavy) {
+@supports not (text-decoration: red wavy underline) {
+  .respec-offending-element:not(pre) {
```
sidvishnoi: Just to confirm, we're not adding `.respec-offending-element` to any `section`, right?

saschanaz: I'm not sure, but haven't seen such case.

saschanaz: @marcoscaceres Thoughts?

marcoscaceres: I’m 99.9% sure we are not.

PR #2419, opened 2019-07-12
feat(core/webidl): point the failing IDL block

saschanaz:
With some error style change:

Before:

image

After:

image

### DIFF
--- a/src/core/webidl.js
+++ b/src/core/webidl.js
@@ -5,11 +5,10 @@
 //  - It could be useful to report parsed IDL items as events
 //  - don't use generated content in the CSS!
 import * as webidl2 from "webidl2";
-import { flatten, showInlineWarning } from "./utils.js";
+import { flatten, showInlineError, showInlineWarning } from "./utils.js";
 import css from "text!../../assets/webidl.css";
 import { findDfn } from "./dfn-finder.js";
 import hyperHTML from "hyperhtml";
-import { pub } from "./pubsubhub.js";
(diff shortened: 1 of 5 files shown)


src/core/webidl.js
```diff
@@ -284,12 +283,13 @@ function renderWebIDL(idlElement) {
...
-      `Failed to parse WebIDL: ${e.message}.
-      <details>
-      <pre>${idlElement.textContent}\n ${e}</pre>
-      </details>`
+    showInlineError(
+      idlElement,
+      `Failed to parse WebIDL: ${e.bareMessage}.`,
+      e.bareMessage,
+      {
+        details: `<pre>${e.context}</pre>`,
```
sidvishnoi: Nit: Can we do this instead?
``` js
{ details: `<pre>${e.context}</pre>` }
```

sidvishnoi:
Why is there that much empty space after the details are open in second screenshot though?

src/core/webidl.js
saschanaz: ESLint makes me do the longer way 🤔

saschanaz: Oh, it seems somehow it allows both ways. I'll apply your suggestion.

saschanaz:
>Why is there that much empty space after the details are open in second screenshot though?

Oh, an empty `<p>` is there, hmm...

saschanaz:
Adding proper newlines removes that empty paragraph. Markdown is tricky 🤣

Issue #2427, opened 2019-07-18
tests/xref: make it easy to update test files


sidvishnoi:
Either add a stable xref `id` in test mode
or send requests directly to respec.org/xref

PR #2431, opened 2019-07-19
tests(core/xref): run tests with respec to respec.org

sidvishnoi:
Closes https://github.com/w3c/respec/issues/2427

ℹ️ Tests failing as: updates pending on respec.org


tests/spec/core/xref-spec.js
```diff
@@ -502,7 +497,10 @@ describe("Core — xref", () => {
         </section>
       </section>
     `;
-    const config = { xref: { url: xrefTestUrl("refs") }, localBiblio };
+    const config = {
+      xref: { url: `${location.origin}/tests/data/xref/refs.json` },
```
sidvishnoi: keeping this one for now, as the file contains fake dataset (due to lack of exported informative references in dataset)

saschanaz:
I think testing with respec.org is good, but I wonder the stub tests still have some use. They are our expectation about how the xref server will work, right?

sidvishnoi:
I fully agree with you. But presently, xref api is changing frequently as we find new use/edge cases and bugs. Updating tests data every time we make a change in api or respec is very painful. So, testing against respec.org until we finalize the xref API and respec integration. Will fallback to stub tests when we've stable xref.

saschanaz:
Okay, totally reasoable.

tests/spec/core/xref-spec.js
```diff
@@ -198,9 +203,9 @@ describe("Core — xref", () => {
 
   it("uses data-cite to disambiguate", async () => {
     const body = `
-      <section id="links" data-cite="service-workers">
-        <p><a>fetch</a> is defined 1 time in service-workers and 2 times in fetch.
-          It uses parent's data-cite (service-workers).</p>
+      <section id="links" data-cite="css-snappoints">
```
saschanaz: Why this and some others switched to another spec?

sidvishnoi: Due to https://github.com/sidvishnoi/respec-xref-route/pull/41

By default, respec doesn't send a for context (unless specified), causing to return unnecessary ambiguity errors - [Example](https://user-images.githubusercontent.com/8426945/61529581-5196a700-aa3f-11e9-80a2-f54eebae9bcc.png)

`fetch` is implicitly (without a `forContext`) defined only in [Fetch Standard](https://fetch.spec.whatwg.org/#concept-fetch). But earlier, xref API returned [3 results for it](https://user-images.githubusercontent.com/8426945/61529500-1a27fa80-aa3f-11e9-87ef-e659da0b25f6.png). Same reason in other tests.

saschanaz: 👍 👍 👍

sidvishnoi:
Hey! Don't approve yet. It's a WIP 😛

PR #2433, opened 2019-07-20
fix(assets/respec2): base.css removed underline from error links

sidvishnoi:
[base.css](https://www.w3.org/StyleSheets/TR/2016/base.css) was overriding broken links (`a.respec-offending-element`) and remove the squiggly underline. 
Not sure if adding `!important` is good idea or not.

image
^^ squiggly underline not shown

### DIFF
--- a/assets/respec2.css
+++ b/assets/respec2.css
@@ -56,8 +56,9 @@ a.bibref {
   animation: pop 0.25s ease-in-out 0s 1;
 }
 
-.respec-offending-element {
-  text-decoration: red wavy underline !important;
+.respec-offending-element,
+a[href].respec-offending-element {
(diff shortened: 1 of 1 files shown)


saschanaz:
I wonder the `text-decoration: none` should be split into a separate rule `a:not(.respec-offending-element) {}` just because `!important` is generally not preferred... 🤔 My suggestion is still not very pretty though.

sidvishnoi:
What about: 
``` diff
- .respec-offending-element {
+ .respec-offending-element, a[href].respec-offending-element  {
  text-decoration: red wavy underline;
}
```

saschanaz:
That seems good, which a comment about the reason.

PR #2439, opened 2019-07-24
fix(core/xref): normalize data-xref-for before fetching

marcoscaceres:
@sidvishnoi, we might want to do the same on the server: "Be conservative in what you send, be liberal in what you accept".


marcoscaceres:
Thanks for the speedy review!

PR #2441, opened 2019-07-24
feat(core/inline-idl-parser): support exceptions

marcoscaceres:
So now:

```
{{ "NotAllowedError" }}
```

Instead of ugly:

```
"{{ NotAllowedError }}"
```


src/core/inline-idl-parser.js
```diff
@@ -4,6 +4,7 @@
 import hyperHTML from "hyperhtml";
 import { showInlineError } from "./utils";
 
+const exceptionRegex = /\B"([^"]*)"\B/; // {{ "SomeException" }}
```
saschanaz: Seems exactly same as `enumValueRegex`, are we okay with it?

marcoscaceres: Yeah, it's the same... I might just kill it the enum-value one. Using `Foo["value"]` is better... alternative, is that we support both, but that's really messy.

saschanaz: Might be better to remove that first, as it seems this shadows enum one...

marcoscaceres: @sidvishnoi, I'm a bit confused as to how `Enum["value"]` becomes just "value" (which is good... but just can't figure it out).

marcoscaceres: Never mind... figured it out 🤪

sidvishnoi:
Just need to remove tests for `enum-value`

src/core/inline-idl-parser.js
```diff
@@ -4,13 +4,13 @@
...
 const methodRegex = /(\w+)\((.*)\)$/;
 const slotRegex = /^\[\[(\w+)\]\]$/;
 // matches: `value` or `[[value]]`
 // NOTE: [[value]] is actually a slot, but database has this as type="attribute"
 const attributeRegex = /^((?:\[\[)?(?:\w+)(?:\]\])?)$/;
-const enumRegex = /^(\w+)\["([\w ]+)"\]$/;
-const enumValueRegex = /\B"([^"]*)"\B/;
+const enumRegex = /^(\w+)\["([\w- ]+)"\]$/;
```
saschanaz: @sidvishnoi Why are some wrapped by `\B` and others not?

sidvishnoi: Interesting. 
Not sure, must have been done to "make things work" at that time.  I think we can remove `\B`

saschanaz:
Maybe we should mark this a breaking change?

marcoscaceres:
It's not exactly a breaking change... the "old way" is still supported, but falls back to xref's _IDL_ magic matcher (so should remain backwards compatible). The quotes makes it explicit that what's inside is an exception.

sidvishnoi:
I think @saschanaz meant by "remove empty enum-value support" in 3057b30

saschanaz:
Not sure I understand...

`{{ "str" }}` previous explicitly represented enum-value, now it explicitly represents exception, and the previous way to write enum-value automatically fallbacks?

> I think @saschanaz meant by "remove empty enum-value support" in [3057b30](https://github.com/w3c/respec/commit/3057b30f2034ed78f2750146f18cac48b3da3847)

Oops, didn't even notice that. How can we represent empty enum value after this PR?

sidvishnoi:
> How can we represent empty enum value after this PR?

Linking a enum-value without the Enum context will not be supported. The syntax for enum/enum-value will be `{{ Enum["enum value"] }}` which will render as `<code><a>"enum value"</a></code>`

marcoscaceres:
> How can we represent empty enum value after this PR?

Do you mean?: 
```
{{ SomeEnum[""] }}
```

marcoscaceres:
> {{ "str" }} previous explicitly represented enum-value, now it explicitly represents exception, and the previous way to write enum-value automatically fallbacks?

Ah, got it. Yes, in that sense, it kinda is a breaking change. I'm conflicted, because xref and the microsyntax is still very much "beta".

src/core/inline-idl-parser.js
sidvishnoi:
```suggestion
const enumRegex = /^(\w+)\["([\w- ]*)"\]$/;
```

for empty enum value (missed it!)

saschanaz:
> 
> 
> > How can we represent empty enum value after this PR?
> 
> Do you mean?:
> 
> ```
> {{ SomeEnum[""] }}
> ```

`/^(\w+)\["([\w- ]+)"\]$/.test('SomeEnum[""]') === false`, because of `+`. I think we shouldn't remove test for the empty enum value in https://github.com/w3c/respec/commit/3057b30f2034ed78f2750146f18cac48b3da3847...

marcoscaceres:
@saschanaz, you are, of course, correct. Adding a test. I think @sidvishnoi just fixed it with his suggestion.

tests/spec/core/xref-spec.js
```diff
@@ -763,19 +763,13 @@ describe("Core — xref", () => {
           enum Foo { "dashed-thing", "" };
           </pre>
           <p id="link1">{{ ServiceWorkerUpdateViaCache["imports"] }}</p>
-          <p id="link3"
-            data-cite="css-layout-api" data-xref-for="ChildDisplayType"
-          >{{ "block" }} {{"block"}} </p>
```
sidvishnoi: this won't be supported.

src/core/inline-idl-parser.js
```diff
@@ -132,6 +132,7 @@ function renderEnum(details) {
     data-xref-type="enum-value"
     data-link-for="${identifier}"
     data-xref-for="${identifier}"
+    data-lt="${!enumValue ? "the-empty-string" : null}"
```
saschanaz: Is this Shepherd compatible? What does Shepherd to for this?

sidvishnoi: Good catch. It isn't.
Need to convert `the-empty-string` back to `""` in core/xref. 
For local linking, need to keep using `the-empty-string`, right? (I used this as we were already using it until 57d9bbc).

tests/spec/core/xref-spec.js
```diff
@@ -762,25 +765,32 @@ describe("Core — xref", () => {
...
-      expect(link1.querySelector("a").textContent).toBe("imports");
+      expect(link1.textContent).toBe(`"no-referrer"`);
+      expect(link1.querySelector("a").textContent).toBe("no-referrer");
       expect(link1.querySelector("a").href).toBe(
-        expectedLinks.get("ServiceWorkerUpdateViaCache.imports")
+        "https://www.w3.org/TR/referrer-policy/#dom-referrerpolicy-no-referrer"
+      );
+
+      const link2 = doc.getElementById("link2");
+      // expect(link2.textContent).toBe(`""`); // TODO
```
sidvishnoi: It's coming out to be `"Referrer Policy"` 😕

marcoscaceres: magic...

PR #2443, opened 2019-07-24
fix(core/webidl): ensure generated dfns get an id

marcoscaceres:


src/core/webidl.js
```diff
@@ -155,7 +157,9 @@ function defineIdlName(escaped, data, parent) {
   if (showWarnings) {
     const styledName = data.type === "operation" ? `${name}()` : name;
     const ofParent = parentName ? ` \`${parentName}\`'s` : "";
-    const msg = `Missing \`<dfn>\` for${ofParent} \`${styledName}\` ${data.type}. [More info](https://github.com/w3c/respec/wiki/WebIDL-thing-is-not-defined).`;
+    const msg = `Missing \`<dfn>\` for${ofParent} \`${styledName}\` ${
```
saschanaz: Maybe you should `npm i` again as AFAIK newer eslint prefers our current styling.

marcoscaceres: trying...

marcoscaceres: Ran `eslint --fix src/core/webidl.js` after update... no change?

src/core/webidl.js
```diff
@@ -139,8 +139,10 @@ function defineIdlName(escaped, data, parent) {
      data-lt="default toJSON operation">${escaped}</a>`;
   }
   if (!data.partial) {
-    return hyperHTML`<dfn data-export data-dfn-type="${linkType}" data-dfn-for="${parent &&
+    const dfn = hyperHTML`<dfn data-export data-dfn-type="${linkType}" data-dfn-for="${parent &&
       parent.name}">${escaped}</dfn>`;
+    decorateDfn(dfn, data, parentName, name);
```
saschanaz: Maybe we could remove `decorateDfn` call in `findDfn` because we will always decorate dfns regardless of existence of links.

src/core/webidl.js
marcoscaceres: oh wait... maybe that's my global one.

src/core/webidl.js
marcoscaceres: yeah... and it's a bit less ugly... "finding" things shouldn't have side effects.... like ids being added. Long wanted to change that.

marcoscaceres: Argh... actually, looking at the code, that change is a bit more involved than I first thought. Might need to do that as a new PR.

marcoscaceres: Filed: https://github.com/w3c/respec/issues/2444

marcoscaceres:
@saschanaz, I think this should be good to go.

src/core/dfn-finder.js
```diff
@@ -150,10 +150,14 @@ function findNormalDfn(defn, parent, name) {
...
+export function decorateDfn(dfn, defn, parent, name) {
   if (!dfn.id) {
-    const middle = parent ? `${parent}-` : "";
-    const last = name.replace(/[()]/g, "").replace(/\s/g, "-");
+    const middle = parent ? `${parent.toLowerCase()}-` : "";
+    let last = name
+      .toLowerCase()
+      .replace(/[()]/g, "")
+      .replace(/\s/g, "-");
+    if (last === "") last = "the-empty-string";
```
saschanaz: This part is always tricky 😄

marcoscaceres: yeah, agree... I thought to just run with it. If it needs something more robust, can fix if it exhibits issues.

saschanaz: Yeah, I'm okay with it for now.

PR #2449, opened 2019-07-26
fix(core/figures): generate tof even when no figcaption

saschanaz:
Currently the whole module fails when there is no valid figcaption, because `getTableOfFiguresListItem` expects a caption element.


saschanaz:
Yay, thanks for the quick review

PR #2450, opened 2019-07-26
fix(core/id-headers): always add header ids

saschanaz:
Regression from #2292 😁


sidvishnoi:
How come we write tests and still miss stuff? 😰 😅

saschanaz:
Review complete even before Travis pass! ⚡😍

saschanaz:
> How come we write tests and still miss stuff?

More and more tests 💪

PR #2456, opened 2019-07-31
feat(core/inlines): add {{ forContext/idl_term }} syntax

sidvishnoi:
This is the simplest way I could think of. Any other ideas are welcome.


tests/spec/core/inlines-spec.js
```diff
@@ -367,4 +367,31 @@ describe("Core - Inlines", () => {
       "https://infra.spec.whatwg.org/#map-iterate"
     );
   });
+
+  it("processes {{ forContext/term }} IDL", async () => {
+    const body = `
+      <section>
+        <p id="link1">{{ Window.event }}</p>
+        <p id="link2">{{ Window/event }}</p>
```
saschanaz: So this is to render `event` while linking to `Window.event`, right? Personally thought Wikipedia-like `{{ Window.event | event }}` but this one looks clean.

sidvishnoi: The Wikipedia-like also looks interesting. 
Went with `{{ Window/event }}` as we already have `[= For/link =]`. 

`{= Window/event =}` hmmm 🤔

sidvishnoi: Wondering if we want to support  this now
```
{{
  Window
  /
  event
}}
```

saschanaz: Do `[= Window/event =]` and `{{ Window/event }}` link to the same target?

sidvishnoi: No. The second one is specific to IDL (and a guarantee that the terms won't be lower cased).

src/core/inline-idl-parser.js
```diff
@@ -29,35 +29,35 @@ function parseInlineIDL(str) {
     if (methodRegex.test(value)) {
       const [, identifier, allArgs] = value.match(methodRegex);
       const args = allArgs.split(/,\s*/).filter(arg => arg);
-      results.push({ type: "method", identifier, args });
+      results.push({ type: "method", identifier, args, str });
```
saschanaz: Could we pass the result of `str.includes("/")` instead, with its field name something like `explicitContext`?

sidvishnoi: or `renderParent`

tests/spec/core/inlines-spec.js
marcoscaceres: Yes, because HTML Tidy might produce some variation of that.

marcoscaceres: Just noting that {{Foo/bar}} will match what BikeShed does also.

src/core/inline-idl-parser.js
```diff
@@ -72,18 +73,19 @@ function parseInlineIDL(str) {
 
 function renderBase(details) {
   // Check if base is a local variable in a section
-  const { identifier } = details;
+  const { identifier, renderParent } = details;
+  if (!renderParent) return document.createDocumentFragment();
```
saschanaz: Maybe should just return without a value?

src/core/inline-idl-parser.js
```diff
@@ -19,45 +19,47 @@ const methodSplitRegex = /\.?(\w+\(.*\)$)/;
 function parseInlineIDL(str) {
   const [nonMethodPart, methodPart] = str.split(methodSplitRegex);
   const tokens = nonMethodPart
-    .split(".")
+    .split(/[./]/)
     .concat(methodPart)
-    .filter(s => s && s.trim());
+    .filter(s => s && s.trim())
```
sidvishnoi: maybe should use `array.reduce` here?

marcoscaceres: 🤔 what’s there looks ok.

sidvishnoi: didn't like `s.trim()` being called twice

tests/spec/core/inlines-spec.js
```diff
@@ -367,4 +367,39 @@ describe("Core - Inlines", () => {
       "https://infra.spec.whatwg.org/#map-iterate"
     );
   });
+
+  fit("processes {{ forContext/term }} IDL", async () => {
```
sidvishnoi: heh.
```suggestion
  it("processes {{ forContext/term }} IDL", async () => {
```

src/core/inline-idl-parser.js
marcoscaceres: Ah, I see. They are semantically slightly different because of the usage context. The filter could regex to .test() matching just whitespace... but trim() does declaratively the same thing.

src/core/inline-idl-parser.js
sidvishnoi: hyperhtml shouts at
```
render`<code>${output}</code>`
```
if `output` contains `undefined`

saschanaz: We still can return without value, and then do not `.push()` when caller gets nothing.

PR #2459, opened 2019-08-01
chore(karma): report when too slow

saschanaz:
Some tests are slower than expected, mainly because of network requests. This PR adds a way to be alerted.

### DIFF
--- a/tests/spec/core/ui-spec.js
+++ b/tests/spec/core/ui-spec.js
@@ -24,21 +24,12 @@ describe("Core - UI", () => {
     expect(window.getComputedStyle(menu).display).toBe("none");
     doc.getElementById("respec-pill").click();
     // spin the event loop
-    await new Promise(resolve => {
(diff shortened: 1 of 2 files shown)


tests/spec/core/ui-spec.js
```diff
@@ -24,21 +24,12 @@ describe("Core - UI", () => {
...
-    });
-    // Allow time to fade in
-    await new Promise(resolve => {
-      setTimeout(() => {
-        expect(window.getComputedStyle(menu).display).toBe("none");
-        resolve();
-      }, 500);
-    });
-    // give it time to fade out
+    await new Promise(resolve => setTimeout(resolve));
```
sidvishnoi: should we move this as a util in SpecHelper.js?

saschanaz: If we get some more use cases, sure.

Issue #2288, opened 2019-04-18
xref linking fails in Firefox private window


sidvishnoi:
## Important info

* URL to affected spec: https://w3c.github.io/payment-request/
* ReSpec version: 24.17.0
* [x] I did a "hard refresh", but it's still busted. 

Firefox doesn't allow to use IDB in private window, so xref linking breaks as stuff just throws (`"A mutation operation was attempted on a database that did not allow mutations."`). We should catch that error and fetch from network directly if IDB fails.
Related Firefox bug: https://bugzilla.mozilla.org/show_bug.cgi?id=781982

PR #2463, opened 2019-08-02
refactor(core/xref): split its cache feature

saschanaz:
Helps #2187 to make caching parts compatible with Node.js.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
     "@types/clipboard": "^2.0.1",
     "@types/pluralize": "0.0.29",
     "babel-eslint": "^10.0.2",
+    "babel-plugin-dynamic-import-amd": "github:holyzfy/babel-plugin-dynamic-import-amd",
     "chai": "^4.2.0",
     "domReady": "github:requirejs/domReady#d85bdc38a6df868ead52ab3e80715aaf60e68765",
     "eslint": "^6.1.0",
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "@types/clipboard": "^2.0.1",
     "@types/pluralize": "0.0.29",
     "babel-eslint": "^10.0.2",
-    "babel-plugin-dynamic-import-amd": "github:holyzfy/babel-plugin-dynamic-import-amd",
+    "babel-plugin-dynamic-import-amd": "github:holyzfy/babel-plugin-dynamic-import-amd#1bbaf86d4472c2eef0fc01f3199d637a4b6c51e2",
     "chai": "^4.2.0",
     "domReady": "github:requirejs/domReady#d85bdc38a6df868ead52ab3e80715aaf60e68765",
     "eslint": "^6.1.0",
--- a/src/core/xref-db.js
+++ b/src/core/xref-db.js
@@ -1,0 +1,79 @@
+// @ts-check
+import { IDBKeyVal } from "./utils.js";
+
+/**
+ * @typedef {import('core/xref').RequestEntry} RequestEntry
+ * @typedef {import('core/xref').Response} Response
+ * @typedef {import('core/xref').SearchResultEntry} SearchResultEntry
+ */
+
+const CACHE_MAX_AGE = 86400000; // 24 hours
+
+/**
+ * Temporary workaround until browsers get import-maps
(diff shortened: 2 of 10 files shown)


saschanaz:
This includes some magic related to dynamic imports, which can also help #2228 by falling back to relative paths when applicable.

src/core/xref-db.js
```diff
@@ -0,0 +1,79 @@
...
+ */
+async function importIdb() {
+  try {
+    return await import("idb");
+  } catch {
+    return await import("../../node_modules/idb/build/esm/index.js");
+  }
+}
+
+async function getIDBKeyValForXref() {
```
sidvishnoi: Nit: `getXrefCache`?

src/core/xref-db.js
```diff
@@ -0,0 +1,79 @@
...
+ * @param {Map<string, SearchResultEntry[]>} data
+ */
+export async function cacheXrefData(data) {
+  const cache = await getIDBKeyValForXref();
+  // add data to cache
+  await cache.addMany(data);
+  await cache.set("__CACHE_TIME__", Date.now());
+}
+
+export async function clearXrefData() {
```
sidvishnoi: Nit: Maybe remove the `Xref` labels from these functions as the module is `xref-db.js`? It's a bit verbose IMO.

src/core/xref-db.js
```diff
@@ -0,0 +1,79 @@
...
+  return await resolveFromCache(uniqueQueryKeys, cache);
+}
+
+/**
+ * @param {RequestEntry[]} keys
+ * @param {IDBKeyVal} cache
+ * @returns {Promise<Map<string, SearchResultEntry[]>>}
+ */
+async function resolveFromCache(keys, cache) {
+  const cacheTime = await cache.get("__CACHE_TIME__");
```
sidvishnoi: Warning: This can throw if cache in't available. Happens in Firefox private window (See: https://github.com/w3c/respec/issues/2288)

src/core/xref-db.js
saschanaz: It would be a bit ambiguous at the caller part without the labels, and IDEs including VSCode can autoimport better when the name is more specific.

But maybe it's not very important as these functions are mainly used for a specific set of modules. Hmm.

src/core/xref-db.js
saschanaz: Oops, accidentally removed try-catch on callers

src/core/xref-db.js
saschanaz: Bit confusing whether it's for retrieving cache data or an IDB handle...

sidvishnoi: As we have `const cache = await getIDBKeyValForXref();`, so I'll say it gets the cache.

package.json
```diff
@@ -27,6 +27,7 @@
     "@types/clipboard": "^2.0.1",
     "@types/pluralize": "0.0.29",
     "babel-eslint": "^10.0.2",
+    "babel-plugin-dynamic-import-amd": "github:holyzfy/babel-plugin-dynamic-import-amd",
```
sidvishnoi: Why are we getting this from a fork instead of `npm`? Also, maybe lock its version to a tag/branch/commit.

saschanaz: Because the original npm package is so old and incompatible with Babel 7. RequireJS is old anyway...

src/core/xref-db.js
```diff
@@ -0,0 +1,84 @@
...
+  return Date.now() - cachedTime > CACHE_MAX_AGE;
+}
+
+/**
+ * @param {Map<string, SearchResultEntry[]>} data
+ */
+export async function cacheXrefData(data) {
+  const cache = await getIdbCache();
+  // add data to cache
+  await cache.addMany(data);
```
sidvishnoi: Missed this one? It can throw too.. 
Maybe we can get some instance with compatible API from `utils.js@IDBKeyVal`? We can avoid try-catch during calls then.

sidvishnoi: Update: updating `utils.js@IDBKeyVal` won't work (like I was thinking), as the function throws in `getIdbCache`.

PR #2465, opened 2019-08-04
fix(core/inline-idl-parser): better support IDL primitives

marcoscaceres:
We didn't support {{ unrestricted double }} {{ long long }} and friends

### DIFF
--- a/tests/spec/core/xref-spec.js
+++ b/tests/spec/core/xref-spec.js
@@ -799,6 +799,37 @@ describe("Core — xref", () => {
       );
     });
 
+    it("links idl primitives that have spaces", async () => {
+      const body = `
+      <section id="test">
+        {{
+          unsigned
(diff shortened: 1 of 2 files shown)


tests/spec/core/xref-spec.js
```diff
@@ -799,6 +799,37 @@ describe("Core — xref", () => {
...
+      const body = `
+      <section id="test">
+        {{
+          unsigned
+          long
+          long
+        }}
+        {{  unrestricted   float
+        }}
+        {{double}}
```
marcoscaceres: These fall back to "base" type in the inline parser... just here for regression testing.

sidvishnoi:
Is `Uint32Array` possible?

marcoscaceres:
Yes, those are handled by “base”.

PR #2468, opened 2019-08-06
Add until date to formereditor

fycdev:
**In Progress**

Add the ability to render a `retiredDate` property. In addition, if an editor has a `retiredDate` it is moved to the `formerEditors` list.

Still needs:
- [x]  testing
- [x] localisation
- [x] documentation in wiki.


src/w3c/headers.js
```diff
@@ -395,18 +395,22 @@ export function run(conf) {
       }
     }
   };
+  if (!conf.formerEditors) conf.formerEditors = [];
   if (conf.editors) {
     conf.editors.forEach(peopCheck);
+    conf.formerEditors = conf.editors.reduce((collector, person) => {
```
marcoscaceres: Probably want a comment here about what this is actually doing... relocating editor to former editor, iff `retiredDate` member is present.

examples/basic.html
```diff
@@ -15,6 +15,12 @@
         editors: [{
           name: "Your Name",
           url: "https://your-site.com",
+          note: "blah"
```
marcoscaceres: Once this is all working, please revert changes to basic.html

marcoscaceres:
Looking good. We'll need to add some tests :)

examples/basic.html
fycdev: Would you like to include `retiredDate` example in `basic.html`?

marcoscaceres: Probably not. But we need to remember to add it to the wiki when we are ready to land this.

PR #2473, opened 2019-08-19
enhancement(core/xref): improve cache busting strategy

sidvishnoi:
Send a request to https://respec.org/xref/meta/version/ to know the last updated timestamp of data update (which changes on server restart and server data update).

Benefit: prevent client having potentially erroneous data
Trade-off: additional request to server


src/core/xref-db.js
```diff
@@ -52,10 +51,22 @@ async function resolveFromCache(keys, cache) {
...
 /**
+ * Get lastUpdated time from server and bust cache based on that. This way, we
+ * prevent dirty/erroneous/stale data being kept on a client (which is possible
+ * if we use a `MAX_AGE` based caching strategy).
  * @param {number} cachedTime
  */
-function isBustedCache(cachedTime) {
-  return Date.now() - cachedTime > CACHE_MAX_AGE;
+async function isBustedCache(cachedTime) {
+  const url = new URL("meta/versions", API_URL);
```
sidvishnoi: Typo
```suggestion
  const url = new URL("meta/version", API_URL);
```

src/core/xref-db.js
```diff
@@ -52,10 +51,22 @@ async function resolveFromCache(keys, cache) {
 }
 
 /**
+ * Get lastUpdated time from server and bust cache based on that. This way, we
```
sidvishnoi:
```suggestion
 * Get last updated timestamp from server and bust cache based on that. This way, we
```

src/core/xref-db.js
```diff
@@ -52,10 +51,22 @@ async function resolveFromCache(keys, cache) {
...
-function isBustedCache(cachedTime) {
-  return Date.now() - cachedTime > CACHE_MAX_AGE;
+async function isBustedCache(cachedTime) {
+  const url = new URL("meta/version", API_URL);
+  try {
+    const res = await fetch(url);
+    if (!res.ok) throw new Error(res.statusText);
+    const lastUpdated = parseInt(await res.text(), 10);
+    return lastUpdated > cachedTime;
+  } catch {
```
marcoscaceres: Probably best to just console.error(err) here ... silent exceptions "no es bueno"

sidvishnoi: On a second thought, I think this try-catch is redundant. we can return `false` right away on `res.ok` instead of throwing and later catching

marcoscaceres: sounds good.

src/core/xref-db.js
```diff
@@ -52,10 +51,17 @@ async function resolveFromCache(keys, cache) {
...
 /**
+ * Get last updated timestamp from server and bust cache based on that. This way, we
+ * prevent dirty/erroneous/stale data being kept on a client (which is possible
+ * if we use a `MAX_AGE` based caching strategy).
  * @param {number} cachedTime
  */
-function isBustedCache(cachedTime) {
-  return Date.now() - cachedTime > CACHE_MAX_AGE;
+async function isBustedCache(cachedTime) {
+  const url = new URL("meta/versions", API_URL);
```
sidvishnoi:
```suggestion
  const url = new URL("meta/version", API_URL);
```

src/core/xref-db.js
```diff
@@ -52,10 +53,22 @@ async function resolveFromCache(keys, cache) {
 }
```
sidvishnoi: should reset time here. otherwise `VERSION_CHECK_WAIT` later, all `version` requests will be networked

saschanaz:
I wonder this can be done in a single request, where the request would include the timestamp and if the server thinks the client has the latest data, it sends nothing back.

sidvishnoi:
That would require client telling server what data it has - all query ids (cached ones) as well as a timestamp. We need to send query ids as the markup may have changed (for example, while editing).
The request is sent once per 5min (on multiple visits) - so the overhead is small.

saschanaz:
Ah, okay, sounds good then 👍

PR #2475, opened 2019-08-20
enhancement(core/inlines): extend method syntax support

sidvishnoi:
Allows `{{EventListener/handleEvent}}` to be linked as a method


saschanaz:
I assume this is supported by Bikeshed, am I right?

sidvishnoi:
Not sure about Bikeshed compatibility on this one. 
This was faced during https://github.com/w3c/geolocation-api/pull/33 and we looked at alternatives as:
1. add "method" to types for `renderAttribute`, increasing the chances of a potential ambiguity but works as methods and attributes don't generally share same name - this one
2. Link it as `{{EventListener/handleEvent()}}` and let xref API treat this method with no args as same as `{{EventListener/handleEvent}}` (something like [`textVariations`](https://github.com/sidvishnoi/respec-xref-route/blob/3d34fb159ec55ed7e36d14009c821b412b290b8b/utils.ts#L8), but for IDL) - not trivial to implement.
3. `{{ EventListener/handleEvent(event) }}` - not ideal.

marcoscaceres:
I think we should still support 2, as it’s weird that handleEvent doesn’t finish with “()” if it’s a method.

marcoscaceres:
So we could say the current behaviour is correct, in as far that handleEvent is not an attribute or a dictionary member.

sidvishnoi:
> as it’s weird that handleEvent doesn’t finish with “()” if it’s a method.

`handleEvent()` also doesn't sound correct as the [`event` argument is required](https://dom.spec.whatwg.org/#dom-eventlistener-handleevent). But as `handleEvent` is more natural to write, as it's less verbose, we should support that.

I also think that we should support textVariations for IDL, but not sure in which cases.

marcoscaceres:
I don’t mind supporting both. The question is really about how Editors write about methods in specs. Like, one would not write, “the handleEvent(foo) method”... it’s more common to just write “handleEvent()” even if an argument is required.

sidvishnoi:
Close this in favour of https://github.com/sidvishnoi/respec-xref-route/pull/51?

marcoscaceres:
Yep, as long as it works:)

marcoscaceres:
Ok, updated the Geolocation  spec to use “handleEvent()”

marcoscaceres:
Maybe we still want this as a way of explicitly telling xref what we want?

sidvishnoi:
> Maybe we still want this as a way of explicitly telling xref what we want?

Not sure I follow. What would we want to tell? In particular, why we would want to allow `{{ handleEvent }}` to be treated as a method?

marcoscaceres:
> Not sure I follow. What would we want to tell? In particular, why we would want to allow {{handleEvent}} to be treated as a method?

sorry, wasn't clear... I mean to say "as long as {{whatEver()}} works"

sidvishnoi:
Closing as things are working fine already.

PR #2484, opened 2019-08-23
fix(core/dfn-finder): dfn-for should not be lcased

### LINKED ISSUE #2483: Exports should retain case
When IDL exporting definitions, the definitions should retain the same case form. Right now, we are exporting everything as lower case.

marcoscaceres:
closes #2483


marcoscaceres:
Darn... the stupid/legacy lower casing search stuff strikes again. I think we finally need to kill lower case stuff of IDL lookups.

marcoscaceres:
It's a big job, so I'll have to come back to this after web share.... But we this is high priority, as all our exports are to Shepherd are currently wrong.

saschanaz:
Oh, I assume the bikeshed people are angry 😨

marcoscaceres:
Not yet... but they will be.

marcoscaceres:
@saschanaz do you see some sane way of doing this? I tried a bunch of different things but just ended breaking it 😢

saschanaz:
Hmmm I'll look into it

saschanaz:
Fixed tests except the "uniquely links to enum values" test. That thing has been already broken because:

* The previous behavior was to insert `data-dfn-for="test2"` to `<dfn>`s without `data-dfn-for` when a parent has `data-dfn-for="Test2"`, __without lowercasing the parent's attribute__.
* Then the definition finder tries to find the definition for `Test2["enum"]` with `closest('[data-dfn-for="test2"]')`.
* It somehow worked because only one `<dfn>` had `data-dfn-for="test2"` while the common parent had `data-dfn-for="Test2"`
* Now it fails, the finder now does `closest('[data-dfn-for="Test2"]')`.

saschanaz:
Okay, tests pass. We should probably mark this as a breaking change because things are now case sensitive.

marcoscaceres:
Woah, you are the best! 🙇

saschanaz:
@sidvishnoi @marcoscaceres The issue continues. Could you please check whether the repo settings page includes something relevant?

marcoscaceres:
Ok, should be good now 🙂

saschanaz:
@marcoscaceres Great! What was it?

marcoscaceres:
I’d checked too many options in the branch protection section. I unchecked the ones that were stuck. I also gave you both admin access if we need change anything.

PR #2489, opened 2019-08-27
fix: allow linking in multiple ways

marcoscaceres:
The changes from #2484 turned out to be too breaking - in particular, we want to be case strict exclusively for WebIDL, but we want to allow any case for concepts. Otherwise, this would break:

```HTML
  <dfn>a definition</dfn>.  
  
  <a>A definition</a>   
```

Which is super common. 

This supports both approaches and fixes a couple of linking bugs while we are at it.


src/core/dfn-map.js
```diff
@@ -8,10 +8,10 @@ export const definitionMap = Object.create(null);
  * @param {string[]} names Names to register the element by
  */
 export function registerDefinition(dfn, names) {
-  for (const name of names) {
-    if (!definitionMap[name]) {
+  for (const name of names.map(name => name.toLowerCase())) {
+    if (name in definitionMap === false) {
       definitionMap[name] = [dfn];
-    } else if (!definitionMap[name].includes(dfn)) {
+    } else if (definitionMap[name].includes(dfn) === false) {
```
saschanaz: I understand the desire to use `===` in `in` expression as it can remove parentheses, but with `includes()`? Not sure...

src/core/dfn-finder.js
```diff
@@ -52,15 +51,29 @@ function tryFindDfn(defn, parent, name) {
...
+    case "operation": {
+      // Allow linking to both "method()" and "method" name.
+      const asMethodName = `${name}()`;
+      const asFullyQualifiedName = `${asQualifiedName}()`;
+      return [asFullyQualifiedName, asQualifiedName, asMethodName, name];
+    }
+    case "attribute":
+      return [asQualifiedName, name];
+  }
+  return;
```
saschanaz:
```suggestion
```

Because it's redundant 😜

src/core/link-to-dfn.js
```diff
@@ -17,36 +17,58 @@ import { pub } from "./pubsubhub.js";
...
+    },
+    duplicateTitle: "This is defined more than once in the document.",
   },
 };
 const lang = defaultLang in l10n ? defaultLang : "en";
 
-export async function run(conf) {
-  document.normalize();
+class CaseInsensitiveMap extends Map {
+  constructor(/** entries: Array<[String, HTMLElement]> */ entries = []) {
```
saschanaz:
```suggestion
  /**
   * @param {Array<[String, HTMLElement]>} [entries]
   */
  constructor(entries = []) {
```

src/core/dfn-map.js
```diff
@@ -8,10 +8,10 @@ export const definitionMap = Object.create(null);
  * @param {string[]} names Names to register the element by
  */
 export function registerDefinition(dfn, names) {
-  for (const name of names) {
-    if (!definitionMap[name]) {
+  for (const name of names.map(name => name.toLowerCase())) {
```
saschanaz: Are we disabling case sensitivity altogether here?

saschanaz:
We should also make sure that lowercased IDL names fail to link, as I mistakenly didn't add that test...

src/core/dfn-map.js
marcoscaceres: I added the `false`s for (my own) legibility. As this code runs, each of the checks are somewhat complex and having the explicit `false` was just to helped me with cognitive load. I'm ok to remove them tho.

src/core/dfn-map.js
marcoscaceres: Yeah... unfortunately, I think we are kinda blocked by `definitionMap` being an Object because of legacy reasons. Eventually we need to get the ARIA specs to move off `conf.definitionMap`  being and object and we can do  something more clever.

saschanaz: IMO no need to be blocked, because... we have ES2015 Proxies.

marcoscaceres: True. We could def do that as a followup.

marcoscaceres:
> We should also make sure that lowercased IDL names fail to link, as I mistakenly didn't add that test...

Will add

marcoscaceres:
Actually, these will link any IDL case ones too - but that's ok for now. The immediate concern is that we ship a fix for `data-dfn-for` being in the correct case form, as that's poisoning the data on Shepherd.

saschanaz:
I feel we could do this without breaking IDL linking...

BTW, IMO this includes some useful refactors as well as the fixes. Could we split into two PRs so that I can see which part is a fix and which part is just a refactor?

marcoscaceres:
> I feel we could do this without breaking IDL linking...

Possible, but unfortunately I'm constrained for time to keep working on it much this week. 

> Could we split into two PRs so that I can see which part is a fix and which part is just a refactor?

Will try.

marcoscaceres:
Hmm... I was hoping src/core/link-to-dfn.js could be treated as a refactor, but unfortunately it breaks if I pull it out individually :(

marcoscaceres:
I undid some of the variable renames to make the patch a bit smaller... I think that's as much as I can do tho.

src/core/dfn-finder.js
```diff
@@ -174,11 +176,18 @@ function getDfns(dfnForArray, parent, originalName, type) {
...
-  // an explicitly empty [for], try <dfn> that inherited a [for].
+  const dfns = [];
+  for (const dfn of dfnForArray) {
+    /** @type {HTMLElement} */
+    const closestDfnFor = dfn.closest(`[data-dfn-for]`);
+    if (!closestDfnFor) {
+      continue;
+    }
+    if (closestDfnFor.dataset.dfnFor === parent) dfns.push(dfn);
+  }
```
saschanaz: I prefer the reverse of b879864043a616057919492054ad1206a1b38bf8:

```js
const dfns = dfnForArray.filter(dfn =>
  dfn.closest(`[data-dfn-for="${parent}"]`)
)
```

src/core/dfn.js
```diff
@@ -9,13 +9,9 @@ export const name = "core/dfn";
 
 export function run() {
   document.querySelectorAll("dfn").forEach(dfn => {
-    /** @type {HTMLElement} */
-    const closestDfn = dfn.closest("[data-dfn-for]");
-    if (closestDfn && closestDfn !== dfn && !dfn.dataset.dfnFor) {
-      dfn.dataset.dfnFor = closestDfn.dataset.dfnFor;
```
saschanaz: I feel this might affect Shepherd to miss `data-dfn-for`, are we sure this is okay?

src/core/dfn-finder.js
marcoscaceres: As long as the result is the same, I don’t mind. Feel free to update.

src/core/dfn.js
marcoscaceres: Ah, good point. We might indeed need to transplant data-dfn-for explicitly.

marcoscaceres: Note, we can't do this for all definitions yet... only for IDL ones for now, because concept definitions are currently not "for" anything.

saschanaz: > Note, we can't do this for all definitions yet... only for IDL ones for now, because concept definitions are currently not "for" anything.

IDL ones get their `data-dfn-for` from `decorateDfn()` which means this should be okay. 👍

marcoscaceres: ok cool. Remove this conversation?

saschanaz: Oh wait, why do we need ``dfn.closest(`[data-dfn-for="${parent}"]`)`` then?

marcoscaceres: having a look...

marcoscaceres: Do you mean in src/core/dfn-finder.js? that's exclusively used by core/webidl.

saschanaz: Ah yes, `findDfn()` uses `closest()` to find definitions and then assign the right value by `decorateDfn()`. 👍

src/core/link-to-dfn.js
```diff
@@ -173,9 +176,8 @@ function findLinkTarget(target, anchor, titleToDfns, possibleExternalLinks) {
       anchor.classList.add("internalDFN");
     }
   }
-  // add a bikeshed style indication of the type of link
   if (!anchor.hasAttribute("data-link-type")) {
-    anchor.dataset.linkType = "dfn";
+    anchor.dataset.linkType = "idl" in dfn.dataset ? "idl" : "dfn";
```
saschanaz: Why do we have `data-link-type`? The bikeshed documentation only mentions `data-link-type=biblio` and nothing more...

marcoscaceres: The `data-link-type` classifies things into the appropriate types... do a view-source on https://fetch.spec.whatwg.org and you should see the different types. 

For the `biblio` types, we should add support for that too.

marcoscaceres: Filed https://github.com/w3c/respec/issues/2492 so we don't forget.

marcoscaceres: @saschanaz here is the link types: https://tabatkins.github.io/bikeshed/#link-types

saschanaz: Thanks! But funny that the list does not have `biblio`.

saschanaz: Does this mean that currently core/webidl is assigning wrong values on data-link-type? It assigns IDL type instead of `dfn`/`idl`.

marcoscaceres: potentially, yes. We are using "idl" as a catch all. However, we should fix that as a followup if it's not compatible.

src/core/utils.js
```diff
@@ -639,18 +630,21 @@ export function getLinkTargets(elem) {
...
+  const results = titles.reduce((result, title) => {
+    // supports legacy <dfn>Foo.Bar()</dfn> definitions
     const split = title.split(".");
     if (split.length === 2) {
       // If there are multiple '.'s, this won't match an
       // Interface/member pair anyway.
       result.push({ for: split[0], title: split[1] });
     }
-    result.push({ for: "", title });
+    result.push({ for: linkFor, title });
```
saschanaz: Why reordering? Is this for some kind of priority thing or just readability?

tests/spec/core/pluralize-spec.js
```diff
@@ -18,7 +18,6 @@ describe("Core - Pluralize", () => {
 
     const dfnFoo = doc.querySelectorAll("#section dfn")[0];
     expect(dfnFoo.id).toBe("dfn-foo");
-    expect(dfnFoo.dataset.lt).toBeFalsy();
```
saschanaz: Do we need to make sure `data-lt` exists?

src/core/dfn-finder.js
```diff
@@ -174,11 +176,18 @@ function getDfns(dfnForArray, parent, originalName, type) {
...
   if (dfns.length === 0 && parent === "" && dfnForArray.length === 1) {
-    return dfnForArray;
+    // Make sure the name exactly matches
+    return dfnForArray[0].textContent === originalName ? dfnForArray : [];
```
saschanaz: Is this just to make sure or did you see some mismatches?

src/core/dfn.js
```diff
@@ -9,13 +9,9 @@ export const name = "core/dfn";
...
-    }
-    // TODO: we should probably use weakmaps and weaksets here to avoid leaks.
-    const titles = getDfnTitles(dfn, { isDefinition: true });
+    const titles = getDfnTitles(dfn);
     registerDefinition(dfn, titles);
+    if (titles.length) dfn.dataset.lt = titles.join("|");
+    if (!dfn.dataset.dfnType) dfn.dataset.dfnType = "dfn";
```
saschanaz: Review note for myself: moved from `getDfnTitles` to reduce side effects 👍

src/core/link-to-dfn.js
```diff
@@ -121,30 +139,15 @@ function collectDfns(title) {
...
 
-/**
- * @param {HTMLElement} dfn
- * @param {string} title
- */
-function assignDfnId(dfn, title) {
-  if (!dfn.id) {
-    const { dfnFor } = dfn.dataset;
-    if (dfn.dataset.idl) {
-      addId(dfn, "dom", (dfnFor ? `${dfnFor}-` : "") + title);
```
saschanaz: I assume this line never ran, probably because `decorateDfn()` ran first, right?

src/core/dfn-finder.js
marcoscaceres: I seem to recall getting mismatches.

src/core/utils.js
marcoscaceres: Just readability. `linkFor` is already initialized to the empty string, so might as well use it.

tests/spec/core/pluralize-spec.js
marcoscaceres: Not as part of this module. It's set by `dfn.js` module IIRC.

src/core/link-to-dfn.js
marcoscaceres: correct. At least, I was never able to trigger it.

src/core/dfn.js
```diff
@@ -9,13 +9,9 @@ export const name = "core/dfn";
...
-    /** @type {HTMLElement} */
-    const closestDfn = dfn.closest("[data-dfn-for]");
-    if (closestDfn && closestDfn !== dfn && !dfn.dataset.dfnFor) {
-      dfn.dataset.dfnFor = closestDfn.dataset.dfnFor;
-    }
-    // TODO: we should probably use weakmaps and weaksets here to avoid leaks.
-    const titles = getDfnTitles(dfn, { isDefinition: true });
+    const titles = getDfnTitles(dfn);
     registerDefinition(dfn, titles);
+    if (titles.length) dfn.dataset.lt = titles.join("|");
```
saschanaz: Found that this unexpectedly add `data-lt` everywhere, because `titles` nearly always contains an item here unless `data-lt-noDefault`.

marcoscaceres: I think that's ok. The `lt`'s now include the normalized text content, which matches what BikeShed does, I think.

saschanaz: Not sure I can find an example that bikeshed greedily adds `data-lt`, Inspecting Web IDL spec says it doesn't.

marcoscaceres: I was looking at fetch, for instance:
```
<dfn class="dfn-paneled" data-dfn-type="dfn" data-export data-lt="collect an HTTP quoted string|collecting an HTTP quoted string" id="collect-an-http-quoted-string">collect an HTTP quoted string</dfn>
```

saschanaz: That's 👍 and it needs `data-lt` to do that, but the ReSpec behavior here is:

```html
<dfn>hey</dfn>
<!-- becomes... -->
<dfn data-lt="hey">hey</dfn>
```

Which is weird.

marcoscaceres: Ah, maybe it only does it for `data-export` things?

marcoscaceres: The question is, is there harm in ReSpec always adding `data-lt`? As an editor, I only ever want to write the alternative linking names, but not the text content. So:

```
<dfn data-lt="yellow fruit">banana</dfn>
<!-- becomes... -->
<dfn data-lt="yellow fruit|banana">banana</dfn>
``` 

Would be ok, no? Or would that be harmful somehow?

saschanaz: I think that's okay to do (regardless of what bikeshed do for that) and it's needed for Shepherd, but in the `hey` case Shepherd won't need it.

marcoscaceres: Ok, I'll omit it to avoid the duplication.

saschanaz:
So this included/includes several great things but seems to be lacking tests for the case insensitivity things.

marcoscaceres:
Forgot how untested the original code was. Will add a few tomorrow.

marcoscaceres:
Thanks for your help @saschanaz! You are the best!

PR #2491, opened 2019-08-27
refactor(core/dfn-finder): move addAlternativeNames to decorateDfn

saschanaz:
The second wave of refactors in #2489.


saschanaz:
This was just to be a refactoring... and turns out to fix a bug that IDL self-definitions lack proper dfn registration.

Issue #394, opened 2015-01-23
Link constructors in WebIDL


jyasskin:
Forked from https://github.com/w3c/respec/pull/373#discussion_r22060777. Link the `[Constructor]` extended attribute to `<dfn title="InterfaceName" for="InterfaceName">`, and give that the id `InterfaceName-InterfaceName`. I'd guess that `[NamedConstructor=Name]` should go to `<dfn title="Name">` and have the id `dom-Name`, but @tabatkins may point out that [Bikeshed](https://github.com/tabatkins/bikeshed/blob/master/docs/idl.md) does something different.

shubhshrma:
@marcoscaceres what are starting steps with this in ```webidl.js```? I'm trying to include this in my gsoc tasks list.

marcoscaceres:
@shubhshrma, basically, when we see a "constructor", it needs to create an entry in `conf.definitionMap` that can be linked to as: "InterfaceName.IntefaceFunction()", as described in https://github.com/w3c/respec/issues/394#issuecomment-71237207  

You probably don't need to go into this level of detail in you proposal, just describe generally what needs to be fixed overall for WebIDL.

marcoscaceres:
Blocked on https://github.com/heycam/webidl/issues/636 

We might be getting new syntax for this.

PR #2499, opened 2019-09-02
feat(core/webidl): support constructor linking

### LINKED ISSUE #2494: Support new constructor IDL syntax
We need to make sure we properly link and `code` constructors in a way that is backwards compatible with the old `Constructor=` extended attribute syntax. In particular `data-lt="InterfaceName.InterfaceName()"` should work, as well as `{{Foo/constructor()}}`.

saschanaz:
Closes https://github.com/w3c/respec/issues/394 
Closes #2494

### DIFF
--- a/examples/basic.html
+++ b/examples/basic.html
@@ -58,14 +58,18 @@
         <dfn>Foo</dfn> interface
       </h2>
       <pre class="idl">
-      [Exposed=Window, Constructor]
+      [Exposed=Window]
       interface Foo {
+        constructor();
         attribute DOMString bar;
         void doTheFoo();
       };
       </pre>
       <p>
         The <a>Foo</a> interface is nice. Lets you do stuff.
+      </p>
+      <p>
+        The <dfn>Foo.constructor()</dfn> creates a Foo instance.
       </p>
       <p>
         The <dfn>bar</dfn> attribute, returns 🍺.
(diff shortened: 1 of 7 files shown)


examples/basic.html
```diff
@@ -58,15 +58,19 @@ <h2>
...
+        constructor();
         attribute DOMString bar;
         void doTheFoo();
       };
       </pre>
       <p>
         The <a>Foo</a> interface is nice. Lets you do stuff.
       </p>
+      <p>
+        The <dfn>Foo.constructor()</dfn> creates a Foo instance.
```
marcoscaceres: We still want to be able to define it as `<dfn>constructor()</dfn>` like a regular operation.

saschanaz: That should work too. Will add a relevant test.

tests/spec/core/webidl-spec.js
```diff
@@ -301,6 +301,34 @@ describe("Core - WebIDL", () => {
...
+        </pre>
+        <dfn>SuperStar.constructor()</dfn>
+        <dfn>HyperStar.constructor()</dfn>
+      </section>
+    `;
+    const ops = makeStandardOps(null, body);
+    const doc = await makeRSDoc(ops);
+    const pre = doc.querySelector("pre");
+    expect(
+      pre.querySelector("a[href=\\#dom-superstar-constructor]")
```
marcoscaceres: what's with the "\\#dom"?

saschanaz: That's just a common prefix we have for IDL links.

Issue #2498, opened 2019-09-02
Support aliasing with [=linking syntax=]


marcoscaceres:
Sometimes it's awkward to use some exported terms in prose, so it's useful to be able to alias them. For example:

```HTML
Return the result of [=complex task|performing the complex task=] once p is settled, with the fullfilment handler which takes an argument |x|:
```

Where that would become:
```HTML
... <a data-lt="complex task">performing the complex task</a>...
```

PR #2504, opened 2019-09-03
feat(core/inlines): Support aliasing with [=linking syntax=]

sidvishnoi:
Closes https://github.com/w3c/respec/issues/2498


marcoscaceres:
Any chance you can update the docs (if not already done)?

sidvishnoi:
We don't have anything for inline syntax in wiki yet, right?

sidvishnoi:
Lets create a page for linking? (and we can extend it with xref linking)

Issue #2496, opened 2019-09-02
Reference errors flagged by respec


dontcallmedom:
## Important info

* URL to affected spec: https://w3c.github.io/webrtc-pc/webrtc.html
* ReSpec version: 24.30.5
* [x] I did a "hard refresh", but it's still busted. 

## Description of problem

What happened (e.g., it crashed)?: 
When trying to upgrade the WebRTC spec above (whose online version is currently tied to a static version of ReSpec 24.3.0), it triggers 30 new errors bound to interfaces or names it does not recognize, even though these names are indeed defined in normatively referenced specifications.

For instance, it complains:
> Couldn't match "EventTarget" to anything in the document or in any other document cited in this specification: bundle, ice, jsep, sdp, sdp-simulcast, trickle-ice, webidl. See how to cite to resolve the error at: 1.

I'm not sure why it picks that random subset of references; we do normatively reference DOM where EventTarget is defined. Similar error messages get generated for a bunch of other interfaces.

More bizarrely, a similar error message is generated for "Constructor":
> Couldn't match "Constructor" to anything in the document or in any other document cited in this specification: bundle, ice, jsep, sdp, sdp-simulcast, trickle-ice, webidl. See how to cite to resolve the error

The How-to-cite links don't really help; I noticed that some well-known names (e.g. VoidFunction) don't appear in the auto-complete box there though.

But what I was expecting?:
No errors for interfaces that are normatively referenced; no error for using the Constructor extended attribute.

marcoscaceres:
> I'm not sure why it picks that random subset of references; we do normatively reference DOM where EventTarget is defined. Similar error messages get generated for a bunch of other interfaces.

References are contextual (i.e., it looks in the DOM tree)... so, to fix, in theory, you need to add  `data-cite="DOM"` where the `<pre class="IDL">` is defined. 

> More bizarrely, a similar error message is generated for "Constructor": 

As of like 3  days ago, Constructor is no longer a thing in WebIDL so it's getting confused. We are working to add support for `constructor()` right now :)  

Will jump over to the WebRTC pull request and give you a hand soon!

dontcallmedom:
ah, thanks for the `data-cite` trick, that helps solve most of the reference issues.

Besides the `Constructor` error, respec still doesn't recognize the following names:
* `AlgorithmIdentifier` defined in Web Crytpo
* `VoidFunction` defined in WebIDL

Neither of those appear in the autocompletion box in https://respec.org/xref/

[...]

PR #2505, opened 2019-09-03
fix(core/pluralize): normalize to lowercase

sidvishnoi:
Regression from https://github.com/w3c/respec/pull/2484, https://github.com/w3c/respec/pull/2489. Should fix https://github.com/w3c/respec/issues/2496#issuecomment-527560437
Makes sense as one won't be pluralizing IDL and concept-dfn can be lowercased.

[Fixes `json-ld-syntax`](https://respec-preview.netlify.com/?spec=https://w3c.github.io/json-ld-syntax/&version=https://deploy-preview-2505--respec-pr.netlify.com/respec-w3c-common.js). Will try to confirm if any exports break.

### DIFF
--- a/src/core/pluralize.js
+++ b/src/core/pluralize.js
@@ -46,7 +46,7 @@ function getPluralizer() {
   /** @type {NodeListOf<HTMLAnchorElement>} */
   const reflessAnchors = document.querySelectorAll("a:not([href])");
   reflessAnchors.forEach(el => {
-    const normText = normalize(el.textContent);
+    const normText = normalize(el.textContent).toLowerCase();
     links.add(normText);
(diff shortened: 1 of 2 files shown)


sidvishnoi:
Tests failed with timeout. Maybe we should increase `jasmine.DEFAULT_TIMEOUT_INTERVAL` to 6 seconds?

sidvishnoi:
> It’s the first time it’s happened

Has [happened before](https://github.com/w3c/respec/runs/202354022#step:5:347) - since we started using GitHub Actions.

marcoscaceres:
Ah, ok, let’s boost to 6 seconds

PR #2508, opened 2019-09-05
feat: allow concepts to be associated via "data-dfn-for"

marcoscaceres:
This would allow defining and linking both concepts and IDL things: 
```
      <pre class="idl">
      [Exposed=Window]
      interface Foo {
        attribute DOMString bar;
      };
      </pre>
      <p>
        The <dfn data-dfn-for="Foo" data-dfn-type="dfn">bar</dfn> concept.
        The <dfn>bar</dfn> attribute, returns 🍺.
      </p>
      <p>
        Links to the concept of [=Foo/bar=].
      </p>
      <p>
        Links to the attribute {{Foo/bar}}.
      </p>
```


src/core/inline-idl-parser.js
```diff
@@ -169,6 +169,7 @@ function renderAttribute(details) {
   const { parent, identifier, renderParent } = details;
   const { identifier: linkFor } = parent || {};
   const html = hyperHTML`${renderParent ? "." : ""}<a
+      data-link-type="idl"
```
saschanaz: Is this documented anywhere?

marcoscaceres: https://tabatkins.github.io/bikeshed/#link-types

marcoscaceres: @sidvishnoi, this could get more specific, per BikeShed/Shepherd. Like IDL-name or whatever.

sidvishnoi: There are lots of them: https://tabatkins.github.io/bikeshed/#dfn-types (Links have the same types as definitions)
I'm not sure if being this explicit will be helpful for? It's good in core/inline-idl-parser though, as we can remove `data-xref-type`

sidvishnoi:
Closing in favour of https://github.com/w3c/respec/pull/2854

src/core/inline-idl-parser.js
sidvishnoi: Will work on this part after #2854

PR #2515, opened 2019-09-15
fix(core/webidl): add !overload to idl-def ids

### LINKED ISSUE #2513: Duplicate ids in markup for overloaded constructors
## Important info

* URL to affected spec: https://w3c.github.io/mediacapture-main/getusermedia.html in https://github.com/w3c/mediacapture-main/pull/622
* ReSpec version: 24.32.1

## Description of problem

The `MediaStream` has 3 overloaded constructors; each of these constructors get assigned the id `idl-def-mediastream-constructor` in its IDL wrapping markup, which makes the document no longer pubrules compliant.

But what I was expecting?:
A different id for each constructor.

saschanaz:
This "fixes" the duplicated id bug, but I think we need to overhaul the overload things, because `!overload` is weird.

Currently operations that do not have overloads get ids like `idl-def-parentname-operationname-argname1-argname2` but this also potentially causes duplicated id problem where argument names can be same.

Closes #2513


marcoscaceres:
> This "fixes" the duplicated id bug, but I think we need to overhaul the overload things, because !overload is weird.

Strongly agree... I think we should look c++ for inspiration, but with slightly less ugly IDs. We can also look at aligning with Bikeshed on the IDs. We might need to deprecate the old IDs, but find a way to keep existing IDs for older specs.

PR #2516, opened 2019-09-15
feat(core/webidl): suggest IDL fix from autofix

saschanaz:
This shows autosuggestion like below:

image

image


saschanaz:
Didn't just apply the autofix intentionally because some people didn't like adding `[Exposed=Window]` and instead wanted to consider exposing to more contexts.

PR #2519, opened 2019-09-17
presentation: slides

marcoscaceres:


examples/tpac_2019.html
```diff
@@ -0,0 +1,424 @@
...
+  - We've had 37 new contributors project.
+  - Supports numerous students - GSoC, Outreachy
+
+
+
+## Now in two flavours
+ReSpec now comes into flavors:
+
+  - 🍏 respec-w3c.js
+    - 327Kb - (93Kb) gzip
```
saschanaz: Hmm, if we ship ReSpec with brotli we could even get < 80 KiB of size. Firefox supports it since version 44, maybe time to support the new algorithm. https://tools.paulcalvano.com/compression.php

image

marcoscaceres: I thought this would be automatically done on the W3C server (once they add support)? We could ask for support if it’s easy to add. I’m not sure what they use.

PR #2521, opened 2019-09-26
fix(assets/respec2): make links white in .simple table

### LINKED ISSUE #2522: links in the registry table headers lack contrast
The use of links in the [registry's table header](https://www.w3.org/TR/timing-entrytypes-registry/#registry) make them very hard to read (low contrats). 

screenshot of the header, with low contrast in a couple columns

It’s probably worth fixing in the spec’s stylesheet?

marcoscaceres:
closes #2522

### DIFF
--- a/assets/respec2.css
+++ b/assets/respec2.css
@@ -134,6 +134,12 @@ table.simple {
   text-align: left;
 }
 
+.simple th a:link {
+  color: #fff;
+  padding: 3px 5px;
+  text-align: left;
+}
+
 .simple th[scope="row"] {
   background: inherit;
   color: inherit;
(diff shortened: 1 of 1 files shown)


assets/respec2.css
```diff
@@ -134,6 +134,12 @@ table.simple {
   text-align: left;
 }
 
+.simple th a:link {
```
saschanaz: Should we remove `:link`? Because otherwise clicking links will restore them as blue.

marcoscaceres: Ah true. I should check :visited too I guess.

marcoscaceres:
did a rebase.. sorry for the `-f` push @saschanaz 🙏

PR #2543, opened 2019-10-12
chore: use @ts-check in all src files

### LINKED ISSUE #1905: Use @ts-check wherever appropriate

NotWoods:
Fixes #1905

Finishes off the last few files without typechecking 😄


src/geonovum/style.js
```diff
@@ -12,6 +13,7 @@ function attachFixupScript(doc, version) {
     "load",
     () => {
       if (window.location.hash) {
+        // @ts-ignore
         window.location = window.location.href;
```
sidvishnoi: I guess it would be fine if we do the following?
```suggestion
         window.location.href = window.location.href;
```

src/type-helper.d.ts
```diff
@@ -18,12 +18,41 @@ declare module "text!*" {
...
+    show(): void;
+    hide(): void;
+    enable(): void;
+    addCommand(
+      label: string,
+      handler: () => void,
+      keyShort?: string,
+      icon?: string
+    ): any;
+    error(mg: string): void;
```
sidvishnoi: Nit: 
```suggestion
    error(msg: string): void;
```

src/geonovum/style.js
NotWoods: Yeah, but then the linter complains...

saschanaz:
Merging because LGTM. Thanks! 🎉

PR #2553, opened 2019-10-27
refactor: migrate to pure es modules

### LINKED ISSUE #1975: Migrate to webpack/rollup and ditch requirejs

saschanaz:
Finally some part of #2187 😄

Closes #1975


marcoscaceres:
Very exciting to see this!

saschanaz:
The work is done, but maybe this is too big to review... 👀

src/core/data-type.js
```diff
@@ -5,17 +5,27 @@
...
  */
-import tooltipStyle from "text!../../assets/datatype.css";
+import { fetchAsset } from "./text-loader.js";
 
 export const name = "core/data-type";
 
-export function run(conf) {
+const tooltipStylePromise = loadStyle();
+
+async function loadStyle() {
```
sidvishnoi: I see this being used multiple times. Can we move this to utils?
``` js
async function loadStyle(filename) {
  try {
    return (await import(`text!../../assets/${filename}`)).default;
  } catch {
    return fetchAsset(filename);
  }
}
// loadStyle("datatype.css");
```

src/core/xref-db.js
```diff
@@ -12,13 +12,13 @@ import { importIdb } from "./idb.js";
 const VERSION_CHECK_WAIT = 5 * 60 * 60 * 1000; // 5 min
 
 async function getIdbCache() {
-  const { openDB } = await importIdb();
-  const idb = await openDB("xref", 1, {
+  const { openDB } = idb;
+  const db = await openDB("xref", 1, {
```
sidvishnoi: Nit:
```suggestion
  const db = await idb.openDB("xref", 1, {
```

src/core/data-type.js
saschanaz: No, dynamic import with dynamic value will prevent bundler from merging assets into one file, because they are not smart enough 😂

sidvishnoi: (╯°□°）╯︵ ┻━┻

saschanaz: I hope there is a better way... but this is my current best.

sidvishnoi: Why do we need them as dynamic import? Can't we do static import and then inject style?

saschanaz: Static import will break native es module system because importing raw string is not in the spec. 🤷‍♀️

sidvishnoi: Can we use something like webpack-dev-server for dev so that we can use ESM, but bundle them in production?

saschanaz: I'm not sure how would that work?

sidvishnoi: I've a potentially really bad idea. Convert CSS files to JS during builddeps or similar step.
``` js
export default `#i-am-css {}`
```

sidvishnoi: Fine with the above `loadStyle()` way also though.

saschanaz: Haven't thought that but should be doable. Interesting 👀

saschanaz: Some objecting opinions:

* Additional code to generate css-js is not quite less than the current method
* CSS generation requires additional build method after modification, while the current method allows you to just refresh on the browser to get the latest bits (if you use ES module version to test)

sidvishnoi: Let's stick with the present way then :+1:

src/core/import-maps.js
```diff
@@ -0,0 +1,22 @@
+// @ts-check
+// Temporary workaround until browsers get real import-maps
```
marcoscaceres: So much sadness... but ok :)

marcoscaceres: What's going to be the long term strategy so we don't need this?

src/core/webidl-clipboard.js
```diff
@@ -6,16 +6,30 @@
...
 export const name = "core/webidl-clipboard";
 
-// This button serves a prototype that we clone as needed.
-const copyButton = document.createElement("button");
-copyButton.innerHTML = svgClipboard;
-copyButton.title = "Copy IDL to clipboard";
-copyButton.classList.add("respec-button-copy-paste", "removeOnSave");
+const copyButtonPromise = createButton();
+
+async function loadSVG() {
```
marcoscaceres: I wonder if we should generalize this into a util function... `loadTextAsset()` or something. I imagine we eventually want to kill these methods once we address whatever is requiring them.

src/core/webidl.js
```diff
@@ -336,7 +335,17 @@ function renderWebIDL(idlElement, index) {
   return parse;
 }
 
-export function run() {
+const cssPromise = loadStyle();
+
+async function loadStyle() {
```
marcoscaceres: as above, `loadTextAsset()`?

src/core/worker.js
```diff
@@ -20,9 +20,22 @@ const hint = {
 const link = createResourceHint(hint);
 document.head.appendChild(link);
 
-const workerURL = URL.createObjectURL(
-  new Blob([workerScript], { type: "application/javascript" })
-);
-export const worker = new Worker(workerURL);
+async function loadWorkerScript() {
```
marcoscaceres: And same here... `loadTextAsset(asset, path)` or something.

marcoscaceres:
Ok, looks like no dev server :( but that might be ok. https://github.com/rollup/rollup/issues/2385#issuecomment-499869919 

We might be able to work around it somehow. There are a few options.

src/core/import-maps.js
saschanaz: Official import maps implementation can replace this with traditional `import x as "module"`.

src/core/webidl-clipboard.js
saschanaz: That's essentially https://github.com/w3c/respec/pull/2553#discussion_r339940372, and we can't remove it as I replied there:

>No, dynamic import with dynamic value will prevent bundler from merging assets into one file, because they are not smart enough 😂

saschanaz:
> Ok, looks like no dev server :( but that might be ok. [rollup/rollup#2385 (comment)](https://github.com/rollup/rollup/issues/2385#issuecomment-499869919)
> 
> We might be able to work around it somehow. There are a few options.

Have we used webpack dev server? 👀

marcoscaceres:
> Have we used webpack dev server? 👀

Not in this project, no (we use `"serve"`). But use it a lot on other projects and it's wonderful.

package.json
```diff
@@ -62,34 +58,33 @@
...
     "karma": "karma start --single-run",
     "lint": "tsc -p src/jsconfig.json && eslint .",
-    "prepare": "npm run snyk-protect && npm run build:components",
+    "prepare": "npm run snyk-protect && npm run builddeps",
     "release": "node ./tools/release.js",
     "server": "npm start",
     "snyk-protect": "snyk protect",
     "snyk": "snyk",
-    "start": "npm run build:components && serve",
+    "start": "serve",
```
marcoscaceres: This is where the "dev server" would come in... basically, what we want here is for rollup to be watching files and automatically running babel for us.

marcoscaceres: oh, and also automatically building the "development" bundle.

marcoscaceres: https://rollupjs.org/guide/en/#rollupwatch

saschanaz: We don't need babel anymore here, so you just want to refresh your browser and that's all. (Only if you are using non-built version)

saschanaz:
>what's the magic to get that to happen? I'd love to quickly try this out.

Everything should be same as before, you build the bundle and insert breakpoints 👍

marcoscaceres:
Ok, no babel is awesome - so that’s fantastic. But the rebuilding step is no good as it makes testing in Karma super cumbersome. That’s the last critical thing I’d really like to address. It seems it’s doable with the watch thing above.

saschanaz:
> Ok, no babel is awesome - so that’s fantastic. But the rebuilding step is no good as it makes testing in Karma super cumbersome. That’s the last critical thing I’d really like to address. It seems it’s doable with the watch thing above.

Not sure what's the rebuilding step here. If you mean `builddeps`, that's just a replacement of previous `copydeps`.

sidvishnoi:
Slightly concerned about the bundle size increase. 

``` bash
curl https://www.w3.org/Tools/respec/respec-w3c | wc -c                      
# 328923
$ curl https://deploy-preview-2553--respec-pr.netlify.com/respec-w3c.js | wc -c
# 389695
# +60722 bytes added
```

With gzip:
``` bash
$ curl -H 'Accept-encoding: gzip' https://www.w3.org/Tools/respec/respec-w3c | wc -c
# 94909
$ curl -H 'Accept-encoding: gzip' https://deploy-preview-2553--respec-pr.netlify.com/respec-w3c.js | wc -c
# 121864
# +26955 bytes added

saschanaz:
>Slightly concerned about the bundle size increase.

Oops, the jquery commit did something wrong. It actually decreases without it.

sidvishnoi:
Happy with the 28043 byte reduction in size (3481 bytes in gzip) :tada: :shipit:

sidvishnoi:
@saschanaz We still need to `build:w3c` before running karma, right?

saschanaz:
> @saschanaz We still need to `build:w3c` before running karma, right?

Right.

marcoscaceres:
Right, so ideally we should never need to run "build:w3c" or the other builders... the watcher should be doing that automatically for us.

saschanaz:
That sound great, and then it sounds like an extra feature. Maybe it can be implemented after this PR?

marcoscaceres:
Yes, definitely could be an extra feature (and new pull request)... but what need need to confirm is that it's possible (sounds like it is, given the watcher thing) - otherwise, we will need to later switch to React.

marcoscaceres:
Lol, not react! webpack.

saschanaz:
Seems there are also third-party rollup dev server libraries, so I think it should be possible to implement.

saschanaz:
Migration to React will be even more difficult than to nanohtml 😂

PR #2556, opened 2019-10-28
fix(tools/copydeps): gracely ignore when no folder

saschanaz:
`errno` is different in Windows, so it's more safe to use `code` instead.


saschanaz:
`"ENOENT"` is same on both, but somehow the error number is 4xxx on Windows.

Issue #2546, opened 2019-10-16
Top level IDL things should have data-dfn-for=""


marcoscaceres:
If no data-dfn-for and top level WebIDL thing (e.g, dictionary, interface, etc), we should add data-dfn-for="" to be clear it's not for anything.

sidvishnoi:
Is this the only thing that needs to be done? :thinking: 
https://github.com/w3c/respec/blob/c4ffceaf59a78279df5538ccf6788acfd3bebe49/src/core/dfn-finder.js#L144

marcoscaceres:
But it depends on parent... it's makes me sad that the "finder" has side effects, but whatever.

sidvishnoi:
cc: @saschanaz

saschanaz:
>we should add data-dfn-for="" to be clear it's not for anything.

Reasonable. I wonder whether it caused some trouble?

>it's makes me sad that the "finder" has side effects

`decorateDfn()` is in dfn-finder but has been split from "finder" functions (in other words, nothing in dfn-finder calls `decorateDfn()` anymore).

sidvishnoi:
>  I wonder whether it caused some trouble?

Somehow, Shepherd (hence bikeshed-data, hence xref) is treating top level things like: https://respec.org/xref/?term=Performance&specs=hr-time&types=interface
Saying `Performance` is defined "for Performance" and is not top level by itself.

saschanaz:
Oops, that's problematic indeed!

sidvishnoi:
@saschanaz I'm not sure whether https://github.com/w3c/respec/issues/2546#issuecomment-543021791 would be enough or it'll break stuff.

PR #2558, opened 2019-10-29
fix(core/dfn-finder): explicit empty dfn-for on top level things

sidvishnoi:
Fixes https://github.com/w3c/respec/issues/2546


marcoscaceres:
I think we want to test a couple of top-level things... might be quick just to dig into the IDL tests and find a:

 * dictionary
 * enum
 * interface 
 
And make sure that attribute and methods are still `for` things.

Issue #1724, opened 2018-06-29
Add support for “changes since”


marcoscaceres:
It would be nice to show substantive since some tag (e.g., “CR”).

sidvishnoi:
Thoughts:
``` js
var respecConfig = {
  changesSince: {
    tag: "CR", // required
    filter: globalFilterFunctionName, // if not provided (not recommeded), use default: show all
    url: BASE_API_URL_FOR_TESTS, // internal
  }
};
// const el = $("#changes-since");
// if el, appends ul>li* to it
// else: error
```

What should be the signature for filter function?
``` ts
type Filter = (commits: Commit[]) => Commit[];
// gives more power and more responsibility to user
// more power: access to next/prev commits if that makes sense

type Filter = (commit: Commit) => boolean;
// easier and simpler
```

marcoscaceres:
> tag: "CR",
can we use a commit as well?

sidvishnoi:
> can we use a commit as well?

yes. any commit-ish.

PR #2567, opened 2019-11-11
feat(core/rs-changelog): list changes between two commits

sidvishnoi:
Wiki: https://github.com/w3c/respec/wiki/rs-changelog
Closes https://github.com/w3c/respec/issues/1724

---

Need additional feedback on:
1. Module name and config parameter names.
2. Should we add it to geonovum profile too?
3. Should `defaultFilter` do something more?


marcoscaceres:
> * Module name and config parameter names.
Will respond inline.. 

> * Should we add it to geonovum profile too?

Nah. 

> Should defaultFilter do something more?

No... it's fine.

src/core/changes-since.js
```diff
@@ -0,0 +1,82 @@
...
+ */
+import { pub } from "./pubsubhub.js";
+
+export const name = "core/changes-since";
+
+export async function run(conf) {
+  const { changesSince } = conf;
+  if (!changesSince) return; // nothing to do
+
+  const el = document.querySelector("ul#changes-since");
```
marcoscaceres: I kinda like this, because it sidesteps the l10n issues and us having to add a heading.... but at the same time, it's different to most of the other ones, which rely on `<section id=>`... I guess in a way this is better because it gives more flexibility.

js/profile-w3c-common.js
```diff
@@ -59,6 +59,7 @@ const modules = [
   import("../src/core/data-type.js"),
   import("../src/core/algorithms.js"),
   import("../src/core/anchor-expander.js"),
+  import("../src/core/changes-since.js"),
```
marcoscaceres: Incentive to move to new profile 🥕?
```suggestion
```

src/core/changes-since.js
marcoscaceres: OMG OMG OMG!  wait!!!!! what if we don't have a configuration option at all! But instead:

```HTML 
<ul id='changes-since' data-from="commit" data-to="tag"></ul>
```


That would be awesome! So, imagine:

```
### Changes between FPWD and CR
<ul id='changes-since' data-from="FPWD" data-to="CR"></ul>

### Changes since FPWD
<ul id='changes-since' data-from="cava2fdasdfawsd3" data-to="FPWD"></ul>
```

🤩🤩🤩🤩🤩🤩

marcoscaceres: Made some additional changes above.

marcoscaceres: One more... what if we make this out first custom element? 

```HTML
<rs-changelog from="" to="" filter=""></rs-changelog>
```

js/profile-w3c-common.js
marcoscaceres: Actually, let's not be mean :) Let's keep the old profile updated too.

marcoscaceres: If we go with custom element, we should probably have:

```
../src/core/custom-elements/rs-whatever.js 
```

src/core/changes-since.js
```diff
@@ -0,0 +1,82 @@
...
+
+/**
+ * @param {Commit[]} commits
+ * @param {(Commit) => boolean} filter
+ * @param {HTMLElement} element
+ */
+function showCommits(commits, filter, element) {
+  const commitsToShow = commits.filter(filter);
+  const nodes = commitsToShow.map(commit => {
+    const el = document.createElement("li");
```
marcoscaceres: We probably want this to be a link to the actual commit on github. And we could be clever and do what github does with the (#number): 

Screenshot 2019-11-12 15 09 43

tests/data/commits
```diff
@@ -0,0 +1,22 @@
...
+    "hash": "a0c831d",
+    "message": "Remove requirement to reject after document is inactive (#875)"
+  },
+  { "hash": "c77ebf8", "message": "chore: tidy" },
+  {
+    "hash": "6f2565b",
+    "message": "Add PaymentRequest.prototype.hasEnrolledInstrument() (#833)"
+  },
+  { "hash": "72de3ba", "message": "chore: enable MDN annotations" },
+  { "hash": "4d8ed23", "message": "refactor: use | instead of <var> (#863)" }
```
marcoscaceres: We probably want to make sure we test that `<var>` doesn't get treated as HTML... I can see that it won't, but we might want to explicitly check for that regardless.

src/core/changes-since.js
sidvishnoi: I did consider this approach. Wasn't sure as the filter function would need to be registered globally, whereas with respecConfig, it can be a block level or even inline. Might be a non issue.

sidvishnoi: Also, the `to` field is presently not supported on respec.org. Do we want to support it?

marcoscaceres: > I did consider this approach.

I don't doubt it! 

> Might be a non issue.

I think it's a non-issue. 

> Also, the to field is presently not supported on respec.org. Do we want to support it?

It be great if we could... doing a range would be amazing.

src/core/changes-since.js
sidvishnoi: will need to break into two `<a>`.. all cool

sidvishnoi:
The centralization part (`custom-elements/index.js`) can be a lot better. Could not figure out how to make use of a promise `ready` there, so tried with event listener.
Also, we need to defined custom element within that module (in `run()`) as `conf` might be needed during class definition (as in case of `rs-changelog`).
This one is a simple implementation, can improve if this (centralization) is the way forward.

src/core/custom-elements/rs-changelog.js
```diff
@@ -0,0 +1,95 @@
...
+ * the user, it is used to filter the commits that are to be shown. Otherwise,
+ * all commits are shown.
+ */
+import { done } from "./index.js";
+import { pub } from "../pubsubhub.js";
+
+export const name = "rs-changelog";
+
+export function run(conf) {
+  class Changelog extends HTMLElement {
```
sidvishnoi: Cried Illegal constructor with `HTMLUListElement`

src/core/custom-elements/rs-changelog.js
```diff
@@ -0,0 +1,95 @@
...
+
+/**
+ * @param {string} from commit-ish
+ * @param {string} to commit-ish
+ * @param {string} apiURL
+ * @typedef {{ hash: string, message: string }} Commit
+ */
+async function getCommits(from, to, apiURL) {
+  const url = new URL("commits", apiURL);
+  url.searchParams.set("from", from);
```
sidvishnoi: Typo: presently the API accepts "since"

src/core/custom-elements/rs-changelog.js
```diff
@@ -0,0 +1,95 @@
...
+        filter:
+          typeof window[this.getAttribute("filter")] === "function"
+            ? window[this.getAttribute("filter")]
+            : defaultFilter,
+        /** @type {Commit[]} */
+        commits: [],
+      };
+    }
+
+    connectedCallback() {
```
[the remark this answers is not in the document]
sidvishnoi: Yes. I came to my mind before sleeping. Will do

src/core/github.js
```diff
@@ -50,19 +58,23 @@ export async function run(conf) {
...
-    pub("error", `\`respecConf.github\` is not a valid URL? (${ghURL})`);
+    const msg = `\`respecConf.github\` is not a valid URL? (${ghURL})`;
+    pub("error", msg);
+    apiURLRejector(new Error(msg));
     return;
   }
   if (ghURL.origin !== "https://github.com") {
     const msg = `\`respecConf.github\` must be HTTPS and pointing to GitHub. (${ghURL})`;
     pub("error", msg);
+    apiURLRejector(new Error(msg));
```
sidvishnoi: This doesn't seem to be a scalable approach :disappointed:

tests/data/commits
sidvishnoi: Updated data and test to take that into account.

src/core/custom-elements/rs-changelog.js
sidvishnoi: Will send PR to fix on respec.org

src/core/custom-elements/rs-changelog.js
```diff
@@ -0,0 +1,94 @@
...
+    this.firstChild.remove();
+
+    if (!commits) return;
+
+    const { repoURL } = await github;
+    const nodes = commits.map(commit => {
+      const [message, prNumber = null] = commit.message.split(/\(#(\d+)\)/, 2);
+      const commitURL = `${repoURL}commit/${commit.hash}`;
+      const prURL = prNumber ? `${repoURL}pull/${prNumber}` : null;
+      const pr = prNumber && hyperHTML` <a href="${prURL}">(#${prNumber})</a>`;
```
sidvishnoi: maybe?
```suggestion
      const pr = prNumber && hyperHTML` (<a href="${prURL}">#${prNumber}</a>)`;
```

src/core/github.js
marcoscaceres: Can you help me understand your thinking? I was thinking that in the dependent module this prevents the module from gracefully proceeding.

src/core/custom-elements/rs-changelog.js
marcoscaceres: 👍

src/core/custom-elements/rs-changelog.js
```diff
@@ -0,0 +1,94 @@
...
+import { pub } from "../pubsubhub.js";
+
+let readyResolver;
+const readyPromise = new Promise(resolve => {
+  readyResolver = resolve;
+});
+
+export const name = "rs-changelog";
+
+export default class Changelog extends HTMLElement {
```
marcoscaceres: Maybe: 
```suggestion
export default class ChangelogElement extends HTMLElement {
```

src/core/custom-elements/rs-changelog.js
```diff
@@ -0,0 +1,94 @@
...
+      to: this.getAttribute("to") || "HEAD",
+      /** @type {(commit: Commit) => boolean} */
+      filter:
+        typeof window[this.getAttribute("filter")] === "function"
+          ? window[this.getAttribute("filter")]
+          : () => true,
+    };
+  }
+
+  get ready() {
```
marcoscaceres: Should this be static?

sidvishnoi: Static won't be able to tell ready for each instance of component. Will be okay if we want each component to be used only once.

src/core/github.js
sidvishnoi: By salability, I meant: if we want to make more things available outside, we need to export them - each having its own promise and having a reject call - which becomes cumbersome.
For this one, I decided to export an object, instead of adding a new promise for each thing.

src/core/custom-elements/rs-changelog.js
sidvishnoi: Gave it another thought during nap. We're using the same promise, so it's essentially bugged. 🐛

marcoscaceres: Why I wondered if this should be static :)

marcoscaceres: We could just make each element have its own `.ready` (we should really call this `.done`) property. Then we can make `render()` private and we should be able to query for the elements, then `await` for each to be `.done`.

src/core/custom-elements/index.js
```diff
@@ -0,0 +1,64 @@
...
+
+/** @param {HTMLElement} el */
+export function ready(el) {
+  el.dispatchEvent(new CustomEvent("ready"));
+}
+
+/** @param {HTMLElement} el */
+function toReadyPromise(el) {
+  return new Promise(resolve => {
+    const timoutId = setTimeout(failure, 3000);
```
sidvishnoi: the simplest case is `el.addEventListener("ready", resolve, { once: true })`, but it doesn't handle if we forgot to call `ready(this)` (like due to some error occurred before that)

src/core/custom-elements/index.js
```diff
@@ -0,0 +1,66 @@
...
+    el.addEventListener("ready", success, { once: true });
+    function success() {
+      clearTimeout(timeoutId);
+      resolve(true);
+    }
+    function failure() {
+      el.removeEventListener("ready", success);
+      const name = el.matches("[is]") ? el.getAttribute("is") : el.localName;
+      pub("error", `Plugin \`core/custom-elements/${name}\` took too long.`);
+      resolve(false);
```
marcoscaceres: maybe this should reject? This seems pretty exceptional.

src/core/custom-elements/index.js
```diff
@@ -0,0 +1,66 @@
...
+ * In your custom element definition, call `ready(this)`.
+ */
+export function ready(el) {
+  el.dispatchEvent(new CustomEvent("ready"));
+}
+
+/** @param {HTMLElement} el */
+function toReadyPromise(el) {
+  return new Promise(resolve => {
+    const timeoutId = setTimeout(failure, 3000);
```
marcoscaceres: Hmm... do we _really_ need this?

src/core/custom-elements/rs-changelog.js
```diff
@@ -0,0 +1,83 @@
...
+      throw new Error(
+        `Request to ${url} failed with status code ${res.status}`
+      );
+    }
+    /** @type {Commit[]} */
+    const commits = await res.json();
+    if (!commits.length) {
+      throw new Error(`No commits between ${from}..${to}.`);
+    }
+    return commits.filter(filter);
```
marcoscaceres: Just a style nit... I wonder if the `return` can happen at the end of the function instead?..

sidvishnoi: Can do. I just personally don't like `let` variables outside try-catch block

src/core/custom-elements/rs-changelog.js
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  connectedCallback() {
+    const { from, to, filter } = this.props;
+    hyperHTML.bind(this)`
+      <ul>
+      ${{
+        any: fetchCommits(from, to, filter)
+          .then(commits => toHTML(commits))
+          .catch(error => pub("error", error.message))
```
marcoscaceres: oh, we have an element... so we should probably use `showInlineError()` from utils

sidvishnoi: Good point!

sidvishnoi: the element is empty though.. so, we don't get error highlighting :D
can append an error message to element maybe

src/core/custom-elements/rs-changelog.js
```diff
@@ -0,0 +1,85 @@
...
+          : () => true,
+    };
+  }
+
+  connectedCallback() {
+    const { from, to, filter } = this.props;
+    hyperHTML.bind(this)`
+      <ul>
+      ${{
+        any: fetchCommits(from, to, filter)
```
marcoscaceres: Nit: this is still a bit of an eyesore for me... could we move it to a function?

marcoscaceres: actually, never mind... it uses a lot of `this`.

marcoscaceres:
As this includes a significant architectural addition  (custom-elements/index.js) @sidvishnoi and I think @saschanaz should also take look... I'm really happy with the design that Sid came up with, but I'd like a second opinion and to make sure Kagami also likes it.

saschanaz:
Will review this, and GITHUB DROP ICE

saschanaz:
@sidvishnoi I still don't know how to use the netlify preview, could you give me some help?

src/core/github.js
```diff
@@ -42,6 +50,7 @@ export async function run(conf) {
       "Config option `[github](https://github.com/w3c/respec/wiki/github)` " +
       "is missing property `repoURL`.";
     pub("error", msg);
+    githubRejector(new Error(msg));
```
saschanaz: Bit weird that a function name is a noun 👀

And could we merge `githubRejector` and `pub("error")` here?

src/core/custom-elements/rs-changelog.js
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * @typedef {{message: string, hash: string}} Commit
+ */
+import { github } from "../github.js";
+import { hyperHTML } from "../import-maps.js";
+import { showInlineError } from "../utils.js";
+
+export const name = "rs-changelog";
+
+export default class ChangelogElement extends HTMLElement {
```
saschanaz: Bit weird to use `default` when we don't really do `import element from "./rs-changelog.js"`. How about `export const element = class ...`?

src/core/custom-elements/index.js
```diff
@@ -0,0 +1,41 @@
...
+
+export const name = "core/custom-elements";
+
+export async function run() {
+  // prepare and register elements
+  CUSTOM_ELEMENTS.forEach(el => {
+    customElements.define(
+      el.name,
+      el.default,
+      el.is ? { extends: el.is } : undefined
```
saschanaz: There is no `is` now, do you have an immediate plan to implement one? If not how about deferring this until we actually need it?

src/core/custom-elements/rs-changelog.js
```diff
@@ -0,0 +1,85 @@
...
+  constructor() {
+    super();
+    this.props = {
+      from: this.getAttribute("from"),
+      to: this.getAttribute("to") || "HEAD",
+      /** @type {(commit: Commit) => boolean} */
+      filter:
+        typeof window[this.getAttribute("filter")] === "function"
+          ? window[this.getAttribute("filter")]
```
saschanaz: This is where I feel weirdest... Could we make this attribute as regex? I think a regex should be sufficient for this purpose.

src/core/custom-elements/index.js
sidvishnoi: No yet, we don't. Lets defer this until needed.

src/core/custom-elements/rs-changelog.js
sidvishnoi: We're using `import *` and then `.default`. Agreed that naming it `element` is better.

src/core/custom-elements/rs-changelog.js
sidvishnoi: We decided to not go with regex as a function will be simpler (and less prone to mistakes) and more flexible (can use something more `commit.message` if needed).

src/core/github.js
sidvishnoi: > Bit weird that a function name is a noun :eyes:

`rejectGithubPromise`?

> And could we merge githubRejector and pub("error") here?

Good idea!

sidvishnoi:
>  I still don't know how to use the netlify preview, could you give me some help?

@saschanaz You can get link to a deployment of ./build by clicking on the "netlify/respec-pr/deploy-preview" check's details (like https://deploy-preview-2567--respec-pr.netlify.com/).
Then you can use use that script locally or through https://respec-preview.netlify.com/ (respec-preview doesn't make sense in this PR unless you've a document that is using this new stuff). I'll try to make the PR workflow simpler in respec-preview.

src/core/custom-elements/rs-changelog.js
sidvishnoi: Might be a good idea to support both?

saschanaz: What I feel weird is that this allows calling arbitrary global function, which is a bit scary... No proof that this can be a problem, though. Just me being scared.

saschanaz:
> @saschanaz You can get link to a deployment of ./build by clicking on the "netlify/respec-pr/deploy-preview" check's details (like https://deploy-preview-2567--respec-pr.netlify.com/).
> Then you can use use that script locally or through https://respec-preview.netlify.com/ (respec-preview doesn't make sense in this PR unless you've a document that is using this new stuff). I'll try to make the PR workflow simpler in respec-preview.

Thanks, I'm sure you taught me once and I forgot 😂😂😂

src/core/github.js
```diff
@@ -41,7 +53,7 @@ export async function run(conf) {
     const msg =
       "Config option `[github](https://github.com/w3c/respec/wiki/github)` " +
       "is missing property `repoURL`.";
-    pub("error", msg);
+    rejectGithubPromise(new Error(msg));
```
saschanaz: Might be cleaner to just pass `msg` and create `Error` inside the function.

src/core/custom-elements/rs-changelog.js
sidvishnoi: It's not arbitrary global function per se, author will mention which function will be called.

src/core/github.js
sidvishnoi: Good idea.

src/core/custom-elements/rs-changelog.js
saschanaz: Hmmm, given that other frameworks also do some similar thing, I would say this is okay. At least we don't use `eval()` 😂

sidvishnoi: How about something like: :thinking: 
``` html
<rs-changelog from="tag">
  <script type="module">
    export default function filter(){}
  </script>
</rs-changelog>
```

Issue #2568, opened 2019-11-11
Citing academic papers: Crossref integration


wetneb:
ReSpec has a fantastic system to cite other specs, built on top of the [SpecRef](https://www.specref.org/) database. However it is geared towards specs only: there does not seem to be a simple way to cite academic papers since the data model for the references is not adapted to them. Given ReSpec's scope, I understand the restriction to specs, but I think citing academic papers could still be useful in non-normative sections, to motivate choices behind a spec.

[Crossref](https://search.crossref.org/) is a database of about 100 million papers, identified by their Digital Object Identifier (DOI). It provides [an HTTP API](https://api.crossref.org/) which can be used to search for papers and retrieve their metadata, just like SpecRef does. Therefore, it should be possible to use it just like SpecRef. For instance, I would like to be able to write something like this:

    … has been extensively studied [[doi:10.1007/978-3-319-93417-4_5]]. However, …

which would fetch citation metadata from Crossref and insert it in the bibliography.
It would generate a citation and link to the corresponding paper at https://doi.org/10.1007/978-3-319-93417-4_5.

Would you be interested in this feature? If you think this is a good idea I could provide a pull request for that.

marcoscaceres:
@wetneb, we'd totally take a PR for this! That sounds amazing. We probably want to discuss architecturally how to integrate this, but it should be fairly straight forward. 

As the "doi" scheme is somewhat different to the bibref scheme, you probably want to add a matching rule here:
https://github.com/w3c/respec/blob/develop/src/core/inlines.js#L30

And then check for `startsWith([["doi:"):
https://github.com/w3c/respec/blob/develop/src/core/inlines.js#L246

And then add a function that writes out the reference, similar to:
https://github.com/w3c/respec/blob/develop/src/core/inlines.js#L96

You can then validate the "doi:" entry there before fetching it. 

And maybe also support "expansions", so [[[doi:10.1007/978-3-319-93417-4_5]]] expands out to the title of an article (or whatever is appropriate):
https://github.com/w3c/respec/blob/develop/src/core/inlines.js#L31
 
You can probably also hook into "biblio.js", "biblio-db.js", and "render-biblio.js":
https://github.com/w3c/respec/blob/develop/src/core/biblio.js
https://github.com/w3c/respec/blob/develop/src/core/biblio-db.js
https://github.com/w3c/respec/blob/develop/src/core/render-biblio.js

(or write equivalents to the above - specifically for doi citations)

[...]

PR #2569, opened 2019-11-15
Crossref integration proof of concept

wetneb:
This demonstrates the approach I had in mind for #2568. I understand this is not the architecture you prefer, I am just opening this PR for @tobie to get an idea of what rendering code can look like (since it seems that the rendering format must be agreed on first).

As far as I am concerned this fits the bill, but if you guys want to have that upstream, it would be great of course!


wetneb:
Closing as this is just a proof of concept whose architecture does not follow the one sketched in #2568.

wetneb:
Reopening to solve linting issues (so that @tobie can read the diffs).

An example of the feature in action can be found at:
https://reconciliation-api.github.io/specs/#bib-beek-2018

PR #2572, opened 2019-11-18
chore(package): update deps

marcoscaceres:


saschanaz:
It's failing 👀

marcoscaceres:
@saschanaz seems to be crawling node_modules?

marcoscaceres:
The typescript checker, that is. That do we need to exclude that in the typescript config file?

saschanaz:
The new version of TS seems buggy 👀

PR #2577, opened 2019-11-20
docs(w3c/headers): w3c-software-doc is the default

### LINKED ISSUE #2575: Is the "w3c" license supported?
## Important info

* URL to affected spec: [w3c-license-bug.html.gz](https://github.com/w3c/respec/files/3864459/w3c-license-bug.html.gz)
* ReSpec version: 24.34.1

## Description of problem

What happened (e.g., it crashed)?: when setting the `license` to `"w3c"` and enabling JSON-LD generation via `doJsonLd: true`, no JSON-LD is generated.

But what I was expecting?:

JSON-LD to be generated.

## Source of the problem

Although the wiki claims that a `"w3c"` license is supported, there does not seem to be a corresponding object in `src/w3c/header.js`, so the JSON-LD code is unable to extract the corresponding license info.

Discrepancy between
https://github.com/w3c/respec/blob/25ef75e9b0c87646a871c64c700207f58e877d68/src/w3c/headers.js#L84-L92
and
https://github.com/w3c/respec/blob/25ef75e9b0c87646a871c64c700207f58e877d68/src/w3c/headers.js#L185-L207

The "w3c" license identifier is supposed to be valid according to #1604.

marcoscaceres:
closes #2575

### DIFF
--- a/src/w3c/headers.js
+++ b/src/w3c/headers.js
@@ -82,13 +82,12 @@
 //          - href: a URL for the value (e.g., "https://foo.com/issues"). Optional.
 //          - class: a string representing CSS classes. Optional.
 //  - license: can be one of the following
-//      - "w3c", currently the default (restrictive) license
 //      - "cc-by", which is experimentally available in some groups (but likely to be phased out).
 //          Note that this is a dual licensing regime.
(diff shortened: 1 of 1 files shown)


marcoscaceres:
updated https://github.com/w3c/respec/wiki/license also.

PR #2578, opened 2019-11-20
fix(core/webidl): escape IDL code for fix suggestions

### LINKED ISSUE #2563: Suggested fix for WebIDL should write out entities
Right now, the suggested fix for IDL in the ReSpec Pill are being treated as HTML... meaning that `<Whatever>` in  `Promise<Whatever>` (or any generic) are being treated as a HTML tag. 

Actual:
```
[Exposed=(DedicatedWorker,Window), SecureContext]
interface SerialPort {
  Promise open(optional SerialOptions options = {});
};
```

Expected:

```
[Exposed=(DedicatedWorker,Window), SecureContext]
interface SerialPort {
  Promise&lt;void&gt; open(optional SerialOptions options = {});
};
```

saschanaz:
Closes #2563

Can't test error messages yet, so just a fix.

### DIFF
--- a/src/core/webidl.js
+++ b/src/core/webidl.js
@@ -5,7 +5,12 @@
 //  - It could be useful to report parsed IDL items as events
 //  - don't use generated content in the CSS!
 import { decorateDfn, findDfn } from "./dfn-finder.js";
-import { flatten, showInlineError, showInlineWarning } from "./utils.js";
+import {
+  flatten,
+  showInlineError,
+  showInlineWarning,
+  xmlEscape,
+} from "./utils.js";
 import { hyperHTML, webidl2 } from "./import-maps.js";
 import { fetchAsset } from "./text-loader.js";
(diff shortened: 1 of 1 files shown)


src/core/webidl.js
```diff
@@ -365,8 +370,10 @@ export async function run() {
     let details = `<pre>${validation.context}</pre>`;
     if (validation.autofix) {
       validation.autofix();
+      const idlToFix = webidl2.write(astArray[validation.sourceName]);
+      const escaped = xmlEscape(idlToFix);
```
marcoscaceres: finally we get to use xmlEscape 🥳

marcoscaceres: XML IS NOT DEAD!

saschanaz:
>Should we add a test?

Can't test an error message for now, no? We really should make a way to do this.

marcoscaceres:
Oops, missed the PR message. All good :)

PR #2579, opened 2019-11-20
fix: put code inside anchors for IDL expansions

marcoscaceres:


src/core/link-to-dfn.js
```diff
@@ -212,7 +212,10 @@ function wrapAsCode(ant, dfn) {
   // only add code to IDL when the definition matches
   const term = ant.textContent.trim();
   const isIDL = dfn.dataset.hasOwnProperty("idl");
-  const needsCode = shouldWrapByCode(dfn, term);
+  const needsCode =
+    ant.firstElementChild && ant.firstElementChild.localName === "code"
```
sidvishnoi: handle this case in `shouldWrapByCode`?

tests/spec/core/xref-spec.js
```diff
@@ -675,23 +676,29 @@ describe("Core — xref", () => {
...
+      expect(link1b.firstElementChild.localName).toBe("code");
       const vars1 = [...doc.querySelectorAll("#link1 var")];
       expect(vars1.length).toBe(2);
       expect(vars1[0].textContent).toBe("type");
       expect(vars1[1].textContent).toBe("callback");
 
-      const [link2a, link2b] = [...doc.querySelectorAll("#link2 code a")];
+      const [link2a, link2b] = [...doc.querySelectorAll("#link2 a")];
       expect(link2a.href).toBe(expectedLinks.get("ChildNode"));
+      expect(link2a.firstElementChild.localName).toBe("code");
```
sidvishnoi: so these tests (localName ones) are all kinda same.. maybe do them once per type?

marcoscaceres:
> Not sure why we need it though. Some recommendation from tr-design?

The styles/markup is inconsistent with the local ones, so it makes them somewhat confusing... this makes them all the same, but with the ability for us to style them a bit differently as needed.

tests/spec/core/xref-spec.js
marcoscaceres: I wanted to keep the original structure we were testing, which structurally was "#link code a" for all of them (now just swapping a and code). I know it's a little redundant, but I think it's ok. More chance to catch any regressions.

PR #2585, opened 2019-11-24
feat(w3c/seo): Improve JSON-LD export of citations

### LINKED ISSUE #2584: Improve citation export in JSON-LD output
Currently, the citations are only exported in JSON-LD in a pretty minimal way: only the title, url and specref id are exposed:

```
{
      "id": "https://tools.ietf.org/html/rfc8174",
      "type": "TechArticle",
      "name": "Ambiguity of Uppercase vs Lowercase in RFC 2119 Key Words",
      "url": "https://tools.ietf.org/html/rfc8174"
}
```
This could be improved by adding other fields such as authors, publishers, publication date and so on.

wetneb:
This exports a few more fields from Specref's data model to the JSON-LD export of the bibliography.

Closes #2584.

### DIFF
--- a/src/w3c/seo.js
+++ b/src/w3c/seo.js
@@ -170,10 +170,24 @@ function addPerson({ name, url, mailto, company, companyURL }) {
  */
 function addRef(ref) {
   const { href: id, title: name, href: url } = ref;
-  return {
+  let jsonld = {
     id,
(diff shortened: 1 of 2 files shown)


marcoscaceres:
Thanks @gkellogg for checking this also.

PR #2587, opened 2019-11-25
fix(w3c/seo): Resolve aliases for JSON-LD bibliography export

### LINKED ISSUE #2586: Citation export in JSON-LD does not honour aliases
Suppose I have a bibliographic record which is an alias of another:

```
localBiblio: {
  testAlias: { aliasOf: 'testRecord' },
  testRecord: { id: 'myTest', title: 'Pride and Prejudice' },
}
```

Then if I cite `[[testAlias]]` in a spec, it will be exported in JSON-LD without any metadata:

``{
   type: 'TechReport'
}``

Whereas it should contain metadata from the record:
``{
   type: 'TechReport',
   name: 'Pride and Prejudice'
}``

wetneb:
Fixes #2586, by using the existing `resolveRef` function to get reference metadata, instead of retrieving them directly from the bibliographical store.


wetneb:
I'm not sure how to solve the failing tests, I'm struggling to debug them as the tested code is run in an iframe.

@marcoscaceres, is there anything obvious to you just by looking at my changes?

sidvishnoi:
This looks like a race condition to me. Every time, different tests are failing. Let me confirm.

sidvishnoi:
Yep. Race condition. Fixed.

@wetneb can you look into the failing assertion?

---

To debug: 
``` bash
# run following each time ./src is changed
npm run build:w3c -- --debug
npm run karma -- start --browsers ChromeHeadless
```

Add `debugger` statements in code or tests and visit http://localhost:9876/debug.html and you can access browser console there.

wetneb:
Thank you so much @sidvishnoi! That's brilliant. I will have a look at the last failing test.

wetneb:
So I had to make sure `core/seo` is run before the JSON-LD export in `w3c/seo` since the latter relies on the abstract being added as a `<meta>` tag. Not being familiar with the details of respec's synchronicity I am not sure if this is the right way to solve this - perhaps `w3c/seo` could instead call the same extraction method itself directly? Or could I have avoided any changes in synchronicity with a synchronous variant of `resolveRef`?

src/core/seo.js
```diff
@@ -6,16 +6,14 @@
...
-  // This is not critical, so let's continue other processing first
-  (async () => {
-    await document.respecIsReady;
-    const firstParagraph = document.querySelector("#abstract p:first-of-type");
-    if (!firstParagraph) {
-      return; // no abstract, so nothing to do
-    }
-    insertMetaDescription(firstParagraph);
-  })();
+export async function run() {
```
marcoscaceres:
```suggestion
export function run() {
```

src/w3c/seo.js
```diff
@@ -135,8 +134,14 @@ async function addJSONLDInfo(conf, doc) {
...
 
   // normative and informative references
-  jsonld.citation = [...conf.normativeReferences, ...conf.informativeReferences]
-    .map(ref => biblio[ref])
+  const citationIds = [
+    ...conf.normativeReferences,
+    ...conf.informativeReferences,
+  ];
+  const citationContents = await Promise.all(
+    citationIds.map(async ref => await resolveRef(ref))
```
sidvishnoi:
```suggestion
    citationIds.map(ref => resolveRef(ref))
```

src/core/seo.js
```diff
@@ -6,16 +6,14 @@
...
+  // This needs to run before JSON-LD export runs, because it relies on
```
marcoscaceres: We probably don't need these comments (as we treat the modules independently)... this can just happen sync, with no mention of JSON-LD.

src/core/seo.js
```diff
@@ -7,15 +7,13 @@
...
-    }
-    insertMetaDescription(firstParagraph);
-  })();
+  // This needs to run before JSON-LD export runs, because it relies on
+  // the meta tag being present to extract the description itself.
+  const firstParagraph = document.querySelector("#abstract p:first-of-type");
+  if (!firstParagraph) {
+    return; // no abstract, so nothing to do
+  }
+  insertMetaDescription(firstParagraph);
```
marcoscaceres: insertMetaDescription can probably be inlined now.

src/core/biblio.js
```diff
@@ -107,14 +107,15 @@ export async function run(conf) {
   conf.biblio = biblio;
   const localAliases = Array.from(Object.keys(conf.localBiblio))
     .filter(key => conf.localBiblio[key].hasOwnProperty("aliasOf"))
-    .map(key => conf.localBiblio[key].aliasOf);
+    .map(key => conf.localBiblio[key].aliasOf)
+    .filter(key => !conf.localBiblio.hasOwnProperty(key));
```
marcoscaceres: I'm unsure about this change... can you help me understand it a bit?

wetneb: These localAliases are added later on to the list of references to fetch from the network / db, since they could refer to external references. The filter I added ensures we only fetch those, and not the local aliases of local references. For tests where the entire bibliography is local, this avoids lookups.

Alternatively, we could also keep `localAliases` as before but swap the order of the `concat` and `filter` in the computation of `neededRefs`.

marcoscaceres: Thanks for the clarification. I’m ok to go with this.

marcoscaceres:
@sidvishnoi, what say you? Look ok?

marcoscaceres:
Will release tomorrow! Thanks for working on this @wetneb!

Issue #2559, opened 2019-10-30
Use rollup dev server or watcher


marcoscaceres:
having to build the profiles when testing with karma is extremely painful, because `<reasons>`. It would be nice if rollup's watcher just did that for us... then it would work like webpack, where it just rebuilds on save, leading to a much nicer development experience.

marcoscaceres:
Talked to @sidvishnoi about this... I'm currently using `watch` and Sid is using `nodemon`. But what's annoying is that you still need to spin up multiple terminal windows:
 
1. for Karma
1. for watching files
1. for `npm run`
1. for doing actual work

As `watch` (and `nodemon`) and `serve` (our dev http server) all provide an API, we should be able to build a very light weight server that at least watches the `src` folder and regenerates the  debug builds. 

A bonus would be to also spin up Karma at the same time and just route the output to the console. That way, we can work with basic.html AND have karma running on the side.... that's the dream set up... and only two terminal tabs needed!

PR #2588, opened 2019-11-25
chore(dev): improve dev mode

sidvishnoi:
Lets see how this works out :smile: 
image

- Starts serve so one can play with examples/basic.html
- Starts karma server so when can go to localhost:9876/debug.html and have fun there.
- Starts a file watcher, which on change: creates a debug build and runs karma tests

Closes https://github.com/w3c/respec/issues/2559
cc: @marcoscaceres @saschanaz 

TODO:
- ~add command line args~
- ~fix that welcome message banner (it's very hacky, spent unnecessarily long time on it :sweat_smile: )... maybe we can use [boxen](https://www.npmjs.com/package/boxen) if we want to be fancy here~
- ~Maybe not always build and run tests (sometimes we just want to prototype with basic.html). Pass a CLI flag to do so.~ Added interactive mode
- ~run particular tests~ (supports --grep) or only those whose source changed? (maybe some other PR)


package.json
```diff
@@ -26,6 +26,7 @@
     "@types/pluralize": "0.0.29",
     "babel-eslint": "^10.0.3",
     "chai": "^4.2.0",
+    "chokidar": "^3.3.0",
```
marcoscaceres: Interested to hear why you chose this one over the other ones?

tools/dev.js
```diff
@@ -0,0 +1,87 @@
...
+        box.vert,
+        colors.white.bold(`${msg[0]}:`),
+        colors.white(msg[1]),
+        box.vert,
+      ].join("\t");
+    }),
+    ` ${box.bottomLeft}${box.hr.repeat(hrWidth)}${box.bottomRight}`,
+  ];
+  console.log(colors.bgBlack.green(lines.map(line => `\t${line}`).join("\n")));
+  console.log();
```
marcoscaceres: Nit: would `console.log(colors.bgBlack.green(lines.map(line => `\t${line}`).join("\n")), "");` do the same? Then we don't need this empty one?

tools/dev.js
```diff
@@ -0,0 +1,87 @@
...
+    console.error(colors.error(err.stack));
+  }
+}
+
+function onError(err) {
+  console.error(colors.error(err.stack));
+  process.exit(1);
+}
+
+function printWelcomeMessage() {
```
marcoscaceres: 😁, lol slightly overkill... but ok.

tools/dev.js
```diff
@@ -0,0 +1,87 @@
...
+  {
+    browsers: ["ChromeHeadless"],
+    autoWatch: false,
+    port: KARMA_PORT,
+    logLevel: karma.constants.LOG_WARN,
+  }
+);
+const karmaServer = new karma.Server(karmaConfig);
+
+main();
```
marcoscaceres: Main looks kinda redundant, tbh... maybe we should just ditch main() and just inline all this?

tools/dev.js
```diff
@@ -0,0 +1,87 @@
...
+const serveConfig = require("../serve.json");
+const { Builder } = require("./builder");
+
+const KARMA_PORT = 9876;
+const SERVE_PORT = 5000;
+
+const karmaConfig = karma.config.parseConfig(
+  path.join(__dirname, "../karma.conf.js"),
+  {
+    browsers: ["ChromeHeadless"],
```
saschanaz: But...but...👀

tools/dev.js
```diff
@@ -0,0 +1,87 @@
...
+  watcher.on("all", onFileChange);
+  watcher.on("error", onError);
+
+  printWelcomeMessage();
+}
+
+async function onFileChange(_event, _file) {
+  try {
+    await Builder.build({ name: "w3c", debug: true });
+    karma.runner.run({ port: KARMA_PORT, autoWatch: true }, () => {});
```
saschanaz: Can we cancel the previous run when file changes happen but the tests are not finished yet?

sidvishnoi: Couldn't figure that one out. Will try again

tools/dev.js
sidvishnoi: This will be cli arg. Using this during prototyping as it doesn't open a new window

package.json
sidvishnoi: Nodemon: API wasn't nice
Watch: too raw
Chokidar: nice API, have used previously, claims to be a wrapper that fixes node's fs events
(Will write a detailed one when I am on laptop)

tools/dev.js
saschanaz: Yes, but...but... why not Firefox! 😂

tools/dev.js
sidvishnoi: that empty one is for new line though, so will add a `"\n"` in end instead

saschanaz: And Marcos' code will insert a new line too!

tools/dev.js
sidvishnoi: So the karma API allows to stop the server, but doesn't allow stopping a run in between. One way could be to stop the server so the run fails and restart the server again.

saschanaz: That sounds reasonable enough.

sidvishnoi: Didn't work :smile: 
Want to give it a try? https://karma-runner.github.io/4.0/dev/public-api.html

tools/dev.js
sidvishnoi: first: with `"\n"`, second: with `""`
image
(talking about an empty line after box)

saschanaz: Ooops you're right, somehow `console.log("abc\n")` and `console.log("abc")` is equivalent, the last explicit newline is ignored. 🤯

tools/dev.js
sidvishnoi: karmaServer is in EventEmitter :tada: 
``` js
karmaServer.emit("run_complete", [], {});
```

sidvishnoi:
Added command line args to:
- build specific profile (`--profile PROFILE`)
- specify karma browsers (`--browsers FirefoxHeadless ChromeHeadless`)
- run in interactive mode (disables file watching, press `t` to build and run tests) (`--interactive`)

The interactive mode might not work on Windows (TTY needed) @saschanaz can you play with it?

saschanaz:
> The interactive mode might not work on Windows (TTY needed) @saschanaz can you play with it?

Works great 👍

wetneb:
This looks amazing! Speaking of which, [the wiki currently recommends](https://github.com/w3c/respec/wiki/Developers-Guide) `npm run babel:watch` which does not seem to work:
```
npm ERR! missing script: babel:watch
```

saschanaz:
> This looks amazing! Speaking of which, [the wiki currently recommends](https://github.com/w3c/respec/wiki/Developers-Guide) `npm run babel:watch` which does not seem to work:

Oops, you don't need babel anymore, as we removed babel dependency and migrated to native ES module 👍

wetneb:
> Oops, you don't need babel anymore, as we removed babel dependency and migrated to native ES module

OK, I'm not sure what to replace it with (or delete?) so I'm not editing it myself. I think there are a couple of other references to running `babel:` things in the rest of the wiki / repo which might be worth cleaning up too.

tools/dev-server.js
```diff
@@ -0,0 +1,158 @@
...
+  stdin.on("data", async key => {
+    switch (key) {
+      case "\u0003":
+      case "q": {
+        return karma.stopper.stop(karmaConfig, code => {
+          setTimeout(() => process.exit(code), 0);
+        });
+      }
+      case "t":
+        return await buildAndTest();
```
saschanaz: Seems that typing `t` multiple times causes building multiple time in parallel 👀

tools/dev-server.js
```diff
@@ -0,0 +1,158 @@
...
+      case "h":
+        return printWelcomeMessage(args);
+      default:
+        process.stdout.write(key);
+    }
+  });
+}
+
+async function onFileChange(_event, _file) {
+  await buildAndTest();
```
saschanaz: Could the karma.stopper also be used here?

tools/dev-server.js
sidvishnoi: Added an active flag to let only single build run at a time. Try again?

tools/dev-server.js
sidvishnoi: Not sure how. karma.stopper stops the server altogether, and based on my attempts, it fails to restart :confused:

PR #2590, opened 2019-11-26
chore(tools/dev-server): run tests on test file change

sidvishnoi:
Tests were not running if we edited test file :man_facepalming: 
Also, allow running tests without building.


tools/dev-server.js
```diff
@@ -115,15 +118,18 @@ function registerStdinHandler() {
...
 
-async function onFileChange(_event, _file) {
-  await buildAndTest();
+async function onFileChange(_event, file) {
+  const preventBuild = file.startsWith("tests");
+  await buildAndTest(preventBuild);
 }
 
-async function buildAndTest() {
+async function buildAndTest(preventBuild = false) {
```
marcoscaceres: boolean trap :) let's use `{preventBuild = false}`

sidvishnoi: /me agrees

Issue #1217, opened 2017-05-17
markdown: Support backticks for webidl


marcoscaceres:
Would be nice be able to use triple backticks to denote a WebIDL pre.

OrionStar25:
@marcoscaceres I'd certainly like to work on this! But can you please elaborate a bit more? We need `''' <some WEbIDL pre> ''' ` to show finally onto the doc, right?

marcoscaceres:
@OrionStar25, that would be correct. However, I'm trying to find you something easier. This one has too many dependencies, so might be too challenging if you are just starting out.

chaals:
Seems like however marked.js picks up things marked with <pre>```</pre> doesn't get the language identifier for extra languages added in #1835, either

PR #2591, opened 2019-11-26
enhancement(core/markdown): support backticks for webidl

sidvishnoi:
Closes https://github.com/w3c/respec/issues/1217


tests/spec/core/markdown-spec.js
```diff
@@ -241,6 +241,45 @@ describe("Core - Markdown", () => {
...
+        attribute DOMString bar;
+        void doTheFoo();
+      };
+      \`\`\`
+
+      \`\`\` js
+      console.log("hey")
+      \`\`\`
+
+      \`\`\
```
marcoscaceres: Should this be? 
```suggestion
      \`\`\`
```

sidvishnoi: Indeed. I wonder how tests passed

src/core/markdown.js
```diff
@@ -86,6 +86,14 @@ const inlineElems = new Set([
   "var",
 ]);
 
+const renderer = new marked.Renderer();
+renderer.code = function(code, language, isEscaped) {
+  if (language === "webidl") {
+    return `<pre class="idl">${code}</pre>`;
+  }
+  return marked.Renderer.prototype.code.call(this, code, language, isEscaped);
```
marcoscaceres: 🤮 ... I wonder if there is a better way of doing this? Wonder if something like this would work:

```JS
class IDLRender extends marked.Renderer {
   code(code, language, isEscaped) {
     super.code(code, language, isEscaped);
   }
}
```

I'm not sure...

tests/spec/core/markdown-spec.js
marcoscaceres: Computers!

src/core/markdown.js
sidvishnoi: Worked :D

Issue #1596, opened 2018-03-29
Confirm that tests exist in WPT


marcoscaceres:
Using the [github api for repo contents]( https://developer.github.com/v3/repos/contents/) we should confirm that `data-tests` files exist.

sidvishnoi:
In this repo: https://github.com/w3c/web-platform-tests/, right? 
Or the repo and branch will be available in `conf`?

Can/Should I take this?

marcoscaceres:
we should plan out the details and see what useful things we get back from the API. Main problem we want to solve is the 404 case, where a spec points to a missing test in `data-tests`.

sidvishnoi:
I looked into the Github Trees API. But it gives complete repo tree (not inside some path) which means the response is unnecessarily large.

sidvishnoi:
With https://bit.ly/wpt-transition, I think the Github trees API might be a good solution.
Should we wait for transition before adding this feature or look into the Content API (if we need it earlier)?

marcoscaceres:
I don't see any harm in us investigating a bit. We could document here first how it would be done before we start coding anything.

sidvishnoi:
Got a local version (running on payment-request):
![image](https://user-images.githubusercontent.com/8426945/69664979-6d761d80-10af-11ea-9702-fe4a50b8cab0.png)

Linter API is not flexible enough to provide informative warnings. It cannot tell which test is missing on which `[data-tests]`. Linter API can only tell: "these elements have failed this lint rule" (though nothing stops us `showInlineWarning` withing linter rule). I think it makes sense to add this to `core/data-tests`. What do you think?

PR #2593, opened 2019-11-26
feat(linter-rules/tests-exist): confirm that tests exist in WPT

sidvishnoi:
Closes https://github.com/w3c/respec/issues/1596


src/core/linter-rules/tests-exist.js
```diff
@@ -0,0 +1,106 @@
...
+      const re = /web-platform-tests\/wpt\/(.+)/;
+      wptDirectory = testSuiteURL.pathname.match(re)[1].replace(/\//g, "");
+    } else {
+      wptDirectory = testSuiteURL.pathname.replace(/\//g, "");
+    }
+  } catch {
+    const msg = "Failed to parse WPT directory from testSuiteURI";
+    pub("warn", msg);
+    return null;
+  }
```
sidvishnoi: This is a bit problematic. Can add WPT directory to lint config as `lint: { "tests-exist": wptDirectory }`?

src/core/defaults.js
```diff
@@ -11,6 +11,7 @@ import { rule as localRefsExist } from "./linter-rules/local-refs-exist.js";
 import { rule as noHeadinglessSectionsRule } from "./linter-rules/no-headingless-sections.js";
 import { rule as noHttpPropsRule } from "./linter-rules/no-http-props.js";
 import { rule as privsecSection } from "./linter-rules/privsec-section.js";
+import { rule as testsExist } from "./linter-rules/tests-exist.js";
```
sidvishnoi: This should probably only go in `w3c/defaults.js`

src/core/linter-rules/tests-exist.js
```diff
@@ -0,0 +1,106 @@
...
+  let wptDirectory;
+  try {
+    const testSuiteURL = new URL(testSuiteURI);
+    if (testSuiteURL.pathname.startsWith("/web-platform-tests/wpt/")) {
+      const re = /web-platform-tests\/wpt\/(.+)/;
+      wptDirectory = testSuiteURL.pathname.match(re)[1].replace(/\//g, "");
+    } else {
+      wptDirectory = testSuiteURL.pathname.replace(/\//g, "");
+    }
+  } catch {
```
marcoscaceres: I think we should still console.error out here... it could be a bad URL or something else in the future. This would just be mostly for us.

src/core/linter-rules/tests-exist.js
marcoscaceres: Let’s start with this and see how we go.

src/core/linter-rules/tests-exist.js
sidvishnoi: Done.

sidvishnoi:
1. Can we do better with linter rule name?
2. Should we enable it by default or after some beta testing?

PR #2594, opened 2019-11-27
refactor(core/github): provide a more flexible apiBase

sidvishnoi:
Basically lets customize repository for GitHub API.
Can access web-platform-tests/wpt easily now, which caused trouble with https://github.com/w3c/respec/pull/2593


src/core/github.js
```diff
@@ -112,7 +112,7 @@ export async function run(conf) {
     ],
   };
   // Assign new properties, but retain existing ones
-  let githubAPI = `https://respec.org/github/${org}/${repo}/`;
+  let githubAPI = `https://respec.org/github/`;
```
sidvishnoi: This is the most important bit of this PR.

marcoscaceres:
```suggestion
  let githubAPI = "https://respec.org/github/";
```

PR #2595, opened 2019-11-27
Convert all raw javascript strings to hyperHTML templated string

### LINKED ISSUE #2551: Remove raw HTML string interpolation

SuyashSalampuria:
Fixes #2551 
Found raw HTML at several places in the codebase. I converted all of them to hyperHTML templated string everywhere I found. Many places had hyperHTML from before.


src/core/issues-notes.js
```diff
@@ -245,7 +245,7 @@ function makeIssueSectionSummary(issueList) {
   ) {
     issueSummaryElement.insertAdjacentHTML(
       "afterbegin",
-      `<h2>${l10n.issue_summary}</h2>`
+      hyperHTML `<h2>${l10n.issue_summary}</h2>`
```
saschanaz: This should be `` element.prepend(hyperHTML`...`) ``, because the current insertAdjacentHTML receives a string.

src/core/mdn-annotation.js
```diff
@@ -86,8 +86,8 @@ function buildBrowserSupportTable(support) {
   let innerHTML = "";
   function addMDNBrowserRow(browserId, yesNoUnknown, version) {
     const displayStatus = yesNoUnknown === "Unknown" ? "?" : yesNoUnknown;
-    const classList = `${browserId} ${yesNoUnknown.toLowerCase()}`;
-    const browserRow = `
+    const classList = hyperHTML `${browserId} ${yesNoUnknown.toLowerCase()}`;
```
saschanaz: This specific line is okay to be string, because it will be escaped by the next line.

src/core/mdn-annotation.js
```diff
@@ -86,8 +86,8 @@ function buildBrowserSupportTable(support) {
...
+    const browserRow = hyperHTML `
```
saschanaz: This is good but but need more changes because `innerHTML += browserRow` will be broken. I suggest replacing `innerHTML` with an array and push to it instead.

src/core/webidl.js
```diff
@@ -367,12 +367,12 @@ export async function run() {
...
   const validations = webidl2.validate(astArray);
   for (const validation of validations) {
-    let details = `<pre>${validation.context}</pre>`;
+    let details = hyperHTML `<pre>${validation.context}</pre>`;
     if (validation.autofix) {
       validation.autofix();
       const idlToFix = webidl2.write(astArray[validation.sourceName]);
       const escaped = xmlEscape(idlToFix);
-      details += `Try fixing as:
+      details += hyperHTML `Try fixing as:
```
saschanaz: `+=` won't work, this should be `` details.append(hyperHTML`...`) ``.

src/geonovum/conformance.js
```diff
@@ -3,13 +3,14 @@
...
 import { hyperHTML as html } from "../core/import-maps.js";
+import hyperHTML from "hyperhtml/esm";
 export const name = "geonovum/conformance";
 
 /**
  * @param {Element} conformance
  */
 function processConformance(conformance) {
-  const content = html`
+  const content = hyperHTML`
```
saschanaz: I think this doesn't need to change? We already are using hyperHTML here.

src/ui/about-respec.js
```diff
@@ -39,7 +39,7 @@ function show() {
         entries.push(entry);
       });
   }
-  render`
+  render hyperHTML `
```
saschanaz: This one is okay as it is, as `render` is already an hyperHTML thing 😁

src/core/inline-idl-parser.js
```diff
@@ -183,7 +183,7 @@ function renderAttribute(details) {
 function renderMethod(details) {
   const { args, identifier, type, parent, renderParent } = details;
   const { identifier: linkFor } = parent || {};
-  const argsText = args.map(arg => `<var>${arg}</var>`).join(", ");
+  const argsText = args.map(arg => hyperHTML `<var>${arg}</var>`).join(", ");
```
saschanaz: `.join(", ")` won't work anymore because hyperHTML will return an HTMLElement here. We need a clever way to get an array with `", "` between each item, and then we can pass it instead of `[argText]` at line 193.

src/w3c/headers.js
```diff
@@ -429,15 +429,15 @@ export function run(conf) {
         alt.hasOwnProperty("lang") && alt.lang ? ` hreflang='${alt.lang}'` : "";
       optional +=
         alt.hasOwnProperty("type") && alt.type ? ` type='${alt.type}'` : "";
-      return `<a rel='alternate' href='${alt.uri}'${optional}>${alt.label}</a>`;
+      return hyperHTML `<a rel='alternate' href='${alt.uri}'${optional}>${alt.label}</a>`;
```
saschanaz: We need a new `joinAnd`-like function but for `HTMLElement`s instead of strings.

src/w3c/headers.js
```diff
@@ -429,15 +429,15 @@ export function run(conf) {
...
     });
   if (conf.bugTracker) {
     if (conf.bugTracker.new && conf.bugTracker.open) {
-      conf.bugTrackerHTML = `<a href='${conf.bugTracker.new}'>${conf.l10n.file_a_bug}</a> ${conf.l10n.open_parens}<a href='${conf.bugTracker.open}'>${conf.l10n.open_bugs}</a>${conf.l10n.close_parens}`;
+      conf.bugTrackerHTML = hyperHTML `<a href='${conf.bugTracker.new}'>${conf.l10n.file_a_bug}</a> ${conf.l10n.open_parens}<a href='${conf.bugTracker.open}'>${conf.l10n.open_bugs}</a>${conf.l10n.close_parens}`;
```
saschanaz: We should make sure that any current use of `[conf.bugTrackerHTML]` should be `conf.bugTrackerHTML` instead. (Because the former one is for raw strings)

src/w3c/headers.js
```diff
@@ -532,20 +532,20 @@ export function run(conf) {
   if (Array.isArray(conf.wg)) {
     conf.multipleWGs = conf.wg.length > 1;
     conf.wgHTML = joinAnd(conf.wg, (wg, idx) => {
-      return `the <a href='${conf.wgURI[idx]}'>${wg}</a>`;
+      return hyperHTML `the <a href='${conf.wgURI[idx]}'>${wg}</a>`;
```
saschanaz: This also need a new joinAnd-like function

src/w3c/headers.js
```diff
@@ -532,20 +532,20 @@ export function run(conf) {
...
     });
     const pats = [];
     for (let i = 0, n = conf.wg.length; i < n; i++) {
       pats.push(
-        `a <a href='${conf.wgPatentURI[i]}' rel='disclosure'>` +
+        hyperHTML `a <a href='${conf.wgPatentURI[i]}' rel='disclosure'>` +
           `public list of any patent disclosures  (${conf.wg[i]})</a>`
       );
     }
     conf.wgPatentHTML = joinAnd(pats);
```
saschanaz: Here too, because `pats` is not a string array anymore.

src/core/issues-notes.js
saschanaz:
```suggestion
      hyperHTML`<h2>${l10n.issue_summary}</h2>`
```

Our code style does not insert a space 😁

SuyashSalampuria:
Thanks for your reviews @saschanaz . I will implement the changes as soon as possible

src/core/issues-notes.js
marcoscaceres: @SuyashSalampuria, you can quickly fix these kinds of things with: `npm run lint -- --fix`

marcoscaceres:
I'm still worried about trying to do these all at once, as we risk breaking existing specs that are currently injecting raw HTML. We need to evaluate which ones are safe to break or which are going to affect existing specs (and go around potentially fixing those specs).

saschanaz:
Reasonable. @suyashsalampuria Could you split this PR to smaller bits so that we can test them in the long term?

SuyashSalampuria:
Sure @saschanaz , i will do these by 2nd December

src/core/issues-notes.js
SuyashSalampuria: @saschanaz Should I remove insertAdjacentHTML and use prepend there instead

saschanaz: Using hyperHTML means we use elements instead, so yes we should use .prepend there.

src/w3c/headers.js
SuyashSalampuria: Should i write a new joinAnd functioin or change the existing one. If a new has to be written, where should I write it

saschanaz: The new one should go `src/core/utils.js` 👍

PR #2605, opened 2019-12-04
Convert all raw String interpolations to HyperHTML templated Strings

SuyashSalampuria:
Fixes: #2551 
Convert all raw string interpolations to hyperHTML templated strings.
Create a empty array and then push all the hyperHTML elements into it since strings were used earlier and they don't operate in the same way
Create a joinAndHyper function in core/utils which is used in same way as joinAnd is used for Strings


SuyashSalampuria:
Please bear with this PR. Next Pr's will be small enough.

Issue #2607, opened 2019-12-04
Syntax highlighting with markdown backticks doesn't work


sidvishnoi:
* ReSpec version: 24.35.2

Also, the indentation gets messed up.

![image](https://user-images.githubusercontent.com/8426945/70174816-a0e22900-16fb-11ea-8a5a-f6c1c2c06c15.png)

Minimal example:
``` html
<!DOCTYPE html>
<html>
<head>
  <meta charset='utf-8'>
  <title>mememe Specification</title>
  <script src="https://www.w3.org/Tools/respec/respec-w3c"></script>
  <script class='remove'>
    var respecConfig = {
      specStatus: "ED",
      editors: [],
      format: "markdown",
    };
  </script>
</head>

<body>
<pre class="css">
  .class {
    border: 2px solid red;
  }
</pre>

\``` css
.class {
  border: 2px solid red;
}
\```
</body>
</html>
```

mimoo:
in the mean time this might be helpful to people trying to make it work:

```py
import sys

toReplaces = ['```rust', '```']

started = False

f = open(sys.argv[1])
toWrite = ""
for line in f.readlines():
    trimmed = line.strip()
    if trimmed[:3] != "```":
        toWrite += line
    else:
        if not started:
            started = True
            for toReplace in toReplaces:
                line = line.replace(toReplace, '<pre class="rust">')
            toWrite += line
        else:
            toWrite += line.replace('```', '</pre>')
            started = False

print(toWrite)
```

PR #2608, opened 2019-12-05
fix(core/markdown): syntax highlight in fenced blocks

sidvishnoi:
Fixes https://github.com/w3c/respec/issues/2607

`getLanguageHint` in `core/highlight.js` expects languages without any prefix.
Alternative way to solve this could be to remove `"language-"` prefixes from classList, but this is one simpler (don't add `language-` prefix)

### DIFF
--- a/src/core/markdown.js
+++ b/src/core/markdown.js
@@ -214,6 +214,7 @@ export function markdownToHtml(text) {
     sanitize: false,
     gfm: true,
     headerIds: false,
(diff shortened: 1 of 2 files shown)


saschanaz:
Should we add a new test for this?

sidvishnoi:
@saschanaz Added tests for looking for hljs and webidl specific `span` elements

PR #2610, opened 2019-12-05
refactor: use `html` as the template string function

saschanaz:
Currently we are mixing the names `hyperHTML` and `html`. As using the shorter de facto standard name `html` enables syntax highlighting and linting from third party tools, this PR replaces the name `hyperHTML` into `html`.

### DIFF
--- a/src/core/best-practices.js
+++ b/src/core/best-practices.js
@@ -5,7 +5,7 @@
 // Best practices are marked up with span.practicelab.
 import { addId, getIntlData, makeSafeCopy } from "./utils.js";
 import { lang as defaultLang } from "../core/l10n.js";
-import { hyperHTML } from "./import-maps.js";
+import { html } from "./import-maps.js";
 import { pub } from "./pubsubhub.js";
 
 export const name = "core/best-practices";
@@ -31,14 +31,13 @@ export function run() {
   const summaryItems = bpSummary ? document.createElement("ul") : null;
   [...bps].forEach((bp, num) => {
     const id = addId(bp, "bp");
-    const localizedBpName = hyperHTML`
-      <a class="marker self-link" href="${`#${id}`}"><bdi lang="${lang}">${
-      l10n.best_practice
-    }${num + 1}</bdi></a>`;
+    const localizedBpName = html` <a class="marker self-link" href="${`#${id}`}"
+      ><bdi lang="${lang}">${l10n.best_practice}${num + 1}</bdi></a
+    >`;
 
     // Make the summary items, if we have a summary
     if (summaryItems) {
-      const li = hyperHTML`
+      const li = html`
         <li>
           ${localizedBpName}: ${makeSafeCopy(bp)}
         </li>
@@ -55,12 +54,12 @@ export function run() {
 
     // Make the advisement box
     container.classList.add("advisement");
-    const title = hyperHTML`${localizedBpName.cloneNode(true)}: ${bp}`;
+    const title = html`${localizedBpName.cloneNode(true)}: ${bp}`;
     container.prepend(...title.childNodes);
   });
   if (bps.length) {
     if (bpSummary) {
-      bpSummary.appendChild(hyperHTML`<h2>Best Practices Summary</h2>`);
+      bpSummary.appendChild(html`<h2>Best Practices Summary</h2>`);
       bpSummary.appendChild(summaryItems);
     }
   } else if (bpSummary) {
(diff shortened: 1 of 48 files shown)


src/core/best-practices.js
```diff
@@ -49,12 +50,18 @@ export function run() {
...
     container.prepend(...title.childNodes);
   });
   if (bps.length) {
     if (bpSummary) {
-      bpSummary.appendChild(hyperHTML`<h2>Best Practices Summary</h2>`);
+      bpSummary.appendChild(
+        html`
+          <h2>Best Practices Summary</h2>
+        `
+      );
```
saschanaz: I'm not quite sure I like this though, because there are certainly some situations where a single line becomes 5 lines. I wonder whether we have any option to make it any shorter. (Not requiring multiline for short-enough string would be great.)

src/core/custom-elements/rs-changelog.js
```diff
@@ -82,7 +82,13 @@ async function toHTML(commits) {
     const [message, prNumber = null] = commit.message.split(/\(#(\d+)\)/, 2);
     const commitURL = `${repoURL}commit/${commit.hash}`;
     const prURL = prNumber ? `${repoURL}pull/${prNumber}` : null;
-    const pr = prNumber && hyperHTML` (<a href="${prURL}">#${prNumber}</a>)`;
-    return hyperHTML`<li><a href="${commitURL}">${message.trim()}</a>${pr}</li>`;
+    const pr =
+      prNumber &&
+      html`
+        (<a href="${prURL}">#${prNumber}</a>)
```
saschanaz: This is particularly breaking, as hyperHTML does not trim the newlines when the string starts with a plaintext. Maybe worth filing a bug? 🤔

src/core/best-practices.js
sidvishnoi: Is prettier forcing this to be multi-line?

saschanaz: Yes. 😢

src/core/custom-elements/rs-changelog.js
saschanaz: There is an existing issue: https://github.com/prettier/prettier/issues/7096

src/core/best-practices.js
sidvishnoi: https://github.com/prettier/prettier/issues/5588

saschanaz: Filed https://github.com/prettier/prettier/issues/7101, because disabling the whole HTML formatting sound bad for me 👀

sidvishnoi:
The only deal breaker here is unnecessary multiline as mentioned in https://github.com/w3c/respec/pull/2610#discussion_r354288077.
I'm in favor of using `html` instead of `hyperHTML` (shorter, replaceable, syntax highlighting and formatting).

If it's just for syntax highlighting, I've been using `webreflection.literally-html` VS Code extension. It doesn't do the formatting.
image

saschanaz:
I'll wait until we see a fix for https://github.com/prettier/prettier/issues/7096.

marcoscaceres:
@saschanaz, maybe we can do this in smaller parts?

saschanaz:
> @saschanaz, maybe we can do this in smaller parts/PRs?

Not sure any partial PRs can change much, although I can change the exported name as `html` in `core/import-maps` to prefer the new name whenever preferable. Sounds good?

marcoscaceres:
As this is just a refactor, I’m ok to just go with all this.

saschanaz:
https://github.com/prettier/prettier/pull/7208 is merged, waiting for a version bump.

marcoscaceres:
Awesome work over on the other repo btw!

sidvishnoi:
@saschanaz Would be better create a new PR for this... too many conflicts

src/core/webidl.js
```diff
@@ -326,9 +330,8 @@ function renderWebIDL(idlElement, index) {
   }
   // we add "idl" as the canonical match, so both "webidl" and "idl" work
   idlElement.classList.add("def", "idl");
-  const html = webidl2.write(parse, { templates });
-  const render = html.bind(idlElement);
-  render`${html}`;
+  const highlights = webidl2.write(parse, { templates });
+  html.bind(idlElement)`${highlights}`;
```
saschanaz: There should be a native function to do this work, something like `idlElement.replaceChildren(...newChildren)`

marcoscaceres: Worth chatting to Annevk about it.

Issue #2596, opened 2019-12-03
Treat webIDL language tag as case insensitive in Markdown


marcoscaceres:
## Important info
* URL to affected spec: https://github.com/WICG/serial/pull/75

## Description of problem

What happened (e.g., it crashed)?: 

When using three backticks, I was expecting `WebIDL` to work.. 

But what I was expecting?:

To be able to use `WebIDL` or `webIDL` or just `IDL`... or even WEBIDL. 

Optional, steps to reproduce:
 1. Set `format: "markdown"` in the config. 
 1. Make a code block:

```WebIDL
interface Foo {};
```

Tag it with `WebIDL`.

sidvishnoi:
Relevant code: https://github.com/w3c/respec/blob/2e7e0e19035304f9816107ad45acf4f41e5fb487/src/core/markdown.js#L90-L95

janiceshiu:
Fixed in #2611

PR #2611, opened 2019-12-06
fix(core/markdown): more forgiving WebIDL language tag

janiceshiu:
Fixes issue [#2596](https://github.com/w3c/respec/issues/2596).

Thanks for guiding me through respec's test workflow and also for general help, @marcoscaceres!

### DIFF
--- a/src/core/markdown.js
+++ b/src/core/markdown.js
@@ -89,7 +89,7 @@ const inlineElems = new Set([
 class Renderer extends marked.Renderer {
   code(code, language, isEscaped) {
     // regex to check whether the language is webidl
-    if (/(^webidl$|^idl$)/i.test(language)) {
+    if (/(^webidl$)/i.test(language)) {
       return `<pre class="idl">${code}</pre>`;
     }
     return super.code(code, language, isEscaped);
--- a/src/core/markdown.js
+++ b/src/core/markdown.js
@@ -95,7 +95,7 @@ class Renderer extends marked.Renderer {
   }
 
   isWebidlLanguageTag(tag) {
-    return /(^webidl$|^idl$)/i.test(tag);
+    return /(^(web)?idl$)/i.test(tag);
   }
 }
 
--- a/src/core/markdown.js
+++ b/src/core/markdown.js
@@ -88,14 +88,11 @@ const inlineElems = new Set([
 
 class Renderer extends marked.Renderer {
   code(code, language, isEscaped) {
-    if (this.isWebidlLanguageTag(language)) {
+    // regex to check whether the language is webidl
+    if (/(^webidl$|^idl$)/i.test(language)) {
       return `<pre class="idl">${code}</pre>`;
     }
     return super.code(code, language, isEscaped);
-  }
-
-  isWebidlLanguageTag(tag) {
-    return /(^(web)?idl$)/i.test(tag);
   }
 }
 
--- a/src/core/markdown.js
+++ b/src/core/markdown.js
@@ -88,10 +88,14 @@ const inlineElems = new Set([
 
 class Renderer extends marked.Renderer {
   code(code, language, isEscaped) {
-    if (language === "webidl") {
+    if (this.isWebidlLanguageTag(language)) {
       return `<pre class="idl">${code}</pre>`;
     }
     return super.code(code, language, isEscaped);
+  }
+
+  isWebidlLanguageTag(tag) {
+    return /(^webidl$|^idl$)/i.test(tag);
   }
 }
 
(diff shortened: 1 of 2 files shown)


src/core/markdown.js
```diff
@@ -88,11 +88,15 @@ const inlineElems = new Set([
...
   code(code, language, isEscaped) {
-    if (language === "webidl") {
+    if (this.isWebidlLanguageTag(language)) {
       return `<pre class="idl">${code}</pre>`;
     }
     return super.code(code, language, isEscaped);
   }
+
+  isWebidlLanguageTag(tag) {
+    return /(^webidl$|^idl$)/i.test(tag);
```
sidvishnoi: Lets do this inline instead of a new method.

sidvishnoi: The `idl` is common, so `web` can be treated as optional
```suggestion
    return /(^(web)?idl$)/i.test(tag);
```

tests/spec/core/markdown-spec.js
```diff
@@ -284,6 +284,93 @@ describe("Core - Markdown", () => {
...
+
+      \`\`\` idlidl
+      [Exposed=Window]
+      interface FooIdlIdl {};
+      \`\`\`
+
+      \`\`\` webidlwebidl
+      [Exposed=Window]
+      interface FooWebidlWebidl {};
+      \`\`\`
```
sidvishnoi: I would suggest removing these tests, as they seem a bit redundant. 
If the user writes an incorrect language tag, they'll see a `<pre>` without any highlighting - which is enough indication to show they wrote something wrong. Supporting case-insensitivity allows user to not remember/lookup whether it was "WebIDL" or "webIDL" - which is good to have. If they're writing "webID", it's on them :slightly_smiling_face:

saschanaz: >which is enough indication to show they wrote something wrong.

The fact that we give enough indication does not mean we don't need a test, does it? 👀

I still think this can somehow be simplified by creating a loop.

sidvishnoi: A loop is a nice middle ground.

saschanaz: Hmmmm..... Reading this again, I don't think we need the latter four. (I thought Sid was about the whole test.)

janiceshiu:
@sidvishnoi @saschanaz addressed y'all's requested changes!
@marcoscaceres thanks for explaining some respec conventions.

marcoscaceres:
This looks good to me. It covers the cases we are primarily interested in. @sidvishnoi, let us know if you are happy with it.

saschanaz:
BTW, do we really want to support ```` ```idl ````? Why?

marcoscaceres:
> BTW, do we really want to support ```idl? Why?

We currently do `<pre class="idl">`, so it's a nice analog to that.

saschanaz:
Currently just `idl` loads completely different syntax highlighter on GitHub, though. Not sure what it is.

```idl
interface Foo {
  void foo();
};
```

```webidl
interface Foo {
  void foo();
};
```

saschanaz:
Might become a bit ambiguous if TC39 decides to introduce non-WebIDL IDL... https://github.com/tc39/proposal-idl

marcoscaceres:
@saschanaz makes a strong case. I wasn't aware that the WebIDL/JSIDL discussions had restarted.  

Ok, let's play it safe and go with just `WebIDL` (and follow GitHub also). ReSpec's position is always to stay neutral (though I personally think JSIDL is a waste of time given how much infrastructure browsers have for WeBIDL... bla bla bla).

src/core/markdown.js
```diff
@@ -88,7 +88,8 @@ const inlineElems = new Set([
 
 class Renderer extends marked.Renderer {
   code(code, language, isEscaped) {
-    if (language === "webidl") {
+    // regex to check whether the language is webidl
+    if (/(^webidl$|^idl$)/i.test(language)) {
       return `<pre class="idl">${code}</pre>`;
```
janiceshiu: regarding @saschanaz's point about `idl` and `webidl` -
to confirm - it is okay that we're currently tagging `webidl` stuff as `<pre class="idl">`, right?

cc @marcoscaceres

saschanaz: Yes, for backward compatibility. Should we also `pre.webidl`? @marcoscaceres

marcoscaceres: > Should we also allow `pre.webidl`?

Yes, that would be fine. Should we do that as a followup? that would require changing core/webidl.

marcoscaceres: @janiceshiu, could you please file a follow-up bug to support the above?

saschanaz:
Great 🎉🎉 

Don't forget closing #2596 too! (It will automatically close if you say "Fixes #2596"` on the PR description. "Fixes issue #2596" is unfortunately not supported...)

sidvishnoi:
Nicely done! Congrats on getting your first PR merged in Respec! 🎉

src/core/markdown.js
janiceshiu: [filed!](https://github.com/w3c/respec/issues/2614)

PR #2613, opened 2019-12-07
refactor(core/utils): add `htmlJoinAnd` utility

### LINKED ISSUE #2551: Remove raw HTML string interpolation

SuyashSalampuria:
Fixes #2551 
Written the htmlJoinAnd function and replaced joinAnd with it at one place in w3c/headers.js
`joinAnd` is still used in w3c/headers.js  at line 432 and line 544. 
In line 544, only one argument(array) is passed in joinAnd. The function to convert it to html elements is not there. 
In line 432, the tests are written, which expects it to be a string. So, htmlJoinAnd can't be used


src/core/utils.js
```diff
@@ -494,6 +495,30 @@ export function flatten(collector, item) {
...
+      return items[0];
+    case 2: // x and y
+      return hyperHTML`${items[0]} and ${items[1]}`;
+    default: {
+      // x, y, and z
+      let joinedItems = ``;
+      for (let i = 0; i < array.length - 1; i++) {
+        joinedItems = `${joinedItems}, ${items[i]}`;
+      }
+      joinedItems = `${joinedItems} and ${items[length - 1]}`;
```
saschanaz: The result of this loop is a bit broken by two ways.

* Consider we have `["x", "y", "z"]` in the array. `joinedItems` will first be `""`, and then `", x"`, `", x, y"`, and finally `", x, y and z"`, which is not what we want. The result should be: `"x, y, and z"`.
* This still does interpolation *before* we pass it into hyperHTML. We should do interpolation *with* hyperHTML so that it can process `HTMLElement`s properly, not stringifying it into `[object HTMLElement]`.

Tests are passing anyway, as it turns out there is no test for `conf.wgHTML`...

But still, this PR is now small enough to review and doesn't break any other things, which is a big 👍! 😁

SuyashSalampuria: @saschanaz I have updated the PR. Please review it once more. 
Thanks for helping me so much

src/core/utils.js
```diff
@@ -494,6 +495,30 @@ export function flatten(collector, item) {
...
+      return items[0];
+    case 2: // x and y
+      return hyperHTML`${items[0]} and ${items[1]}`;
+    default: {
+      // x, y, and z
+      let joinedItems = `${items[0]}`;
+      for (let i = 1; i < items.length - 1; i++) {
+        joinedItems = hyperHTML`${joinedItems}, ${items[i]}`;
+      }
+      joinedItems = hyperHTML`${joinedItems} and ${items[length - 1]}`;
```
saschanaz: * This still treats `items[0]` as a string 👀, we should also pass it into hyperHTML.
* Too easy to forget `hyperHTML` here, so how about:

   ```js
   const joinedItems = items.slice(-1).map(item => hyperHTML`${item}, `);
   return hyperHTML`${joinedItems} and ${items[items.length - 1]}`;
   ```

   IMO much shorter in this way. (The value of `joinedItems` never changes which leads to more readable code, and also adds a required comma before "and".)

* Note that `length` here is undefined, should be `items.length` instead.

SuyashSalampuria: items is an array of hyperHTML elements. So items[0] should be an HTML element here.

I will just change length to items.length

SuyashSalampuria: Also 
`const joinedItems = items.slice(-1).map(item => hyperHTML`${item}, `);
return hyperHTML'${joinedItems} and ${items[items.length - 1]}';`
will return the last element twice separated by both `,` and `and`
`const joinedItems = items.slice(0,-1).map(item => hyperHTML`${item}, `);
return hyperHTML'${joinedItems} and ${items[items.length - 1]}';`
will have the same problem as my previous code had.

SuyashSalampuria:
@saschanaz is this PR fine or I need to make more changes

SuyashSalampuria:
> Sorry for the delay! We still have a remaining work: adding a test, although the code itself looks good now +1
> 
> Since there is no test for `htmlJoinAnd`, we want to add one in [`tests/spec/core/utils-spec.js`](https://github.com/w3c/respec/blob/develop/tests/spec/core/utils-spec.js).
> 
> Basically the test would be similar with:
> 
> https://github.com/w3c/respec/blob/b9dab188f0eb460980fcabac718e85cb421fed47/tests/spec/core/utils-spec.js#L324-L337
> 
> But the input here should be `HTMLElement` arrays and we should make sure that the return value contains those elements intact. (That means we can't simply compare with a string here.)
> 
> Since you already did a good job, maybe I can do the test part (please tell me if you want me to do so), but if you want to continue your challenge, please do! grin

Please let me take up this challenge :)

saschanaz:
BTW, make sure you don't commit `builds/` as we only actually update them when getting a new release. 😁

saschanaz:
One more request while you're at it:

https://github.com/w3c/respec/blob/24a54cad846a0f833d095395d369beda8cfb8e0c/src/w3c/templates/sotd.js#L333

Please replace `[conf.wgHTML]` here with `conf.wgHTML` as the array becomes redundant with this change 👍 (It's a weird hyperHTML way to mark the input as a raw string.)

SuyashSalampuria:
The code is almost ready. I think that I will be able to commit it by tomorrow :)
Thanks @saschanaz for helping me so much

SuyashSalampuria:
@saschanaz  Please help me in this
_case 2: // x and y
      return `${items[0]} and ${items[1]}`;_
goes fine with the test:
     _expect(utils.htmlJoinAnd(["x", "x"])).toEqual(`x and x`);_

But 
_case 2: // x and y
      return hyperHTML`${items[0]} and ${items[1]}`;_
gives an error with the test
      _expect(utils.htmlJoinAnd(["x", "x"])).toEqual(hyperHTML`x and x`);_
The error is
 _Expected Pe({ childNodes: [ HTMLNode, HTMLNode, HTMLNode, HTMLNode, HTMLNode ], firstChild: HTMLNode, lastChild: HTMLNode, ownerDocument: HTMLNode, _: null }) to equal HTMLNode._
 If the test is 
   _expect(utils.htmlJoinAnd(["x", "x"]).firstChild).toEqual(hyperHTML`x and x`);_
The error is 
  _Expected HTMLNode to equal HTMLNode._

SuyashSalampuria:
Also, if any test is written for any hyperHTML element, please send it . 
It will be much helpful

saschanaz:
Because with hyperHTML we return a hyperHTML object instead of a string.

We probably want something like:

```js
const div = document.createElement("div");
const render = hyperHTML.bind(div);
render`${utils.htmlJoinAnd(["<x>", "<x>"], item => hyperHTML`<a>${item}</a>`)}`
expect(div.innerHTML).toBe("<a>&lt;x&gt;</a> and <a>&lt;x&gt;</a>");
```

(Might include some error, @sidvishnoi could you take a look? I'm out now, sorry 😅)

sidvishnoi:
@SuyashSalampuria The tests suggested by @saschanaz are the way to go. Don't forget to import `hyperHTML` in `utils-spec.js` though.

Also, instead of testing innerHTML, try using `querySelector` and `textContent`, as hyperHTML will insert some HTML comments which will change on each test run. Something like following maybe:

``` js
      const div = document.createElement("div");
      hyperHTML.bind(div)`${utils.htmlJoinAnd(
        ["<x>", "<y>"],
        item => hyperHTML`<a>${item}</a>`
      )}`;
      expect(div.textContent).toBe("<x> and <y>");
      const [linkA, linkB] = div.querySelectorAll("a");
      expect(linkA.textContent).toBe("<x>");
      expect(linkB.textContent).toBe("<y>");
```

SuyashSalampuria:
Thanks @sidvishnoi  @saschanaz  a lot for your help

tests/spec/core/utils-spec.js
```diff
@@ -453,6 +454,47 @@ describe("Core - Utils", () => {
     });
   });
 
+  describe("htmlJoinAnd", () => {
+    it("joins with proper commas and 'and'", () => {
+      const div = document.createElement("div");
+
+      hyperHTML.bind(div)`${utils.htmlJoinAnd(
```
saschanaz:
```js
const render  = hyperHTML.bind(div);
render`...`
```

... will be make this shorter 👍

tests/spec/core/utils-spec.js
```diff
@@ -453,6 +454,47 @@ describe("Core - Utils", () => {
...
+        [],
+        item => hyperHTML`<a>${item}</a>`
+      )}`;
+      expect(div.textContent).toBe("");
+
+      hyperHTML.bind(div)`${utils.htmlJoinAnd(
+        ["x"],
+        item => hyperHTML`<a>${item}</a>`
+      )}`;
+      expect(div.textContent).toBe("x");
```
saschanaz: Not just "x", we should try inserting some HTML and check it escapes well. `"<x>"` may work.

tests/spec/core/utils-spec.js
```diff
@@ -453,6 +454,42 @@ describe("Core - Utils", () => {
     });
   });
 
+  describe("htmlJoinAnd", () => {
+    it("joins with proper commas and 'and'", () => {
+      const div = document.createElement("div");
+      const render = hyperHTML.bind(div);
+
+      render`${utils.htmlJoinAnd([], item => hyperHTML`<a>${item}</a>`)}`;
+      expect(div.textContent).toBe("");
```
saschanaz: We additionally need to check we get `<a>` as expected. Just `expect(div.getElementsByTagName("a").length).toBe(...)` would be fine.

SuyashSalampuria:
Thanks @saschanaz a lot. 
Such a small task seemed to be so big because of me.
I hope I will get better and will be able to contribute a lot now

saschanaz:
> Such a small task seemed to be so big because of me.

No way! You did a good job even when you didn't have hyperHTML experience, which is amazing 👍

> I hope I will get better and will be able to contribute a lot now

Yay, please do 😍

src/core/utils.js
```diff
@@ -494,6 +495,25 @@ export function flatten(collector, item) {
 
 // --- DOM HELPERS -------------------------------
 
+// Takes an array and returns a string that separates each of its items with the proper commas and
```
sidvishnoi: I think there is some room for improvement in this comment.

src/core/utils.js
```diff
@@ -494,6 +495,25 @@ export function flatten(collector, item) {
...
+// "and". The second argument is a mapping function that can convert the items before they are
+// joined
+// Finally converts to hyperHTML while joining them
+// htmlJoinAnd(["x"])).toEqual(hyperHTML`x`);
+export function htmlJoinAnd(array = [], mapper = item => hyperHTML`${item}`) {
```
sidvishnoi: I don't think array can be undefined. If it is, it's a usage error.
Maybe use JSDoc/ts-check to ensure that it's an error.

src/core/utils.js
saschanaz: Agree, and this means our existing comment for `joinAnd` is also bad... 😱

Maybe some JSDoc will allow removing some prose.

tests/spec/core/utils-spec.js
```diff
@@ -453,6 +454,49 @@ describe("Core - Utils", () => {
...
+      expect(div.getElementsByTagName("a").length).toBe(3);
+
+      render`${utils.htmlJoinAnd(
+        ["<x>", "<x>", "<x>", "<x>"],
+        item => hyperHTML`<a>${item}</a>`
+      )}`;
+      expect(div.textContent).toBe("<x>, <x>, <x>, and <x>");
+      expect(div.getElementsByTagName("a").length).toBe(4);
+
+      render`${utils.htmlJoinAnd(
```
sidvishnoi: Maybe skip this one test and add a `<X>` in previous test.

src/core/utils.js
saschanaz: TS strict mode is needed to check `undefined`. Not sure we are ready to be strict though 😥

sidvishnoi: 159 errors with `strictNullChecks` :grin: 
I think it's fine to add JSDoc here enforcing `any[]`. It's cleaner in terms of this util API (and we're the sole users of this util)

src/core/utils.js
saschanaz: My suggestion:

```js
/**
 * Separates each item with proper commas and "and".
 * @param {string[]} array
 * @param {(str: string) => object} mapper
 */
```

Anything else looks redundant.

sidvishnoi: LGTM.

marcoscaceres: Suggested comment looks good.

SuyashSalampuria: Shall I try using JSDoc

src/core/utils.js
saschanaz: Final nit (sorry to add again and again 😭):

```suggestion
export function htmlJoinAnd(array, mapper = item => item) {
```

Removing array default value as @sidvishnoi mentioned, and reverting mapper default value because `` hyperHTML`${item}` `` doesn't really do anything here (it's near no-op).

src/core/utils.js
saschanaz: Yup, https://github.com/w3c/respec/pull/2613#discussion_r356013231 uses JSDoc syntax. (Note the use of `/**` and `@param`.)

SuyashSalampuria:
@sidvishnoi  Thanks for commit of reduntant test :)
I had just began working on it though.

src/core/utils.js
```diff
@@ -494,6 +495,25 @@ export function flatten(collector, item) {
...
+ * Separates each item with proper commas and "and".
+ * @param {string[]} array
+ * @param {(str: string) => object} mapper
+ */
+export function htmlJoinAnd(array, mapper = item => item) {
+  const items = array.map(mapper);
+  switch (items.length) {
+    case 0:
+    case 1: // "x"
+      return hyperHTML`${items[0]}`;
```
saschanaz:
```suggestion
      return items[0];
```

Another final nit (will commit this myself)

PR #2615, opened 2019-12-09
refactor(core/markdown): parse blocks in native way

saschanaz:
Previously the module parsed leaf elements as markdown and then parsed their parents again including the same children, resulting overheads (and harder-to-understand code).

Now this change utilizes [native Markdown syntax](https://spec.commonmark.org/0.29/#example-157) to allow block level Markdown:

```markdown
<section>

# Markdown allows this if double newlines exist after an HTML start tag

__Yay!__

</section>
```

```markdown
<section>
<!--
This is incompatible markdown, we previously manually querySelectorAll'ed and parsed this.
Now the PR just adds newline for marked to properly detect this block as markdown.
 -->
# title
</section>
```

No additional tests needed as this is already covered.


saschanaz:
Or we may just drop this and force specs to add proper newlines...

saschanaz:
>I think if an editor uses markdown, then we should do no special handling (stick to the standards!). 

@marcoscaceres Does that mean we should drop this complexity? This PR still does same special handling as we did but with less overhead.

marcoscaceres:
> @marcoscaceres Does that mean we should drop this complexity? This PR still does same special handling as we did but with less overhead.

The only way we can drop the complexity is if we won't break specs. We would need to check... but if specs are depending on it, then we can't drop it.

saschanaz:
Per https://github.com/WICG/serial/pull/75 I think we need this.

Issue #2566, opened 2019-11-11
Add support for `data-local-lt`


marcoscaceres:
Right now, we have a situation where we are exporting, for example:
```
<dfn data-lt=
        "installing|installed|installation" data-export="">install</dfn>
```

We should provide a means to exclude some terms from being exported. BikeShed uses `data-local-lt`, I believe, to achieve the same thing. Would be good if we supported the same thing.

saschanaz:
Relevant Bikeshed doc: https://tabatkins.github.io/bikeshed/#local-lt

marcoscaceres:
Added documentation:
https://github.com/w3c/respec/wiki/data-local-lt

@sidvishnoi, want to check it?

sidvishnoi:
Nice. Thanks!

PR #2616, opened 2019-12-09
enhancement(core/dfn): add support for `data-local-lt`

sidvishnoi:
Closes https://github.com/w3c/respec/issues/2566


sidvishnoi:
I'll look for things that can be moved from `lt` to `local-lt`

PR #2619, opened 2019-12-10
chore(ci): allow skipping builds

sidvishnoi:


sidvishnoi:
If `event_type` is `"pull_request"`, we don't have access to `head_commit` (hence commit message) :sob: 
Can add a bash script in action, but not worth it maybe. Re-open if this interests you.

Issue #1103, opened 2017-02-21
Steal WHATWG/HTML's dfn popup thing


marcoscaceres:
We've been missing that feature for too long.

PR #2626, opened 2019-12-11
feat(core/dfn-panel): add panel with local dfn references

sidvishnoi:
Closes https://github.com/w3c/respec/issues/1103 (finally! :tada:)
:sparkles: Check out the demo at respec-preview (in checks below or [here](https://deploy-preview-2626--respec-pr.netlify.com/)) :sparkles:
Markup and style are same as bikeshed for consistency.

Should we add a resize handler to reposition the panel?
 
// TODO
- [x] Add tests

### DIFF
--- a/assets/dfn-panel.css
+++ b/assets/dfn-panel.css
@@ -66,6 +66,7 @@ dfn {
   top: unset;
   bottom: 2em;
   margin: 0 auto;
-  max-width: calc(100vw - 2.5em);
+  /* 0.75em from padding, 0.5em from left position */
+  max-width: calc(100vw - (0.75em + 0.5em) * 2);
   max-height: 30vh;
 }
--- a/assets/dfn-panel.css
+++ b/assets/dfn-panel.css
@@ -52,6 +52,7 @@ dfn {
   display: inline-block;
   position: fixed;
   left: 0.5em;
+  top: unset;
   bottom: 2em;
   margin: 0 auto;
   max-width: calc(100vw - 1.5em - 0.4em - 0.5em);
--- a/assets/dfn-panel.css
+++ b/assets/dfn-panel.css
@@ -2,6 +2,7 @@
 dfn {
   cursor: pointer;
 }
+
 .dfn-panel {
   position: absolute;
   left: var(--left); /* set via JS */
@@ -18,36 +19,46 @@ dfn {
   color: black;
   border: outset 0.2em;
 }
+
 .dfn-panel * {
   margin: 0;
   padding: 0;
   text-indent: 0;
 }
+
 .dfn-panel > b {
   display: block;
 }
+
 .dfn-panel a[href] {
   color: black;
 }
+
 .dfn-panel a[href].self-link {
   display: inline-block;
 }
+
 .dfn-panel a:not(:hover) {
   text-decoration: none !important;
   border-bottom: none !important;
 }
+
 .dfn-panel a[href]:hover {
   border-bottom-width: 1px;
 }
+
 .dfn-panel > b + b {
   margin-top: 0.25em;
 }
+
 .dfn-panel ul {
   padding: 0;
 }
+
 .dfn-panel li {
   list-style: inside;
 }
+
 .dfn-panel.minimized {
   display: inline-block;
   position: fixed;
--- a/assets/dfn-panel.css
+++ b/assets/dfn-panel.css
@@ -30,12 +30,8 @@ dfn {
   display: block;
 }
 
-.dfn-panel a[href] {
+.dfn-panel ul a[href] {
   color: black;
-}
-
-.dfn-panel a[href].self-link {
-  display: inline-block;
 }
 
 .dfn-panel a:not(:hover) {
--- a/assets/dfn-panel.css
+++ b/assets/dfn-panel.css
@@ -53,7 +53,7 @@ dfn {
   list-style: inside;
 }
 
-.dfn-panel.minimized {
+.dfn-panel.docked {
   display: inline-block;
   position: fixed;
   left: 0.5em;
--- a/assets/dfn-panel.css
+++ b/assets/dfn-panel.css
@@ -22,8 +22,6 @@ dfn {
 
 .dfn-panel * {
   margin: 0;
-  padding: 0;
-  text-indent: 0;
 }
 
 .dfn-panel > b {
--- a/assets/dfn-panel.css
+++ b/assets/dfn-panel.css
@@ -27,6 +27,9 @@ dfn {
 .dfn-panel a[href] {
   color: black;
 }
+.dfn-panel a[href].self-link {
+  display: inline-block;
+}
 .dfn-panel a:not(:hover) {
   text-decoration: none !important;
   border-bottom: none !important;
--- a/assets/dfn-panel.css
+++ b/assets/dfn-panel.css
@@ -66,6 +66,6 @@ dfn {
   top: unset;
   bottom: 2em;
   margin: 0 auto;
-  max-width: calc(100vw - 1.5em - 0.4em - 0.5em);
+  max-width: calc(100vw - 2.5em);
   max-height: 30vh;
 }
--- a/assets/dfn-panel.css
+++ b/assets/dfn-panel.css
@@ -4,6 +4,8 @@ dfn {
 }
 .dfn-panel {
   position: absolute;
+  left: var(--left); /* set via JS */
+  top: var(--top); /* set via JS */
   z-index: 35;
   height: auto;
   width: max-content;
--- a/assets/dfn-panel.css
+++ b/assets/dfn-panel.css
@@ -66,7 +66,7 @@ dfn {
   top: unset;
   bottom: 2em;
   margin: 0 auto;
-  /* 0.75em from padding, 0.5em from left position */
-  max-width: calc(100vw - (0.75em + 0.5em) * 2);
+  /* 0.75em from padding (x2), 0.5em from left position, 0.2em border (x2) */
+  max-width: calc(100vw - 0.75em * 2 - 0.5em - 0.2em * 2);
   max-height: 30vh;
 }
--- a/assets/dfn-panel.css
+++ b/assets/dfn-panel.css
@@ -1,0 +1,56 @@
+dfn {
+  cursor: pointer;
+}
+.dfn-panel {
+  position: absolute;
+  z-index: 35;
+  height: auto;
+  width: max-content;
+  max-width: 300px;
+  max-height: 500px;
+  overflow: auto;
+  padding: 0.5em 0.75em;
+  font: small Helvetica Neue, sans-serif, Droid Sans Fallback;
+  background: #dddddd;
+  color: black;
+  border: outset 0.2em;
+}
+.dfn-panel:not(.on) {
+  display: none;
+}
+.dfn-panel * {
+  margin: 0;
+  padding: 0;
+  text-indent: 0;
+}
+.dfn-panel > b {
+  display: block;
+}
+.dfn-panel a[href] {
+  color: black;
+}
+.dfn-panel a:not(:hover) {
+  text-decoration: none !important;
+  border-bottom: none !important;
+}
+.dfn-panel a[href]:hover {
+  border-bottom-width: 1px;
+}
+.dfn-panel > b + b {
+  margin-top: 0.25em;
+}
+.dfn-panel ul {
+  padding: 0;
+}
+.dfn-panel li {
+  list-style: inside;
+}
+.dfn-panel.activated {
+  display: inline-block;
+  position: fixed;
+  left: 0.5em;
+  bottom: 2em;
+  margin: 0 auto;
+  max-width: calc(100vw - 1.5em - 0.4em - 0.5em);
+  max-height: 30vh;
+}
--- a/assets/dfn-panel.css
+++ b/assets/dfn-panel.css
@@ -1,3 +1,4 @@
+/* dfn popup panel that list all local references to a dfn */
 dfn {
   cursor: pointer;
 }
@@ -14,9 +15,6 @@ dfn {
   background: #dddddd;
   color: black;
   border: outset 0.2em;
-}
-.dfn-panel:not(.on) {
-  display: none;
 }
 .dfn-panel * {
   margin: 0;
@@ -45,7 +43,7 @@ dfn {
 .dfn-panel li {
   list-style: inside;
 }
-.dfn-panel.activated {
+.dfn-panel.minimized {
   display: inline-block;
   position: fixed;
   left: 0.5em;
--- a/profiles/w3c.js
+++ b/profiles/w3c.js
@@ -56,6 +56,7 @@ const modules = [
   import("../src/core/data-tests.js"),
   import("../src/core/list-sorter.js"),
   import("../src/core/highlight-vars.js"),
+  import("../src/core/dfn-panel.js"),
   import("../src/core/data-type.js"),
   import("../src/core/algorithms.js"),
   import("../src/core/anchor-expander.js"),
--- a/src/core/dfn-panel.js
+++ b/src/core/dfn-panel.js
@@ -70,7 +70,7 @@ function createPanel(dfn) {
   /** @type {HTMLElement} */
   const panel = hyperHTML`
     <aside class="dfn-panel" id="dfn-panel">
-      <b><a class="self-link" href="${href}">${href}</a></b>
+      <b><a class="self-link" href="${href}">Permalink</a></b>
       ${
         links.length
           ? hyperHTML`<b>Referenced in:</b>${referencesToHTML(id, links)}`
--- a/src/core/dfn-panel.js
+++ b/src/core/dfn-panel.js
@@ -99,11 +99,11 @@ function referencesToHTML(id, links) {
   });
 
   /**
-   * Takes an "entry" from `titleToIDs`, and converts it to a list that is
-   * easier to render in `listItemToHTML`. The first list item contains title
-   * from `getReferenceTitle`, rest of items contain strings like `(2)`, `(3)`
-   * as title.
-   * @param {[string, string[]]} entry
+   * Returns a list that is easier to render in `listItemToHTML`.
+   * @param {[string, string[]]} entry an entry from `titleToIDs`
+   * @returns {{ title: string, id: string }[]} The first list item contains
+   * title from `getReferenceTitle`, rest of items contain strings like `(2)`,
+   * `(3)` as title.
    */
   const toLinkProps = ([title, ids]) => {
     return [{ title, id: ids[0] }].concat(
--- a/src/core/dfn-panel.js
+++ b/src/core/dfn-panel.js
@@ -68,8 +68,7 @@ function parseAction(clickTarget) {
 function createPanel(dfn) {
   const { id } = dfn;
   const href = `#${id}`;
-
-  const links = [...document.querySelectorAll(`a[href="${href}"]`)];
+  const links = document.querySelectorAll(`a[href="${href}"]`);
 
   /** @type {HTMLElement} */
   const panel = hyperHTML`
@@ -87,7 +86,7 @@ function createPanel(dfn) {
 
 /**
  * @param {string} id dfn id
- * @param {HTMLLinkElement[]} links
+ * @param {NodeListOf<HTMLLinkElement>} links
  * @returns {HTMLUListElement}
  */
 function referencesToHTML(id, links) {
--- a/src/core/dfn-panel.js
+++ b/src/core/dfn-panel.js
@@ -148,11 +148,11 @@ function displayPanel(dfn, panel) {
     left = dfnRect.left - (panelWidth + 5);
     if (left < 0) {
       left = dfnRect.left;
-      top = top + dfnRect.height;
+      top += dfnRect.height;
     }
   }
 
-  // "docked" state can independently set its own position in CSS.
+  // Allows ".docked" rule to override the position, unlike `style.left = left`.
   panel.style.setProperty("--left", `${left}px`);
   panel.style.setProperty("--top", `${top}px`);
 }
--- a/src/core/dfn-panel.js
+++ b/src/core/dfn-panel.js
@@ -5,12 +5,10 @@ import { fetchAsset } from "./text-loader.js";
 import { hyperHTML } from "./import-maps.js";
 import { norm } from "./utils.js";
 
-const cssPromise = loadStyle();
-
 export const name = "core/dfn-panel";
 
 export async function run() {
-  const css = await cssPromise;
+  const css = await loadStyle();
   document.head.insertBefore(
     hyperHTML`<style>${css}</style>`,
     document.querySelector("link")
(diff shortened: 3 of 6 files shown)


profiles/w3c.js
```diff
@@ -56,6 +56,7 @@ const modules = [
   import("../src/core/data-tests.js"),
   import("../src/core/list-sorter.js"),
   import("../src/core/highlight-vars.js"),
+  import("../src/core/dfn-panel.js"),
```
sidvishnoi: Will add to other profile also. Not sure if this should be module of it's own or a part of `core/dfn`

sidvishnoi:
Adding some screenshots, as it's fun :smile: 

By default, goes to right side of dfn.
image

If not enough space on right side, moves left
image

If not enough space on either side, moves down
image

Gets pinned down if some reference (other than self link) is clicked
image

Positioning made simple with a "state machine"  :tm:

src/core/dfn-panel.js
```diff
@@ -0,0 +1,163 @@
...
+    if (left < 0) {
+      left = dfnRect.left;
+      top = top + dfnRect.height;
+    }
+  }
+
+  Object.assign(panel.style, { left: `${left}px`, top: `${top}px` });
+}
+
+async function loadStyle() {
```
marcoscaceres: we should make this a util function.

saschanaz: As previously noted, doing so will confuse the bundler as it can't track dynamic import expression with dynamic path value. What we need is a way to import CSS as mentioned in [a recent TC39 proposal](https://github.com/tc39/proposal-module-attributes/blob/master/README.md#motivation).

src/core/dfn-panel.js
```diff
@@ -0,0 +1,162 @@
...
+
+const cssPromise = loadStyle();
+
+export const name = "core/dfn-panel";
+
+export async function run() {
+  const css = await cssPromise;
+  document.head.insertBefore(
+    hyperHTML`<style>${css}</style>`,
+    document.querySelector("link")
```
marcoscaceres:
```suggestion
```

assets/dfn-panel.css
```diff
@@ -0,0 +1,56 @@
+dfn {
```
marcoscaceres: As this is being inserted into the document, we should add comment just explaining that this is the dfn-popup.

src/core/dfn-panel.js
```diff
@@ -0,0 +1,161 @@
...
+  );
+
+  document.body.addEventListener("click", event => {
+    /** @type {HTMLElement} */
+    const el = event.target;
```
marcoscaceres:
```suggestion
    const { target: el } = event;
```

src/core/dfn-panel.js
```diff
@@ -0,0 +1,161 @@
...
+    const panel = document.getElementById("dfn-panel");
```
marcoscaceres: Maybe move this to the outer function? Then only needs to do the lookup once?

src/core/dfn-panel.js
```diff
@@ -0,0 +1,161 @@
...
+
+    const action = parseAction(el);
```
marcoscaceres: Total nit... but we are not really "parsing" and action... move deriving an action... so `deriveAction()`... I know, my feedback is amazing.

src/core/dfn-panel.js
sidvishnoi: Can't. We remove panel from DOM instead of just hiding it (as we recreate entire content - which is simpler)
This reminds me, we can simplify CSS due to that.

src/core/dfn-panel.js
```diff
@@ -0,0 +1,161 @@
...
+    switch (action) {
+      case "SHOW": {
```
marcoscaceres: More nits... maybe lowercase these? uppercase is so loud 🙉

src/core/dfn-panel.js
```diff
@@ -0,0 +1,161 @@
...
+      }
+    }
+  });
+}
+
+/** @param {HTMLElement} clickTarget */
+function parseAction(clickTarget) {
+  const hitALink = !!clickTarget.closest("a");
+  if (clickTarget.closest("dfn")) {
+    return hitALink ? "NOOP" : "SHOW";
```
marcoscaceres: Maybe just: 

```suggestion
    return hitALink ? null : "SHOW";
```

As it's literally an unhandled case.

src/core/dfn-panel.js
```diff
@@ -0,0 +1,161 @@
...
+    switch (action) {
+      case "SHOW": {
+        if (panel) panel.remove();
+        const dfn = el.closest("dfn");
+        displayPanel(dfn, createPanel(dfn));
+        break;
+      }
+      case "ACTIVATE": {
+        panel.classList.add("activated");
+        panel.style.left = null;
```
marcoscaceres: no way to do this with CSS, I guess?

src/core/dfn-panel.js
```diff
@@ -0,0 +1,161 @@
...
+      const clickedSelfLink = clickTarget.classList.contains("self-link");
+      return clickedSelfLink ? "HIDE" : "ACTIVATE";
+    }
+    const panel = clickTarget.closest("#dfn-panel");
+    return panel.classList.contains("activated") ? "HIDE" : "NOOP";
+  }
+  if (document.getElementById("dfn-panel")) {
+    return "HIDE";
+  }
+  return "NOOP";
```
marcoscaceres: as above... or maybe just ""?

src/core/dfn-panel.js
sidvishnoi: nothing beats inline styles added via JS

src/core/dfn-panel.js
```diff
@@ -0,0 +1,161 @@
...
+ * @param {string} id dfn id
+ * @param {NodeListOf<HTMLLinkElement>} links
+ * @returns {HTMLUListElement}
+ */
+function referencesToHTML(id, links) {
+  /** @type {Map<string, string[]>} */
+  const titleToIDs = new Map();
+  links.forEach((link, i) => {
+    const linkID = link.id || `ref-for-${id}-${i + 1}`;
+    if (!link.id) link.id = linkID;
```
marcoscaceres: Maybe do this before... then you don't need to build linkID if it's not going to be used?

src/core/dfn-panel.js
sidvishnoi: Would've to change JSDoc definition this way

src/core/dfn-panel.js
sidvishnoi: Removed the "on" state as if it exists, it's either minimzed (activated) or visible,

src/core/dfn-panel.js
```diff
@@ -0,0 +1,161 @@
...
+    const title = getReferenceTitle(link);
+    const ids = titleToIDs.get(title) || titleToIDs.set(title, []).get(title);
+    ids.push(linkID);
+  });
+
+  /** @param {[string, string[]]} entry */
+  const toLinkProps = ([title, ids]) => {
+    return [{ title, id: ids[0] }].concat(
+      ids.slice(1).map((id, i) => ({ title: `(${i + 2})`, id }))
```
marcoscaceres: This one is a bit fancy... maybe add a comment explaining it?

src/core/dfn-panel.js
```diff
@@ -0,0 +1,161 @@
...
+    )}</li>`;
+
+  return hyperHTML`<ul>${[...titleToIDs.entries()].map(listItemToHTML)}</ul>`;
+}
+
+/** @param {HTMLAnchorElement} link */
+function getReferenceTitle(link) {
+  const section = link.closest("section");
+  if (!section) return null;
+  const heading = section.querySelector("h1, h2, h3, h4, h5");
```
marcoscaceres: No h6? 
```suggestion
  const heading = section.querySelector("h1, h2, h3, h4, h5, h6");
```

tests/spec/core/dfn-panel-spec.js
```diff
@@ -0,0 +1,168 @@
...
+        <p>[=many=] [=many=]</p>
+      </section>
+      <p>[= many =]</p>
+    </section>
+  `;
+  const ops = makeStandardOps(null, body);
+
+  describe("dfnPanel state", () => {
+    it("opens panel on dfn click", async () => {
+      const doc = await makeRSDoc(ops);
```
marcoscaceres: Hmm... I wonder if we can just create one `doc` and then just use that to check all the states?

marcoscaceres: Or is this because random order is fun? :::)

marcoscaceres: I guess after each test, you could put the dfn box back into its initial state?

sidvishnoi: Would've to close the popup before each test case though.... 
That should be easy though (came to my mind while writing this reply)

src/core/dfn-panel.js
sidvishnoi: went with `null`

src/core/dfn-panel.js
```diff
@@ -0,0 +1,161 @@
...
+    return "HIDE";
+  }
+  return "NOOP";
+}
+
+/** @param {HTMLElement} dfn */
+function createPanel(dfn) {
+  const { id } = dfn;
+  const href = `#${id}`;
+  const links = document.querySelectorAll(`a[href="${href}"]`);
```
marcoscaceres: Maybe just? 

```suggestion
  const links = document.querySelectorAll(`a[href="#${dfn.id}"]`);
```

src/core/dfn-panel.js
saschanaz: It actually makes the line longer, what would be the reason to prefer the latter?

tests/spec/core/dfn-panel-spec.js
sidvishnoi: Random order causing issues :sob:

src/core/dfn-panel.js
sidvishnoi: oh oh.. we can use CSS variables :smile:

assets/dfn-panel.css
```diff
@@ -0,0 +1,59 @@
...
+  text-indent: 0;
+}
+.dfn-panel > b {
+  display: block;
+}
+.dfn-panel a[href] {
+  color: black;
+}
+.dfn-panel a[href].self-link {
+  display: inline-block;
```
sidvishnoi: With `display: inline-block`, longer self-link text is slightly harder to select (block padding)
image

Without inline-block, small text wrap around unnecessarily
image

`white-space: nowrap` helps with small text:
image

but adds scroll on longer texts:
image

Anyone with better ideas?

src/core/dfn-panel.js
sidvishnoi: Added.

src/core/dfn-panel.js
sidvishnoi: Made some improvements in https://github.com/w3c/respec/pull/2626/commits/e562d736e3fa3492446988c76c146ef0da214531

assets/dfn-panel.css
marcoscaceres: Scrollable might be ok. People generally click, and then copy the URL. We should think of how to avoid showing the fragment entirely in the future. We could just have a “#”.

sidvishnoi: TC39 calls this label as just "Permalink". Maybe we can use that

assets/dfn-panel.css
```diff
@@ -0,0 +1,60 @@
...
+  list-style: inside;
+}
+.dfn-panel.minimized {
+  display: inline-block;
+  position: fixed;
+  left: 0.5em;
+  top: unset;
+  bottom: 2em;
+  margin: 0 auto;
+  max-width: calc(100vw - 1.5em - 0.4em - 0.5em);
```
saschanaz: Why this calculation? Some comment will be useful for future maintenance.

assets/dfn-panel.css
```diff
@@ -0,0 +1,60 @@
+/* dfn popup panel that list all local references to a dfn */
+dfn {
+  cursor: pointer;
+}
+.dfn-panel {
```
saschanaz: Should we add a newline between each rule?

assets/dfn-panel.css
sidvishnoi: I mostly copy-pasted style from bikeshed :sweat_smile:

assets/dfn-panel.css
sidvishnoi: Good point. For consistency.

assets/dfn-panel.css
saschanaz: Hmm, maybe just replace it with precalculated value if we don't know it? 🤔

sidvishnoi: Understood. 
```
0.75em from padding, 0.5em from left position
0.75em * 2 = 1.5em
0.5em * 2 = 1em
(total 2.5em - it was 2.4 for some reason)
```

saschanaz: Not sure why we should do left * 2, maybe the 0.4em was from border * 2.

sidvishnoi: Oh. good point!

sidvishnoi: Corrected.

marcoscaceres:
Could you check on mobile? Tried in Safari on my phone but the dfn panel doesn’t seem to want to apper?

sidvishnoi:
Worked for me in Firefox and Chrome on Android

src/core/dfn-panel.js
```diff
@@ -0,0 +1,164 @@
...
+  if (left + panelWidth > document.body.scrollWidth) {
+    // Reposition, because the panel is overflowing
+    left = dfnRect.left - (panelWidth + 5);
+    if (left < 0) {
+      left = dfnRect.left;
+      top = top + dfnRect.height;
+    }
+  }
+
+  panel.style.setProperty("--left", `${left}px`);
```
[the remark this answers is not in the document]
sidvishnoi: This way, I avoided the requirement to unset left/right with JS. See: https://github.com/w3c/respec/pull/2626#discussion_r357025402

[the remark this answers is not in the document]
saschanaz: Ah got it, so that `.dfn-panel.minimized` can get its own left value. 👍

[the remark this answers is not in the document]
saschanaz: Perhaps we can add a comment about it?

sidvishnoi:
This looks good to me to be checked-in. Any other opinions?

assets/dfn-panel.css
marcoscaceres: I'm happy with that.

tests/spec/core/dfn-panel-spec.js
marcoscaceres: ok, let's go with this for now... hopefully the random issue is not indicative of actual issues.

tests/spec/core/dfn-panel-spec.js
```diff
@@ -0,0 +1,168 @@
...
+      expect(doc.getElementById("dfn-panel")).toBeNull();
+      const dfn = doc.querySelector("dfn");
+      dfn.click();
+      const panel = doc.getElementById("dfn-panel");
+      expect(panel).not.toBeNull();
+      expect(panel.classList.contains("minimized")).not.toBe(true);
```
saschanaz:
```suggestion
      expect(panel.classList).not.toContain("minimized")
```

tests/spec/core/dfn-panel-spec.js
```diff
@@ -0,0 +1,168 @@
...
+
+      panel.click();
+      expect(doc.getElementById("dfn-panel")).toBe(panel);
+    });
+
+    it("minimizes open panel on reference click", async () => {
+      const doc = await makeRSDoc(ops);
+      doc.querySelector("dfn").click();
+      const panel = doc.getElementById("dfn-panel");
+      expect(panel.classList.contains("minimized")).toBe(false);
```
saschanaz:
```suggestion
      expect(panel.classList).not.toContain("minimized");
```

tests/spec/core/dfn-panel-spec.js
```diff
@@ -0,0 +1,168 @@
...
+      panel.querySelector("ul a").click();
+      expect(panel.classList.contains("minimized")).toBe(true);
```
saschanaz:
```suggestion
      expect(panel.classList).toContain("minimized");
```

tests/spec/core/dfn-panel-spec.js
```diff
@@ -0,0 +1,168 @@
...
+    });
+
+    it("closes minimized panel on panel click", async () => {
+      const doc = await makeRSDoc(ops);
+      doc.querySelector("dfn").click();
+      const panel = doc.getElementById("dfn-panel");
+      panel.querySelector("ul a").click();
+      expect(panel.classList.contains("minimized")).toBe(true);
```
saschanaz:
```suggestion
      expect(panel.classList).toContain("minimized");
```

tests/spec/core/dfn-panel-spec.js
sidvishnoi: Got into habit of using `contains` as:
``` js
// el.classList == "foo-minimized-bar"
expect(el.classList).toContain("minimized") // pass
expect(el.classList.contains("minimized")).toBe(true); // fail
```

Should I change here?

src/core/dfn-panel.js
```diff
@@ -0,0 +1,168 @@
...
+      }
+    }
+  });
+}
+
+/** @param {HTMLElement} clickTarget */
+function deriveAction(clickTarget) {
+  const hitALink = !!clickTarget.closest("a");
+  if (clickTarget.closest("dfn")) {
+    return hitALink ? null : "show";
```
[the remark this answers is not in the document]
sidvishnoi: We have a plan for triggering this on hover, as an enhancement (similar to what Wikipedia does)

assets/dfn-panel.css
```diff
@@ -0,0 +1,68 @@
...
+  font: small Helvetica Neue, sans-serif, Droid Sans Fallback;
+  background: #dddddd;
+  color: black;
+  border: outset 0.2em;
+}
+
+.dfn-panel * {
+  margin: 0;
+  padding: 0;
+  text-indent: 0;
```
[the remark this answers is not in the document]
sidvishnoi: Probably for the `<ul>`, `<li>`. Do you want me to inspect?

assets/dfn-panel.css
```diff
@@ -0,0 +1,68 @@
...
+
+.dfn-panel ul {
+  padding: 0;
+}
+
+.dfn-panel li {
+  list-style: inside;
+}
+
+.dfn-panel.minimized {
```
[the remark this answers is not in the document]
sidvishnoi: Bikeshed calls it "activated". "docked" sounds better to me indeed. :+1:

tests/spec/core/dfn-panel-spec.js
saschanaz: Couldn't confirm such behavior. Could you try it again? https://jsfiddle.net/saschanaz/3vt7ew15/1/

image

sidvishnoi: Whoa. Thanks for the test! Will update

assets/dfn-panel.css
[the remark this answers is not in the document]
saschanaz: Disabling those two in the inspector didn't introduce any visual difference, I think we can skip them 👀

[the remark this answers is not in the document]
sidvishnoi: I had added a `.dfn-panel ul { padding: 0 }` later in same file. :sweat_smile:

src/core/dfn-panel.js
```diff
@@ -0,0 +1,166 @@
...
+
+  /** @type {HTMLElement} */
+  const panel = hyperHTML`
+    <aside class="dfn-panel" id="dfn-panel">
+      <b><a class="self-link" href="${href}">Permalink</a></b>
+      <b>Referenced in:</b>
+      ${referencesToHTML(id, links)}
+    </aside>
+  `;
+  return document.body.appendChild(panel);
```
saschanaz: Should this line be in displayPanel?

src/core/dfn-panel.js
```diff
@@ -0,0 +1,166 @@
...
+ * @param {HTMLElement} dfn
+ * @param {HTMLElement} panel
+ */
+function displayPanel(dfn, panel) {
+  const dfnRect = dfn.getBoundingClientRect();
+  const panelRect = panel.getBoundingClientRect();
+  const panelWidth = panelRect.right - panelRect.left;
+
+  let top = window.scrollY + dfnRect.top;
+  let left = dfnRect.left + dfnRect.width + 5;
```
saschanaz: What does the 5 mean here?

src/core/dfn-panel.js
```diff
@@ -0,0 +1,166 @@
...
+  if (left + panelWidth > document.body.scrollWidth) {
+    // Reposition, because the panel is overflowing
+    left = dfnRect.left - (panelWidth + 5);
+    if (left < 0) {
+      left = dfnRect.left;
+      top = top + dfnRect.height;
```
saschanaz:
```suggestion
      top += dfnRect.height;
```

src/core/dfn-panel.js
```diff
@@ -0,0 +1,166 @@
...
+  links.forEach((link, i) => {
+    const linkID = link.id || `ref-for-${id}-${i + 1}`;
+    if (!link.id) link.id = linkID;
+    const title = getReferenceTitle(link);
+    const ids = titleToIDs.get(title) || titleToIDs.set(title, []).get(title);
+    ids.push(linkID);
+  });
+
+  /**
+   * Takes an "entry" from `titleToIDs`, and converts it to a list that is
```
saschanaz: Maybe move some sentences to `@param` and `@return` for readability?

src/core/dfn-panel.js
```diff
@@ -0,0 +1,166 @@
...
+  if (left + panelWidth > document.body.scrollWidth) {
+    // Reposition, because the panel is overflowing
+    left = dfnRect.left - (panelWidth + 5);
+    if (left < 0) {
+      left = dfnRect.left;
+      top = top + dfnRect.height;
+    }
+  }
+
+  // "docked" state can independently set its own position in CSS.
```
saschanaz:
```suggestion
  // Allows ".docked" rule to override the position, unlike `style.left = left`.
```

IMO better, what do you think?

src/core/dfn-panel.js
sidvishnoi: It's a little padding so that the popup isn't just at the end of `dfn` text. Not sure if this needs a comment as it's clear that it's a position?

src/core/dfn-panel.js
sidvishnoi: Makes sense.

src/core/dfn-panel.js
saschanaz: Ah, understood. I thought it might be related to some existing margins/paddings 👀

sidvishnoi:
Looks like it's time for a release!

PR #2634, opened 2019-12-14
chore(tests): skip building profile in dev env

sidvishnoi:
TODO: remove build step from tools/dev-server.js, which should make tools/dev-server.js essentially redundant.

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -46,6 +46,10 @@ module.exports = function(config) {
         included: false,
       },
       {
+        pattern: "profiles/**/*.*",
+        included: false,
+      },
+      {
         pattern: "node_modules/idb/**/*.js",
         included: false,
       },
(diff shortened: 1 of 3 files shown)


karma.conf.js
```diff
@@ -164,6 +169,11 @@ module.exports = function(config) {
       args: ["--grep", config.grep || ""],
     },
   };
+
+  if (process.env.GITHUB_WORKFLOW) {
+    options.files.unshift("tests/use_bundled_profile.js");
+  }
+
   if (process.env.TRAVIS) {
```
saschanaz: We don't need this anymore, right?

sidvishnoi: Yep. Will remove in a cleanup (along with `detectBrowsers` - unless someone uses it)

saschanaz: Maybe @marcoscaceres ?

marcoscaceres: Should be ok to remove.

saschanaz: Yay, we can finally remove mentioning of IE

sidvishnoi:
Closing as I think it's better if we do not diverge between CI and local env. Might get weird "works on my machine" bugs. Also, it doesn't save a lot of time either - but adds its complexity.

PR #2636, opened 2019-12-15
refactor: replaced joinAnd with htmlJoinAnd

SuyashSalampuria:
Replaces joinAnd with htmlJoinAnd in headers.js and replalced [keywords] with just keywords


SuyashSalampuria:
@saschanaz Sorry for disturbing you so much.
Can you please help me why headless tests are failing even though the karma tests are passing
:(

saschanaz:
Oh wow. It means that it currently unexpectedly creates invalid self-referencing anchor if there is no `conf.wgPatentURI` value set.

@marcoscaceres What would be the expected behavior if without `conf.wgPatentURI`? An error, or just skipping the patent part?

marcoscaceres:
Hmm... I think it depends if the document is on the recommendation track. If it's on the recommendation track, the it must include a patent URL.

PR #2639, opened 2019-12-15
chore(.github): point OpenCollective in Sponsors

saschanaz:
This way we can get a nice Sponsor button at the top of the GitHub page, [as parceljs has](https://github.com/parcel-bundler/parcel/blob/v2/.github/FUNDING.yml). 

image


saschanaz:
Yay, got it immediately!

PR #2640, opened 2019-12-15
adding informative class to non-conformance sections

ridhishjain:
#2580
if section id is not "conformance", then informative class is added to the body.

### DIFF
--- a/src/core/inlines.js
+++ b/src/core/inlines.js
@@ -182,6 +182,7 @@ export function run(conf) {
   document.normalize();
   if (!document.querySelector("section#conformance")) {
     // make the document informative
+    // throw an error like "Conformance not found"
     document.body.classList.add("informative");
   }
   conf.normativeReferences = new InsensitiveStringSet();
--- a/src/w3c/conformance.js
+++ b/src/w3c/conformance.js
@@ -55,6 +55,9 @@ export function run(conf) {
   const conformance = document.querySelector("section#conformance");
   if (conformance) {
     processConformance(conformance, conf);
+  } else {
+    document.body.classList.add("Informative");
+    // throw an error like "Conformance not found"
   }
   // Added message for legacy compat with Aria specs
   // See https://github.com/w3c/respec/issues/793
--- a/src/w3c/conformance.js
+++ b/src/w3c/conformance.js
@@ -55,9 +55,14 @@ export function run(conf) {
   const conformance = document.querySelector("section#conformance");
   if (conformance) {
     processConformance(conformance, conf);
+    if (document.body.classList.contains("informative")) {
+      // case when both section#conformance and .informative is present
+      // to be dealt with
+    }
   } else {
-    document.body.classList.add("Informative");
-    // throw an error like "Conformance not found"
+    if (!document.body.classList.contains("informative")) {
+      pub("Informative class not found!", name);
+    }
   }
   // Added message for legacy compat with Aria specs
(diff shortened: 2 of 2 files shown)


ridhishjain:
What I understood necessary doing:
1. If the conformance section is missing, ReSpec should report an error like "No conformance section found" along with changing the class of the body from normative to informative.

Am I right?
If I am, I added the informative class in `src/w3c/conformace.js`. Only dealing with the error part is remaining, @sidvishnoi, a little guidance regarding that would do.

saschanaz:
From https://github.com/w3c/respec/issues/2580#issuecomment-556977843:

>That is, instead of silently assuming that the content is informative when the conformance section is missing, ReSpec could actually report an error such as "No conformance section found for this normative specification. If the spec is meant to be informative-only, please add an informative class to <body> (or set noRecTrack)" or something like that.

I think we should:

* Remove the current automatic addition of `informative`
* Add a logic that:
   1. Check if the doc have `.informative` class in body
   2. If the doc lacks both the class and `section#conformance`, then throw error via `pub("error")`
   3. ?? What if there is `.informative` and still we have `section#conformance`? @marcoscaceres, thoughts?

ridhishjain:
> From [#2580 (comment)](https://github.com/w3c/respec/issues/2580#issuecomment-556977843):
> 
> > That is, instead of silently assuming that the content is informative when the conformance section is missing, ReSpec could actually report an error such as "No conformance section found for this normative specification. If the spec is meant to be informative-only, please add an informative class to  (or set noRecTrack)" or something like that.
> 
> I think we should:
> 
> * Remove the current automatic addition of `informative`
> * Add a logic that:
>   
>   1. Check if the doc have `.informative` class in body
>   2. If the doc lacks both the class and `section#conformance`, then throw error via `pub("error")`
>   3. ?? What if there is `.informative` and still we have `section#conformance`? @marcoscaceres, thoughts?

@saschanaz, all I'm saying is along with throwing error, we should also add `.informative` class in the body if it's not present already
In case if both `.informative` and `section#conformance` are present, why not just change the `.informative` to`.normative`

saschanaz:
>we should also add .informative class in the body if it's not present already

Not sure, because we'll throw an error anyway.

>In case if both .informative and section#conformance are present, why not just change the .informative to.normative

The issue is about doing things implicitly, so we shouldn't do that implicitly either.

sidvishnoi:
> What if there is .informative and still we have section#conformance?

Assuming that `.informative` is on body, that's an error - a fatal one - stop processing altogether, as author's intention is unclear.

marcoscaceres:
I think what we have at the moment is fine. All we want is to show an *warning* if RFC2119 keywords are used. Conformance relates specifically to RFC2119 keywords and how they are to be interpreted. Another thing we can check for is if a reference is added as `[[!normative]]`, which is indicative that the author is intending the document to be normative.

saschanaz:
Maybe src/inlines.js then?

ridhishjain:
@marcoscaceres I am quite confused here 😓, can u guide me a little more regarding what exactly to do?

marcoscaceres:
Apologies this is confusing @ridhishjain, as it really requires understanding how specs work and what an author's intent is. I'd suggest leaving this one, as it's a tricky one to get right.

sidvishnoi:
Fixed in https://github.com/w3c/respec/pull/2659

PR #2641, opened 2019-12-16
refactor(core/markdown): drop paragraph removal

saschanaz:
...and a bit of additional refactoring.

I'm not sure empty paragraphs still happen in the latest version of marked, at least it doesn't happen in wicg/serial. I don't think we should actively workaround marked bugs in ReSpec anyway...

### DIFF
--- a/src/core/markdown.js
+++ b/src/core/markdown.js
@@ -221,14 +221,23 @@ export function markdownToHtml(text) {
   return result;
 }
 
-function processElements(selector) {
+/**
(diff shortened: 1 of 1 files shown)


sidvishnoi:
> I'm not sure empty paragraphs still happen in the latest version of marked

I didn't see any at https://marked.js.org/demo/

marcoscaceres:
Be amazing if this is finally fixed... it was a nightmare.

PR #2642, opened 2019-12-16
enhancement(core/inlines): add `data-link-type="biblio"` to inline citations

### LINKED ISSUE #2492: Biblio links should have a `data-link-type="biblio"`

ridhishjain:
fixes #2492 
reviews @sidvishnoi @marcoscaceres

### DIFF
--- a/tests/spec/core/inlines-spec.js
+++ b/tests/spec/core/inlines-spec.js
@@ -234,10 +234,10 @@ describe("Core - Inlines", () => {
           [= \`link\` element =] and [= some \`Coded\` thing =]
         </p>
         <p id="multiline">
-        [=environment 
+        [=environment
             settings
-          object / 
-          responsible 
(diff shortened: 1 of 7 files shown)


tests/spec/core/inlines-spec.js
```diff
@@ -435,4 +435,19 @@ describe("Core - Inlines", () => {
       "#dom-referrerpolicy-no-referrer"
     );
   });
+
+  it("checks data-link-type of inline citations", async () => {
```
sidvishnoi: Might be better to add these tests within: https://github.com/w3c/respec/blob/6a18cf58d1e9853e92344753983ec9502d09caeb/tests/spec/core/inlines-spec.js#L38-L47

A single test like: `expect(links[0].dataset.linkType).toBe("biblio")` should be enough.

PR #2643, opened 2019-12-17
fix(core/markdown): indent idl correctly after a list

saschanaz:
Fixing a regression found in wicg/serial:

image

Lack of newlines before HTML end tags caused this.

### DIFF
--- a/src/core/markdown.js
+++ b/src/core/markdown.js
@@ -246,10 +246,10 @@ function enableBlockLevelMarkdown(element, selector) {
   for (const element of elements) {
     // Double newlines are needed to be parsed as Markdown
     const { innerHTML } = element;
-    if (!innerHTML.match(/^\n\s*\n/)) {
+    if (!/^\n\s*\n/.test(innerHTML)) {
       element.prepend("\n\n");
     }
-    if (!innerHTML.match(/\n\s*\n$/)) {
+    if (!/\n\s*\n$/.test(innerHTML)) {
       element.append("\n\n");
     }
   }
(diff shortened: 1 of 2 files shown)


src/core/markdown.js
```diff
@@ -249,9 +249,13 @@ function enableBlockLevelMarkdown(element, selector) {
   const elements = element.querySelectorAll(selector);
   for (const element of elements) {
     // Double newlines are needed to be parsed as Markdown
-    if (!element.innerHTML.match(/^\n\s*\n/)) {
+    const { innerHTML } = element;
+    if (!innerHTML.match(/^\n\s*\n/)) {
```
sidvishnoi: Nit: Use `re.test()` as we're not interested in match results?

tests/spec/core/markdown-spec.js
```diff
@@ -417,4 +419,25 @@ describe("Core - Markdown", () => {
...
   });
+
+  describe("Markdown-inside-block backward compatibility", () => {
+    it("parses indented <pre> after a list", async () => {
+      const idl = `dictionary Indented {\n  any shouldBeIndented();\n};`;
+      const body = `
+        <section>
+        1. I lack double newlines between HTML block
+        </section>
+        <pre class="idl" id="pre">
```
sidvishnoi: Just to confirm, triple backtick blocks work fine?

saschanaz: Yup, as triple backtick requires adding an extra newline after any HTML tags.

saschanaz:
Removed trimRight as we do need the newlines here.

PR #2644, opened 2019-12-17
refactor(w3c/templates/headers): contain its own l10n strings

ridhishjain:
refactor issue #2084 (templates/header.js)
test is still to be added


ridhishjain:
The check is failing because `FormerEditor:` is absent in `es` object in `core/l10n.js`. 
Screenshot (61)_LI
`l10n` object is overridden in `templates/headers.js` by copying from `core/l10.js`......therefore `l10n.es.FormerEditor` is missing....which is why the check is failing

sidvishnoi:
@ridhishjain Can you incorporate changes from #2646 ([example](https://github.com/w3c/respec/pull/2646/files#diff-7867d74ee4f786f89e7251b1f0abdea3))

ridhishjain:
@sidvishnoi , I don't think, separate test for `editors`, `former editors` and `authors` need to be created as they are already included in `describe("editors", ...`

tests/spec/w3c/headers-spec.js
```diff
@@ -1605,4 +1606,22 @@ describe("W3C — Headers", () => {
...
+      config: makeBasicConfig(),
+      htmlAttrs: {
+        lang: "nl",
+      },
+    };
+    const doc = await makeRSDoc(ops);
+
+    const terms = doc.querySelectorAll("dt");
+    expect(terms[0].textContent).toBe("Deze versie:");
+    expect(terms[0].nextElementSibling.localName).toBe("dd");
```
sidvishnoi: Lets remove these `"dd"` tests.

sidvishnoi:
Looks good. Just a nit: lets remove those `"dd"` tests.

src/w3c/templates/headers.js
```diff
@@ -13,6 +14,66 @@ const legalDisclaimer =
...
+    authors: "Authors:",
+    editor: "Editor:",
+    editors: "Editors:",
+    former_editor: "Former editor:",
+    former_editors: "Former editors:",
+    latest_editors_draft: "Latest editor's draft:",
+    latest_published_version: "Latest published version:",
+    this_version: "This version:",
+  },
+  ko: {
```
marcoscaceres: @saschanaz any chance you can provide the missing ones? 🙇

src/w3c/templates/headers.js
```diff
@@ -13,6 +14,66 @@ const legalDisclaimer =
...
+    author: "저자:",
+    authors: "저자:",
```
saschanaz:
```suggestion
    authors: "저자:",
    editor: "편집자:",
    editors: "편집자:",
    former_editor: "이전 편집자:",
    former_editors: "이전 편집자:",
    latest_editors_draft: "최신 편집 초안:",
```

saschanaz:
This should also remove existing code in l10n.js, right?

sidvishnoi:
> This should also remove existing code in l10n.js, right?

Ideally yes. But we can do it end the end when all modules are done.

saschanaz:
>Ideally yes. But we can do it end the end when all modules are done.

We can, but why? Incremental removal should be safer.

ridhishjain:
@sidvishnoi , removal of localization strings from `src/core/l10n.js` should be done after adding them into `cgbg-headers.js` also (error generation otherwise for each string removal). It is targeted in my next PR (which is for refactoring `cgbg-headers.js`)

sidvishnoi:
@ridhishjain As headers and cgbg-headers are related, I would prefer if we do that refactoring within this PR.

src/w3c/templates/headers.js
```diff
@@ -13,6 +14,71 @@ const legalDisclaimer =
...
+    authors: "Autores:",
+    editor: "Editor:",
+    editors: "Editores:",
+    latest_editors_draft: "Borrador de editor mas reciente:",
+    latest_published_version: "Versión publicada mas reciente:",
+    this_version: "Ésta versión:",
+  },
+};
+
+const l10n = getIntlData(localizationStrings);
```
sidvishnoi: Lets export this and reuse it in `src/w3c/templates/cgbg-headers.js`
```suggestion
export const l10n = getIntlData(localizationStrings);
```

PR #2646, opened 2019-12-17
fix(core/l10n): fallback to 'en' for missing l10n keys

sidvishnoi:
We were doing l10n migration to modules wrong :fire:  
We were assuming if we have some translations for a language, we have all translations for that language.
This was taken into account when we had `core/l10n`, where we added default as `en` by Object spread.

Consider:
``` js
const localizationStrings = {
  en: { key1, key2, key3 },
  ja: { key1 } // we assumed `lang` as "ja" even when key2, key3 are undefined 
}
```

Also, a little refactor as we don't have to `const lang = defaultLang in l10n ? defaultLang : "en";` in every module.

Alternate approach considered: `{...en, ...[lang]}`. Decided to go against it as it creates a new clone (not sure how bad it could be though).
Proxy approach also allows us to show a missing translation/invalid key warning if we want to.

### DIFF
--- a/src/core/data-tests.js
+++ b/src/core/data-tests.js
@@ -9,11 +9,11 @@
  *
  * Docs: https://github.com/w3c/respec/wiki/data-tests
  */
-import { lang as defaultLang } from "./l10n.js";
+import { getIntlData } from "./l10n.js";
 import { hyperHTML } from "./import-maps.js";
 import { pub } from "./pubsubhub.js";
 import { showInlineWarning } from "./utils.js";
-const l10n = {
+const localizationStrings = {
   en: {
     missing_test_suite_uri:
       "Found tests in your spec, but missing '" +
@@ -23,9 +23,9 @@ const l10n = {
   },
 };
 
-export const name = "core/data-tests";
+const l10n = getIntlData(localizationStrings);
 
-const lang = defaultLang in l10n ? defaultLang : "en";
+export const name = "core/data-tests";
 
 function toListItem(href) {
   const emojiList = [];
@@ -82,7 +82,7 @@ export function run(conf) {
     return;
   }
   if (!conf.testSuiteURI) {
-    pub("error", l10n[lang].missing_test_suite_uri);
+    pub("error", l10n.missing_test_suite_uri);
     return;
   }
 
@@ -105,7 +105,7 @@ function toTestURLs(tests, testSuiteURI) {
       try {
         return new URL(test, testSuiteURI).href;
       } catch {
-        pub("warn", `${l10n[lang].bad_uri}: ${test}`);
+        pub("warn", `Bad URI: ${test}`);
       }
     })
     .filter(href => href);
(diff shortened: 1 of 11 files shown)


src/core/data-tests.js
```diff
@@ -105,7 +105,7 @@ function toTestURLs(tests, testSuiteURI) {
       try {
         return new URL(test, testSuiteURI).href;
       } catch {
-        pub("warn", `${l10n[lang].bad_uri}: ${test}`);
+        pub("warn", `Bad URI: ${test}`);
```
sidvishnoi: This was a missing l10n string in all languages :tada: 
Kinda caught by TypeScript (not lint error sadly, but shows up as "any" instead of "string" and does't appear in VSCode's suggestions).

marcoscaceres:
Fancy proxy is fancy!

src/core/l10n.js
```diff
@@ -171,6 +171,22 @@ l10n["zh-cn"] = l10n.zh;
 
 export const lang = html && html.lang in l10n ? html.lang : "en";
 
+/**
+ * @typedef {Record<string, Record<string, string|Function>>} LanguageData
+ * @type {<T extends LanguageData>(obj: T) => T[keyof T]}
```
saschanaz: Can we use `@template`, `@param` and `@return` instead?

src/core/l10n.js
```diff
@@ -171,6 +171,22 @@ l10n["zh-cn"] = l10n.zh;
...
+ */
+export function getIntlData(languageData) {
+  // Proxy return type is a known bug:
+  // https://github.com/Microsoft/TypeScript/issues/20846
+  // @ts-ignore
+  return new Proxy(languageData, {
+    /** @param {string} key */
+    get(data, key) {
+      return data[lang][key] || data.en[key] || key;
```
saschanaz: What would `key` mean? Should it throw or just undefined?

marcoscaceres: Hmm... missed that. If key is not found it should probably throw, so we can fix that.

src/core/l10n.js
sidvishnoi: @saschanaz Can you help me with this one? Specifically the `T extends` part?

saschanaz: AFAIK:

```suggestion
 * @template {Languagedata} T
 * @param {T} obj
 * @return {T[keyof T]}
```

sidvishnoi: This is same as what I came up with, but:
image

sidvishnoi: Looks like narrowing the type isn't supported: https://github.com/Microsoft/TypeScript/issues/1178#issuecomment-348755243

saschanaz:
```suggestion
 */
/**
 * @template {LanguageData} T
 * @param {T} localizationStrings
 * @return {T[keyof T]}
```

No idea why but splitting works 🤣

sidvishnoi: :rofl: :shipit:

PR #2652, opened 2019-12-22
remove raw string interpolations

### LINKED ISSUE #2551: Remove raw HTML string interpolation

SuyashSalampuria:
Fixes #2551 . Removed most of the raw string interpolations in src/core


sidvishnoi:
These were supposed to stay as strings, mate. HyperHTML is for HTML 🙂

SuyashSalampuria:
Oh! I am sorry for this

PR #2654, opened 2019-12-23
BREAKING CHANGE: Remove core/requirements

SuyashSalampuria:
Since requirements.js is no longer used, it has been removed!!


profiles/w3c.js
```diff
@@ -28,7 +28,6 @@ const modules = [
   import("../src/core/pluralize.js"),
   import("../src/core/examples.js"),
   import("../src/core/issues-notes.js"),
-  import("../src/core/requirements.js"),
```
saschanaz: Could you also remove this from profiles/genovum and profiles/w3c-common?

SuyashSalampuria: I will remove it for sure. 
Is it the only this making test case fail
Because I am getting error
Screenshot from 2019-12-23 15-57-04
There is no error till your commit of **_use reindent() instead_**
I am still guessing  what happened there.:(
I will fix it out by tonight for sure

saschanaz: Maybe you'll want to do `npm run build:w3c` again?

PR #2655, opened 2019-12-23
test(core/markdown): <pre> should not parse as markdown

saschanaz:
Our wiki currently says:

> ## Known issues
> 
> In some cases, the markdown processor gets confused. For example, given: 
> 
> ```HTML
> <pre class="example">
> 1 * 2 * 3;
> </pre>
> ``` 
> 
> The "`*`"s get converted to an `<em>`. Obviously, that's not ideal. A workaround is to wrap the "*" in back-ticks, like this:
> 
> ```HTML
> <pre class="example">
> 1 `*` 2 `*` 3;
> </pre>
> ```

This is not true anymore, so this PR makes sure that won't happen.

### DIFF
--- a/tests/spec/core/markdown-spec.js
+++ b/tests/spec/core/markdown-spec.js
@@ -330,6 +330,22 @@ function getAnswer() {
     expect(pre.textContent).toBe(idl);
   });
 
+  it("does not parse texts inside <pre> as markdown", async () => {
+    const code = `1 * 2 * 3;`;
+    const body = `
+      <pre class="example">
+      ${code}
+      </pre>
+    `;
+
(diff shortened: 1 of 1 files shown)


tests/spec/core/markdown-spec.js
```diff
@@ -330,6 +330,22 @@ function getAnswer() {
     expect(pre.textContent).toBe(idl);
   });
 
+  it("does not parse texts inside <pre> as markdown", async () => {
+    const code = `1 * 2 * 3;`;
```
marcoscaceres: Any reason to not inline this?

saschanaz: This line is reused in line 346, IMO better as a separate value 👀

PR #2657, opened 2019-12-24
BREAKING CHANGE: remove refNote support

### LINKED ISSUE #817: Deprecate refNote
Where it is possible to add markup into a document, we should deprecate things link refNote... instead, we should check for special element+id combos, and if those elements are present, use those instead.

marcoscaceres:
Closes #817 

As only one spec uses `refNote`, and it's a potential xss venerability, this removes it. 

This patch adds support for and explicit `<section id="reference">`, where spec editors can now add whatever they like:

```HTML
<section>
  <h2>Custom header</h2>
  <p>Custom content...</p>
  <!-- generated informative and normative refs sections go here... localized --> 
</section>
```


marcoscaceres:
Sent PR to affected spec: https://github.com/w3c/webmediaapi/pull/240

PR #2658, opened 2019-12-25
fix(core/ui): update aria-label as errors/warnings are added

marcoscaceres:


sidvishnoi:
Wouldn't this result in too many error announcements, which would be annoying? Maybe we should announce errors collectively at the end (n Respec errors, m Respec warnings)

marcoscaceres:
I'm not sure, tbh. We are updating the label, but we are not asking the AT to announce anything? I don't imagine it randomly starts speaking if a label changes - but I'm not sure.

sidvishnoi:
I'll check just to be sure 🙂

sidvishnoi:
Checked with NVDA. It wasn't randomly announced like we were thinking. 
On hovering over error count though, it just announced the number, without any mention of what number it is. Focus worked as intended. Probably an NVDA issue

PR #2659, opened 2019-12-25
feat(w3c/conformance): allow override of conformance section

### LINKED ISSUE #273: WG dependent Conformance templates
The conformance section of the CSS WG is significant longer than the predefined one. 

Is it for example possible to upload the template for the CSS WG, which is just chosen when one of the WG strings in the meta data section is `CSS Working Group` or `SVG Working Group` for instance?

### LINKED ISSUE #2580: Why do some specs end up with a class="informative" on body?

marcoscaceres:
closes #273 
closes #2580

### DIFF
--- a/tests/spec/w3c/conformance-spec.js
+++ b/tests/spec/w3c/conformance-spec.js
@@ -5,6 +5,7 @@ import {
   makeBasicConfig,
   makeDefaultBody,
   makeRSDoc,
+  makeStandardOps,
 } from "../SpecHelper.js";
 
 describe("W3C — Conformance", () => {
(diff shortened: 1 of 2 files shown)


tests/spec/w3c/conformance-spec.js
```diff
@@ -62,4 +63,29 @@ describe("W3C — Conformance", () => {
...
     expect(doc.querySelectorAll("#conformance .rfc2119").length).toBe(0);
   });
+
+  it("allows conformance section to be completely overridden via .override css class", async () => {
+    const body = `
+      <section>
+        <h2>Plain section</h2>
+        <p>MUST SHOULD MAY.</p>
+      </section>
+      <section id="conformance" class="override">
```
saschanaz: Too late but shouldn't this be more like an attribute, like `data-override`?

marcoscaceres: I prefer using CSS classes for none “data” things, tbh.

PR #2661, opened 2019-12-25
feat: allow overriding copyright using HTML

### LINKED ISSUE #263: Allow overriding copyright
We currently have additionalCopyrightHolders for that, but 1) it's a horrible name (due to the fact that we're reusing an option designed for something else), and 2) if people switch the specStatus they might forget to remove it, in which case it will have bad side-effects.

We need a "copyright" setting that only applies to unofficial drafts.

### LINKED ISSUE #2243: Making copyright configurable for non-W3C specs
## Important info

* URL to affected spec: N/A
* ReSpec version: 24.10.2

## Description of problem

_Apologies in advance if there's already a solution for this problem_.

If used for non-W3C specs, the only way I know to modify the copyright holder(s) is to use `overrideCopyright`.

This, however, has two main issues:
1. The content of the `overrideCopyright` replaces both copyright notice and licence
2. The copyright notice is static, so the start / end year for the copyright needs to be hard-coded

A possible option could be to add a config param `copyright`, with the following properties:
- `url` (optional): the URL pointing to a copyright page
- `copyrightHolder` (mandatory): an array of objects with the following properties:
  - `name` (mandatory): the name of the copyright holder
  - `url` (optional): the URL pointing to the Web site of the copyright holder

marcoscaceres:
This allows editors to provide their own copyright markup via:

```HTML
<p class="copyright">
  My custom copyright.  
</p>
```

Closes #263
Closes #2243

### DIFF
--- a/src/w3c/templates/cgbg-headers.js
+++ b/src/w3c/templates/cgbg-headers.js
@@ -114,7 +114,7 @@ export default conf => {
           `
         : ""}
       ${existingCopyright
-        ? [existingCopyright]
+        ? existingCopyright
         : html`
             <p class="copyright">
               <a href="https://www.w3.org/Consortium/Legal/ipr-notice#Copyright"
--- a/src/w3c/templates/cgbg-headers.js
+++ b/src/w3c/templates/cgbg-headers.js
@@ -5,6 +5,10 @@ import showLogo from "./show-logo.js";
 import showPeople from "./show-people.js";
 
 export default conf => {
+  const existingCopyright = document.querySelector(".copyright");
+  if (existingCopyright) {
+    existingCopyright.remove();
+  }
   return html`
     <div class="head">
       ${conf.logos.map(showLogo)}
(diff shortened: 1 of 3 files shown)


src/w3c/templates/cgbg-headers.js
```diff
@@ -109,41 +113,48 @@ export default conf => {
...
-          : html`
-              <a href="https://www.w3.org/community/about/agreements/cla/"
-                >W3C Community Contributor License Agreement (CLA)</a
-              >. A human-readable
-              <a href="https://www.w3.org/community/about/agreements/cla-deed/"
-                >summary</a
+      ${existingCopyright
+        ? [existingCopyright]
+        : html`
+            <p class="copyright">
```
marcoscaceres: This is just a whitespace change...

src/w3c/templates/cgbg-headers.js
```diff
@@ -109,41 +113,48 @@ export default conf => {
...
-              is available.
-            `
-          : html`
-              <a href="https://www.w3.org/community/about/agreements/cla/"
-                >W3C Community Contributor License Agreement (CLA)</a
-              >. A human-readable
-              <a href="https://www.w3.org/community/about/agreements/cla-deed/"
-                >summary</a
+      ${existingCopyright
+        ? [existingCopyright]
```
sidvishnoi: Can we use it without []?

marcoscaceres: Yes, oops. Forgot about that.

PR #2663, opened 2019-12-26
feat(core/webidl): allow linking to WebIDL blocks

### LINKED ISSUE #444: Link button in IDL definitions
Is there anyway we can add link buttons to IDL definitions to get the url to the IDL faster? Every time I want to reference the IDL I'm looking at, I have to open the inspector copy the id and paste it into the address bar.

marcoscaceres:
closes #444


src/core/utils.js
```diff
@@ -480,6 +494,18 @@ export function htmlJoinAnd(array, mapper = item => item) {
   }
 }
 
+/**
+ * Creates and sets an ID to an element (elem) by hashing the text content.
+ *
+ * @param {HTMLElement} elem element to hash from
+ * @param {String} prefix prefix to append to the generated id
```
sidvishnoi:
```suggestion
 * @param {String} prefix prefix to prepend to the generated id
```

src/core/webidl-clipboard.js
```diff
@@ -30,12 +30,15 @@ async function createButton() {
 export async function run() {
   // This button serves a prototype that we clone as needed.
   const copyButton = await copyButtonPromise;
-  for (const pre of document.querySelectorAll("pre.idl")) {
+  for (const idlHeader of document.querySelectorAll("div.idlHeader")) {
     const button = copyButton.cloneNode(true);
+    const { textContent: idlText } = idlHeader
```
sidvishnoi: Can we use `idlHeader.nextElementSibling`?
Might be better to iterate over `pre.idl` and then find `idlHeader` and `span` though

src/core/webidl.js
```diff
@@ -315,7 +316,11 @@ function renderWebIDL(idlElement, index) {
   idlElement.classList.add("def", "idl");
   const html = webidl2.write(parse, { templates });
   const render = hyperHTML.bind(idlElement);
-  render`${html}`;
+  const anchor = hyperHTML`<a>WebIDL</a>`;
+  addHashId(idlElement, "webidl");
+  anchor.href = `#${idlElement.id}`;
+  anchor.classList.add("self-link");
```
sidvishnoi: Maybe we should do this like:
``` js
const anchor = hyperHTML`<a class="self-link" href="#${idlElement.id}">WebIDL</a>`;
```

sidvishnoi:
I think IDL index should not have these buttons as interfaces are clickable here anyway:
image

marcoscaceres:
Yeah, that’s pretty ugly. Will fix.

src/core/webidl-clipboard.js
marcoscaceres: Just realized there could be multiple `<span>`s 

```
<pre>
dict Foo {};
interface Y {};
</pre>
```

src/core/utils.js
```diff
@@ -8,6 +8,20 @@ import { pub } from "./pubsubhub.js";
 export const name = "core/utils";
 
 const dashes = /-/g;
+/**
+ * Hashes a string from char code.
+ * https://stackoverflow.com/questions/7616461/generate-a-hash-from-string-in-javascript
+ *
+ * @param {String} text
+ */
+function hashString(text) {
```
sidvishnoi: This can return negative numbers also. Not a problem, but a note.

src/core/webidl-index.js
```diff
@@ -63,5 +63,8 @@ export function run() {
     });
   // Remove duplicate IDs
   pre.querySelectorAll("*[id]").forEach(elem => elem.removeAttribute("id"));
+  pre
```
sidvishnoi: Can we instead copy everything else other than `.idlHeader` instead of removing them?

src/core/webidl-index.js
```diff
@@ -63,5 +63,8 @@ export function run() {
...
+    .querySelectorAll(".idlHeader:not(:first-child)")
```
sidvishnoi: Also, not removing the `:first-child` means the button will link to first interface only (which doesn't make sense).

marcoscaceres: Lol, woops! Good point. I think i need to add a function to add these headers as needed.

sidvishnoi: How about we do something like:
``` js
document.querySelectorAll("pre.webidl :not(.idlHeader), pre.idl :not(.idlHeader)")
  .forEach(idls => idlIndex.append(idls));
```

src/core/utils.js
marcoscaceres: Added note.

src/core/webidl-index.js
marcoscaceres: Went with the function instead. It was a bit simpler.

tests/spec/core/webidl-spec.js
```diff
@@ -17,6 +17,40 @@ describe("Core - WebIDL", () => {
...
+        <pre class="webidl">
+          dictionary Bar {};
+        </pre>
+      `;
+      const ops = makeStandardOps(null, body);
+      const doc = await makeRSDoc(ops);
+      const [idl1, idl2] = doc.querySelectorAll("pre");
+      expect(idl1.id).toBeTruthy();
+      expect(idl2.id).toBeTruthy();
+      expect(idl1.id).not.toBe(idl2.id);
```
sidvishnoi: Nicely avoided hard-coding :ok_hand:

foolip:
I believe this caused https://github.com/tidoust/reffy/issues/205, and was a bit tricky to pin down with specs using https://www.w3.org/Tools/respec/respec-w3c directly. The fix should be straightforward, though.

If it's not overly burdensome, when the structure of the IDL blocks changes in the future, it'd great with an issue on https://github.com/tidoust/reffy. Figuring out if it *will* break requires understanding the implementation in some detail so I wouldn't want to ask for that.

sidvishnoi:
Sorry about the breakage.

> If it's not overly burdensome, when the structure of the IDL blocks changes in the future, it'd great with an issue

Will keep this in mind next time!

saschanaz:
Why do we need this when we can link to definitions directly? 👀

sidvishnoi:
Probably for cases like this: https://w3c.github.io/manifest/#webidl-x1223078318?

saschanaz:
That one is about `BeforeInstallPromptEvent` so a link to that can be used. Not sure there is a good use case that can't link to the definitions.

foolip:
In Blink's IDL files I always try to link to the IDL block as a whole since linking directly to an interface will usually scroll past the extended attributes.

PR #2665, opened 2019-12-26
feat(core/structure): prefer `heading.id` over `section.id`

sidvishnoi:
Prefer `heading.id` for TOC linking and self linking.

### DIFF
--- a/src/core/id-headers.js
+++ b/src/core/id-headers.js
@@ -13,9 +13,13 @@ export function run(conf) {
     `section:not(.head):not(.introductory) h2, h3, h4, h5, h6`
   );
   for (const h of headings) {
-    addId(h);
+    // prefer for ID: heading.id > parentElement.id > newly generated heading.id
+    let id = h.id;
+    if (!id) {
+      addId(h);
+      id = h.parentElement.id || h.id;
+    }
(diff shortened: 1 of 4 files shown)


src/core/id-headers.js
```diff
@@ -13,9 +13,13 @@ export function run(conf) {
     `section:not(.head):not(.introductory) h2, h3, h4, h5, h6`
   );
   for (const h of headings) {
-    addId(h);
+    // prefer for ID: heading.id > parentElement.id > newly generated heading.id
+    let id = h.id;
```
marcoscaceres: I think we still want to id every heading regardless (I think it might be a W3C PubRules requirement), even if we prefer the other.

sidvishnoi: We are adding an id if it doesn't exist (next line)

marcoscaceres: resolving based on discussions in slack

sidvishnoi: <details>
<summary>Conversation log (in case we forget what we discussed)</summary>
<pre>
Marcos:  do we need `let id = h.id`;
sidvishnoi: we don't have a way to know if it's already existing id or newly added one (so we test it first) ... so "preference"</pre>
</details>

PR #2667, opened 2019-12-27
fix(core/ui): trap focus in ReSpec menu to improve a11y

SuyashSalampuria:
Added aria-label on search-speref.js and also added aria-live regions for searches
Added tabindex=-1 to on the pill itself
used the inert feature while opening and closing dialogues.


SuyashSalampuria:
Please help me some of the features
import "../../node_modules/wicg-inert/dist/inert.js";
at the beginning of the page gives an error 
The 'this' keyword is equivalent to 'undefined' at the top level of an ES module, and has been rewritten 
during build probably for 
`const ariaMap = new Map([["labelledby", headingId]]);
    ariaDecorate(modal, ariaMap);
    document.body.append(overlay, modal);
    overlay.addEventListener("click", () => this.closeModal(currentOwner));
    overlay.classList.toggle("respec-show-overlay");
    modal.hidden = false;
  },
`
in freshModal function in src/core/ui.js

SuyashSalampuria:
Also I dont feel that any close button is needed because clicking the Respec button again closes the dialogue box

SuyashSalampuria:
@marcoscaceres @saschanaz  reviews please !!

saschanaz:
I'm not sure we should use `inert` now, is the specification stable enough? AFAIK there is no implementation, and that suggests even the polyfill should be considered highly experimental.

SuyashSalampuria:
Just saw a few videos to understand [pollyfill and Allycasts](https://www.youtube.com/watch?v=fGLp_gfMMGU). It seems stables although. 
But since it is not much old
https://github.com/WICG/inert
I also felt that it should not be used
But as mentioned in this [comment](https://github.com/w3c/respec/issues/884#issuecomment-242375658), I did them. should I remove them?

marcoscaceres:
Yeah, I'm also a bit worried about using the polyfill as it adds another dependency. @SuyashSalampuria please remove it. If it becomes standardized across browsers, then we can add support.

SuyashSalampuria:
>     * then, focus should be handled to remain inside the dialog/overlay (so keyboard user isn't tabbing out of the dialog and back into the page behind the dimmed overlay);
I am leaving out this part as for now since we dont have to use pollyfill

>   also add an explicit "Close" button to those dialogs - in addition to allowing user to close with ESC
Should I add this because clicking on respec button again closes the dialogue button

src/core/ui.js
```diff
@@ -56,9 +56,10 @@ const buttons = {};
 sub("start-all", () => document.body.prepend(respecUI), { once: true });
 sub("end-all", () => document.body.prepend(respecUI), { once: true });
 
-const respecPill = hyperHTML`<button id='respec-pill' disabled>ReSpec</button>`;
+const respecPill = hyperHTML`<button id='respec-pill' tabindex="-1" disabled>ReSpec</button>`;
```
marcoscaceres: When checking the preview, I can't seem to tab to the menu at all? also, for whatever reason, it doesn't seem to be focusing? As you seeing the same?

marcoscaceres:
> Should I add this because clicking on respec button again closes the dialogue button

Let's do that in a new pull request, after this one gets merged.

src/core/ui.js
SuyashSalampuria: Sorry for late reply @marcoscaceres .
As far as I understood from the [comment(5th point)](https://github.com/w3c/respec/issues/884#issuecomment-242375658), we have to make the pill unreachable by pressing tab, And once, The pill is clicked by mouse, focus goes to the pill and things behave as they were earlier. 
Please correct me if I am wrong

marcoscaceres: I don't know - tbh 🥺 We might need to double check with @patrickhlauke if he is around. 

However, while we wait for him to respond, we can add the close button to the dialogs.

sidvishnoi: As I understand, what we need to do is "when a popup opens, make pill and rest of page unreachable", and not make pill unreachable as this PR is doing.

SuyashSalampuria: > As I understand, what we need to do is "when a popup opens, make pill and rest of page unreachable", and not make pill unreachable as this PR is doing.
@sidvishnoi This seems quite good to me. Shall I go on do this?

SuyashSalampuria: Also, pressing enter key on the pill, does not open the dialogue box, Shall I add this feature too?

SuyashSalampuria: Now, the control remains inside the overlay only. If this is right, Can I add the close button too?

SuyashSalampuria: Also, should I add a feature so that pressing esc closes the dialogue/overay

sidvishnoi: >  Also, should I add a feature so that pressing esc closes the dialogue/overay

This already exists I believe.

SuyashSalampuria: > This already exists I believe.
Can you please confirm if it exists. because when I am pressing esc, the dialogue is not closing:(

sidvishnoi: Looking at https://w3c.github.io/payment-request/, I can close it by escape already

SuyashSalampuria: Must be some issue with my system. I am unable to close it here too

src/core/ui.js
```diff
@@ -130,6 +134,30 @@ function createWarnButton(butName, arr, title) {
   ariaDecorate(button, ariaMap);
   return button;
 }
+window.addEventListener("load", () => {
+  ui.addCommand(`Close`, ui.closeModal, "Esc", "❌");
+  menuLength = menu.children.length;
+});
+
+menu.addEventListener("keydown", e => {
```
marcoscaceres: This is close, but not quite right. It should be possible to trap the tabbing just based the fact that the menu is "open".... i.e., it tabbing tries to leave the menu with a shift, it focuses the last, otherwise, if it tries to leave via the last, it goes to first.

src/core/ui.js
```diff
@@ -130,6 +134,30 @@ function createWarnButton(butName, arr, title) {
...
+  if (
+    e.target == menu.getElementsByTagName("li")[0].firstElementChild &&
```
marcoscaceres: you can probably just cache the first and last children in variables, or just look them up more cleanly via a CSS selector (e.g:, `menu.querySelector("li:first-child")`)

src/core/ui.js
```diff
@@ -59,6 +60,8 @@ sub("end-all", () => document.body.prepend(respecUI), { once: true });
 const respecPill = hyperHTML`<button id='respec-pill' disabled>ReSpec</button>`;
 respecUI.appendChild(respecPill);
 respecPill.addEventListener("click", e => {
+  menuLength = menu.children.length;
+  respecPill.focus();
```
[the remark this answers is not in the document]
SuyashSalampuria: Oh!! I added it when there was tab index on it. removing it now.

src/core/ui.js
SuyashSalampuria: Unable to get you @marcoscaceres .
 Thats what I have done. When it tries to leave the last via Tab, it reaches the first.
When It tries to leaves the first via Shift+Tab, it reaches the last

src/core/ui.js
marcoscaceres: Just confirming also the "esc" works for me.

src/core/ui.js
marcoscaceres: Understood, the logic is right, but I think the code could be simplified. Maybe worth adapting:
https://hiddedevries.nl/en/blog/2017-01-29-using-javascript-to-trap-focus-in-an-element

src/core/ui.js
marcoscaceres: if we do this right, we should be able to use the same code for the other popup dialogs. Basically, there should be a generic trap focus function that traps the focus inside a given context until some action is performed.

src/core/ui.js
SuyashSalampuria: Oh !  I was asking for pressing Esc on this stage
Screenshot from 2019-12-31 12-44-34
Offcourse , it closes when we press esc after selecting one of these options/opening a model

marcoscaceres: Ah, ok - sorry for the misunderstanding. Yes, you are right. It indeed doesn't close there, and probably should.

src/core/ui.js
```diff
@@ -130,6 +134,30 @@ function createWarnButton(butName, arr, title) {
   ariaDecorate(button, ariaMap);
   return button;
 }
+window.addEventListener("load", () => {
+  ui.addCommand(`Close`, ui.closeModal, "Esc", "❌");
```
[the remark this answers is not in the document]
marcoscaceres: Let's not do this in this pull request, otherwise this pull request will get out of hand.

src/core/ui.js
SuyashSalampuria: @marcoscaceres I have made the changes, is it right now or changes are needed

src/core/ui.js
```diff
@@ -70,6 +70,13 @@ respecPill.addEventListener("click", e => {
   respecPill.setAttribute("aria-expanded", String(menu.hidden));
   menu.hidden = !menu.hidden;
 });
+respecPill.addEventListener("keydown", e => {
+  if (e.key === "Escape") {
+    e.preventDefault();
+    respecPill.click();
```
saschanaz: Do we want to reopen the menu by tapping Escape twice? Is it expected elsewhere?

src/core/ui.js
```diff
@@ -131,6 +138,31 @@ function createWarnButton(butName, arr, title) {
...
+menu.addEventListener("keydown", e => {
+  if (e.key === "Escape") {
+    e.preventDefault();
+    menu.classList.add("respec-hidden");
+    menu.classList.remove("respec-visible");
+    respecPill.setAttribute("aria-expanded", String(menu.hidden));
+    menu.hidden = true;
+    respecPill.focus();
+  } else if (
+    e.target == menu.querySelector("li:first-child").firstElementChild &&
```
saschanaz: This can be even shorter by `li:first-child *` without firstElementChild 👍

src/core/ui.js
```diff
@@ -131,6 +138,31 @@ function createWarnButton(butName, arr, title) {
...
+    e.shiftKey &&
+    e.key === "Tab"
```
saschanaz: This tab key check is repeated, maybe we can do `if (tab) { if (e.target ...) ... else if ... }` for readability?

src/core/ui.js
SuyashSalampuria: Yes , I wanted this happen and also, it is making the code the shorter.
Should I remove this?

saschanaz: Reopening already works with Space key, not sure we need this. I think we instead need to focus the first element when users tab Space key, as currently ReSpec just keeps its focus on the pill button.

SuyashSalampuria: Is closing it by esc key needed?

saschanaz: Yes, as otherwise no way to close the menu as we trap the focus within it.

src/core/ui.js
SuyashSalampuria: The event.target is that button inside li tags. So the condition won't be true in without using firstElementChild. May be, we have a better alternative to firstElementChild.

saschanaz: `li:first-child *` selects the first element child inside of first `li`, so it should be equivalent here.

SuyashSalampuria: I think that `li:first-child` select the first li which is inside any parent. 
But I will check it again 
Thanks for guidance

saschanaz: Note the trailing ` *`, it selects any first element child of the parent (in this case its the first li).

src/core/ui.js
marcoscaceres: Again, this remains way overly specific and it interferes with the eventing model... here a generalized alternative I quickly slapped together (it's incomplete, but should give you an idea):

https://jsfiddle.net/j7205dsf/11/

What we are trying to get to is something that should work with any tree structure, menu or anything else, as it only cares if the focus has tried to break out of the container. It if it breaks out, the focus is returned to the first focusable element.

src/core/ui.js
```diff
@@ -69,7 +69,7 @@ respecPill.addEventListener("click", e => {
   }
   respecPill.setAttribute("aria-expanded", String(menu.hidden));
   menu.hidden = !menu.hidden;
-  menu.querySelector("li:first-child").firstElementChild.focus();
+  menu.querySelector("li:first-child *").focus();
```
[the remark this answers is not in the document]
saschanaz: `li` itself is not focusable, is it?

Maybe better to use `button` instead of `*` to be more specific, though.

src/core/ui.js
saschanaz: (Please don't insert tabindex=0 into `li` elements, which will cause focusing a single item twice)

marcoscaceres: Yes, what @saschanaz said. The code above is just to illustrate a somewhat generalized way of solving this. 

Again, the problem to solve is:

  * given _any_ element (menu, list, Iframe, whatever), how can I trap the focus inside it allowing people to tab around to any focusable element.
 
The same solution must be usable in both the ReSpec menu and any of the popups dialogs.

SuyashSalampuria: Thanks for your help @marcoscaceres . I am working on a genric solution to this and come out with it soon!!

src/core/ui.js
marcoscaceres: Appreciate your efforts. This is a really hard problem and definitely not "good first issue" territory. I think we are getting closer to a solution however. It might be worth Googling a bit more, however as this has to be solved problem. We just need to adapt someone else's solution.

src/core/ui.js
marcoscaceres: Appreciate your efforts. This is a really hard problem and definitely not "good first issue" territory. I think we are getting closer to a solution however. It might be worth Googling a bit more, however as this has to be solved problem. We just need to adapt someone else's solution.

SuyashSalampuria: > Again, the problem to solve is:
given any element (menu, list, Iframe, whatever), how can I trap the focus inside it allowing people to tab around to any focusable element.

I think that https://hiddedevries.nl/en/blog/2017-01-29-using-javascript-to-trap-focus-in-an-element is an excellent solution and I have implemented on it. Sending the first and last focussable elements are an excellent way offcourse.
Does this seem right to you

src/core/ui.js
```diff
@@ -131,6 +133,35 @@ function createWarnButton(butName, arr, title) {
   return button;
 }
 
+function focusTrapper(e, firstFocussable, lastFocussable) {
```
marcoscaceres: This is better, but it's still not quite right... as I mentioned previously, is should be possible to use "focusTrapper" independently of anything else... I suggest doing this: 

 1. copy/paste the exact code from [https://hiddedevries.nl](https://hiddedevries.nl/en/blog/2017-01-29-using-javascript-to-trap-focus-in-an-element) to core/utils.js without modifying it and `export function trapFocus(element){}` (remove the namespace, it's not used).
 1. import that function into here. 
 1. Only when the menu opens, call `trapFocus(menu);`
 1. When to menu closes, see if `trapFocus()` knows to release the focus.

src/core/ui.js
```diff
@@ -131,6 +133,35 @@ function createWarnButton(butName, arr, title) {
...
+      e.preventDefault();
+      lastFocussable.focus();
+    } else if (e.target == lastFocussable && !e.shiftKey) {
+      e.preventDefault();
+      firstFocussable.focus();
+    }
+  }
+}
+
+menu.addEventListener("keydown", e => {
```
marcoscaceres: We shouldn't need any of this, I don't think.

src/core/ui.js
saschanaz: We should keep related functions in a relevant module if possible. I would keep that function here as we only use that function in this module.

src/core/ui.js
saschanaz: Escaping is needed because trapping focus otherwise won't allow to turn the menu off with keyboard buttons.

saschanaz: For a real field example, check out what 9-square button do in https://google.com.

src/core/ui.js
marcoscaceres: My idea is to use it for all the dialogs, as they all require the same thing... let me check where they come from.

src/core/ui.js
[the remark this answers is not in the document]
marcoscaceres: Right, we want to focus on the first focusable element.

src/core/ui.js
```diff
@@ -79,6 +81,19 @@ document.documentElement.addEventListener("click", () => {
...
+  if (e.key === "Escape" && !menu.hidden) {
+    menu.classList.add("respec-hidden");
+    menu.classList.remove("respec-visible");
+    respecPill.setAttribute("aria-expanded", String(menu.hidden));
+    respecPill.focus();
+    menu.hidden = true;
+  }
+});
+window.addEventListener("load", () => {
```
marcoscaceres: Consider, we only want to trap focus when the menu is visible, otherwise we need to release focus. See where the click even to open the menu happens.

src/core/ui.js
```diff
@@ -79,6 +81,19 @@ document.documentElement.addEventListener("click", () => {
 });
 respecUI.appendChild(menu);
 
+menu.addEventListener("keydown", e => {
```
marcoscaceres: This still seems out of place. This should all be handled by the menu closing (i.e., we should already have much of this code here, or it should be part of ui.hide() or .close() (whichever it is below... on my phone so hard to check)

src/core/ui.js
SuyashSalampuria: In that case, the menu.removeEventListener has also to be written?

src/core/ui.js
```diff
@@ -101,17 +101,17 @@ function trapFocus(element) {
   const lastFocusableEl = focusableEls[focusableEls.length - 1];
 
   element.addEventListener("keydown", e => {
-    if (e.key !== "Tab" || e.keyCode !== /*tab keycode*/ 9) {
+    if (e.key !== "Tab" || e.keyCode !== /* tab keycode */ 9) {
```
saschanaz: Why keyCode? We don't support oooold IE, right?

src/core/ui.js
```diff
@@ -69,7 +69,10 @@ respecPill.addEventListener("click", e => {
   }
   respecPill.setAttribute("aria-expanded", String(menu.hidden));
   menu.hidden = !menu.hidden;
+  trapFocus(menu);
```
saschanaz: Adding event listener every time when clicking pill doesn't sound good... 👀

marcoscaceres: Agree... it needs unregister itself later when closing happens.

src/core/ui.js
marcoscaceres: No, we don't... if this is just for IE, we can ditch.

src/core/ui.js
```diff
@@ -79,6 +82,44 @@ document.documentElement.addEventListener("click", () => {
...
+    menu.classList.remove("respec-visible");
+    respecPill.setAttribute("aria-expanded", String(menu.hidden));
+    respecPill.focus();
+    menu.hidden = true;
+  }
+});
+
+// Code adapted from https://hiddedevries.nl/en/blog/2017-01-29-using-javascript-to-trap-focus-in-an-element
+function trapFocus(element) {
+  const focusableEls = element.querySelectorAll(
```
saschanaz: We don't need this long list at least for now, although I see you want this function to be general...

src/core/ui.js
```diff
@@ -79,6 +82,44 @@ document.documentElement.addEventListener("click", () => {
...
+    if (e.key !== "Tab") {
+      return;
+    }
+    // shift + tab
+    if (e.shiftKey) {
+      if (document.activeElement === firstFocusableEl) {
+        lastFocusableEl.focus();
+        e.preventDefault();
+      }
+      return;
```
saschanaz: switch statement may simplify this 👀

src/core/ui.js
saschanaz: Can we modify this to be called just after line 49? That won't require un-registration then.

SuyashSalampuria:
> Can we modify this to be called just after line 49? That won't require un-registration then.

The menu will not be fully loaded till then. So, the focusEls array will go empty. Either, we wait for the menu to fully load there. Or add an un registration

src/core/ui.js
marcoscaceres: on the one hand, this covers every possible case ... hence it's quite nice for `trapFocus(modal)` below, as it means we don't need to maintain the function or think about it. On the other hand, there are definitely some things here that are not needed right now.

Related... I wonder if we could do simplify `input` to just:
```CSS 
input:not([disabled])
```

src/core/ui.js
```diff
@@ -79,6 +84,44 @@ document.documentElement.addEventListener("click", () => {
...
+    respecPill.setAttribute("aria-expanded", String(menu.hidden));
+    respecPill.focus();
+    menu.hidden = true;
+  }
+});
+
+// Code adapted from https://hiddedevries.nl/en/blog/2017-01-29-using-javascript-to-trap-focus-in-an-element
+function trapFocus(element) {
+  const focusableEls = element.querySelectorAll(
+    'a[href]:not([disabled]), button:not([disabled]), textarea:not([disabled]), input[type="text"]:not([disabled]), input[type="radio"]:not([disabled]), input[type="checkbox"]:not([disabled]), select:not([disabled])'
```
```suggestion
    'a[href]:not([disabled]), button:not([disabled]), textarea:not([disabled]), input:not([disabled]), select:not([disabled])'
```

src/core/ui.js
marcoscaceres: what would we switch on? do you mean for `key`?

marcoscaceres:
This is working really nicely not for "Search SpeRef", but it's not working for others :( 
https://respec-preview.netlify.com/?spec=https%3A%2F%2Fw3c.github.io%2Fpayment-request%2F&version=https%3A%2F%2Fdeploy-preview-2667--respec-pr.netlify.com%2Frespec-w3c.js

I'm not sure why it's not finding the anchors in, for instance, the "Definitions" dialog.

SuyashSalampuria:
> This is working really nicely not for "Search SpecRef", but it's not working for others :(
> https://respec-preview.netlify.com/?spec=https%3A%2F%2Fw3c.github.io%2Fpayment-request%2F&version=https%3A%2F%2Fdeploy-preview-2667--respec-pr.netlify.com%2Frespec-w3c.js
> 
> I'm not sure why it's not finding the anchors in, for instance, the "Definitions" dialog.

I am not sure if I got you properly BUT the Definitions dialogue is working fine
Screenshot from 2020-01-03 11-19-50

marcoscaceres:
@SuyashSalampuria, when you say working, do you mean that it just shows the dialog? I'm able to escape (see W3C logo has stolen the focus):

Screenshot 2020-01-03 16 54 44

SuyashSalampuria:
> @SuyashSalampuria, when you say working, do you mean that it just shows the dialog? I'm able to escape (see W3C logo has stolen the focus):
> 
> Screenshot 2020-01-03 16 54 44

I understood you. Let me see what is causing the problem

SuyashSalampuria:
Actually, it's not on this PR which is causing the problem. The current develop branch is facing the same problem . I think we need a new Issue for it. Or shall I do it in this PR

src/core/ui.js
saschanaz: Oops, misread, sorry 🤦‍♀️

Still I think else-if is better than return + if.

src/core/ui.js
```diff
@@ -47,6 +47,9 @@ function ariaDecorate(elem, ariaMap) {
 
 const respecUI = hyperHTML`<div id='respec-ui' class='removeOnSave' hidden></div>`;
 const menu = hyperHTML`<ul id=respec-menu role=menu aria-labelledby='respec-pill' hidden></ul>`;
+window.addEventListener("load", () => {
```
saschanaz: This doesn't need to wait until `load` as we already have a `menu` defined.

src/core/ui.js
```diff
@@ -79,6 +84,44 @@ document.documentElement.addEventListener("click", () => {
 });
 respecUI.appendChild(menu);
 
+menu.addEventListener("keydown", e => {
+  if (e.key === "Escape" && !menu.hidden) {
+    menu.classList.add("respec-hidden");
+    menu.classList.remove("respec-visible");
+    respecPill.setAttribute("aria-expanded", String(menu.hidden));
+    respecPill.focus();
+    menu.hidden = true;
```
saschanaz: This code partially duplicates Line 79-83. Probably better to add a function and use it here and there.

src/core/ui.js
SuyashSalampuria: Surely, the menu is defined , but the list items of menu are still not loaded. So, the focusEls will go empty and the function won't work well
We have to use something else if we don't want this

src/core/ui.js
SuyashSalampuria: Yes, thats the problem. may be, I can write a function to toggle class

src/core/ui.js
saschanaz: Oh, you're right. We probably want to use `window.respecIsReady` in that case, which indicates every ReSpec module is loaded.

SuyashSalampuria: I have made this change to trapFocus function but I don't understand why focusElvs is going empty
```async function trapFocus(element) {
  await document.respecIsReady;
  const focusableEls = element.querySelectorAll(
    "a[href]:not([disabled]), button:not([disabled]), textarea:not([disabled]), input:not([disabled]), select:not([disabled])"
  ); ```
Any comments @saschanaz

Comment (saschanaz) [reply src/core/ui.js]: First of all, we can't use top level await as we want it to be first supported by browsers.

Second, sorry, `document.respecIsReady` is undefined as `core/ui` somehow runs before `core/respec-ready`.

Let's keep depending on `load` for now.

File: src/core/ui.js (hunk 17 above)
Comment (saschanaz) [reply src/core/ui.js]: I would still want to see this rewritten as else-if 👀

File: src/core/ui.js (hunk 19 above)
Comment (saschanaz) [inline src/core/ui.js]: `window.addEventListener("load", () => trapFocus(menu))` for simplicity 😄

File: src/core/ui.js [hunk 21]
```diff
@@ -60,25 +63,60 @@ const respecPill = hyperHTML`<button id='respec-pill' disabled>ReSpec</button>`;
...
 
+menu.addEventListener("keydown", e => {
+  if (e.key === "Escape" && !menu.hidden) {
+    respecPill.setAttribute("aria-expanded", String(menu.hidden));
+    closeOpenMenu();
+    respecPill.focus();
+  }
+});
+
+function closeOpenMenu() {
```
saschanaz: Maybe `toggleMenu()` ?

src/core/ui.js
saschanaz: `return` is redundant

src/core/ui.js
```diff
@@ -195,6 +234,7 @@ export const ui = {
     overlay.addEventListener("click", () => this.closeModal(currentOwner));
     overlay.classList.toggle("respec-show-overlay");
     modal.hidden = false;
+    trapFocus(modal);
```
saschanaz: I suggest we defer this, as:

1. The focus does not enter the modal window, so not trapped. This one is easy to fix.
2. Trapping doesn't work when the modal window dynamically modifies its tree (specref) or uses iframe (definition search). This one is harder.

This PR is for the pill, I would defer anything irrelevant for future new PRs.

src/core/ui.js
```diff
@@ -102,12 +100,11 @@ function trapFocus(element) {
       return;
     }
     // shift + tab
-    if (e.shiftKey) {
+    else if (e.shiftKey) {
```
saschanaz: We choose `return` or `else if`, not both 👀

src/core/ui.js
SuyashSalampuria: Okay! I am putting on hold and start working on the two above points you just mentioned :)

saschanaz:
```suggestion
```

Let's do the modal thing in the next PRs and complete this one 🎉

src/core/ui.js
sidvishnoi: Can we pass this as a parameter instead?

src/core/ui.js
```diff
@@ -60,25 +61,59 @@ const respecPill = hyperHTML`<button id='respec-pill' disabled>ReSpec</button>`;
...
+    "a[href]:not([disabled]), button:not([disabled]), textarea:not([disabled]), input:not([disabled]), select:not([disabled])"
+  );
+  const firstFocusableEl = focusableEls[0];
+  const lastFocusableEl = focusableEls[focusableEls.length - 1];
+
+  element.addEventListener("keydown", e => {
+    if (e.key !== "Tab") {
+      return;
+    }
+    // shift + tab
```
sidvishnoi: Not liking the repletion here (no big deal though)
Can probably do something like?
``` js
let focusableEl = null;
if (e.shiftKey && document.activeElement === firstFocusableEl) {
  focusableEl = lastFocusableEl;
} else if (!e.shiftKey && document.activeElement === lastFocusableEl) {
  focusableEl = firstFocusableEl;
}
if (focusableEl) {
  focusableEl.focus();
  e.preventDefault();
}
```

Or a slightly cryptic version:
``` js
let focusableEl = null;
const focusEls = [firstFocusableEl, lastFocusableEl];
if (document.activeElement === focusEls[Number(e.shiftKey)]) {
  focusableEl = focusEls[Number(!e.shiftKey)];
}
if (focusableEl) {
  focusableEl.focus();
  e.preventDefault();
}
```

src/core/ui.js
```diff
@@ -60,25 +61,59 @@ const respecPill = hyperHTML`<button id='respec-pill' disabled>ReSpec</button>`;
...
-    menu.classList.remove("respec-hidden");
-    menu.classList.add("respec-visible");
-  } else {
-    menu.classList.add("respec-hidden");
-    menu.classList.remove("respec-visible");
-  }
   respecPill.setAttribute("aria-expanded", String(menu.hidden));
-  menu.hidden = !menu.hidden;
+  toggleMenu();
+  menu.querySelector("li:first-child *").focus();
```
sidvishnoi: Can probably replace it it `"li:first-child button"` as we know what's in menu.

src/core/ui.js
saschanaz: Let's not be cryptic 😂

src/core/ui.js
SuyashSalampuria: Do you mean passing button in this case as parameter

src/core/ui.js
SuyashSalampuria: Is it fine presently or should I revert the changes

src/core/ui.js
sidvishnoi: I mean the selector. When we're trapping focus in some element, we generally know what focusable elements are in that particular element.

src/core/ui.js
sidvishnoi: Keep them.

src/core/ui.js
SuyashSalampuria: Won't make much change here.
But passing the selector everytime will make things tough for us next time we use this function
Still if you insist , I will do this

sidvishnoi: @saschanaz WDYT?

saschanaz: I vote to keep it so that we can forget what's actually in there. This should have been much easier when CSS had jQuery's `:focusable` 😔

src/core/ui.js
```diff
@@ -60,25 +61,59 @@ const respecPill = hyperHTML`<button id='respec-pill' disabled>ReSpec</button>`;
...
-    menu.classList.remove("respec-hidden");
-    menu.classList.add("respec-visible");
-  } else {
-    menu.classList.add("respec-hidden");
-    menu.classList.remove("respec-visible");
-  }
   respecPill.setAttribute("aria-expanded", String(menu.hidden));
-  menu.hidden = !menu.hidden;
+  toggleMenu();
+  menu.querySelector("li:first-child build").focus();
```
sidvishnoi:
```suggestion
  menu.querySelector("li:first-child button").focus();
```

saschanaz:
A long journey, thank you for the hard work! 🎉🎉🥳

saschanaz:
Marcos won't respond for weeks, maybe we should force merge this.

sidvishnoi:
I merged before seeing your comment :1st_place_medal: :smiley_cat:

PR #2670, opened 2019-12-29
fix(core/dfn-finder): dont export local-lt things

marcoscaceres:


src/core/dfn-finder.js
```diff
@@ -71,13 +107,15 @@ function findOperationDfn(defn, parent, name) {
 
 /**
  * @param {HTMLElement} dfn
- * @param {string[]} names
+ * @param {Object} names
```
sidvishnoi:
```suggestion
 * @param {Record<"local" | "exportable", string[]>} names
```

src/core/dfn-finder.js
```diff
@@ -40,21 +40,57 @@ export function findDfn(defn, name, { parent = "" } = {}) {
...
     case "attribute":
-      return [asQualifiedName, name];
+      return {
+        local: [asQualifiedName],
+        exportable: [name],
+      };
   }
 }
 
+function deriveMethodWithArgs(operationName, argsAst) {
```
sidvishnoi: I think a comment to explain what this function does might be better. It's not clear from function name

src/core/dfn-finder.js
```diff
@@ -123,50 +161,54 @@ function findNormalDfn(defn, parent, ...names) {
...
+    !dfnElem.querySelector("code") &&
+    !dfnElem.closest("code") &&
+    dfnElem.children
+  ) {
+    wrapInner(dfnElem, dfnElem.ownerDocument.createElement("code"));
   }
 
-  // Add data-lt values and register them
-  switch (defn.type) {
+  // Add data-lt and data-lt values and register them
```
sidvishnoi: Typo: `data-local-lt`

sidvishnoi:
Didn't see some tests were failing. But changes are good anyway

src/core/dfn-finder.js
marcoscaceres: yeah, you are right ... will both rename and document.

PR #2675, opened 2019-12-30
feat(core/markdown): allow custom ID for headings

sidvishnoi:
Allows passing custom headings like:
``` markdown
## Heading {#custom-id}

Foo title {#foo}
================

```


sidvishnoi:
> Maybe just check a couple more heading levels in the tests

We'll be testing whether marked.js picks headings correctly that way, instead of heading ids.

PR #2650, opened 2019-12-19
fix(core/markdown): use reindent() instead

saschanaz:
Fixes #2609

PR #2678, opened 2019-12-30
Revert "fix(core/markdown): use reindent() instead (#2650)"

marcoscaceres:
This reverts commit d33de51ec5100987ce3b6aef95888a3ab0c29fea.

@saschanaz, sorry, we had to revert this :( it turns out a lot of specs are depending on this behavior. We'll need to go through specs that use `format: "markdown"` and fix them manually (or find some other alternative) before we can again merge this.


saschanaz:
Could you list the examples?

marcoscaceres:
See:
 https://github.com/search?q=org%3Aw3c+format%3A+markdown&type=Code

And see:
 https://w3c.github.io/trace-interchange-format/

We fixed, then we realized lots broke:
 https://w3c.github.io/resource-hints/

saschanaz:
I think the trace-interchange-format depends on `data-include` which conflicts with new indentation. Could we try fixing it?

sidvishnoi:
Not just data-include-format=markdown. https://w3c.github.io/preload/ doesn't use it, still broken.

marcoscaceres:
>  Could we try fixing it before reverting?

We can. You up for a late night marathon hacking session? :)

saschanaz:
> Not just data-include-format=markdown. https://w3c.github.io/preload/ doesn't use it, still broken.

Indentation of `<pre>` looks broken there, probably written before ReSpec autoprocess them.

The search result Marcos gave shows a lot of specs depends on data-include, so fixing it probably significantly reduce the broken spec list. What do you think?

marcoscaceres:
It's basically any spec that declares "markdown" and then has bad indentation, I think.

saschanaz:
>probably written before ReSpec autoprocess them.

(I think HTML needs a native way to not require awkward indentations in `<pre>`... 🤔

saschanaz:
>You up for a late night marathon hacking session? :)

Will try 😂

sidvishnoi:
Another example without data-include-format=markdown : https://w3c.github.io/performance-timeline/
Indentation seems fine there (unless I missed something)

marcoscaceres:
@saschanaz, I'll stay up for another hour (2am my time). Ping me on Slack if it's looking too broken. 

Otherwise, please ping @sidvishnoi and check how long he will be up for.

sidvishnoi:
I've ~3 hours. Lets collect the broken specs, revert this commit (and release) (and sleep if needed), fix the specs using https://unpkg.com/respec@25.0.0/builds/respec-w3c.js, fix respec, release.

saschanaz:
Got a title: Web breaker

marcoscaceres:
When we come back to this:
 https://github.com/search?q=org%3AWICG+markdown&type=Code

Issue #2438, opened 2019-07-23
tests/xref: refactor xref tests


sidvishnoi:
There are too many xref tests, a lot of them being duplicates or redundant. A refactor would be nice.

SuyashSalampuria:
@sidvishnoi  Shall I try on this

ridhishjain:
@SuyashSalampuria are u working on some other issue??
if so, @sidvishnoi can I take this?

marcoscaceres:
@ridhishjain, yes, feel free to take it.

sidvishnoi:
Closing as discussed in https://github.com/w3c/respec/pull/2681#issuecomment-570612971

PR #2681, opened 2020-01-02
refactor(test/core/xref-spec)

ridhishjain:
issue #2438 
`"shows error if external term doesn't exist"`
this test is merged into 
`"adds link to unique external terms and shows error if external term doesn't exist"`

Do we need the file to be refactored in this manner?


saschanaz:
Tests are intentionally granular even when it requires longer code, because granular tests allows maintainers to easily get where exactly is broken. So... IMO this change is not needed.

ridhishjain:
In that case, can u tell me where exactly do we need refactoring?

saschanaz:
I have no idea for now, @sidvishnoi do you have any?

sidvishnoi:
> Tests are intentionally granular even when it requires longer code, because granular tests allows maintainers to easily get where exactly is broken. So... IMO this change is not needed.

This made introspect. I agree, we don't need these changes.
Earlier, I was thinking that we can merge some tests to improve testing speed, maintainability and reduce xref requests on server.

ridhishjain:
then should I close this PR?

PR #2682, opened 2020-01-03
refactor: convert definitionMap to CaseInsensitiveMap

marcoscaceres:


saschanaz:
I think we want to make IDL definitions case sensitive, and this sounds conflicting with that... What do you think?

marcoscaceres:
We are not quite there yet. This just the start towards better classification of the definitions.

src/core/dfn-finder.js
```diff
@@ -145,20 +145,18 @@ function addAlternativeNames(dfn, names) {
 function findNormalDfn(defn, parent, ...names) {
   for (const name of names) {
     let resolvedName =
-      defn.type === "enum-value" && name === ""
-        ? "the-empty-string"
-        : name.toLowerCase();
-    let dfnForArray = definitionMap[resolvedName];
+      defn.type === "enum-value" && name === "" ? "the-empty-string" : name;
+    let dfnForArray = [...(definitionMap.get(resolvedName) || [])];
```
saschanaz: Can we do:

```suggestion
    let dfnForArray = definitionMap.get(resolvedName) || [];
```

?

sidvishnoi: It's a set as per dfn-map.js

saschanaz: @sidvishnoi Hmm, in that case how about just do `definitionMap.get(resolvedName)` as we did before, instead of converting a set to an array every time? Line 156 then can be `dfnForArray && dfnForArray.size === 1` (or hopefully `dfnForArray?.size === 1` soon)

saschanaz: (Yay no Marcos, everything is ours, party time 🎉)

sidvishnoi: I think the motivation to use a Set comes from registerDefinition in dfn-map.js
https://github.com/w3c/respec/commit/8c0e56933356addf052c278a81860e9cec2ea280#diff-dbc281fb8c3069b68bed379e67cda5e4

saschanaz: I think so too.

sidvishnoi: Though we can move this dfnForArray inside the if block

marcoscaceres: @saschanaz, hopefully less ugly now. Let me know what you think!

src/ui/dfn-list.js
```diff
@@ -41,7 +41,7 @@ ul.addEventListener("click", ev => {
 });
 
 function show() {
-  const definitionLinks = Object.entries(definitionMap)
+  const definitionLinks = Array.from(definitionMap.entries())
```
saschanaz:
```suggestion
  const definitionLinks = Array.from(definitionMap)
```

`.entries()` is redundant 👀

marcoscaceres:
Thanks @saschanaz for the review and suggestions! This is much better now.

PR #2685, opened 2020-01-03
fix(ui/dfn-list): close popup only if link is clicked

sidvishnoi:


src/ui/dfn-list.js
```diff
@@ -18,6 +18,9 @@ ul.classList.add("respec-dfn-list");
 const render = hyperHTML.bind(ul);
 
 ul.addEventListener("click", ev => {
+  if (ev.target instanceof HTMLElement && !ev.target.matches("a")) {
```
saschanaz: This potentially closes the ui when the target is not HTMLElement. It won't happen, but then maybe we don't need that specific condition?

sidvishnoi: Typescript was crying eventTarget has no matches()
Thought this was better than ts-ignore
Open to other ways

saschanaz: Oh. How about `if (ev.target instanceof HTMLElement && ev.target.matches("a")) {` and wrapping the code then? Becasue reading this line makes me think whether `Element` is intentionally allowed or not 😂

sidvishnoi: That's cool by me. Updating

PR #2689, opened 2020-01-04
Update l10n.js - Korean

pensivej:
I am working on translating w3c using respec, and TOC is not rendered due to lack of I10n when I change to lang=ko.
So I have added translations into korean

### DIFF
--- a/src/core/l10n.js
+++ b/src/core/l10n.js
@@ -44,6 +44,11 @@ const base = {
 
 const ko = {
   abstract: "요약",
+  participate: "참여",
+  references: "참조",
+  toc: "목차",
+  example: "예시",
+  informative: "정보",
 };
 
 const zh = {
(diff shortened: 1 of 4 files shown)


saschanaz:
Hello @pensivej, thank you for working on this!

Do you have a public translation that I can check out now?

src/core/l10n.js
```diff
@@ -44,6 +44,11 @@ const base = {
 
 const ko = {
   abstract: "요약",
+  participate: "참여",
+  references: "참조",
+  toc: "목차",
+  example: "예시",
+  informative: "정보",
```
saschanaz: These two should go `src/core/example.js` and `src/core/informative.js`.

pensivej:
Here it is:
https://ssimeetupkorea.github.io/did-core/

For now, it is set to html=en

src/core/l10n.js
sidvishnoi: I should probably finish up https://github.com/w3c/respec/issues/2084 first

pensivej: resolved

PR #2690, opened 2020-01-04
fix(core/l10n): do not throw if l10n strings don't exist for lang

sidvishnoi:
Came across this bug from https://github.com/w3c/respec/pull/2689
~TODO: Move this to utils and add tests~

### DIFF
--- a/src/core/l10n.js
+++ b/src/core/l10n.js
@@ -6,7 +6,6 @@
  * to manage the config.l10n object so that other parts of the system can
  * localize their text.
  */
-import { getIntlData } from "./utils.js";
 
 export const name = "core/l10n";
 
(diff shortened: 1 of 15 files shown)


src/core/l10n.js
```diff
@@ -123,26 +125,7 @@ l10n["zh-cn"] = l10n.zh;
...
-    get(data, key) {
-      const result = data[lang][key] || data.en[key];
-      if (!result) {
-        throw new Error(`No l10n data for key: "${key}"`);
-      }
-      return result;
-    },
-  });
-}
+export { getIntlData };
```
sidvishnoi: Should I avoid the re-export and rename all imports now or in some other PR?
I can do it in https://github.com/w3c/respec/issues/2084

saschanaz: I would prefer now 👍

sidvishnoi:
Thanks for quick review!

PR #2691, opened 2020-01-04
refactor: modules to contain their own localization data

sidvishnoi:
Closes https://github.com/w3c/respec/issues/2084

TODO:
- [x] Move localization data to modules
- [x] Resolve issue mentioned in https://github.com/w3c/respec/pull/2691#issuecomment-570805732 (Decided to revert for now)
- [ ] Support language aliases in getIntlData (`l10n["zh-hans"] = l10n.zh; l10n["zh-cn"] = l10n.zh`) (https://github.com/w3c/respec/pull/2692)
- [x] Remove `core/l10n.js` (Decided to keep it for a while for compatiblity, emptied it mostly)


sidvishnoi:
https://github.com/w3c/respec/pull/2691/commits/2ad96e3e55ec83ec56c084406b4f8fde1ee4d756 is problematic. We use this "additions" way to extend localization strings as a `[profile]/l10n.js` plugin, which extends `conf.l10n`. Here, we're using `l10n.status_at_publication`, which will break this particular l10n for geonovum :disappointed:

saschanaz:
Might want to keep it for compatibility at least for a while.

src/core/l10n.js
```diff
@@ -18,97 +18,17 @@ if (html && !html.hasAttribute("lang")) {
...
-  search_specref: "Search Specref",
-  toc: "Table of Contents",
-  warning: "Warning",
-};
+const base = {};
 
-const ko = {
-  abstract: "요약",
-};
+const ko = {};
```
saschanaz: No need to keep these local variables 👀

sidvishnoi: Was removing things incrementally... reduce -> empty -> delete

PR #2692, opened 2020-01-04
fix(core/utils): allow lang aliases with getIntlData

sidvishnoi:


src/core/utils.js
```diff
@@ -292,12 +292,24 @@ export function norm(str) {
   return str.trim().replace(/\s+/g, " ");
 }
 
+/**
+ * @param {string} lang
+ */
+function resolveLanguageAlias(lang) {
+  const aliases = {
+    "zh-hans": "zh",
+    "zh-cn": "zh",
```
saschanaz: Has there been a request for this?

sidvishnoi: No. But it existed in core/l10n.js, so decided to keep. I think we can ignore this. (Can always add later if someone needs it)

sidvishnoi: A search through github.com/W3C found only one document using it: https://github.com/w3c/danmaku

saschanaz: Ah, I agree to keep it then. Interesting API BTW.

sidvishnoi: Marcos would've asked for a switch-case :stuck_out_tongue_closed_eyes:

saschanaz: I would've asked for generalized algorithm for `ko-kr` etc, but it's just an existing code, no need to make it complex.

sidvishnoi:
@saschanaz Release v25.0.3 after this?

saschanaz:
Sounds good 👍

But you do 😂

PR #2694, opened 2020-01-05
chore(tools/release): fix staging before checkout

sidvishnoi:
Fix release yet again :balloon:


sidvishnoi:
> Good catch!

Didn't catch, it broke :sob:

saschanaz:
> Didn't catch, it broke 😭

😅😅😅

sidvishnoi:
Broke again :balloon: :tada: 
> you need to resolve your current index first

PR #2696, opened 2020-01-05
chore(tools/release): create commit instead of amend

sidvishnoi:
`npm version --no-git-tag-version` does not create a commit (we believed it just didn't create tag, as name implies), so amending does not make sense.


sidvishnoi:
Unfortunately, only way to detect these bugs is by releasing :disappointed:

PR #2697, opened 2020-01-05
feat(core/ui): add close button to modals

SuyashSalampuria:


assets/ui.css
```diff
@@ -1,3 +1,14 @@
+#close-button{
```
saschanaz:
```suggestion
.close-button {
```

assets/ui.css
```diff
@@ -1,3 +1,14 @@
...
+  font-weight: bold;
+  float: left;
+  cursor: pointer;
+  margin-left: 0.5%;
+}
+
+#close-button:hover {
```
saschanaz:
```suggestion
.close-button:hover {
```

src/core/ui.js
```diff
@@ -47,13 +47,15 @@ function ariaDecorate(elem, ariaMap) {
 
 const respecUI = hyperHTML`<div id='respec-ui' class='removeOnSave' hidden></div>`;
 const menu = hyperHTML`<ul id=respec-menu role=menu aria-labelledby='respec-pill' hidden></ul>`;
+const closeButton = hyperHTML`<span id="close-button">❌</span>`;
```
saschanaz:
```suggestion
const closeButton = hyperHTML`<span class="close-button">❌</span>`;
```

`class` is preferred since we don't use the id anywhere else. Also I would prefer putting this line at line 51 as currently line 49 and 50 are related and should be together.

src/core/ui.js
```diff
@@ -47,13 +47,15 @@ function ariaDecorate(elem, ariaMap) {
...
 window.addEventListener("load", () => trapFocus(menu));
 let modal;
 let overlay;
 const errors = [];
 const warnings = [];
 const buttons = {};
 
+closeButton.addEventListener("click", () => ui.closeModal());
```
saschanaz: Maybe move to line 52? I would prefer putting the relevant lines together.

src/core/ui.js
saschanaz: Can we use `button` here rather than `span`, so that screen readers can see whether this is a button without `aria-role`?

assets/ui.css
```diff
@@ -1,3 +1,14 @@
+.close-button{
```
saschanaz:
```suggestion
.close-button {
```

assets/ui.css
```diff
@@ -1,3 +1,14 @@
...
+  font-weight: bold;
+  float: left;
+  cursor: pointer;
+  margin-left: 0.5%;
```
saschanaz:
```suggestion
  margin-left: 0.5%;
  border: none;
  background: transparent;
```

assets/ui.css
```diff
@@ -1,3 +1,14 @@
+.close-button{
+  font-weight: bold;
+  float: left;
+  cursor: pointer;
```
saschanaz: Not sure this one is needed, @sidvishnoi what do you think?

SuyashSalampuria: float:left would be fine. I think.

saschanaz: Yeah, I mean `cursor: pointer` 👀

SuyashSalampuria: Also, is this margin for the close button fine or i shall decrease it
Screenshot from 2020-01-09 13-48-19

assets/ui.css
saschanaz: Maybe a fixed value, as there is no need to change the value based on window size.

assets/ui.css
saschanaz: IMO it's okay as-is.

sidvishnoi: I prefer to keep `cursor:pointer` and remove the scaling animation.

sidvishnoi:
Need to add `background: none; border: none` on button
image

src/core/ui.js
```diff
@@ -47,7 +47,9 @@ function ariaDecorate(elem, ariaMap) {
 
 const respecUI = hyperHTML`<div id='respec-ui' class='removeOnSave' hidden></div>`;
 const menu = hyperHTML`<ul id=respec-menu role=menu aria-labelledby='respec-pill' hidden></ul>`;
+const closeButton = hyperHTML`<button class="close-button">❌</button>`;
```
sidvishnoi:
```suggestion
const closeButton = hyperHTML`<button class="close-button" aria-label="Close">❌</button>`;
```

src/core/ui.js
```diff
@@ -222,7 +224,7 @@ export const ui = {
     const id = `${currentOwner.id}-modal`;
     const headingId = `${id}-heading`;
     modal = hyperHTML`<div id='${id}' class='respec-modal removeOnSave' role='dialog'>
-      <h3 id="${headingId}">${title}</h3>
+      <h3 id="${headingId}">${closeButton}${title}</h3>
```
sidvishnoi: I'm not sure of the accessibility implications of adding close button next to heading. It'll be announced as: "heading level 3 button close Search definitions"
Might be better to move this out of the heading and add a `aria-labelledby="${headingID}"` on `div[role='dialog']` above.

src/core/ui.js
```diff
@@ -47,7 +47,9 @@ function ariaDecorate(elem, ariaMap) {
 
 const respecUI = hyperHTML`<div id='respec-ui' class='removeOnSave' hidden></div>`;
 const menu = hyperHTML`<ul id=respec-menu role=menu aria-labelledby='respec-pill' hidden></ul>`;
+const closeButton = hyperHTML`<button class="close-button" aria-label="Close">❌</button>`;
```
saschanaz:
```suggestion
const closeButton = hyperHTML`<button class="close-button" title="Close">❌</button>`;
```

`title` also allows visual readers to see the button role by mouse hovering, while screen readers will still perceive it as `aria-label`.

sidvishnoi: That's even better!

src/core/ui.js
SuyashSalampuria: Should the close button be at the same place?
Screenshot from 2020-01-09 13-48-19
Also, shall I write it something like this:
```
<div id="${headingId}">${closeButton} <h3>${title}</h3></div>
```

src/core/ui.js
```diff
@@ -221,13 +223,14 @@ export const ui = {
     overlay = hyperHTML`<div id='respec-overlay' class='removeOnSave'></div>`;
     const id = `${currentOwner.id}-modal`;
     const headingId = `${id}-heading`;
-    modal = hyperHTML`<div id='${id}' class='respec-modal removeOnSave' role='dialog'>
+    modal = hyperHTML`<div id='${id}' class='respec-modal removeOnSave' role='dialog' aria-labelledby='${headingId}' >
```
saschanaz:
```suggestion
    modal = hyperHTML`<div id='${id}' class='respec-modal removeOnSave' role='dialog' aria-labelledby='${headingId}' >
      ${closeButton}
```

I think this was the suggestion 👀

SuyashSalampuria: I had tried that. But a white line was added above the heading line and that is offcourse not desirable
Screenshot from 2020-01-09 19-18-46

saschanaz: Hmm, we might want to do `position: absolute` in that case. @sidvishnoi What do you think?

sidvishnoi: The markup is what we want, so yes, position it with CSS.

SuyashSalampuria:
> Why these changes?

<h3> always changes the line. So we need an inline display element

assets/ui.css
```diff
@@ -231,7 +243,6 @@
   background: #fff;
   border: 5px solid #666;
   min-width: 20%;
-  width: 79%;
```
sidvishnoi: Why was this removed? I'm seeing modal like this now:
image

assets/ui.css
```diff
@@ -245,6 +256,7 @@
 }
 
 .respec-modal h3 {
+  width: 100%;
```
sidvishnoi: This is causing an overflow-x. Any reason to add it?

assets/ui.css
```diff
@@ -1,3 +1,15 @@
+.close-button {
+  position: absolute;
+  z-index: inherit;
+  padding-top: .2em;
+  font-weight: bold;
+  float: left;
```
sidvishnoi: We have `position: absolute` already
```suggestion
```

assets/ui.css
```diff
@@ -1,3 +1,15 @@
+.close-button {
+  position: absolute;
+  z-index: inherit;
+  padding-top: .2em;
```
sidvishnoi:
```suggestion
  padding: .2em;
```

assets/ui.css
SuyashSalampuria: There was a small gap between the heading and the boundary in some of the modals(search definitions). I wanted to remove that actually. Correcting this!

src/core/ui.js
```diff
@@ -47,7 +47,9 @@ function ariaDecorate(elem, ariaMap) {
 
 const respecUI = hyperHTML`<div id='respec-ui' class='removeOnSave' hidden></div>`;
 const menu = hyperHTML`<ul id=respec-menu role=menu aria-labelledby='respec-pill' hidden></ul>`;
+const closeButton = hyperHTML`<button class="close-button" title="Close">❌</button>`;
 window.addEventListener("load", () => trapFocus(menu));
+closeButton.addEventListener("click", () => ui.closeModal());
```
sidvishnoi: Can we do this inline?
``` js
hyperHTML`<button class="close-button" title="Close" onclick={() => ui.closeModal()}>❌</button>`
```

assets/ui.css
```diff
@@ -1,3 +1,14 @@
+.close-button {
+  position: absolute;
```
sidvishnoi: Should we move this to right side? Feels more natural to me. WDYT @saschanaz ?

SuyashSalampuria: I was asked to add this at that place by @marcoscaceres

assets/ui.css
```diff
@@ -245,6 +256,7 @@
 }
 
 .respec-modal h3 {
+  /* width: 100%; */
```
sidvishnoi:
```suggestion
```

assets/ui.css
```diff
@@ -1,3 +1,14 @@
+.close-button {
```
sidvishnoi: Maybe just to be safe?
```suggestion
.respec-modal .close-button {
```

src/core/ui.js
```diff
@@ -47,6 +47,8 @@ function ariaDecorate(elem, ariaMap) {
 
 const respecUI = hyperHTML`<div id='respec-ui' class='removeOnSave' hidden></div>`;
 const menu = hyperHTML`<ul id=respec-menu role=menu aria-labelledby='respec-pill' hidden></ul>`;
+const closeButton = hyperHTML`<button class="close-button" onclick=${() =>
```
sidvishnoi: Nit: can we avoid creating another function?
```suggestion
const closeButton = hyperHTML`<button class="close-button" onclick=${ui.closeModal} title="Close">❌</button>`;
```

SuyashSalampuria: We actually need a new function because in that case, ui will be used directly before its declaration

SuyashSalampuria:
Can this be merged if it is ready because I need the close button for trapping focus inside modals

sidvishnoi:
Missed that. Merged now :slightly_smiling_face:

PR #2687, opened 2020-01-03
feat: add German translations

sebastian-meyer:
This adds German translations for all ReSpec sources (except `geonovum` and error messages).

It also fixes a small issue where in `src/ui/about-respec.js` a hard-coded string was used although a translatable constant was defined (see https://github.com/w3c/respec/commit/3ce9e50c9f7fab0fa8ad5535da5767df6ef14584).

PR #2699, opened 2020-01-06
chore: add Japanese translation

himorin:
- not sure whether adding to core/link-to-dfn.js is ok or not.
- until in w3c/templates/show-people.js could be mis-aligned in sentence on pre-/post- issue


src/w3c/templates/show-people.js
```diff
@@ -14,6 +14,9 @@ const localizationStrings = {
   es: {
     until: "Hasta",
   },
+  ja: {
+    until: "以前",
```
saschanaz: This will be shown as e.g. "Jane Doe (以前 2020-01-07)", would it be good enough? Should we add a way to do "2020-01-07 以前"?

himorin: If possible, adding a way is the best. (I hope user can understand even if that is not easy...)

saschanaz: Done! Could you rebase and use the new function pattern? Thanks!

himorin: Thank you! rebased and updated.

himorin: Thank you! and rebased / updated.

src/core/figures.js
```diff
@@ -21,7 +21,7 @@ const localizationStrings = {
     fig: "Figure ",
   },
   ja: {
-    fig: "図",
+    fig: "図 ",
```
saschanaz: This is causing test failure, could you also revise the relevant test in `tests/spec/core/figures-spec.js:52:33`?

src/ui/about-respec.js
```diff
@@ -26,6 +26,9 @@ const localizationStrings = {
   nl: {
     about_respec: "Over",
   },
+  ja: {
+    about_respec: "これについて",
```
saschanaz: #2687 will use `about_respec` in the respec pill menu item, so this will be shown as e.g. "これについて 2.4.10". Not sure it would be good enough, what do you think?

marcoscaceres:
@himorin, let us know if we can be of any help. This looks like good addition.

himorin:
sorry, quite busy for other staff recently. le tme back within this weekend (or early next week)...

src/ui/about-respec.js
himorin: Yes for this point.
Also I've gone through all translations, and I think this kind of items does not exist no more (other than this).

src/core/figures.js
himorin: fixed.

PR #2701, opened 2020-01-07
fix(core/webidl): use valid markup inside pre

sidvishnoi:
We had invalid markup... causing [errors with respec-validator](https://travis-ci.com/w3c/screen-orientation/builds/143352826#L705)
image

### DIFF
--- a/src/core/webidl.js
+++ b/src/core/webidl.js
@@ -349,10 +349,10 @@ function renderWebIDL(idlElement, index) {
  */
 export function addIDLHeader(pre) {
   addHashId(pre, "webidl");
-  const header = hyperHTML`<div class="idlHeader"><a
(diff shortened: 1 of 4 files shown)


src/core/webidl.js
```diff
@@ -349,10 +349,10 @@ function renderWebIDL(idlElement, index) {
  */
 export function addIDLHeader(pre) {
   addHashId(pre, "webidl");
-  const header = hyperHTML`<div class="idlHeader"><a
+  const header = hyperHTML`<span class="idlHeader"><a
```
saschanaz: Do we need this span? Can we merge this with the child anchor?

sidvishnoi: We later add the webidl clipboard button in this wrapper

PR #2704, opened 2020-01-08
feat(show-people): add Korean translation

saschanaz:
Should help #2699.

### DIFF
--- a/src/w3c/templates/show-people.js
+++ b/src/w3c/templates/show-people.js
@@ -9,10 +9,25 @@ import { hyperHTML as html } from "../../core/import-maps.js";
 
 const localizationStrings = {
   en: {
-    until: "Until",
+    until(date) {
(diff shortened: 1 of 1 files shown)


src/w3c/templates/show-people.js
```diff
@@ -9,10 +9,25 @@ import { hyperHTML as html } from "../../core/import-maps.js";
 
 const localizationStrings = {
   en: {
-    until: "Until",
+    until(date) {
+      return html`
```
sidvishnoi: Why do we need `html`?

saschanaz: Because `data` is `HTMLTimeElement` 👀

sidvishnoi: Oops. Didn't see that.

PR #2709, opened 2020-01-09
feat(core/ui): trap focus inside modals

SuyashSalampuria:


SuyashSalampuria:
The only problem I am facing now is the security issue
```
Permission denied to access property "document" on cross-origin object
```
when trying to get the contentDocument of Iframe. I will get through it by tommorow

SuyashSalampuria:
@saschanaz @sidvishnoi I researched a lot but came to know that Cross site contents can not be read by javascript. No major browser will allow you that.So, it won't be possible to trap focus for an iframe

saschanaz:
We might want to add invisible focusable elements to allow trapping.

sidvishnoi:
Button added by https://github.com/w3c/respec/pull/2697 will be focussable, right?

saschanaz:
> Button added by #2697 will be focusable, right?

Yes but it won't be sufficient as we also need an element *after* iframe.

SuyashSalampuria:
Shall I add something like 
<button style="background:transparent; border:none; color:transparent;">invisible</button>
at the end of the modal.
The problem with this way is that the focus will be lost according to the user when it will be focussing on this button.
Do you have a better way @sidvishnoi  @saschanaz

sidvishnoi:
> the focus will be lost according to the user

Correct. 

Can we add `iframe` to `focusableEls` in `trapFocus(element)`? `iframe` is focusable, so we should be able to cycle focus without adding invisible element. Can you try?

sidvishnoi:
...that might not work. One way could be: after tabbing, if `document.activeElement` is our invisible element, immediately call `firstFocusableEl.focus()`

SuyashSalampuria:
> ...that might not work. One way could be: after tabbing, if `document.activeElement` is our invisible element, immediately call `firstFocusableEl.focus()`

Any children of elements with display:none  are not in the focus order.
So, we need to check using different styles actually

SuyashSalampuria:
Although ,I have pushed it, But I dont find satisfactory behaviour at all. Please suggest if you find better ways!!

saschanaz:
IMO we don't need any special handling for iframe (as it won't work). Maybe insert a footer, e.g. "File an issue" button?

SuyashSalampuria:
Is this kind of behaviour acceptable ? I will refactor the code well

saschanaz:
LGTM for the behavior. I think we should insert the additional focusable item only in the definition search, though.

src/core/ui.js
```diff
@@ -234,6 +236,17 @@ export const ui = {
     overlay.addEventListener("click", () => this.closeModal(currentOwner));
     overlay.classList.toggle("respec-show-overlay");
     modal.hidden = false;
+    if (content.tagName === "IFRAME") {
+      modal.insertAdjacentHTML(
```
sidvishnoi: I've an idea - add a footer on iframe one.: a link that asks: "open this UI in new tab"
WDYT @saschanaz ?

saschanaz: That sounds good.

saschanaz: BTW, I'd prefer to insert the whatever footer in ui/search-xref rather than core/ui.

sidvishnoi: @SuyashSalampuria 
Added a link after iframe with https://github.com/w3c/respec/pull/2712, no need to add invisible button anymore.

SuyashSalampuria:
Closes #884

src/core/ui.js
```diff
@@ -96,7 +96,9 @@ function trapFocus(element) {
   );
   const firstFocusableEl = focusableEls[0];
   const lastFocusableEl = focusableEls[focusableEls.length - 1];
-
+  if (focusableEls) {
```
sidvishnoi: `focusableEls` is always truthy
```suggestion
  if (firstFocusableEl) {
```

sidvishnoi:
Focus is not getting trapped in `About ReSpec` popup. Can you confirm?
Overall, this is really awesome! Great work @SuyashSalampuria :100:

SuyashSalampuria:
> Focus is not getting trapped in `About ReSpec` popup. Can you confirm?
> Overall, this is really awesome! Great work @SuyashSalampuria 100

Indeed it is not . Let  me check why?

sidvishnoi:
I just checked. When `trapFocus` is called, the modal only has the close button. It takes a few moments to render rest of it, so trapping doesn't work.

sidvishnoi:
Can you try moving https://github.com/w3c/respec/blob/4b6ef28013c2caaecbd856143282cde1840596e8/src/ui/about-respec.js#L36 to be after the "render`" call?

SuyashSalampuria:
> Can you try moving
> 
> https://github.com/w3c/respec/blob/4b6ef28013c2caaecbd856143282cde1840596e8/src/ui/about-respec.js#L36
> to be after the "render`" call?

It worked absolutely!!

src/ui/about-respec.js
```diff
@@ -33,7 +33,7 @@ const button = ui.addCommand(
 );
 
 function show() {
-  ui.freshModal(`${l10n.about_respec} - ${window.respecVersion}`, div, button);
+  // ui.freshModal(`${l10n.about_respec} - ${window.respecVersion}`, div, button);
```
sidvishnoi:
```suggestion
```

SuyashSalampuria:
@saschanaz @sidvishnoi I see that the time taken for loading the respec fill is nearly 3 seconds, which is too high. Any idea on how we can decrease it

saschanaz:
>I see that the time taken for loading the respec fill is nearly 3 seconds, which is too high. Any idea on how we can decrease it

You might want to click the About menu item and see what modules take too much time. Maybe there can be some lines to optimize.

Alternatively, we might want to run some modules in parallel, as currently all modules run in serial way.

PR #2710, opened 2020-01-12
l10n(core/inlines): RFC2119 for German specs (2119de)

sebastian-meyer:
This adds support for German keywords when checking for RFC2119 usage.

German keywords according to https://github.com/adfinis-sygroup/2119/blob/master/2119de.rst


src/w3c/conformance.js
```diff
@@ -1,13 +1,56 @@
...
+      "nicht normativ. Alle andere Angaben sind normativ.",
+    keywordInterpretation(keywords, plural) {
+      return html`
+        <p>
+          ${plural ? "Die Schlüsselwörter" : "Das Schlüsselwort"} ${keywords} in
+          diesem Dokument ${plural ? "sind" : "ist"} gemäß
+          <a href="https://tools.ietf.org/html/bcp14">BCP 14</a>
+          ${renderInlineCitation("RFC2119")} ${renderInlineCitation("RFC8174")}
+          und unter Berücksichtigung von
+          <a href="https://goo.gl/6QZH4J">2119de</a>
```
saschanaz:
```suggestion
          <a href="https://github.com/adfinis-sygroup/2119/blob/master/2119de.rst">2119de</a>
```

No need to use shortened URL here.

BTW, it seems it's not an official W3C translation, am I right? I think we need an official translation, but I would wait for @marcoscaceres to make sure.

sebastian-meyer: Is there an official translation? I'd rather use that, but I didn't find any.

src/w3c/conformance.js
```diff
@@ -1,13 +1,59 @@
...
+        <p>
+          ${plural ? "Die Schlüsselwörter" : "Das Schlüsselwort"} ${keywords} in
+          diesem Dokument ${plural ? "sind" : "ist"} gemäß
+          <a href="https://tools.ietf.org/html/bcp14">BCP 14</a>
+          ${renderInlineCitation("RFC2119")} ${renderInlineCitation("RFC8174")}
+          und unter Berücksichtigung von
+          <a
+            href="https://github.com/adfinis-sygroup/2119/blob/master/2119de.rst"
+            >2119de</a
+          >
```
sebastian-meyer: This looks really weird now, but that's the only way the linter is happy with...

marcoscaceres: Probably just a prettier bug... no big deal.

src/w3c/conformance.js
marcoscaceres: Yeah, it's a bit unfortunate that there is no official translation. However, we would need to bug the IETF about that. I'm ok with going with this for now tho.

PR #2711, opened 2020-01-13
feat(ui/dfn-list): mark exported and unused definitions

### LINKED ISSUE #904: Lint for definitions that are defined but never used
It should flag things that are defined but never used as potential errors.

### LINKED ISSUE #2686: Exported definitions should be marked
In `ui/dfn-list` (Definitions), exported definitions should be marked differently than non-exported definitions. The UI for marking is for discussion.

janiceshiu:
Closes #2686
Incorporates some part of #904

Thanks for your help, @sidvishnoi!


tests/spec/core/ui-spec.js
```diff
@@ -89,4 +89,40 @@ describe("Core - UI", () => {
...
+    const dfnListButton = doc.getElementById("respec-button-definitions");
+    dfnListButton.click();
+    await new Promise(resolve => setTimeout(resolve));
+
+    const dfns = doc.querySelectorAll("ul.respec-dfn-list li");
+    expect(dfns.length).toBe(4);
+
+    const [dfnBar, dfnBaz, dfnFoo, dfnFoobar] = dfns;
+
+    expect(dfnBar.innerText.includes("(exported) (unused)")).toBeTrue();
```
sidvishnoi: Nit:
```suggestion
    expect(dfnBar.textContent).toContain("(exported) (unused)");
```

tests/spec/core/ui-spec.js
```diff
@@ -89,4 +89,40 @@ describe("Core - UI", () => {
...
+
+    expect(dfnBaz.innerText.includes("(unused)")).toBeTrue();
+    expect(dfnBaz.innerText.includes("(exported)")).toBeFalse();
```
sidvishnoi:
```suggestion
    expect(dfnBaz.textContent).not.toContain("(exported)");
```

tests/spec/core/ui-spec.js
janiceshiu: @sidvishnoi ah - was looking for something like `toContain` and didn't find it in the docs. thanks!

could you explain why you prefer `textContent` over `innerText`?

src/ui/dfn-list.js
```diff
@@ -49,9 +49,34 @@ function show() {
...
   ui.freshModal(l10n.list_of_definitions, ul, button);
 }
+
+/**
+ * If a definition is exported, label it accordingly
+ * @param {HTMLElement} dfn a definition
+ */
+function labelDfnIfExported(dfn) {
+  if (dfn.hasAttribute("data-export")) {
+    return hyperHTML`(exported)`;
```
sidvishnoi: So we can style this.
```suggestion
    return hyperHTML`<span class="exported">(exported)</span>`;
```

tests/spec/core/ui-spec.js
sidvishnoi: `textContent` vs `innerText`: https://stackoverflow.com/a/35213639
tl;dr: textContent is more performant, innerText capabilities are not needed here. We use textContent in most (all?) tests, so it's also consistent.

janiceshiu: fixed, thanks for explaining!

src/core/link-to-dfn.js
```diff
@@ -193,9 +193,18 @@ function findLinkTarget(target, anchor, titleToDfns, possibleExternalLinks) {
...
+  markDfnAsUsed(dfn);
   return true;
 }
 
+/**
+ * Mark a definition as used once it has been used at least once within a spec
+ * @param {HTMLElement} dfn a definition
+ */
+function markDfnAsUsed(dfn) {
+  dfn.setAttribute("definition-used", "");
```
saschanaz: I would create a weakmap and export it rather than adding an attribute.

src/ui/dfn-list.js
```diff
@@ -49,9 +49,34 @@ function show() {
...
+    return null;
+  }
+}
+
+/**
+ * If a definition is unused, label it accordingly
+ * @param {HTMLElement} dfn a definition
+ */
+function labelDfnIfUnused(dfn) {
+  if (dfn.hasAttribute("definition-used")) {
```
sidvishnoi: I would prefer if we do this test here only, and not in `core/link-to-dfn`, although it's a but more efficient to that way.
Why? We use `isUnused` here only, and a user may not even open the dfn-list, so it makes sense to keep that logic localized here (makes it hard to reason about) and not do unnecessary work in link-to-dfn.

We can do something like:
``` js
const isUnused = !document.querySelector(`a[href^="#${dfn.id}"]`);
if (isUnused) {
  return null;
}
return hyperHTML`<span class="unused">(unused)</span>`;
```

janiceshiu: @sidvishnoi do we expect all definitions in specs to have an `id`? 

* the sample spec in `basic.html` only has `<dfn>Foo</dfn>`.
* `<a>Foo</a>` results in `<a href="#dom-foo" class="internalDFN" data-link-type="idl"><code>Foo</code></a>` after respec has done its thing. 
* thus, using code above would mean that terms referenced in other parts of the spec would be incorrectly marked as `unused`

sidvishnoi: > do we expect all definitions in specs to have an id?

Yes. If user doesn't specify an id, ReSpec auto-generates one.

> thus, using code above would mean that terms referenced in other parts of the spec would be incorrectly marked as unused

As UI is enabled after all plugins have processed, we are certain that ID generation and linking have been done already.

janiceshiu: @sidvishnoi ah i see the issue. XD your code should've been `if (isUnused) { return hyperHTML(unused) etc}` that's why my test started failing as well. XD

janiceshiu:
Note: I thought that the changes in https://github.com/w3c/respec/pull/2711/commits/0a0eb9154eb3b2453aa113f1f3c76663264f5741 were just a refactoring. However, I noticed that those changes weren't a refactor - they also changed functionality.

for example, in the sample spec in `basic.html`,
* using the old method of marking the term with an attribute if `findLinkTarget` returns `true`
  * the terms `bar`, `Foo.constructor()` and `doTheFoo()` are marked as `(unused)`
* using the new method of `document.querySelector` for `a href`
  * no terms are marked as `(unused)`

@sidvishnoi mentioned he'd look into it. commenting here in case anyone else has an explanation. ^^

assets/dfn-panel.css
```diff
@@ -3,6 +3,25 @@ dfn {
   cursor: pointer;
 }
 
+span.exported, span.unused {
```
sidvishnoi: These show go in https://github.com/w3c/respec/blob/develop/assets/ui.css

assets/ui.css
```diff
@@ -426,3 +426,22 @@
...
+
+.respec-dfn-list li span.exported, span.unused {
+  margin-left: 8px;
+  padding: 0.1em;
+  text-align: center;
+  white-space: nowrap;
+}
+
+.respec-dfn-list li span.exported {
+  background: hsl(205.7, 70%, 96.1%);
```
sidvishnoi: Nit: Lets round-off all the colors.
```suggestion
  background: hsl(206, 70%, 96%);
```

src/ui/dfn-list.js
```diff
@@ -49,9 +49,33 @@ function show() {
...
   ui.freshModal(l10n.list_of_definitions, ul, button);
 }
+
+/**
+ * If a definition is exported, label it accordingly
+ * @param {HTMLElement} dfn a definition
+ */
+function labelDfnIfExported(dfn) {
+  if (dfn.hasAttribute("data-export")) {
+    return hyperHTML`<span class="exported">(exported)</span>`;
```
sidvishnoi: Maybe remove the parenthesis? 
```suggestion
    return hyperHTML`<span class="exported">exported</span>`;
```

assets/ui.css
```diff
@@ -426,3 +426,22 @@
 #xref-ui:not(.ready) {
   background: url("https://respec.org/xref/loader.gif") no-repeat center;
 }
+
+.respec-dfn-list li span.exported, span.unused {
+  margin-left: 8px;
```
sidvishnoi: Nit: for consistency with rest of styles.
```suggestion
  margin-left: 0.5em;
```

assets/ui.css
saschanaz:
```suggestion
  background: #eef6fc;
```

src/ui/dfn-list.js
janiceshiu: imho keeping the parenthesis is good in case the styling doesn't jive with whatever the user is using to read the specs. otherwise they might think that `exported` or `unused` is part of the definition name.

sidvishnoi: I don't have a strong opinion on this. Happy to keep them, happy to remove them. 
BTW, if styling didn't work, most likely JS didn't work either.

janiceshiu: hm. also thinking of colour-blindness and stuff. for example, there's a type of colour-blindness called `Achromatopsia` where vision is essentially monochromatic. folk like that would probably appreciate the parentheses.

assets/ui.css
janiceshiu: changed the lot to hex in https://github.com/w3c/respec/pull/2711/commits/2e75a7b1b4d11d14de299f49fa1691915ddf1547

assets/ui.css
```diff
@@ -427,6 +427,25 @@
   background: url("https://respec.org/xref/loader.gif") no-repeat center;
 }
 
+.respec-dfn-list li span.exported, span.unused {
```
sidvishnoi:
```suggestion
.respec-dfn-list li span.exported, .respec-dfn-list li span.unused {
```
or:
``` suggestion
.respec-dfn-list li span {
```

src/ui/dfn-list.js
```diff
@@ -49,9 +49,33 @@ function show() {
...
+  }
+  return null;
+}
+
+/**
+ * If a definition is unused, label it accordingly
+ * @param {HTMLElement} dfn a definition
+ */
+function labelDfnIfUnused(dfn) {
+  const isUsed = document.querySelector(`a[href^="#${dfn.id}"]`);
```
sidvishnoi: Nit: maybe use this for consistency with labelDfnIfExported
```js
  const isUsed = document.querySelector(`a[href^="#${dfn.id}"]`);
  if (!isUsed) {
    return hyperHTML`<span class="unused">(unused)</span>`;
  }
  return null;
```

janiceshiu: hm. one `!` so i'm still okay with it. XD

if we're going with that, then for overall consistency, we should change `labelDfnIfExported/1` to
```javascript
function labelDfnIfExported(dfn) {
  const isExported =  dfn.hasAttribute("data-export");
  if (isExported) {
    return hyperHTML`<span class="exported">(exported)</span>`;
  }
  return null;
}
```

src/ui/dfn-list.js
saschanaz: > hm. also thinking of colour-blindness and stuff. for example, there's a type of colour-blindness called `Achromatopsia` where vision is essentially monochromatic. folk like that would probably appreciate the parentheses.

Does the same apply for GitHub labels? Does GitHub do something for them?

assets/ui.css
```diff
@@ -427,6 +427,27 @@
...
+  margin-left: 0.5em;
+  padding: 0.1em;
+  text-align: center;
+  white-space: nowrap;
+  font-size: 90%;
+  border-radius: 0.2em;
+}
+
+.respec-dfn-list li span.exported {
+  background: #fcfbee;
```
saschanaz: image

Hard to tell the difference between unused in no-green/no-red accessibility emulation mode, we might want to adjust colors for better accessibility.

sidvishnoi: I found `#d1edfd` to be a good fit for background for `span.exported`. WDYT?

sidvishnoi: Wow. GitHub shows color preview in comments!

saschanaz: Yes, seems the following works:

```css
.respec-dfn-list li span.unused {
	background: #d1edfd;
	color: #040b1c;
	box-shadow: 0 0 0 0.125em #1ca5f940;
}
```

saschanaz: Can the selector be simplified? I think `.unused` is enough, or maybe `.unused-dfn` for less ambiguity.

sidvishnoi: I think `.respec-dfn-list` based scoping is good idea (makes it easy to keep related code together). We can remove the `li span` part.

janiceshiu:
all updated! @saschanaz if all is good can you approve? thanks!

src/ui/dfn-list.js
janiceshiu: @saschanaz interesting question. i don't know!

saschanaz: Can we go without parentheses for now? Because:

* Wikipedia says total color blindness is not just a normal colorblind but a more serious medical syndrome.
* Even if someone is total colorblind they should still be able to see gray background

So for now I don't think additional visual indication is needed.

assets/ui.css
```diff
@@ -427,6 +427,27 @@
   background: url("https://respec.org/xref/loader.gif") no-repeat center;
 }
 
+.respec-dfn-list li span {
```
saschanaz: I would prefer using a proper classname here rather than `li span`, for example `.dfn-status`.

janiceshiu: @saschanaz makes sense. in case we do have other elements that also match `li span`. added `.dfn-status` in https://github.com/w3c/respec/pull/2711/commits/d5108429caa30bd82e42cc1852b33a35b036fa53!

janiceshiu: @saschanaz makes sense. in case we do have other elements that also match li span. added .dfn-status in d510842!

sidvishnoi:
ReSpec v25.1.0 coming up!

marcoscaceres:
Very cool, @janiceshiu! Nice one.

PR #2715, opened 2020-01-14
fix(core/markdown): retry using reindent() instead

saschanaz:
This reverts commit 903b712465b13f99b8e3c60552d2de2708a46565.


saschanaz:
Specs look okay now 👀, could you double check?

sidvishnoi:
@saschanaz [wot-thing-description](https://respec-preview.netlify.com/?spec=https%3A%2F%2Fw3c.github.io%2Fwot-thing-description%2F&version=https%3A%2F%2Fdeploy-preview-2715--respec-pr.netlify.com%2Frespec-w3c.js) broke :disappointed: 
The case is similar with [payment-request?format=markdown](https://w3c.github.io/payment-request/?format=markdown), both have `format: "markdown"`, but are not using markdown..

saschanaz:
That spec uses arbitrary indentation levels, will open a PR there to fix it.

src/core/utils.js
```diff
@@ -804,6 +805,27 @@ export function children(element, selector) {
...
+  const { previousSibling } = element;
+  if (!previousSibling || previousSibling.nodeType !== Node.TEXT_NODE) {
+    return "";
+  }
+  const index = previousSibling.textContent.lastIndexOf("\n");
+  if (index === -1) {
+    return "";
+  }
+  const slice = previousSibling.textContent.slice(index + 1);
+  if (/[^\s]/.test(slice)) {
```
sidvishnoi:
```suggestion
  if (/\S/.test(slice)) {
```

sidvishnoi:
Checked specs again. Seemed fine.
We've to wait for w3c/wot-thing-description#872 before merging though, as that's the only remaining broken spec, and most impacted also.

src/core/utils.js
```diff
@@ -804,6 +805,27 @@ export function children(element, selector) {
   }
 }
 
+/**
+ * Calculates indentation when the element starts after a newline.
+ * The value will be empty if no newline or any non-whitespace exists after one.
+ * @param {Element} element
+ */
+export function getElementIndentation(element) {
```
sidvishnoi: @saschanaz I'm failing to understand this function. I think the comment can be improved.
Adding some unit tests would be really nice.

sidvishnoi: Ah. Makes sense now. Thanks!

saschanaz:
Merging, as there is only one broken spec with a PR opened.

PR #2716, opened 2020-01-15
chore(tools/builder): use IIFE and remove appendBoilerplate

sidvishnoi:
Reduces uncompressed build size from 320KB to 284KB - thanks to Terser optimizing IIFE better.

### DIFF
--- a/tools/builder.js
+++ b/tools/builder.js
@@ -126,7 +126,7 @@ const Builder = {
     };
     const outputOptions = {
       file: outPath,
-      format: "commonjs",
+      format: "iife",
       sourcemap: true,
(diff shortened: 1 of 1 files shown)


tools/builder.js
```diff
@@ -150,14 +126,15 @@ const Builder = {
     };
     const outputOptions = {
       file: outPath,
-      format: "commonjs",
+      format: "iife",
       sourcemap: true,
+      strict: true,
```
saschanaz: This is redundant as it's true by default per [the documentation](https://rollupjs.org/guide/en/#danger-zone).

PR #2717, opened 2020-01-15
chore(tools/builder): minify html literals

sidvishnoi:
Finds template literals starting with `hyperhtml` (`hyperHTML.bind` / `hyperHTML.wire`), `html` and our commonly used `render`; and minifies them
``` diff
+ 284K    builds/respec-w3c.js // before
- 272K    builds/respec-w3c.js // after
+ 276K    builds/respec-w3c.js // after
```


saschanaz:
Oops, tests collapse 😱

sidvishnoi:
We need to add some whitespace manually... which is kinda right way to do it.

saschanaz:
> We need to add some whitespace manually... which is kinda right way to do it.

I think this is a bug in the minifier, as it should not just remove whitespaces (it should rather collapse it to a single space).

sidvishnoi:
If it requires too many changes on our behalf, we will leave this for now.

saschanaz:
Prettier by default *collapses* whitespace, I hope minifiers provide that option too.

sidvishnoi:
It mostly works now. But the benefits don't seem good as before. Maybe we should leave it as it is?

saschanaz:
There is a pending prettier PR to insert less newlines, which will further reduce the size difference. Maybe we can wait and see.

sidvishnoi:
Lets wait for it then.

sidvishnoi:
Here is the size change (in bytes, gzipped in second column). Worth adding the dependencies?
``` diff
- 286128 91092 
+ 278295 89777
= 7833   1315
```

saschanaz:
I vote to wait as I'm still doing followups as the last PR skipped some modules because of test failures.

marcoscaceres:
I don't mind waiting either.

tools/builder.js
```diff
@@ -128,6 +128,19 @@ const Builder = {
       input: require.resolve(`../profiles/${name}.js`),
       plugins: [
         !debug && require("rollup-plugin-terser").terser(),
+        !debug &&
+          require("rollup-plugin-minify-html-literals").default({
+            options: {
+              shouldMinify({ tag = "" }) {
+                return /^(hyperhtml|html|render)/i.test(tag);
```
marcoscaceres, line 135: Maybe a good opportunity to squash any `hyperhtml` tags?

sidvishnoi:
Current state:
| | Before | After | Diff |
| -- | --: | --: | --: |
| uncompressed | 297965 | 289158 | -8807 |
| gzip | 93900 | 92471 | -1429 |
| GZIP=-9 | 93899 | 92335 | -1565 |

Brotli size is 80143 though (12328 less than gzip version :exploding_head:)

saschanaz:
Yes, do brotli right now! Should we contact someone to do that?

sidvishnoi:
@marcoscaceres should know whom to contact

marcoscaceres:
We'd probably want to talk to @deniak and/or @plehegar.

deniak:
I don't see any problem with supporting brotli. @vivienlacourba WDYT?

deniak:
Systems team is planning to discuss that topic next week. I'll let you know what is the plan about supporting brotli on our servers.

marcoscaceres:
Thanks for the update @deniak!

deniak:
Because of some security concerns, the systeam was reluctant to add brotli support on our entire website but we agreed to enable it for specific resources like respec which is now done.
```sh
$ curl -H "Accept-Encoding: br" -I -L https://www.w3.org/Tools/respec/respec-w3c | grep content
content-location: respec-w3c.js.br
content-length: 81611
content-type: application/javascript
content-encoding: br
```

sidvishnoi:
Thanks @deniak! Can we have https://www.w3.org/Tools/respec/respec-highlight.js with brotli also?

deniak:
> Thanks @deniak! Can we have https://www.w3.org/Tools/respec/respec-highlight.js with brotli also?

Done!

sidvishnoi:
@deniak, would it be possible to support the version with extensions also? ReSpec loads the highlight script with `.js` extension.
If it's not simple, we can do it on our end. 

``` bash
$ curl -H "Accept-Encoding: br" -I -L https://www.w3.org/Tools/respec/respec-highlight
# content-encoding: br
$ curl -H "Accept-Encoding: br" -I -L https://www.w3.org/Tools/respec/respec-highlight.js
# no content encoding
```

deniak:
@sidvishnoi this is a bit tricky. We currently rely Apache `Multiviews` for content negotiation  and this only works when a [file doesn't exist on the server](https://httpd.apache.org/docs/2.4/mod/mod_negotiation.html#multiviews). That's why when you request the `.js` version, the file actually exists and the content negotiation doesn't work.
We could use a map but this is something we are trying to avoid for different reasons.

sidvishnoi:
No problem. We'll remove the extension from ReSpec.
Thanks for adding brotli support!

sidvishnoi:
@saschanaz Should we bring this in? Can you share what changes in other files were left?

saschanaz:
I don't have any objection now, but brotli size comparison should be helpful.

sidvishnoi:
|        | Before |  After | Change |
| ------ | -----: | -----: | -----: |
| Brotli |  83675 |  82739 |   -936 |
| Gzip   |  96840 |  95715 |  -1125 |
| Normal | 304508 | 298280 |  -6228 |

saschanaz:
I'm also leaning to skipping this. I think @marcoscaceres thinks every KB counts, though 👀 (And I don't disagree.)

marcoscaceres:
Yeah, the change seems too small to warrant adding another dependency tbh.

sidvishnoi:
Closing as discussed above.

PR #2718, opened 2020-01-15
fix(assets/ui.css): improve focus styles for export buttons

### LINKED ISSUE #2713: Improve styling of anchor tags on focussing upon them
* URL to affected spec: https://github.com/w3c/respec/blob/develop/src/ui/save-html.js

When the anchor tags namely-HTML, XML, EPUB3 are in focus, they are not clear.
We need better styling . 
(They open on clicking export after respec pill)

ridhishjain:
fixes #2713


assets/ui.css
```diff
@@ -143,6 +147,11 @@
   padding-top: 16px;
 }
 
+.respec-save-button:link:focus {
+  color: white;
+  background: rgb(42, 90, 168);
```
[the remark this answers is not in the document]
ridhishjain: instead of changing gradient, why not toy with `outline` property??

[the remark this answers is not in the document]
sidvishnoi: A darker shade and default outline would be good enough.

sidvishnoi:
@ridhishjain We follow [AngularJS commit message conventions](https://gist.github.com/stephenparish/9941e89d80e2bc58a153#format-of-the-commit-message). Please try to follow same for PR titles and commit messages, as it makes easier to follow what's the scope of changes. Look at previous commits and PRs for reference.

assets/ui.css
```diff
@@ -9,6 +9,10 @@
   background: transparent;
 }
 
+.respec-modal .close-button:focus {
```
saschanaz:
```suggestion
.respec-modal .close-button:focus:not(:focus-visible) {
```

Because we do want to show outline when navigating through keyboard buttons.

ridhishjain: @saschanaz this doesn't seem to solve the problem of close button being focused initially

assets/ui.css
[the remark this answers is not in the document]
ridhishjain: Screenshot (67)

see the XML tag getting focused...looking good?

assets/ui.css
saschanaz: Ah yes, [no browser support yet](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible#Browser_compatibility)

As it's not breaking anything, I would just wait for the browser support rather than just permanently disabling outline.

assets/ui.css
```diff
@@ -9,6 +9,10 @@
   background: transparent;
 }
 
+.respec-modal .close-button:focus() {
```
saschanaz: Please remove this, we'll do this when browsers add [`:focus-visible`](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible#Browser_compatibility) support.

assets/ui.css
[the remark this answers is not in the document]
saschanaz: I don't like the shadow, wouldn't changing background color be more intuitive?

assets/ui.css
saschanaz: Focusing on close button by default sounds weird, but not sure what to do.

assets/ui.css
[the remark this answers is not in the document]
ridhishjain: Screenshot (69)

Looking at this, I'll vote for shadow 😅 ... maybe reducing the effect will be fine?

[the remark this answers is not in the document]
sidvishnoi: Above looks fine to me.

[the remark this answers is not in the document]
ridhishjain: I think the same effect should also be added to mouse hover then??

assets/ui.css
```diff
@@ -143,6 +143,11 @@
   padding-top: 16px;
 }
 
+.respec-save-button:link:focus {
```
saschanaz:
```suggestion
.respec-save-button:focus {
```

No need to use `:link` as every button has one. I think other rules also should remove it.

assets/ui.css
```diff
@@ -143,6 +143,11 @@
...
+  color: white;
```
saschanaz: This seems redundant, the text color is already white without this line.

assets/ui.css
```diff
@@ -143,6 +143,11 @@
...
+  background: rgba(25, 55, 102);
```
saschanaz:
```suggestion
  background: #193766;
```

Shorter 👀

assets/ui.css
ridhishjain: link is necessary here, it won't work otherwise, tested it.

saschanaz: Ooops, that's because of base.css. Okay, let's keep it here.

PR #2720, opened 2020-01-16
refactor(core/utils): no `arguments`

saschanaz:
Can't remove this horrible function as [several specs already are using it](https://github.com/search?l=HTML&q=respec+oninclude+org%3Aw3c&type=Code) 🤦‍♀️

### DIFF
--- a/src/core/utils.js
+++ b/src/core/utils.js
@@ -403,16 +403,11 @@ export function linkCSS(doc, styles) {
  * @this {any}
  * @param {string} [flist]
  */
-export function runTransforms(content, flist) {
-  let args = [this, content];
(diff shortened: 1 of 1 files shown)


src/core/utils.js
```diff
@@ -403,16 +403,11 @@ export function linkCSS(doc, styles) {
  * @this {any}
  * @param {string} [flist]
  */
-export function runTransforms(content, flist) {
-  let args = [this, content];
-  const funcArgs = Array.from(arguments);
-  funcArgs.shift();
-  funcArgs.shift();
-  args = args.concat(funcArgs);
```
sidvishnoi: That looked archaeological :smile:

PR #2722, opened 2020-01-16
perf(core/xref): remove duplicate cite context to reduce query size

sidvishnoi:
On payment-request, the cite/spec context reduced from 1265 specs to 1023 specs and size of entire query reduced from 29833 bytes to 26215 bytes.
This also increases chances of getting common queries across specs on same domain - so we can use more cached results (though still rare).

The basic idea is to make `specs.flat().length` equal to `new Set(specs.flat()).size)`

### DIFF
--- a/src/core/xref.js
+++ b/src/core/xref.js
@@ -130,48 +130,29 @@ function normalizeConfig(xref) {
 function getRequestEntry(elem) {
   const isIDL = "xrefType" in elem.dataset;
 
-  const term = getTermFromElement(elem, isIDL);
-  const specs = getSpecContext(elem);
-  const types = getTypeContext(elem, isIDL);
-  const forContext = getForContext(elem, isIDL);
+  let term = getTermFromElement(elem);
+  if (!isIDL) term = term.toLowerCase();
 
-  return {
-    term,
-    types,
-    ...(specs.length && { specs }),
-    ...(typeof forContext === "string" && { for: forContext }),
-  };
-}
-
-/**
- * @param {HTMLElement} elem
- * @param {boolean} isIDL
- */
-function getTermFromElement(elem, isIDL) {
-  const { lt: linkingText } = elem.dataset;
-  let term = linkingText ? linkingText.split("|", 1)[0] : elem.textContent;
-  term = normalize(term);
-  if (term === "the-empty-string") {
-    term = "";
-  }
-  return isIDL ? term : term.toLowerCase();
-}
-
-/**
- * Get spec context as a fallback chain, where each level (sub-array) represents
- * decreasing priority.
- * @param {HTMLElement} elem
- */
-function getSpecContext(elem) {
   /** @type {string[][]} */
   const specs = [];
-
   /** @type {HTMLElement} */
   let dataciteElem = elem.closest("[data-cite]");
-
-  // If element itself contains data-cite, we don't take inline context into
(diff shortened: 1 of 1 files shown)


sidvishnoi:
@saschanaz ICYMI

src/core/xref.js
```diff
@@ -154,9 +159,11 @@ function getRequestEntry(elem) {
       ? closestSection.querySelectorAll("a.bibref")
       : [];
     const inlineRefs = [...bibrefs].map(el => el.textContent.toLowerCase());
-    const uniqueInlineRefs = [...new Set(inlineRefs)].sort();
+    const uniqueInlineRefs = [...new Set(inlineRefs)].filter(
+      spec => spec && !(specs[specs.length - 1] || []).includes(spec)
```
sidvishnoi:
```suggestion
      spec => spec && !(specs[0] || []).includes(spec)
```

src/core/xref.js
```diff
@@ -139,9 +139,14 @@ function getRequestEntry(elem) {
   let dataciteElem = elem.closest("[data-cite]");
   while (dataciteElem) {
     const cite = dataciteElem.dataset.cite.toLowerCase().replace(/[!?]/g, "");
-    const cites = cite.split(/\s+/).filter(s => s);
-    if (cites.length) {
-      specs.push(cites.sort());
+    const cites = cite.split(/\s+/);
+    // if we already have a spec in previous level of fallback chain, skip it
```
saschanaz: By "previous" it means "higher" right?

sidvishnoi: Lower level actually - as we're traversing up towards root. Want me to change the comment?

> Can this be a separate function

Will do in another PR.

saschanaz: > Lower level actually

But that level has higher priority, right?

>Will do in another PR.

I would prefer doing it here if no blocker 👀

sidvishnoi: Lower level is higher priority - correct.

> I would prefer doing it here if no blocker :eyes:

:+1:

saschanaz: It was initially a bit ambiguous for me what "previous" mean there. Maybe just "subtree"?

sidvishnoi:
@saschanaz Sorry about large refactor - couldn't resist refactoring all of them instead of just `getSpecContext` :sweat_smile: 
I have split it into commits, so it's easier to follow.

src/core/xref.js
```diff
@@ -202,6 +171,52 @@ function getTermFromElement(elem) {
...
+    const cite = dataciteElem.dataset.cite.toLowerCase().replace(/[!?]/g, "");
+    const cites = cite.split(/\s+/).filter(s => s);
+    // Optimization: if we already have a spec in a higher priority level of
+    // fallback chain, skip it from this level
+    const uniqueCites = [...new Set(cites)].filter(
+      spec => !(specs[specs.length - 1] || []).includes(spec)
+    );
+    if (uniqueCites.length) {
+      specs.push(uniqueCites.sort());
+    }
```
saschanaz: I meant a separate function to make the array items unique and thought it could be done within this PR 👀. For unrelated refactors I prefer a separate PR. Sorry to confuse you 😔

sidvishnoi: Oops :man_facepalming: Let me know if the current update is what you mean

PR #2724, opened 2020-01-16
feat(core/inlines): extend bikeshed like section links

### LINKED ISSUE #2708: Short syntax to link to fragment inside other spec
## Important info

* URL to affected spec: https://github.com/w3c/fingerprinting-guidance/pull/39/files
* ReSpec version: 25.0.4

## Description of problem

What happened?: 

The syntax `<a data-cite="?HTML/webstorage.html#user-tracking">[HTML#user-tracking]</a>` links to a fragment inside the HTML spec, but `[[[?HTML/webstorage.html#user-tracking]]]` does not.

But what I was expecting?:

`[[[spec#frag]]]` or `[[spec#frag]]` interprets its content like a `data-cite` attribute.

janiceshiu:
Closes #2708

Allows using `[[[SPEC/PATH#frag]]]` to be used as a shorthand to `<a data-cite="SPEC/PATH#frag">[SPEC/PATH#frag]</a>`. 
*Important*: This feature is recommended only for to link to sections/headings. For linking exported terms, such as definitions and IDL content, please use `[= term =]` and `{{ IDL }}` syntax respectively.

Not supporting `[[[?HTML/webstorage.html#user-tracking | replacement text content]]]` as it doesn't give any benefit over  existing `<a data-cite="?HTML/webstorage.html#user-tracking">replacement text content</a>` syntax.

Thanks for your help again, @sidvishnoi!


src/core/data-cite.js
```diff
@@ -59,7 +59,14 @@ function requestLookup(conf) {
     switch (elem.localName) {
       case "a": {
         if (elem.textContent === "" && elem.dataset.lt !== "the-empty-string") {
-          elem.textContent = title;
+          // support bikeshed-like section links https://tabatkins.github.io/bikeshed/#section-links
+          // eg: [[[spec/path.html#frag]]] becomes <a href="spec/path.html#frag">[spec/path#frag]</a>
+          if (originalKey.includes("#")) {
+            // ! and ? are for respec to check whether something is normative or not. readers do not need it.
```
sidvishnoi:
```suggestion
```

src/core/data-cite.js
```diff
@@ -59,7 +59,14 @@ function requestLookup(conf) {
     switch (elem.localName) {
       case "a": {
         if (elem.textContent === "" && elem.dataset.lt !== "the-empty-string") {
-          elem.textContent = title;
+          // support bikeshed-like section links https://tabatkins.github.io/bikeshed/#section-links
+          // eg: [[[spec/path.html#frag]]] becomes <a href="spec/path.html#frag">[spec/path#frag]</a>
```
sidvishnoi:
```suggestion
          // eg: [[[spec/path.html#frag]]] becomes <a href="url-to-spec/path.html#frag">[spec/path#frag]</a>
```

tests/spec/core/inlines-spec.js
```diff
@@ -224,6 +224,36 @@ describe("Core - Inlines", () => {
     expect(badOne.textContent).toBe("[[[#does-not-exist]]]");
   });
 
+  it("processes normative and informative `[[[spec/path#fragment]]]` into a link with the correct data-cite and text content", async () => {
+    const body = `
+      <p id="output">
+        [[[!HTML/path.html#some-normative-fragment]]]
```
sidvishnoi: Lets use a real world example maybe?
```suggestion
        [[[!HTML/webstorage.html#user-tracking]]]
```

tests/spec/core/inlines-spec.js
```diff
@@ -224,6 +224,36 @@ describe("Core - Inlines", () => {
     expect(badOne.textContent).toBe("[[[#does-not-exist]]]");
   });
 
+  it("processes normative and informative `[[[spec/path#fragment]]]` into a link with the correct data-cite and text content", async () => {
```
sidvishnoi:
```suggestion
  it("supports bikeshed like `[[[spec/path#fragment]]]`section links", async () => {
```

tests/spec/core/inlines-spec.js
```diff
@@ -224,6 +224,36 @@ describe("Core - Inlines", () => {
...
+    `;
+
+    const doc = await makeRSDoc(makeStandardOps(null, body));
+    const anchors = doc.querySelectorAll("#output a");
+
+    expect(anchors.length).toBe(2);
+
+    const [normativeAnchor, informativeAnchor] = anchors;
+
+    expect(normativeAnchor.textContent).toBe(
```
sidvishnoi: Maybe skip the normative/informative part in tests. We're testing the inline section like syntax, not whether normative/informative works here.
```suggestion
    expect(anchors[0].textContent).toBe(
```

janiceshiu: Hm. I like naming it because it explains the different cases we're catering for and why we support them. In this case, the cases are `[[[spec/path#frag]]]`, `[[[!spec/path#frag]]]` and `[[[?spec/path#frag]]]` without us having to explicitly say so.

^ ...and that third case with no `?` or `!` doesn't have a test. Time to add it.

src/core/data-cite.js
janiceshiu: Could you explain why you'd rather remove this? 

I added it because it explains the `why` of `why are we supporting [[[?spec/path#frag]]] and [[[!spec/path#frag]]] in addition to [[[spec/path#frag]]] ?`

sidvishnoi: It's kinda _obvious_, we support this everywhere we have a spec citation.

janiceshiu: @sidvishnoi i'm new to the project and to specs, and it isn't obvious to me. I poked through documentation and the closest I found was

`Normative and informative references work similarly for [[[FOO]]] as they work for [[FOO]], and [[[?FOO]]] can be used to have a non-normative reference in a normative section.`

in the [respec editor's guide](https://github.com/w3c/respec/wiki/ReSpec-Editor's-Guide)

Perhaps the solution is to beef up documentation, since I could not find a reference to `[[!FOO]]` or something similar in the docs. This is something both editors and devs should be aware of.

marcoscaceres: I don't mind these comments so much if they help with the flow of understanding the code. Even though they are obvious, they indeed would help new contributors understand a bit more about what is going on. Having said that, the "!" is sorta/kinda deprecated, because ReSpec does "smart citations" based on context.

janiceshiu: @marcoscaceres hm. in that case should we account for a deprecated feature in a new feature? leaning towards no if we want it to be really deprecated and eventually gone.

marcoscaceres: Not sure I follow? Just to be clear, we can never drop support for "!" because there are potentially hundreds of specs using it (and a lot of those specs are no longer maintained but important nonetheless), but we actively discourage people from using "!" via spec reviews, etc.

tests/spec/core/inlines-spec.js
marcoscaceres: @sidvishnoi  right, we only care about output in this part, not how the reference is classified. We test ? ! classifications in other tests so testing for those is redundant here.

tests/spec/core/inlines-spec.js
```diff
@@ -224,6 +224,44 @@ describe("Core - Inlines", () => {
...
+
+    const doc = await makeRSDoc(makeStandardOps(null, body));
+    const anchors = doc.querySelectorAll("#output a");
+
+    expect(anchors.length).toBe(3);
+
+    const [normativeAnchor, informativeAnchor, generalAnchor] = anchors;
+
+    expect(normativeAnchor.textContent).toBe(
+      "[HTML/webstorage.html#user-tracking]"
```
marcoscaceres: This doesn't look right.... I was expecting just "[HTML]" as we never show fragments or the document to end-users (the link itself is fine though). Do you have a link to a spec that does this or the BikeShed documentation?

marcoscaceres:
Ok, I see... yes, this needs to expand correctly... as seen in the BikeShed documentation: 

```HTML
[[css-flexbox-1#auto-margins]]
[[CSS-access-19990804#Features]]
```

> which renders as:

```HTML
<a href="https://drafts.csswg.org/css-flexbox-1/#auto-margins">CSS Flexbox 1 §8.1 Aligning with auto margins</a>
<a href="http://www.w3.org/1999/08/NOTE-CSS-access-19990804#Features">Accessibility Features of CSS §Features</a>
```

I don't know how we will pull the section names from, for instance, [[[CSS-access-19990804#Features]]].

tests/spec/core/inlines-spec.js
marcoscaceres: Found the docs... commented here https://github.com/w3c/respec/pull/2724#issuecomment-577488110

We would need to expand these out so: 

```
[[css-flexbox-1#auto-margins]]
```

Becomes:

```HTML
<a href="https://drafts.csswg.org/css-flexbox-1/#auto-margins">CSS Flexbox 1 §8.1 Aligning with auto margins</a>
```

@sidvishnoi, do you know if Shepherd has this data?

janiceshiu: fwiw I tried
`bikeshed refs --text=css`
`bikeshed refs --text=flexbox`
`bikeshed refs --text=auto-margins`
in console and didn't get any results.

marcoscaceres: no idea how Bikeshed works :) however, BikeShed pulls data from Shepherd, so in theory this stuff is in the Shepherd DB. If that’s true, we may be able to do this, but will require significant work as we will need to also add support to the respec.org end-points.

sidvishnoi:
We've the data[1], we can add an endpoint as `/xref/by-url` but it's a significant effort for a feature that I'm not sure whether it'll be used much or not.

Though `/xref/by-url` can be useful as a validation effort.

[1] https://github.com/tabatkins/bikeshed-data/tree/master/data/headings

tests/spec/core/inlines-spec.js
janiceshiu: > We test ? ! classifications in other tests
ah okay. wasn't aware about that. i've renamed the variables accordingly.

janiceshiu:
@sidvishnoi @marcoscaceres  I've made the requested changes on tests - let me know if that wasn't what y'all wanted.

Regarding your comments at https://github.com/w3c/respec/pull/2724#issuecomment-577488110 and https://github.com/w3c/respec/pull/2724#issuecomment-577555825, I'm not exactly sure what's the final conclusion - are there additional requirements for this PR? If there are, could you clarify them? Thanks!

marcoscaceres:
I'm going to go ahead and close this, as we don't currently plan to support this. We can reopen if we decide to add server side support.

PR #2727, opened 2020-01-21
fix(core/data-include): follow element indentation

saschanaz:
A trial to prevent the breakage found in #2650.


tests/spec/core/data-include-spec.js
```diff
@@ -81,4 +88,38 @@ describe("Core — Data Include", () => {
     expect(h2.textContent).toBe("1. PASS");
     expect(doc.querySelectorAll("*[data-include]").length).toBe(0);
   });
+
+  it("indents multiline text", async () => {
+    const include = `Blue rose is blue
+
+Red rose is red
+
+Rose of the wasteland is violet
```
marcoscaceres: 🥺😂😭

tests/spec/core/data-include-spec.js
```diff
@@ -81,4 +88,38 @@ describe("Core — Data Include", () => {
...
+      config: {
+        ...makeBasicConfig(),
+        format: "markdown",
+      },
+      body: `
+        ## R
+
+        1. Rose
+
+          <div data-include="${generateDataUrl(
```
marcoscaceres: This is.... insane... I hope no one actually does this in practice! But very cool that it works.

src/core/data-include.js
```diff
@@ -14,28 +14,76 @@ import { runTransforms } from "./utils.js";
...
+function indentTextWithoutFirstLine(text, indent) {
+  const lines = text.split("\n");
+  const firstLine = lines.shift();
+  return `${firstLine}\n${lines.map(line => indent + line).join("\n")}`;
+}
+
+/**
+ * @param {HTMLElement} el
+ * @param {string} data
+ * @param {boolean} replace
```
marcoscaceres: I kinda like it that TypeScript makes boolean traps less bad. Stylistically tho, I wonder if we should still use a property bag here?

tests/spec/core/data-include-spec.js
saschanaz: Insane, yes 😂

Although this makes more sense in more real-world like example:

```
  <section id="sotd">...</section>

  <section data-include="..."></section>
```

Previously no indentation was applied and thus the replacement text went indentation-less, causing incorrect markdown. I couldn't add this test here though, as `normalizePadding` somehow silences it.

marcoscaceres: Make sense.

src/core/data-include.js
```diff
@@ -14,28 +14,77 @@ import { runTransforms } from "./utils.js";
...
+}
+
+/**
+ * @param {string} text
+ * @param {string} indent
+ */
+function indentTextWithoutFirstLine(text, indent) {
+  const lines = text.split("\n");
+  const firstLine = lines.shift();
+  return `${firstLine}\n${lines.map(line => indent + line).join("\n")}`;
```
marcoscaceres: clever :)

marcoscaceres:
(not sure if you were ready for review... but looking pretty good)

saschanaz:
I think it's ready, personally would prefer more tests but probably after replacing normalize Padding.

marcoscaceres:
Should we track down some specs that would make use of this to test?

saschanaz:
Basically nearly all specs that we previously observed with data-include need this before we replace normalize Padding 👀

saschanaz:
@marcoscaceres I tried making this a separate PR but I feel this is half-complete without #2715. Some thoughts:

* I added a test here but it's hypothetical rather than a real-world problem.
* Tried merging this in #2715 and immediately found a bug, of which the fix should be both here and #2715.
* #2715 might become too huge to review when merged

What do you think? Does merging the two PR look okay?

saschanaz:
Merged already there, hope it won't be too huge 🙏

PR #2734, opened 2020-01-29
refactor(core/xref): split getting context into functions

sidvishnoi:


saschanaz:
This removes `dedupeSpecContext`, is it by accident?

src/core/xref.js
```diff
@@ -208,6 +199,28 @@ function getSpecContext(elem) {
...
+  if (!isIDL) {
+    // non-idl terms having for context is not presently supported
+    return null;
+  }
+
+  /** @type {HTMLElement} */
+  const dataXrefForElem = elem.closest("[data-xref-for]");
+  if (dataXrefForElem) {
+    return normalize(dataXrefForElem.dataset.xrefFor);
+  }
```
saschanaz: Seems we can save a `return null` if we move this block into `if (isIDL)` 👀

sidvishnoi:
> This removes dedupeSpecContext, is it by accident?

No. I thought it's better to not add the duplicate specs in first place, instead of removing them later. Added a comment to make it clearer what we're doing and why..

src/core/xref.js
sidvishnoi: Good point.

saschanaz:
> No. I thought it's better to not add the duplicate specs in first place, instead of removing them later. Added a comment to make it clearer what we're doing and why..

I think it's about readability rather than performance. The `getSpecContext` is still long and the deduping can easily be a separate function.

BTW, currently your code creates `uniqueCites`, I think separating that part is enough rather than keeping the mentioned one.

src/core/xref.js
```diff
@@ -156,60 +176,62 @@ function getRequestEntry(elem) {
     const inlineRefs = [...bibrefs].map(el => el.textContent.toLowerCase());
     const uniqueInlineRefs = [...new Set(inlineRefs)].sort();
     if (uniqueInlineRefs.length) {
-      specs.unshift(uniqueInlineRefs);
+      specs.push(uniqueInlineRefs);
```
saschanaz: Does this create a behavior difference?

sidvishnoi: No. Earlier, we were doing this the end of all processing, and then unshifting. Now, we push early. The priority in fallback chain remains same,

sidvishnoi:
Thanks for feedback @saschanaz :100:

Issue #2737, opened 2020-01-29
.hljs-comment, .hljs-quote color does not pass WCAG contrast


carmacleod:
As a follow-on to PR https://github.com/w3c/respec/pull/2683, the `.hljs-comment` color in `highlight.css` does not have enough contrast.

The failing color is `#a0a1a7`, and it fails against a white background, and also against the very light orange `.example` color `#fcfaee` that is used in the ARIA spec, the APG, and possibly others.

```css
.hljs-comment,
.hljs-quote {
  color: #a0a1a7;     // please change this to #717277 - see below for justification
  font-style: italic;
}
```

> Aside: No worries, the previous comment `color: #998;` that was in `github.css` also failed, so this failure was not recently introduced - it has been there for a long time.

The screen snap below shows a comment in example code in the ARIA spec, with DevTools open to see class `.hljs-comment` color `#a0a1a7` failing in the Color Contrast Analyzer.

![image](https://user-images.githubusercontent.com/3331913/73401790-d4839e80-42b9-11ea-8db5-c247a8bfa5a2.png)

If I use the Chrome DevTools Color Picker to help choose a passing color, I can see that color `#717277` would still be considered to be a "light bluish-grey" suitable for comments, but it is more legible, and it passes the color contrast ratio of 4.5:1.

The screen snap below shows the Chrome DevTools Color Picker being used to change the comment color to a passing color.

![image](https://user-images.githubusercontent.com/3331913/73400971-34794580-42b8-11ea-8515-f8024c698540.png)

carmacleod:
Just leaving an FYI note here that there's an [a11y-light theme](https://github.com/highlightjs/highlight.js/blob/b7337391a18276c52b3d55b14117d3b3d6632f77/src/styles/a11y-light.css) on highlight.js now...
and its `.hljs-comment` color is `#696969`... in case you feel like an even darker grey...

```css
.hljs-comment,
.hljs-quote {
  color: #696969;
}
```

marcoscaceres:
Closed via #2738

PR #2738, opened 2020-01-30
Changed .hljs-comment color

Nikhar444:
Fix for #2737


saschanaz:
It indeed does have better contrast, but now it's less distinguishable from normal black text. @sidvishnoi thoughts?

sidvishnoi:
It's slightly less distinguishable, but I think as we've italics, we're good to go.

Nikhar444:
> 
> 
> The new color has much better contrast, thanks!
> 
> Do you mind also adjusting hljs-string?

Thanks
I guess, hljs-string has good enough contrast.

saschanaz:
>hljs-string has good enough contrast.

It also doesn't pass WCAG contrast though:

image

Nikhar444:
@saschanaz @marcoscaceres  Please merge these changes in develop branch.

sidvishnoi:
I found `#0d8601` to be slightly more vibrant than `#42803C`, while still satisfying AA.

saschanaz:
> I found `#0d8601` to be slightly more vibrant than `#42803C`, while still satisfying AA.

Agreed, while it weirdly gets less contrast number (4.57 versus 4.53).

marcoscaceres:
I'm merging this for now and we can fiddle with the color further as needed. Appreciate you doing this @Nikhar444!

marcoscaceres:
@Nikhar444, could you be a super hero and make the same changes to:
https://github.com/highlightjs/highlight.js/blob/master/src/styles/atom-one-light.css

We are just using that one, so we might as well also fix it for everyone at the source.

Nikhar444:
> 
> 
> @Nikhar444, could you be a super hero and make the same changes to:
> https://github.com/highlightjs/highlight.js/blob/master/src/styles/atom-one-light.css
> 
> We are just using that one, so we might as well also fix it for everyone at the source.

Sure @marcoscaceres, I'll make the changes there as well.

Issue #2406, opened 2019-06-27
support "levels"


marcoscaceres:
Specifications often use "levels", such as https://w3c.github.io/page-visibility/ (level 2)... so it would be nice have a `level:` configuration option that: 

 - [ ] combines the level with the shortName (so... page-visibility becomes "page-visibility-2")
 - [ ] automatically appends the "Level X" to the title of the document. 
 - [ ] makes sure all appropriate links also get the level, though these are probably just derived from the shortName.

PR #2741, opened 2020-02-05
Add level config

janiceshiu:
Closes [#2406](https://github.com/w3c/respec/issues/2406)

 - [x] combines the level with the shortName (so... page-visibility becomes "page-visibility-2")
- [x] automatically appends the "Level X" to the title of the document.
- [x] makes sure all appropriate links also get the level, though these are probably just derived from the shortName.

### DIFF
--- a/src/w3c/headers.js
+++ b/src/w3c/headers.js
@@ -273,6 +273,7 @@ export function run(conf) {
 
   if (conf.level) {
     conf.title = `${conf.title} Level ${conf.level}`;
+    conf.shortName = `${conf.shortName}-${conf.level}`;
   }
 
   if (!conf.subtitle) conf.subtitle = "";
--- a/src/w3c/headers.js
+++ b/src/w3c/headers.js
@@ -2,6 +2,7 @@
 // Module w3c/headers
 // Generate the headers material based on the provided configuration.
 // CONFIGURATION
+//  - level: TODO SOME EXPLANATION HERE
 //  - specStatus: the short code for the specification's maturity level or type (required)
 //  - shortName: the small name that is used after /TR/ in published reports (required)
 //  - editors: an array of people editing the document (at least one is required). People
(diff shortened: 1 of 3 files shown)


src/w3c/headers.js
```diff
@@ -2,6 +2,7 @@
 // Module w3c/headers
 // Generate the headers material based on the provided configuration.
 // CONFIGURATION
+//  - level: TODO SOME EXPLANATION HERE
```
janiceshiu: I am still not very clear on what levels are in specs, so I am not sure how to explain it. @marcoscaceres could you suggest something? Thank you!

marcoscaceres: heh, yeah... it's ... complicated... I don't think we need to define what a level *is* because we won't get agreement, more that if a working group wants to add a level to a specification, this is where they can do it.

src/w3c/headers.js
```diff
@@ -269,6 +270,12 @@ export function run(conf) {
   if (document.title && conf.isPreview && conf.prNumber) {
     document.title = `Preview of PR #${conf.prNumber}: ${document.title}`;
   }
+
+  if (conf.level) {
```
marcoscaceres: We probably want to check that level is a decimal, and show an error if it's not.

marcoscaceres: Just noting that we've had Level 0 specs, so we should allow that... basically any sequence of decimals should be fine.

src/w3c/headers.js
janiceshiu: I see. XD I'd still like to add an explanation in this file to follow the convention of all config keys having some explanation. Perhaps something a bit more generic might work?  My suggestion below.
> level: the specification's level. For example, `{level: 2}` (optional)

src/w3c/headers.js
janiceshiu: To clarify - by `decimal` you mean that it is okay to have the level be something like `2.5`, `2.00`, etc, yes?

saschanaz: Probably only zero and positive integers?

janiceshiu:
Closing in favour of [#2758](https://github.com/w3c/respec/pull/2758) because [#2750](https://github.com/w3c/respec/pull/2750) did a lot of refactoring that resulted in lots of conflicts in this branch.

PR #2747, opened 2020-02-12
refactor(core/biblio): improve type support

sidvishnoi:
Would it make sense to add a global `.d.ts` file to annonate `Conf` type?

### DIFF
--- a/src/core/biblio.js
+++ b/src/core/biblio.js
@@ -7,15 +7,9 @@
 /* jshint jquery: true */
 import { biblioDB } from "./biblio-db.js";
 import { createResourceHint } from "./utils.js";
-import { pub } from "./pubsubhub.js";
-export const biblio = {};
 
-/**
- * @typedef {object} Conf
- * @property {Set<string>} informativeReferences
- * @property {Set<string>} normativeReferences
- * @property {Record<string, { aliasOf?: string }>} localBiblio
- */
+/** @type {Conf['biblio']} */
+export const biblio = {};
 
 // for backward compatibity
 export { wireReference, stringifyReference } from "./render-biblio.js";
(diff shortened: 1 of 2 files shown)


marcoscaceres:
> Would it make sense to add a global .d.ts file to annonate Conf type?

That's more a @saschanaz question, I think :)

saschanaz:
Sounds like a good idea.

src/core/biblio.js
```diff
@@ -20,6 +21,7 @@ const bibrefsURL = new URL("https://specref.herokuapp.com/bibrefs?refs=");
 /**
  * Normative references take precedence over informative ones,
  * so any duplicates ones are removed from the informative set.
+ * @param {Conf} conf
```
saschanaz: I wonder we can make modules export a class instead, so that core/base-runner can create instances from modules like `new Biblio(conf)` and we can stop passing `conf` every time we call internal functions.

Just a thought for future refactoring.

src/type-helper.d.ts
```diff
@@ -87,6 +87,14 @@ interface JQuery {
   allTextNodes(exclusions: string[]): Text[];
 }
 
+type BiblioData = { aliasOf?: string; id?: string };
```
saschanaz: I prefer `interface` here, IMO `type` is more about creating aliases or complex types.

src/core/biblio.js
```diff
@@ -92,39 +95,32 @@ export async function resolveRef(key) {
...
-    .concat(localAliases)
-    // remove duplicates
-    .reduce((collector, item) => {
-      if (collector.indexOf(item) === -1) {
-        collector.push(item);
-      }
-      return collector;
-    }, [])
-    .sort();
+  const neededRefs = Array.from(
```
saschanaz: I hope there be something like `Array.prototype.unique()` to do this 🤯

sidvishnoi: I wanted it too! Can we add our own with a name that won't conflict? Like `Array.prototype.rsUnique`

saschanaz: That sounds too hacky 🤔

PR #2748, opened 2020-02-12
chore(tools/netlify): add DINI specs for preview

sidvishnoi:


tools/netlify.js
```diff
@@ -12,11 +12,17 @@ const { DEPLOY_PRIME_URL, COMMIT_REF, REVIEW_ID, REPOSITORY_URL } = process.env;
...
-  "https://w3c.github.io/hr-time/",
-];
+const SPECS = {
+  W3C: [
+    "https://w3c.github.io/payment-request/",
+    "https://w3c.github.io/gamepad/",
+    "https://w3c.github.io/hr-time/",
+  ],
+  DINI: [
+    "https://dini-ag-kim.github.io/oer-service-card/latest/",
```
marcoscaceres: @sebastian-meyer, just a heads up that we will be depending on these URLs. Is that ok?

marcoscaceres: @sebastian-meyer, just to be clear, to see this in action:

Screenshot_2020-02-13_11_28_16

And then, you see the above:

Screenshot 2020-02-13 11 30 21

marcoscaceres: As you add more important specs, we can add them here so you can check your profile.

sidvishnoi: Also, this will work only if you're using w3.org hosted ReSpec scripts.

sebastian-meyer: As far as I know the DINI profile isn't published to w3.org (at least I can't reach it via https://w3.org/Tools/respec/respec-dini), so we currently have a fork at https://github.com/dini-ag-kim/respec which is published via GitHub Pages. We then use the profile by referencing https://dini-ag-kim.github.io/respec/builds/respec-dini.js. Whenever there is a new release of ReSpec, we have to update our fork.

Alternatively, we could directly reference https://w3c.github.io/respec/builds/respec-dini.js, but I wasn't sure, if the GitHub Pages for this repo won't be turned off some day, and wanted to make sure, that linking to the profile would stay persistent.

Or is there any better way of handling this?

sidvishnoi: The DINI profile should've been published to w3.org as soon as we did the release. @marcoscaceres Can you check why it didn't?
https://w3c.github.io/respec/builds/respec-dini.js seems like a good alternative and I don't think the GitHub page will be turned off in a foreseeable future (without us letting know the users). I'll add support for the GitHub pages version in PR preview if we can't add it to w3.org

marcoscaceres: > Can you check why it didn't?

It might be scoped to proxy only specific files on the w3c side.  If we can use gh-pages as an alternative, then we probably should. I'd prefer not to put the profile on the w3c server unless we really have to... but if it's a huge pain, I can ask the W3C to mirror them.

sebastian-meyer: I am OK with using gh-pages. My only concern was having a persistent URL. As long as gh-pages isn't turned off for the w3c/respec Repo this is fine!

sidvishnoi: Updated https://respec-preview.netlify.com/ to support https://w3c.github.io/respec hosted ReSpec scripts (https://github.com/sidvishnoi/respec-preview/commit/f3dd6faf2532a2ded75c44402a19787a5af9d202).

PR #2750, opened 2020-02-12
refactor(core/title): standardise document and h1 title

janiceshiu:
Noticed that if a spec has both a `<title>` and the `h1#title` override, stuff gets inconsistent. To replicate...

* modify `examples/basic.html` and add a `h1#title` element such as `<h1 id="title">override</h1>`
* note that there is already a `<title>Replace me with a real title</title>` in that spec. 
* open the spec in a browser
* The tab bar says `Replace me...` but the dark blue h1 element on the page itself it shows `override`

After discussing with @marcoscaceres and @sidvishnoi, we agreed to remove `conf.title` and update the code to be consistent in its various title forms. During the discussion, I realised that there are a lot of different cases to check so... lots of new specs. ^^

Note: I have not yet modified the DINI profile because it was a very recent addition.


src/w3c/templates/cgbg-headers.js
```diff
@@ -130,8 +129,8 @@ export default conf => {
                     ${[conf.additionalCopyrightHolders]} &amp;
                   `
                 : ""}
-              the Contributors to the ${conf.title} Specification, published by
-              the
+              the Contributors to the ${document.title} Specification, published
```
marcoscaceres: This needs to be a clone of the children of `getSpecTitleElem()`. Otherwise, we lose the `<code>` or other tags.

src/w3c/templates/headers.js
```diff
@@ -89,29 +89,42 @@ const localizationStrings = {
 
 export const l10n = getIntlData(localizationStrings);
 
-function getSpecTitleElem(conf) {
+export function getSpecTitleElem(conf) {
```
marcoscaceres: This function should be moved to utils.js.

src/w3c/templates/headers.js
```diff
@@ -89,29 +89,42 @@ const localizationStrings = {
...
+        "warn",
+        'textContent of h1#title element is "" or whitespace. Falling back to `document.title` or "No Title"'
+      );
+      specTitleElem.textContent = document.title || "No Title";
+    }
+
+    document.title = norm(specTitleElem.textContent);
   } else {
-    specTitleElem.textContent = conf.title;
+    if (document.title) {
```
marcoscaceres: nit - if you ever find you do:

```JS
if(x)
  y = a
else 
  y = b
```

Then, rewrite as:

```
specTitleElem.textContent = document.title ? document.title : "No Title"; 
```

For example.

src/w3c/templates/headers.js
```diff
@@ -89,29 +89,42 @@ const localizationStrings = {
...
+      specTitleElem.textContent = document.title;
+    } else {
+      specTitleElem.textContent = document.title = "No Title";
```
marcoscaceres: Nit: try to avoid these kinds of cascades, as they are unpredictable. For instance, setting document.title has side effects (it strips whitespace). By the end of the chain, you never sure what you are going to get.

src/w3c/templates/headers.js
```diff
@@ -89,29 +89,42 @@ const localizationStrings = {
...
     specTitleElem.id = "title";
   }
+
   if (conf.isPreview && conf.prNumber) {
     const prUrl = conf.prUrl || `${conf.github.repoURL}pull/${conf.prNumber}`;
     const { childNodes } = html`
       Preview of PR <a href="${prUrl}">#${conf.prNumber}</a>:
     `;
     specTitleElem.prepend(...childNodes);
+    document.title = `Preview of PR #${conf.prNumber}: ${document.title}`;
```
marcoscaceres: As above... don't set document.title here... only in one place right at the very end.

src/w3c/templates/headers.js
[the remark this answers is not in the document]
janiceshiu: @marcoscaceres I've done a pretty large refactor on it and added test cases to cover the `cgbg-headers` use case. Remaining things
* warning
  * rewrite it to be more user friendly. You [recommended](https://github.com/w3c/respec/pull/2750#discussion_r378621127) `"The document is missing a title, so defaulting to 'No Title'. To fix this, do X, Y, and Z`. Am still thinking of what to put in the "to fix this" part. The fix is basically `add text in your h1#title element` but I am not sure whether this is user-friendly enough.
  * use `showInlineError()` fn
* move specTitleElem/1 to utils.js
  * will do so once you're okay with the way I refactored stuff so far.
* DINI profile still uses `conf.title`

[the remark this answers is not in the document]
marcoscaceres: For the error message, we generally prefer people use a `<title>`, so we could just say:

> To fix this, please give your document a `<title>`. If you need special markup in the document's title, please use a `<h1 id="title">`.

src/w3c/templates/headers.js
```diff
@@ -89,30 +91,48 @@ const localizationStrings = {
...
 
-function getSpecTitleElem(conf) {
+export function getSpecTitleElem(conf) {
+  /** @type {HTMLElement} */
   const specTitleElem =
     document.querySelector("h1#title") || document.createElement("h1");
-  if (specTitleElem.parentElement) {
+
+  if (specTitleElem.isConnected) {
     specTitleElem.remove();
```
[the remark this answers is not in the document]
janiceshiu: which `if` check? Not following, apologies

src/core/utils.js
```diff
@@ -7,6 +7,65 @@ import { hyperHTML } from "./import-maps.js";
...
+}
+
+function setDocumentTitle(conf, specTitleElem) {
+  let documentTitle;
+
+  if (specTitleElem.textContent.trim() === "") {
+    specTitleElem.textContent = document.title || `${l10n.default_title}`;
+  }
+
+  documentTitle = norm(specTitleElem.textContent);
```
saschanaz: Maybe just `let documentTitle` at this line?

src/w3c/templates/headers.js
```diff
@@ -135,7 +109,7 @@ function getSpecSubTitleElem(conf) {
 export default conf => {
   return html`
     <div class="head">
-      ${conf.logos.map(showLogo)} ${getSpecTitleElem(conf)}
+      ${conf.logos.map(showLogo)} ${updateSpecTitleElem(conf)}
```
saschanaz: I don't like how several modules are "updating" the title. Maybe we should create `core/title` to update it first?

src/w3c/templates/cgbg-headers.js
```diff
@@ -4,16 +4,22 @@ import { l10n } from "./headers.js";
...
 export default conf => {
   const existingCopyright = document.querySelector(".copyright");
   if (existingCopyright) {
     existingCopyright.remove();
   }
+
+  const specTitleElem = updateSpecTitleElem(conf);
+  const specTitleElemClone = specTitleElem.cloneNode(true);
+  const fragment = document.createDocumentFragment();
+  fragment.append(...specTitleElemClone.childNodes);
```
saschanaz: We should be able to directly insert `.childNodes` rather than creating a fragment.

src/w3c/templates/headers.js
marcoscaceres: Yeah, that might be the way to go. Let's get the logic right first, get everything passing, then refactor.

src/w3c/templates/cgbg-headers.js
marcoscaceres: That's a great suggestion! Forgot `append()` will just adopt the nodes.

src/core/utils.js
janiceshiu: @saschanaz didn't get what you meant at first - thought it was change `documentTitle = norm(specTitleElem.textContent);` to `let documentTitle` - but @marcoscaceres explained to me what you meant. ^^ fixed, thanks!

saschanaz: Oh, Marcos just read my mind 🤓

janiceshiu:
@sebastian-meyer hi sebastian! this is touching the dini profile. we made some modifications to how the document's title is set and included. it is much better for localisation now. i have also added your german localisation for the default title.

please let me know if you have any concerns. thank you.

src/core/title.js
```diff
@@ -0,0 +1,78 @@
...
+export function run(conf) {
+  /** @type {HTMLElement} */
+  const h1Elem =
+    document.querySelector("h1#title") || document.createElement("h1");
+
+  // check existing element is ok to use
+  if (h1Elem.isConnected && h1Elem.textContent.trim() === "") {
+    const msg =
+      "The document is missing a title, so using a default title. " +
+      "To fix this, please give your document a `<title>`." +
```
saschanaz: This needs a space after dot. Maybe we should use ``` `` ``` instead?

BTW, TIL `isConnected`

src/core/title.js
```diff
@@ -0,0 +1,78 @@
...
+    default_title: "Kein Titel",
+  },
+};
+
+const l10n = getIntlData(localizationStrings);
+
+export function run(conf) {
+  /** @type {HTMLElement} */
+  const h1Elem =
+    document.querySelector("h1#title") || document.createElement("h1");
```
saschanaz:
```suggestion
    document.querySelector("h1#title") || hyperHTML`<h1 id="title">`;
```

This can skip assigning id below.

src/core/title.js
```diff
@@ -0,0 +1,78 @@
...
+      "To fix this, please give your document a `<title>`." +
+      "If you need special markup in the document's title, " +
+      'please use a `<h1 id="title">`.';
+    const msgTitle = "Document is missing a title";
+    showInlineError(h1Elem, msg, msgTitle);
+  }
+
+  // Decorate the spec title
+  if (!h1Elem.id) h1Elem.id = "title";
+  h1Elem.classList.add("title");
```
saschanaz: Are we sure we don't need `p-name`?

src/w3c/templates/cgbg-headers.js
```diff
@@ -10,10 +10,13 @@ export default (conf, options) => {
   if (existingCopyright) {
     existingCopyright.remove();
   }
+
+  const specTitleElem = document.querySelector("h1#title");
+  const specTitleElemClone = specTitleElem.cloneNode(true);
```
saschanaz: Maybe merge these into one variable?

src/core/title.js
marcoscaceres: yeah, I checked... p-name is for people. No idea why it ended up here.

src/w3c/templates/cgbg-headers.js
marcoscaceres: I think both get used independently, but @janiceshiu can confirm... but it might be possible to remove one of the variables.

profiles/w3c.js
```diff
@@ -18,6 +18,7 @@ const modules = [
   import("../src/core/data-include.js"),
   import("../src/core/markdown.js"),
   import("../src/core/reindent.js"),
+  import("../src/core/title.js"),
```
sidvishnoi: Add to geonovum profile also?

saschanaz: Maintaining multiple profiles is certainly a pain 😛

sidvishnoi: We should ask geonovum folks if they're still using the profile we maintain or their own fork.

src/w3c/templates/cgbg-headers.js
saschanaz: Ah true, they are used independently. Sorry!

profiles/w3c.js
janiceshiu: @sidvishnoi geonovum doesn't use the main `headers.js` plugin that w3c uses and that dini replicated. they use `fix-headers` and `id-headers` which do other things. since it seems like they don't use this feature's functionality, we didn't want to add core/title to the geonovum profile.

sidvishnoi: Apparently, they use a self hosted version (https://github.com/search?q=respec-geonovum.js+org%3AGeonovum&type=Code) and are maintaining their own fork https://github.com/Geonovum/respec/

src/core/title.js
janiceshiu: I used `` but @marcoscaceres changed it to ""

XD

src/core/title.js
janiceshiu: ooh nice. thanks!

profiles/w3c.js
saschanaz: That fork is superseded by the profile as maintaining a fork was too hard.

sidvishnoi: Should've seen last commit date, instead of last open issue date :sweat_smile:

src/core/title.js
janiceshiu: spacing [added](https://github.com/w3c/respec/pull/2750/commits/9eb266ae4b98269d277edf57265438a2fb7f217a)

tests/spec/w3c/headers-spec.js
```diff
@@ -557,37 +557,156 @@ describe("W3C — Headers", () => {
...
+
+      const titleElem = doc.querySelector("title");
+      expect(titleElem).toBeTruthy();
+      expect(titleElem.textContent).toBe("No Title");
+
+      const h1 = doc.querySelector("h1#title");
+      expect(h1.textContent).toBe("No Title");
+    });
+
+    it("uses a default title when <title> contains is an empty string", async () => {
```
saschanaz:
```suggestion
    it("uses a default title when <title> contains an empty string", async () => {
```

tests/spec/w3c/headers-spec.js
```diff
@@ -557,37 +557,156 @@ describe("W3C — Headers", () => {
...
+      const titleElem = doc.querySelector("title");
+      expect(titleElem).toBeTruthy();
+      expect(titleElem.textContent).toBe("Title!!!");
+
+      const h1 = doc.querySelector("h1#title");
+      expect(h1).toBeTruthy();
+      expect(h1.textContent).toBe("Title!!!");
+    });
+
+    it("adds h1#title with correct text and uses a default title when the document excludes a title", async () => {
```
saschanaz:
```suggestion
    it("uses a default title when the document excludes a title", async () => {
```

The "correct text" is a default title, so it seems a bit redundant 👀

sebastian-meyer:
> @sebastian-meyer hi sebastian! this is touching the dini profile. we made some modifications to how the document's title is set and included. it is much better for localisation now. i have also added your german localisation for the default title.
> 
> please let me know if you have any concerns. thank you.

Looks good to me. Thank you!

janiceshiu:
a much longer haul than expected but it's merged now and i'm so happy~~~~

thanks y'all for reviewing and nitpicking and helping. especial thanks to @marcoscaceres XD

Issue #2749, opened 2020-02-12
Refactoring suggestion: Export classes from modules


saschanaz:
From https://github.com/w3c/respec/pull/2747/files#r378205900:

> I wonder we can make modules export a class instead, so that core/base-runner can create instances from modules like `new Biblio(conf)` and we can stop passing `conf` every time we call internal functions.
> 
> Just a thought for future refactoring.

marcoscaceres:
The be really nice... and less gross than today’s 🤮

marcoscaceres:
Closing this refactoring suggestion. After 6 years with no concrete plan or demand, closing as stale.

PR #2752, opened 2020-02-13
refactor(core/biblio): export as a class

saschanaz:
Part of #2749.


src/core/biblio.js
```diff
@@ -137,19 +95,78 @@ export async function run(conf) {
...
-  const externalRefs = split.noData.map(item => item.id);
-  if (externalRefs.length) {
-    // Going to the network for refs we don't have
-    const data = await updateFromNetwork(externalRefs, { forceUpdate: true });
-    Object.assign(biblio, data);
+
+  return idbRefs;
+}
+
+export default class {
```
saschanaz: Maybe we could add a proper name and check that instead. `class` is a keyword so that can't be used, do you have a good idea?

sidvishnoi: `klass` is pretty common

saschanaz: Decided to use `Class` as using uppercase matches our code style 👀

sidvishnoi: Smart

sidvishnoi: Wait. How about `Plugin`?

saschanaz: Every modules are Plugin. I'm not sure they are really plugins, though. (No real plugin API, need to modify profile whenever we need third party modules.)

src/core/base-runner.js
```diff
@@ -51,13 +54,17 @@ function toRunnable(plug) {
   };
 }
 
+function isRunnableModule(plug) {
+  return plug && (plug.run || plug.exportsClass);
```
sidvishnoi:
```suggestion
  return plug && (plug.run || plug.Class);
```

src/core/biblio.js
```diff
@@ -137,19 +93,78 @@ export async function run(conf) {
...
+  /** @param {Conf} conf */
+  constructor(conf) {
+    this.conf = conf;
+  }
+
+  /**
+   * Normative references take precedence over informative ones,
+   * so any duplicates ones are removed from the informative set.
+   */
+  normalizeReferences() {
```
saschanaz: A separate thought: I don't quite like how this affects other modules... 🤔

sidvishnoi: Another separate thought: `normativeReferences` should not exist on `conf` (keep conf for user defined cobfig only?). We should rather `export const normativeReferences = new Set()` from the module where it's created

src/core/biblio.js
marcoscaceres: Although @saschanaz is technically correct about not really being "plugins", we've always referred to these as plugins. The Class thing looks a bit weird, tbh... so I'd prefer we just call them `Plugin` for legacy reasons and hold our noses. Naming things is hard.

src/core/biblio.js
marcoscaceres: agree... of some kind of promise getter.

src/core/biblio.js
saschanaz: Okay, reasonable. Actually I like the Plugin name too, maybe one day we could make them real plugins...

marcoscaceres: I would love that :) The ultimate plugins would all be their own git repos and we'd have a nicer plugin API. What's some nice plugins APIs you've seen?

saschanaz: Not sure about what's "nice", but I would mention the plugins we have in our `package.json` and the APIs they are using (from Babel, ESLint etc.).

PR #2754, opened 2020-02-14
refactor(tools/release): no need to checkout

saschanaz:
These lines does nothing, as the generated builds are already committed.

### DIFF
--- a/tools/release.js
+++ b/tools/release.js
@@ -395,9 +395,6 @@ const run = async () => {
     indicators.get("push-to-server").hide();
     console.log(colors.info(" Publishing to npm... 📡"));
     await npm("publish", { showOutput: true });
-    // publishing generates a new build, which we don't want
-    // on develop branch
-    await git("checkout builds");
     if (initialBranch !== MAIN_BRANCH) {
       await Prompts.askSwitchToBranch(MAIN_BRANCH, initialBranch);
     }


marcoscaceres:
Oh, um... maybe then we should run these earlier? Like right at the start?

saschanaz:
> Oh, um... maybe then we should run these earlier? Like right at the start?

What would be the purpose, when we do want to commit the builds?

marcoscaceres:
> What would be the purpose, when we do want to commit the builds?

When running `npm start`, the developer server constantly regenerates builds/... once someone finishes coding, stuff gets merged, etc. there might still be "dirty/outdated" builds/ files. The idea was to make sure those build files got removed... but maybe you are right... maybe the release script will complain early that there uncommitted files in "develop". I think we were trying to not bother ourselves with the build files, and just wanted them blown away.

sidvishnoi:
FWIW, the `npm start` script undoes changes to `./builds/` on exit.
https://github.com/w3c/respec/blob/264cc6fffcc088a6223141d6124e63c150238f67/tools/dev-server.js#L87-L89

saschanaz:
This is about `tools/release`, not `npm start`. Releasing should regenerate and commit the builds, so there shouldn't be "dirty" one here.

marcoscaceres:
Right, I need to check 🥴. I know these is about release, but `npm start` is the dev workflow that leads to release. So the thing to check is if possible to release after running `npm start`,

PR #2755, opened 2020-02-16
refactor: remove raw HTML insertions

saschanaz:
Part of #2551

Remaining are all `conf.*HTML` or such, which will cause breaking changes when removed.

### DIFF
--- a/src/core/dfn-panel.js
+++ b/src/core/dfn-panel.js
@@ -111,13 +111,16 @@ function referencesToHTML(id, links) {
     );
   };
 
-  const listItems = [...titleToIDs].map(
-    entry =>
-      hyperHTML`<li>${toLinkProps(entry).map(
-        link => hyperHTML`<a href="#${link.id}">${link.title}</a>${" "}`
-      )}</li>`
-  );
+  /**
+   * @param {[string, string[]]} entry
+   * @returns {HTMLLIElement}
+   */
+  const listItemToHTML = entry =>
+    hyperHTML`<li>${toLinkProps(entry).map(
+      link => hyperHTML`<a href="#${link.id}">${link.title}</a>${" "}`
+    )}</li>`;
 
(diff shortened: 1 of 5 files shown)


src/core/dfn-panel.js
```diff
@@ -111,16 +111,14 @@ function referencesToHTML(id, links) {
...
-    )}</li>`;
+  const listItems = [...titleToIDs].map(
+    entry =>
+      hyperHTML`<li>${toLinkProps(entry).map(
+        link => hyperHTML`<a href="#${link.id}">${link.title}</a>${" "}`
+      )}</li>`
+  );
 
-  return hyperHTML`<ul>${[...titleToIDs.entries()].map(listItemToHTML)}</ul>`;
+  return hyperHTML`<ul>${listItems}</ul>`;
```
sidvishnoi: Can we skip this change? I personally prefer the previous version. Not a big deal though.

src/core/utils.js
```diff
@@ -500,8 +506,8 @@ export function htmlJoinAnd(array, mapper = item => item) {
     case 2: // x and y
       return hyperHTML`${items[0]} and ${items[1]}`;
     default: {
-      const joinedItems = items.slice(0, -1).map(item => hyperHTML`${item}, `);
-      return hyperHTML`${joinedItems}and ${items[items.length - 1]}`;
+      const joined = htmlJoinComma(items.slice(0, -1));
```
sidvishnoi: Nice :100:

src/core/dfn-panel.js
marcoscaceres: I also don't have a preference here. I'm happy with either.

saschanaz: I prefer to avoid complex expression within template strings 👀

Should I keep the function separate, if that is what you prefer?

sidvishnoi: Separate `listItemToHtml` would be great

Issue #2743, opened 2020-02-11
Duplicate code in profiles


marcoscaceres:
We should look at reducing the duplication between W3C and DINI profiles. If we have things that are share, those should move to "core/". We should otherwise export things from core/*.js that can be used by DINI. 

See https://github.com/w3c/respec/pull/2739 for specifics.

marcoscaceres:
Closing this self-filed refactoring task. After 6 years with no activity, the profiles work well as-is. If profile duplication becomes a maintenance burden, it can be revisited.

PR #2756, opened 2020-02-17
refactor(w3c/headers): pass alternatesHTML as options

saschanaz:
This removes additional redundant `${[...]}` and also reduces `conf` pollution.

Partly also #2743.

### DIFF
--- a/src/dini/headers.js
+++ b/src/dini/headers.js
@@ -42,9 +42,8 @@
 //      - "cc-by-sa"
 //      - "cc-by"
 //      - "cc0"
-import { ISODate, htmlJoinAnd } from "../core/utils.js";
+import { ISODate } from "../core/utils.js";
 import headersTmpl from "./templates/headers.js";
-import { hyperHTML } from "../core/import-maps.js";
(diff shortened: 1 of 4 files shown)


src/dini/headers.js
```diff
@@ -136,15 +135,6 @@ export function run(conf) {
...
   });
-  conf.multipleAlternates =
-    conf.alternateFormats && conf.alternateFormats.length > 1;
-  conf.alternatesHTML =
-    conf.alternateFormats &&
-    htmlJoinAnd(conf.alternateFormats, alt => {
-      const lang = alt.hasOwnProperty("lang") && alt.lang ? alt.lang : null;
-      const type = alt.hasOwnProperty("type") && alt.type ? alt.type : null;
-      return hyperHTML`<a rel='alternate' href='${alt.uri}' hreflang='${lang}' type='${type}'>${alt.label}</a>`;
-    });
```
sidvishnoi: Is this a breaking change to DINI?

saschanaz: I don't think so, dini templates never use them.

PR #2757, opened 2020-02-17
refactor(l10n): use hyperHTML

saschanaz:


src/w3c/l10n.js
```diff
@@ -2,36 +2,27 @@
...
 // Looks at the lang attribute on the root element and uses it to manage the config.l10n object so
 // that other parts of the system can localise their text
+import { hyperHTML } from "../core/import-maps.js";
 import { l10n } from "../core/l10n.js";
 export const name = "w3c/l10n";
 const additions = {
   en: {
-    status_at_publication:
-      "This section describes the status of this document at the time of its publication. Other documents may supersede this document. A list of current W3C publications and the latest revision of this technical report can be found in the <a href='https://www.w3.org/TR/'>W3C technical reports index</a> at https://www.w3.org/TR/.",
+    status_at_publication: hyperHTML`This section describes the status of this document at the time of its publication. Other documents may supersede this document. A list of current W3C publications and the latest revision of this technical report can be found in the <a href='https://www.w3.org/TR/'>W3C technical reports index</a> at https://www.w3.org/TR/.`,
```
saschanaz: This seems outdated, though. There is no "W3C technical reports index" in https://www.w3.org/TR/.

marcoscaceres: Not sure what you mean by "no W3C technical reports index"? The landing page there lists all the specs. 

I agree that the linking is weird, but it's sadly required by w3c put rules:
https://www.w3.org/pubrules/doc/rules/?profile=FPWD#boilerplateTRDoc

saschanaz: It's "All Standards and Drafts" not "W3C technical reports index" 🤔

It may make sense if we remove `at w3c.org/TR` but with it the phrase makes me find where the exact section is.

marcoscaceres: To be clear, we need to match the requirement from PubRules, or else publication will fail. So we need to update PubRules first, then update anything here.

marcoscaceres:
@saschanaz, if you feel the boilerplate should change. I'm sure the W3C folks would be ok to discuss that. It would also mean changing both BikeShed's templates too and possibly the W3C Process Document (although I've not verified if that template is also there).

saschanaz:
Will merge and explore the possible change later 😁

marcoscaceres:
Ok, cool. Just ping either dontcallmedom and/or plehegar. I'm sure either would be happy to discuss/revise.

PR #2758, opened 2020-02-19
feat(w3c/level): add level configuration

janiceshiu:
Closes [#2406](https://github.com/w3c/respec/issues/2406)

- [x] combines the level with the shortName (so... page-visibility becomes "page-visibility-2")
- [x] automatically appends the "Level X" to the title of the document.
- [x] makes sure all appropriate links also get the level, though these are probably just derived from the shortName.

### DIFF
--- a/profiles/dini.js
+++ b/profiles/dini.js
@@ -17,7 +17,6 @@ const modules = [
   import("../src/core/markdown.js"),
   import("../src/core/reindent.js"),
   import("../src/core/title.js"),
-  import("../src/core/level.js"),
   import("../src/dini/headers.js"),
   import("../src/dini/abstract.js"),
   import("../src/core/data-transform.js"),
--- a/profiles/dini.js
+++ b/profiles/dini.js
@@ -17,6 +17,7 @@ const modules = [
   import("../src/core/markdown.js"),
   import("../src/core/reindent.js"),
   import("../src/core/title.js"),
+  import("../src/core/level.js"),
   import("../src/dini/headers.js"),
   import("../src/dini/abstract.js"),
   import("../src/core/data-transform.js"),
--- a/src/core/level.js
+++ b/src/core/level.js
@@ -21,7 +21,7 @@ export function run(conf) {
 
   if (conf.hasOwnProperty("level")) {
     if (Number.isInteger(conf.level) && conf.level >= 0) {
-      h1Elem.append(`${l10n.level} ${conf.level}`));
+      h1Elem.append(` ${l10n.level} ${conf.level}`);
       document.title = `${document.title} ${l10n.level} ${conf.level}`;
       conf.shortName = `${conf.shortName}-${conf.level}`;
     } else {
--- a/src/core/level.js
+++ b/src/core/level.js
@@ -1,0 +1,27 @@
+/**
+ * This module updates the title and shortName
+ * of a spec when there is a valid level configuration.
+ *
+ * Levels should be integers >= 0.
+ */
+
+import { showInlineWarning } from "./utils.js";
+export const name = "core/level";
+
+export function run(conf) {
+  const h1Elem = document.querySelector("h1#title");
+
+  if (conf.hasOwnProperty("level")) {
+    if (Number.isInteger(conf.level) && conf.level >= 0) {
+      h1Elem.append(document.createTextNode(` Level ${conf.level}`));
+      document.title = `${document.title} Level ${conf.level}`;
+      conf.shortName = `${conf.shortName}-${conf.level}`;
+    } else {
+      showInlineWarning(
+        h1Elem,
+        `The \`level\` configuration option must be a number greater or equal to 0. It is currently set to \`${conf.level}\``,
+        "Invalid level config."
+      );
+    }
+  }
+}
--- a/src/core/level.js
+++ b/src/core/level.js
@@ -5,16 +5,24 @@
  * Levels should be integers >= 0.
  */
 
-import { showInlineWarning } from "./utils.js";
+import { getIntlData, showInlineWarning } from "./utils.js";
 export const name = "core/level";
+
+const localizationStrings = {
+  en: {
+    level: "Level",
+  },
+};
+
+const l10n = getIntlData(localizationStrings);
 
 export function run(conf) {
   const h1Elem = document.querySelector("h1#title");
 
   if (conf.hasOwnProperty("level")) {
     if (Number.isInteger(conf.level) && conf.level >= 0) {
-      h1Elem.append(document.createTextNode(` Level ${conf.level}`));
-      document.title = `${document.title} Level ${conf.level}`;
+      h1Elem.append(document.createTextNode(` ${l10n.level} ${conf.level}`));
+      document.title = `${document.title} ${l10n.level} ${conf.level}`;
       conf.shortName = `${conf.shortName}-${conf.level}`;
     } else {
       showInlineWarning(
--- a/src/core/level.js
+++ b/src/core/level.js
@@ -21,7 +21,7 @@ export function run(conf) {
 
   if (conf.hasOwnProperty("level")) {
     if (Number.isInteger(conf.level) && conf.level >= 0) {
-      h1Elem.append(document.createTextNode(` ${l10n.level} ${conf.level}`));
+      h1Elem.append(`${l10n.level} ${conf.level}`));
       document.title = `${document.title} ${l10n.level} ${conf.level}`;
       conf.shortName = `${conf.shortName}-${conf.level}`;
     } else {
--- a/src/core/title.js
+++ b/src/core/title.js
@@ -75,15 +75,17 @@ function setDocumentTitle(conf, h1Elem) {
     documentTitle = `Preview of PR #${conf.prNumber}: ${documentTitle}`;
   }
 
-  if (validLevel(conf.level)) {
-    h1Elem.innerHTML = `${h1Elem.innerHTML} Level ${Number(conf.level)}`;
-    documentTitle = `${documentTitle} Level ${Number(conf.level)}`;
-  } else {
-    showInlineWarning(
-      h1Elem,
-      `The \`level\` config should be an integer greater than or equal to 0. It is currently set to \`${conf.level}\``,
-      "Invalid level config."
-    );
+  if (conf.hasOwnProperty("level")) {
+    if (validLevel(conf.level)) {
+      h1Elem.innerHTML = `${h1Elem.innerHTML} Level ${Number(conf.level)}`;
+      documentTitle = `${documentTitle} Level ${Number(conf.level)}`;
+    } else {
+      showInlineWarning(
+        h1Elem,
+        `The \`level\` config should be an integer greater than or equal to 0. It is currently set to \`${conf.level}\``,
+        "Invalid level config."
+      );
+    }
   }
 
   document.title = documentTitle;
--- a/src/core/title.js
+++ b/src/core/title.js
@@ -77,7 +77,7 @@ function setDocumentTitle(conf, h1Elem) {
 
   if (conf.hasOwnProperty("level")) {
     if (validLevel(conf.level)) {
-      h1Elem.innerHTML = `${h1Elem.innerHTML} Level ${Number(conf.level)}`;
+      h1Elem.append(document.createTextNode(` Level ${conf.level}`));
       documentTitle = `${documentTitle} Level ${Number(conf.level)}`;
     } else {
       showInlineWarning(
--- a/src/core/title.js
+++ b/src/core/title.js
@@ -10,13 +10,7 @@
  *
  */
 
-import {
-  getIntlData,
-  norm,
-  showInlineError,
-  showInlineWarning,
-  validLevel,
-} from "./utils.js";
+import { getIntlData, norm, showInlineError } from "./utils.js";
 import { hyperHTML } from "./import-maps.js";
 export const name = "core/title";
 
@@ -73,19 +67,6 @@ function setDocumentTitle(conf, h1Elem) {
     `;
     h1Elem.prepend(...childNodes);
     documentTitle = `Preview of PR #${conf.prNumber}: ${documentTitle}`;
-  }
-
-  if (conf.hasOwnProperty("level")) {
-    if (validLevel(conf.level)) {
-      h1Elem.append(document.createTextNode(` Level ${conf.level}`));
-      documentTitle = `${documentTitle} Level ${Number(conf.level)}`;
-    } else {
-      showInlineWarning(
-        h1Elem,
-        `The \`level\` configuration option must be a number greater or equal to 0. It is currently set to \`${conf.level}\``,
-        "Invalid level config."
-      );
-    }
   }
 
   document.title = documentTitle;
--- a/src/core/title.js
+++ b/src/core/title.js
@@ -78,6 +78,12 @@ function setDocumentTitle(conf, h1Elem) {
   if (validLevel(conf.level)) {
     h1Elem.innerHTML = `${h1Elem.innerHTML} Level ${Number(conf.level)}`;
     documentTitle = `${documentTitle} Level ${Number(conf.level)}`;
+  } else {
+    showInlineWarning(
+      h1Elem,
+      `The \`level\` config should be an integer greater than or equal to 0. It is currently set to \`${conf.level}\``,
+      "Invalid level config."
+    );
   }
 
   document.title = documentTitle;
--- a/src/core/title.js
+++ b/src/core/title.js
@@ -10,7 +10,13 @@
  *
  */
 
-import { getIntlData, norm, showInlineError } from "./utils.js";
+import {
+  getIntlData,
+  norm,
+  showInlineError,
+  showInlineWarning,
+  validLevel,
+} from "./utils.js";
 import { hyperHTML } from "./import-maps.js";
 export const name = "core/title";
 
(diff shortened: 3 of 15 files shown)


src/core/utils.js
```diff
@@ -7,6 +7,15 @@ import { hyperHTML } from "./import-maps.js";
 import { pub } from "./pubsubhub.js";
 export const name = "core/utils";
 
+/**
+ * Checks whether a level is valid. Levels should be integers >= 0.
+ *
+ * @param {*} level
+ */
+export function validLevel(level) {
```
marcoscaceres: We probably only want to export a util function if this is used in multiple places.

src/core/title.js
```diff
@@ -69,6 +75,19 @@ function setDocumentTitle(conf, h1Elem) {
...
   }
 
+  if (conf.hasOwnProperty("level")) {
+    if (validLevel(conf.level)) {
+      h1Elem.innerHTML = `${h1Elem.innerHTML} Level ${Number(conf.level)}`;
+      documentTitle = `${documentTitle} Level ${Number(conf.level)}`;
+    } else {
+      showInlineWarning(
+        h1Elem,
+        `The \`level\` config should be an integer greater than or equal to 0. It is currently set to \`${conf.level}\``,
```
marcoscaceres:
```suggestion
        `The \`level\` configuration option must be a number greater or equal to 0. It is currently set to \`${conf.level}\``,
```

src/core/title.js
```diff
@@ -69,6 +75,19 @@ function setDocumentTitle(conf, h1Elem) {
     documentTitle = `Preview of PR #${conf.prNumber}: ${documentTitle}`;
   }
 
+  if (conf.hasOwnProperty("level")) {
+    if (validLevel(conf.level)) {
+      h1Elem.innerHTML = `${h1Elem.innerHTML} Level ${Number(conf.level)}`;
```
marcoscaceres: Using .innerHTML makes me a bit uncomfortable. We are just appending  "Level ${conf.level}", so we can probably just `h1Elem.append()`. Also, we probably don't need to cast it into a Number, as it's just going to end up as a string.

src/core/title.js
```diff
@@ -69,6 +75,19 @@ function setDocumentTitle(conf, h1Elem) {
     documentTitle = `Preview of PR #${conf.prNumber}: ${documentTitle}`;
   }
 
+  if (conf.hasOwnProperty("level")) {
```
marcoscaceres: I think we should move all this to a new plugin ("w3c/levels") that runs after core/title.

src/w3c/headers.js
```diff
@@ -265,6 +265,9 @@ export function run(conf) {
       pub("warn", msg);
     }
   }
+  if (validLevel(conf.level)) {
+    conf.shortName = `${conf.shortName}-${conf.level}`;
```
marcoscaceres: This should happen in the "w3c/levels" plugin - so then it's all in one place (and then we don't need the util function anymore).

src/core/title.js
```diff
@@ -69,6 +75,19 @@ function setDocumentTitle(conf, h1Elem) {
...
+    if (validLevel(conf.level)) {
```
marcoscaceres: we should probably normalize the level here... so: 
```suggestion
    const level = parseInt(conf.level);
    if (validLevel(level)) {
```

src/core/utils.js
janiceshiu: it is currently used in `title.js` for the title and `header.js` for the shortName

src/core/title.js
janiceshiu: `parseInt(num)` has unexpected things such as 
```parseInt("1t")
1```

thus, I'd rather avoid that and stick to `Number.isInteger` and `>=`

Comment (marcoscaceres) [reply src/core/title.js]: I think that's fine. We don't need to overcheck.

File: src/core/title.js [hunk 7]
```diff
@@ -69,6 +75,19 @@ function setDocumentTitle(conf, h1Elem) {
...
+      h1Elem.append(document.createTextNode(` Level ${conf.level}`));
```
marcoscaceres: No need to create a text node... the .append() api does it for free! :) 
```suggestion
      h1Elem.append(` Level ${conf.level}`);
```

marcoscaceres: Note, the word "level" might need to be localized. We can just have the English localization for now.

janiceshiu: haha braintwins. writing that in the level plugin now.

src/core/title.js
janiceshiu: could you help me understand what `parseInt` does that the checks for validLevel (`Number.isInteger` and `>=`) don't? I'm not sure why it is needed in addition to those two checks. thanks!

src/core/level.js
```diff
@@ -0,0 +1,35 @@
...
+    level: "Level",
+  },
+};
+
+const l10n = getIntlData(localizationStrings);
+
+export function run(conf) {
+  const h1Elem = document.querySelector("h1#title");
+
+  if (conf.hasOwnProperty("level")) {
```
marcoscaceres: So, instead of nesting if statements here, we want to return early:

```JS
if (!conf.hasOwnProperty("level")) return; // nothing to do. 
```

src/core/level.js
```diff
@@ -0,0 +1,35 @@
...
+    if (Number.isInteger(conf.level) && conf.level >= 0) {
```
marcoscaceres: I think we should invert this one, so we do the validity check, and it fails, we `showInlineWarning()` and return. Finally, if all is good, we do the append. That will get rid of the else block too.

tests/spec/w3c/headers-spec.js
```diff
@@ -557,6 +557,116 @@ describe("W3C — Headers", () => {
     });
   });
 
+  describe("valid level config", () => {
+    it("updates relevant locations with the level when level is the integer 0", async () => {
+      const body = `
+      <h1 id="title">Spec <code>Marked</code> Up</h1>${makeDefaultBody()}`;
```
marcoscaceres: Good test! 
```suggestion
        <h1 id="title">Spec <code>Marked</code> Up</h1>${makeDefaultBody()}`;
```

tests/spec/w3c/headers-spec.js
```diff
@@ -557,6 +557,116 @@ describe("W3C — Headers", () => {
...
+      expect(terms[1].textContent).toBe("Latest published version:");
+      expect(terms[1].nextElementSibling.localName).toBe("dd");
+      expect(terms[1].nextElementSibling.textContent).toContain("TR/xyz-9870/");
+    });
+  });
+
+  describe("invalid level configs", () => {
+    it("warns the user and does not add the level to the relevant places when level is a string", async () => {
+      const body = `
+      <h1 id="title">Spec <code>Marked</code> Up</h1>${makeDefaultBody()}`;
```
marcoscaceres:
```suggestion
        <h1 id="title">Spec <code>Marked</code> Up</h1>${makeDefaultBody()}`;
```

profiles/dini.js
```diff
@@ -17,6 +17,7 @@ const modules = [
   import("../src/core/markdown.js"),
   import("../src/core/reindent.js"),
   import("../src/core/title.js"),
+  import("../src/core/level.js"),
```
marcoscaceres:
```suggestion
```

marcoscaceres: Levels are a W3C thing... not sure DINI will want them too, but @sebastian-meyer can confirm? 

@sebastian-meyer, just FYI, W3C specs sometimes use a (not well defined at all) concept of API/technology "layering"... it's kinda like versioning... but not really. A "level 2 spec" will usually contain of "level 1"'s text, for example:
https://w3c.github.io/input-events/

 Levels are always backwards compatible. This is different to "versions", which can be incompatible with each other.

src/core/level.js
```diff
@@ -0,0 +1,35 @@
+/**
+ * This module updates the title and shortName
```
marcoscaceres: Let's wait for @sebastian-meyer to respond... but my feeling is that this should be a "w3c/" plugin, not a "core/" plugin.

profiles/dini.js
sebastian-meyer: We don't need levels in the DINI profile. DINI is using a rather classical versioning schema.

src/core/level.js
sebastian-meyer: I agree!

src/core/level.js
marcoscaceres: Please see Slack for why I think we should use `const level = parseInt(conf.level);`

src/core/level.js
```diff
@@ -0,0 +1,35 @@
...
+const l10n = getIntlData(localizationStrings);
+
+export function run(conf) {
+  const h1Elem = document.querySelector("h1#title");
+
+  if (conf.hasOwnProperty("level")) {
+    if (Number.isInteger(conf.level) && conf.level >= 0) {
+      h1Elem.append(` ${l10n.level} ${conf.level}`);
+      document.title = `${document.title} ${l10n.level} ${conf.level}`;
+      conf.shortName = `${conf.shortName}-${conf.level}`;
```
sidvishnoi: I'm a bit concerned that this might break `data-cite` at:
https://github.com/w3c/respec/blob/63fd45b24125edd7795375809800b89d9001d36b/src/core/data-cite.js#L33-L39

Can't find any existing test or use-case for this though

src/core/level.js
janiceshiu: we currently warn on the h1Elem though. would you prefer we warn on another element? `showInlineWarning()` needs an elem.

src/core/level.js
janiceshiu: hm. let's take [navigation timing level 2](https://w3c.github.io/navigation-timing/) as an example.

their shortName is `navigation-timing-2`.

like most specs, that spec relies on other specs and cites them by their shortNames. eg:
```
      xref: {
        url: "https://respec.org/xref",
        specs: [ "service-workers", "hr-time-2", "performance-timeline-2", "resource-timing-2"],
        profile: "web-platform",
      },
```

it seems that if we rely on respec to append the level number to the shortName, we'd remove the `-2` from those specs. respec would then be confused when there's `spec-1` and `spec-2`

profiles/dini.js
janiceshiu: thanks for clarifying, @sebastian-meyer!

src/core/level.js
sidvishnoi: The xref part is fine here, and it should use numbered shortnames. I think there might be a problem with "self citing". Like, we can't do `[[navigation-timing]]`. Not sure if it's used though.
Let me find some examples. Otherwise, we can remove that part from data-cite.js itself

sidvishnoi: Apparently, we won't be able to link stuff like:
``` html
    <section id="test">
      <h2>test</h2>
      <a data-cite="shortname-api#test">link to test</a>
    </section>
``` 
which is good as we want to discourage such usage!

sidvishnoi: cc: @marcoscaceres who might know why we have such condition in data-cite.js

src/core/level.js
marcoscaceres: Ah, good point! So this indeed comes after the first check. Flipping the if statement to get rid of the `else` would be good though.

src/core/level.js
janiceshiu: > The xref part is fine here, and it should use numbered shortnames

hm. my concern is that it using the level config would mean non-numbered shortnames.  as you can see from the [requirements](https://github.com/w3c/respec/issues/2406) in the original issue, 

> combines the level with the shortName (so... page-visibility becomes "page-visibility-2")

Thus, a spec using the levels feature would have `config = { shortName: `navigation-timing`, level: 2` because respec would automatically append `-2` to `navigation-timing`. my understanding of the code that you linked is that the lack of a numbered shortName might break things.

@marcoscaceres could you help us clarify? thanks ^^

src/core/level.js
janiceshiu: kk! as per MDN, `isInteger` returns ["A Boolean indicating whether or not the given value is an integer"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isInteger),
 so I'm gonna adjust your code for that.

src/core/title.js
marcoscaceres: > could you help me understand what parseInt does that the checks for validLevel (Number.isInteger and >=) don't?

So, first thing:

 Number.isInteger("1") === false 

but we are generally pretty liberal with allowing simple values to be canonicalized. The other thing that we get is an simple assurance of a number or NaN:

```
const level = parseInt(conf.level, 10);
```

So it's mostly just for string handling, really.

src/core/level.js
marcoscaceres: Specs that self citing itself is "nonsensical": it doesn't make any sense to ever do that, specially in light of xref and the HTML's self linking capability. So I don't think we should worry about it.

marcoscaceres: Just for reference, the un-numbered short name always points to the highest level. For example:
https://www.w3.org/TR/css-color/  <- is, today, an alias for "css-color-4". 

Similarly, these all exist:
ttps://www.w3.org/TR/css-color-4/
https://www.w3.org/TR/css-color-3/
https://www.w3.org/TR/css-color-2/

So, by W3C policy, we should be fine*.

<small>*Terms and conditions apply. May cause irreversible damage to the liver and kidneys. Please consult your doctor before using levels.</small>.

src/w3c/level.js
```diff
@@ -0,0 +1,37 @@
...
+      h1Elem,
+      `The \`level\` configuration option must be a number greater or equal to 0. It is currently set to \`${level}\``,
+      "Invalid level config."
+    );
+    return;
+  }
+
+  h1Elem.append(` ${l10n.level} ${level}`);
+  document.title = `${document.title} ${l10n.level} ${level}`;
+  conf.shortName = `${conf.shortName}-${level}`;
```
marcoscaceres: Let's finally use the canonical form, in case we need it again. 
 
```suggestion
  conf.level = level;
```

src/w3c/level.js
```diff
@@ -0,0 +1,37 @@
...
+const l10n = getIntlData(localizationStrings);
+
+export function run(conf) {
+  if (!conf.hasOwnProperty("level")) return;
+
+  const h1Elem = document.querySelector("h1#title");
+
+  const level = parseInt(conf.level);
+  if (!Number.isInteger(level) || level < 0) {
+    showInlineWarning(
```
marcoscaceres: As this is a MUST, we probably want a proper error here. 

```suggestion
    showInlineError(
```

src/core/level.js
janiceshiu: Thanks for the explanation and background info both here and in slack @marcoscaceres! It's a lot clearer now.

src/w3c/level.js
janiceshiu: by this you mean that you just want to add an additional line that says `conf.level = level;` right? 

The suggestion also says that you want to remove `conf.shortName = `${conf.shortName}-${level}`;`

marcoscaceres: oh whoops! yes... argh. let me fix that!

tests/spec/w3c/headers-spec.js
```diff
@@ -557,6 +557,105 @@ describe("W3C — Headers", () => {
     });
   });
 
+  describe("valid level config", () => {
+    it("updates relevant locations with the level when level is the integer 0", async () => {
+      const body = `
+        <h1 id="title">Spec <code>Marked</code> Up</h1>${makeDefaultBody()}`;
+      const ops = makeStandardOps({}, body);
+      const newProps = { level: 0, specStatus: "REC", shortName: "abc" };
```
sidvishnoi: We can pass newProps as first argument to makeStandardOps

tests/spec/w3c/headers-spec.js
```diff
@@ -557,6 +557,105 @@ describe("W3C — Headers", () => {
...
+      expect(h1Elem.classList).toContain("respec-offending-element");
+      expect(h1Elem.textContent).toBe("Spec Marked Up");
+
+      const date = new Date();
+
+      const terms = doc.querySelectorAll("dt");
+      expect(terms[0].textContent).toBe("This version:");
+      expect(terms[0].nextElementSibling.localName).toBe("dd");
+      expect(terms[0].nextElementSibling.textContent).toContain(
+        `REC-xxx-${date.getFullYear()}`
```
sidvishnoi: We probably don't need a date here. Can use a regex instead if we want the date part

tests/spec/w3c/headers-spec.js
```diff
@@ -557,6 +557,105 @@ describe("W3C — Headers", () => {
...
+
+      const h1Elem = doc.querySelector("h1#title");
+      // html is not escaped
+      expect(h1Elem.childNodes[1].localName).toBe("code");
+      expect(h1Elem.textContent).toBe("Spec Marked Up Level 9870");
+      expect(doc.title).toBe("Spec Marked Up Level 9870");
+
+      const terms = doc.querySelectorAll("dt");
+      expect(terms[0].textContent).toBe("This version:");
+      expect(terms[0].nextElementSibling.localName).toBe("dd");
```
sidvishnoi: We can probably use 
```
version = terms[0].querySelector(':scope > dd')
version.textContent === ...
```

What I mean is these tests can be simplified a bit to not involve nextSibling like traversals...

sidvishnoi: Never mind.. this looked bad only on GitHub app while reviewing.

tests/spec/w3c/headers-spec.js
sidvishnoi:
``` js
const ops = makeStandardOps({ level: 0, specStatus: "REC", shortName: "abc" }, body);
```

tests/spec/w3c/headers-spec.js
janiceshiu: I need the date part because if level is added, it becomes `REC-shortname-level-YYYYMMDD`. I'll look into a regex. ^^

tests/spec/w3c/headers-spec.js
janiceshiu: nice! fixed it for all the specs.

sidvishnoi:
TODO: Update wiki

janiceshiu:
@sidvishnoi in progress~ ^^

marcoscaceres:
Docs are in place. Great work @janiceshiu 👏👏👏

Issue #2374, opened 2019-05-31
Plan to deprecate EdgeHTML support


saschanaz:
Microsoft won't replace the old Edge soon enough, likely to treat it like IE. We won't support it that long so probably we should get a plan to deprecate it.

So...

Remove Edge support and start using new APIs including `Array.prototype.flat`:

1. Now
1. After the new Edge gets its first stable release

I wanted to keep supporting Edge when we had no Chromium Edge, but now I have no strong opinion and my life will be completely okay without EdgeHTML support.

cc @marcoscaceres @sidvishnoi 

---

Collected from discussion below:

- [x] Replace `trimLeft`, `trimRight` (#2746)
- [x] Use `Object.fromEntries()` where applicable (#2761)
- [x] Use `Array@flat` (#2746)
- [x] Clipboard (#2762)
- [x] [utils@children](https://github.com/w3c/respec/blob/d3ca6fddfc944b7be9bffb27d059926b7e171a53/src/core/utils.js#L770-L796) (#2763)

sidvishnoi:
What are the pain points in supporting current Edge?

saschanaz:
For now the pain is small. Being forced to use `.trimLeft()` instead of `.trimStart()`, being unable to use `Object.fromEntries()` (some code uses `.reduce()` to do this), etc.

marcoscaceres:
I think just to be safe, let's wait until first stable. At least then we have a legitimate to drop support for old Edge, and people who run into trouble have an option to use it (without needing to switch browser vendors).

saschanaz:
One more pain: a polyfill is required to use `navigator.clipboard.writeText()`. (We already have one but anyway.)

sidvishnoi:
Another pain: 'TextEncoder' is not supported, causing entire xref to fail in Edge ([image](https://user-images.githubusercontent.com/8426945/61868793-ce21fd80-aef7-11e9-871e-da994ac38bda.png)). 
TextEncoder is [used in xref](https://github.com/w3c/respec/blob/develop/src/core/xref.js#L427) to create SHA based unique ids.
Version: Microsoft Edge 44.18362.1.0, Microsoft EdgeHTML 18.18362

We can use a polyfill, or use some other way to create hashes (or just deprecate EdgeHTML support :wink:)

marcoscaceres:
Xref is a W3C feature really, so I’d be fine with us ignoring Edge support.

sidvishnoi:
I guess we can drop old Edge now

[...]

PR #2761, opened 2020-02-19
refactor(core/override-config): use Object.fromEntries

sidvishnoi:
Closes https://github.com/w3c/respec/issues/2374

@saschanaz Do you remember any other instances where we can use Object.fromEntries?


sidvishnoi:
> maybe checking remaining .reduce uses?

All seem legitimate `.reduce` uses (other than we can use `for of` or something cleaner there).

PR #2764, opened 2020-02-21
chore(package): update deps

marcoscaceres:


package.json
```diff
@@ -34,9 +34,9 @@
     "eslint-plugin-import": "^2.20.1",
     "eslint-plugin-jasmine": "^4.1.0",
     "eslint-plugin-prettier": "^3.1.2",
-    "handlebars": "^4.7.2",
+    "handlebars": "^4.7.3",
```
saschanaz: I think we are keeping this for forks, but are there any active forks remaining?

marcoscaceres: Geonovum is still active. But I've not coordinated with them for a while. I'll see if we can get them to switch over to template strings.

saschanaz: Haven't they migrated to the profile yet? 👀

marcoscaceres: They have, but I think they are still running their fork as I don't think they finished migrating to the profile. They were sending one piece at a time. I'll need to speak to them about restarting that work.

PR #2765, opened 2020-02-21
refactor(core/issues-notes): separate number getter

saschanaz:
Longer but IMO easier to understand 🧐


src/core/issues-notes.js
```diff
@@ -99,26 +99,19 @@ const l10n = getIntlData(localizationStrings);
...
     const { type, displayType, isFeatureAtRisk } = getIssueType(inno);
     const isIssue = type === "issue";
     const isInline = inno.localName === "span";
     const { number: dataNum } = inno.dataset;
-    /** @type {Partial<Report>} */
     const report = {
       type,
       inline: isInline,
       title: inno.title,
+      number: getIssueNumber(inno),
```
marcoscaceres: is this assured not to throw? createIssueNumberGetter can return undefined.

saschanaz: Yes, tests assure that numberless issues silently gets no number, by checking `report.number !== undefined`.

src/core/issues-notes.js
```diff
@@ -142,10 +135,10 @@ function handleIssues(ins, ghIssues, conf) {
...
       let ghIssue;
       if (isIssue) {
-        if (!hasDataNum) {
-          text += ` ${issueNum}`;
-        } else if (dataNum) {
-          text += ` ${dataNum}`;
+        if (report.number !== undefined) {
+          text += ` ${report.number}`;
+        }
+        if ("number" in inno.dataset) {
```
marcoscaceres: Maybe... saves walking the prototype chain (nit pick optimization that's probably not needed).
  
```suggestion
        if (inno.dataset.hasOwnProperty("number")) {
```

PR #2766, opened 2020-02-21
refactor(core/utils): no special behavior on addId

saschanaz:
~~I don't think this is used anywhere.~~ OMG it's being used!


marcoscaceres:
@saschanaz, how do you want to proceed?

saschanaz:
I think the new logic is more general and less messy, so I vote to merge this.

PR #2771, opened 2020-02-25
Support more than 26 Appendices

sglaser:
Add recursive appendixNumber function instead of indexing into alphabet.

### DIFF
--- a/src/core/structure.js
+++ b/src/core/structure.js
@@ -2,8 +2,6 @@
 // Module core/structure
 //  Handles producing the ToC and numbering sections across the document.
 
-// LIMITATION:
-//  At this point we don't support having more than 26 appendices.
 // CONFIGURATION:
 //  - noTOC: if set to true, no TOC is generated and sections are not numbered
 //  - tocIntroductory: if set to true, the introductory material is listed in the TOC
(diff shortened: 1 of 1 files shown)


sglaser:
Sure. That means I have to figure them out, but that's goodness.

marcoscaceres:
@sglaser if it helps, we have a little bit of documentation here:
https://github.com/w3c/respec/wiki/Developers-Guide#testing

The related tests, which you can probably copy/paste and modify from, related to headers:
https://github.com/w3c/respec/blob/develop/tests/spec/core/structure-spec.js

Happy to help if you have questions!

src/core/structure.js
```diff
@@ -245,3 +242,12 @@ function createTableOfContents(ol) {
   const link = hyperHTML`<p role='navigation' id='back-to-top'><a href='#title'><abbr title='Back to Top'>&uarr;</abbr></a></p>`;
   document.body.append(link);
 }
+
+function appendixNumber(index) {
```
sidvishnoi: Can you confirm that we want 26 to map to 'BA' instead of 'AA'?
I added some tests based on how spreadsheets handle index to column name mapping, which I think is the desirable.

``` js
const tests = [
  [0, "A"],
  [1, "B"],
  [2, "C"],
  [25, "Z"],
  [26, "AA"],
  [27, "AB"],
  [28, "AC"],
  [54, "BC"]
];

for (const [index, expected] of tests) {
  const result = appendixNumber(index);
  console.assert(result === expected, "", { index, result, expected });
}
```

sidvishnoi: I wrote a version which passes above tests. WDYT?
``` js
function appendixNumber(index) {
  index += 1; // index is 0 based
  let s = "";
  while (index > 0) {
    index -= 1;
    s = String.fromCharCode(65 + (index % 26)) + s;
    index = Math.floor(index / 26);
  }
  return s;
}
```

sidvishnoi:
@sglaser Did you get a chance to add tests for these PRs?

PR #2772, opened 2020-02-25
Support formatting of section headers above h6, correct computation of level

sglaser:
Headers above h6 will be of form &lt;h6 class="h7"&gt; so css can do magic.

The level computation assumed section numbers and appendix numbers were a single digit/character


sglaser:
I'll add a test for this. Please review but don't merge yet.

src/core/structure.js
```diff
@@ -202,11 +202,18 @@ function renameSectionHeaders() {
...
+    const depth = parents(header, "section").length + 1;
+    const h = `h${Math.min(depth, 6)}`;
     if (header.localName !== h) {
       renameElement(header, h);
     }
+    if (depth > 6) {
+      if (header.classList) {
+        header.classList.add(`h${depth}`);
+      } else {
+        header.className = `h${depth}`;
```
sidvishnoi: We don't need this as a special case as we support modern browsers (which support `Element.classList`), so the following is enough:
``` js
if (depth > 6) {
  header.classList.add(`h${depth}`);
}
```

sidvishnoi:
I'm a bit concerned about the accessibility implications of this. As per [following page from WAI wiki](https://www.w3.org/WAI/GL/wiki/Using_role%3Dheading_to_identify_headings#Example_2), the recommended way of supporting heading level beyond h6 is:
``` html
<div role="heading" aria-level="7">Jonagold/div>
```
Also see: https://stackoverflow.com/questions/55891025/is-h6-aria-level-7-a-reliable-way-to-create-a-h7-element

If it's fine to treat all headings above h6 as h6, I think this would be fine though.

sglaser:
I understand the implications of deep nesting. I’m faced with an existing specification that I can’t change that uses h7.

I’ll fix the “bug” where I unnecessarily checked for an empty classList.

Steve Glaser
sglaser@sglaser.com

src/core/structure.js
sglaser: Fixed and updated.

marcoscaceres:
@stevefaulkner, we could use your guidance here: if we want a "h7", should we do `<h6 aria-level="7">` or just use a `<div aria-level="7"></div>`?

marcoscaceres:
Closing as stale.

PR #2775, opened 2020-02-27
Fix typo

acka47:


acka47:
> This is fine, but it should be fixed in src/w3c/conformance.js as well.

Done

Issue #2777, opened 2020-03-03
Support accessibility checks via linter 


marcoscaceres:
In a different forum, some folks were asking about supporting a11y checks on specs. We might be able to hook up Axe: 

https://github.com/dequelabs/axe-core

Axe looks pretty trivial to implement. We could ask the w3c to host the javascript library (or we just put it on respec.org), then we can make a simple opt-in option in ReSpec to add the required script tag and we can report out errors/warnings directly in ReSpec.

PR #2779, opened 2020-03-04
feat(core/a11y): add axe-core integration

sidvishnoi:
Closes https://github.com/w3c/respec/issues/2777

- The linter API is not flexible enough with error reporting, so created a new core/a11y module.
- Axe is quite slow on W3C specs. Need to disable a lot of rules, or provide authors a way to select rules. Should be used locally only to fix issues, not on live versions.

### DIFF
--- a/src/core/a11y.js
+++ b/src/core/a11y.js
@@ -1,4 +1,8 @@
 // @ts-check
+/**
+ * Module: core/a11y
+ * Lints for accessibility issues using axe-core package.
+ */
 
 import { pub } from "./pubsubhub.js";
 import { showInlineWarning } from "./utils.js";
--- a/src/core/a11y.js
+++ b/src/core/a11y.js
@@ -9,8 +9,11 @@ const DISABLED_RULES = [
   "color-contrast", // too slow 🐢
 ];
 
-export async function run(_conf) {
-  // TODO: run conditionally based on conf
+export async function run(conf) {
+  if (!conf.a11y) {
+    return;
+  }
+
   const violations = await getViolations();
   for (const violation of violations) {
     /**
--- a/src/core/a11y.js
+++ b/src/core/a11y.js
@@ -1,0 +1,86 @@
+import { showInlineWarning } from "./utils.js";
+
+export const name = "core/a11y";
+
+const DISABLED_RULES = [
+  "landmark-one-main", // need to add a <main>, else it marks entire page as errored
+  "landmark-unique",
+  "region",
+  "color-contrast", // too slow 🐢
+];
+
+export async function run(_conf) {
+  // TODO: run conditionally based on conf
+  const violations = await getViolations();
+  for (const violation of violations) {
+    /**
+     * We're grouping by failureSummary as it contains hints to fix the issue.
+     * For example, with color-constrast rule, it tells about the present color
+     * contrast and how to fix it. If we don't group, errors will be repetitive.
+     * @type {Map<string, HTMLElement[]>}
+     */
+    const groupedBySummary = new Map();
+    for (const node of violation.nodes) {
+      const { failureSummary, element } = node;
+      const elements =
+        groupedBySummary.get(failureSummary) ||
+        groupedBySummary.set(failureSummary, []).get(failureSummary);
+      elements.push(element);
+    }
+
+    const { id, help, description, helpUrl } = violation;
+    const title = `a11y/${id}: ${help}`;
+    for (const [failureSummary, elements] of groupedBySummary) {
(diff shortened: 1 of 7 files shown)


src/core/a11y.js
```diff
@@ -0,0 +1,86 @@
...
+const DISABLED_RULES = [
+  "landmark-one-main", // need to add a <main>, else it marks entire page as errored
+  "landmark-unique",
+  "region",
+  "color-contrast", // too slow 🐢
+];
+
+export async function run(_conf) {
+  // TODO: run conditionally based on conf
+  const violations = await getViolations();
```
marcoscaceres: Try/catch, as we discussed on Slack.

src/core/a11y.js
```diff
@@ -0,0 +1,101 @@
...
+    elementRef: true,
+    resultTypes: ["violations"],
+    reporter: "v1", // v1 includes a `failureSummary`
+  };
+
+  let axe;
+  try {
+    axe = await importAxe();
+  } catch (error) {
+    pub("error", "Failed to load a11y linter.");
```
marcoscaceres:
```suggestion
    pub("error", "Failed to load a11y linter. See developer console for details.");
```

sidvishnoi: Heh. I added it and then removed it for no reason.

sidvishnoi:
Demo: [Manifest spec with color-contrast rule](https://respec-preview.netlify.com/?spec=https%3A%2F%2Fw3c.github.io%2Fmanifest%2F&version=https%3A%2F%2Fdeploy-preview-2779--respec-pr.netlify.com%2Frespec-w3c.js&respecConfig=respecConfig.a11y+%3D+%7B+runOnly%3A+%5B%27color-contrast%27%5D+%7D)
Warning: color-contrast rule is slow and will hand the document. Shown in example to demonstrate rules can be overridden if needed.

src/core/a11y.js
```diff
@@ -0,0 +1,103 @@
+import { pub } from "./pubsubhub.js";
```
marcoscaceres: Might be nice to have a short pre-amble just stating what the module does and how it works.

src/core/a11y.js
```diff
@@ -0,0 +1,103 @@
...
+
+function importAxe() {
+  return new Promise((resolve, reject) => {
+    const script = document.createElement("script");
+    script.classList.add("remove");
+    script.src = "https://unpkg.com/axe-core@3/axe.min.js";
+    script.onload = () => resolve(window.axe);
+    script.onerror = reject;
+    document.head.appendChild(script);
+  });
```
marcoscaceres: Nit: 

```suggestion
  const script = document.createElement("script");
  script.classList.add("remove");
  script.src = "https://unpkg.com/axe-core@3/axe.min.js";
  document.head.appendChild(script);
  return new Promise((resolve, reject) => {
    script.onload = () => resolve(window.axe);
    script.onerror = reject;
  });
```

src/core/a11y.js
```diff
@@ -0,0 +1,107 @@
...
+}
+
+/** @returns {Promise<typeof window.axe>} */
+function importAxe() {
+  const script = document.createElement("script");
+  script.classList.add("remove");
+  script.src = "https://unpkg.com/axe-core@3/axe.min.js";
+  document.head.appendChild(script);
+  return new Promise((resolve, reject) => {
+    script.onload = () => resolve(window.axe);
```
saschanaz: So it works as a global script? Bit sad...

sidvishnoi: :crying_cat_face:
They do plan to make it a ES module though.

marcoscaceres:
I can write the docs if you’d like.

src/core/a11y.js
saschanaz: Maybe rebuild as a module, as we do for jQuery and pluralize? Bit different as they are already CJS but should still be doable.

sidvishnoi: That would mean we either bundle this ~400KB library into ReSpec, or ask W3C to host it separately. unpkg seems fine to me, given how infrequently this feature would be used (won't be used in live versions for sure, as it's slow... but only during local spec authoring)

sidvishnoi:
I wrote a draft at https://github.com/w3c/respec/wiki/a11y

marcoscaceres:
Thanks for updating the docs! I did a little cleanup. I think the expectation is run Axe on CI, but also on normal specs. I'll let the a11y folks know about the potential pitfalls of enabling too many features - or that they should only enable certain features in CI.

PR #2785, opened 2020-03-09
fix(core/a11y): allow overriding rules while keeping disabled ones

sidvishnoi:
Earlier, if one added `rules` option, it'll run even the ones we have disabled. One can still enable the rules we've disabled, but would need to be explicit.


sidvishnoi:
Updated docs.

PR #2787, opened 2020-03-09
fix(assets/webidl.css): fix color contrast in IDL header

sidvishnoi:


assets/respec.css
```diff
@@ -79,7 +79,7 @@ cite .bibref {
 }
 
 code {
-  color: #c83500;
+  color: #c63501;
```
sidvishnoi: - `#c83500` (or `hsl(15.9, 100%, 39.2%)`) (color contrast: 4.48, fail)
- `#c63501` (or `hsl(15.8, 99%, 39.0%)`) (color contrast: 4.54, pass)
(at least the a11y linter is not labeling 20% of page as error)

assets/webidl.css
```diff
@@ -24,7 +24,6 @@ pre.idl {
 }
 
 .idlHeader a.self-link {
-   color: #fff;
```
sidvishnoi: It will use the default dark blue-ish link color from base.css

assets/webidl.css
```diff
@@ -14,7 +14,7 @@ pre.idl {
 .idlHeader {
   display: block;
   width: 150px;
-  background: #90b8de;
+  background: #8ccbf2;
```
sidvishnoi: Same as border color or IDL block

sidvishnoi: Before:
image

After:
image

marcoscaceres: hehe, 👓... totally see the difference!

marcoscaceres: in all seriousness, I did prefer the white... but can live with the dark.

PR #2800, opened 2020-03-22
chore: update prettier, style changes

sidvishnoi:
Closes https://github.com/w3c/respec/pull/2799 (which uses prettier 2.0.0)

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -1,7 +1,7 @@
 // Karma configuration
 // Generated on Fri Feb 26 2016 13:09:51 GMT+1100 (AEDT)
 "use strict";
-module.exports = function(config) {
+module.exports = function (config) {
   const options = {
     // base path that will be used to resolve all patterns (eg. files, exclude)
     basePath: "./",
--- a/src/core/dfn-finder.js
+++ b/src/core/dfn-finder.js
@@ -182,10 +182,7 @@ export function decorateDfn(dfnElem, idlAst, parent, name) {
   if (!dfnElem.id) {
     const lCaseParent = parent.toLowerCase();
     const middle = lCaseParent ? `${lCaseParent}-` : "";
-    let last = name
-      .toLowerCase()
-      .replace(/[()]/g, "")
-      .replace(/\s/g, "-");
+    let last = name.toLowerCase().replace(/[()]/g, "").replace(/\s/g, "-");
     if (last === "") last = "the-empty-string";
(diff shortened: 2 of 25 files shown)


karma.conf.js
```diff
@@ -1,7 +1,7 @@
 // Karma configuration
 // Generated on Fri Feb 26 2016 13:09:51 GMT+1100 (AEDT)
 "use strict";
-module.exports = function(config) {
+module.exports = function (config) {
```
sidvishnoi: Not a fan of these changes - a space after "function" keyword

src/core/examples.js
```diff
@@ -62,9 +62,7 @@ function makeTitle(elem, num, report) {
   if (report.title) elem.removeAttribute("title");
   const number = num > 0 ? ` ${num}` : "";
   const title = report.title
-    ? html`
-        <span class="example-title">: ${report.title}</span>
-      `
+    ? html` <span class="example-title">: ${report.title}</span> `
```
saschanaz: Need to check we want the space or not. I think generally we don't.

karma.conf.js
saschanaz: Might be worth to check the relevant discussion...

saschanaz: https://github.com/prettier/prettier/pull/3903 is the culprit, and seemingly the majority of people liked it.

src/core/dfn-finder.js
```diff
@@ -182,10 +182,7 @@ export function decorateDfn(dfnElem, idlAst, parent, name) {
   if (!dfnElem.id) {
     const lCaseParent = parent.toLowerCase();
     const middle = lCaseParent ? `${lCaseParent}-` : "";
-    let last = name
-      .toLowerCase()
-      .replace(/[()]/g, "")
-      .replace(/\s/g, "-");
+    let last = name.toLowerCase().replace(/[()]/g, "").replace(/\s/g, "-");
```
saschanaz: No needless newlines 🎉

src/core/examples.js
```diff
@@ -62,9 +62,7 @@ function makeTitle(elem, num, report) {
...
     : "";
   return html`
     <div class="marker">
```
saschanaz: With prettier 2.0 this can be:

```js
html`<div>
  ...
</div>`;
```

This will reduce one indentation level, what do you think @sidvishnoi?

sidvishnoi: Sounds good to me. :+1:

PR #2801, opened 2020-03-22
refactor: remove redundant spaces around elements

saschanaz:
Removed spaces around templates with a single root element, as those will always be trimmed by hyperHTML and thus completely redundant.


src/core/examples.js
```diff
@@ -81,11 +79,9 @@ export async function run() {
...
   const css = await cssPromise;
   document.head.insertBefore(
-    html`
-      <style>
-        ${css}
-      </style>
-    `,
+    html`<style>
+      ${css}
+    </style>`,
```
sidvishnoi: Will this work?
```suggestion
    html`<style>${css}</style>`,
```

saschanaz: With whatever reason, no. 🤔

src/core/examples.js
```diff
@@ -62,14 +62,12 @@ function makeTitle(elem, num, report) {
...
-    ? html` <span class="example-title">: ${report.title}</span> `
+    ? html`<span class="example-title">: ${report.title}</span>`
     : "";
-  return html`
-    <div class="marker">
-      <a class="self-link">${l10n.example}<bdi>${number}</bdi></a
-      >${title}
-    </div>
-  `;
+  return html`<div class="marker">
```
sidvishnoi: On a second thought, maybe extra indentation was better. Looks a bit sticky. WDYT?

saschanaz: I personally like this as:

* Consistency. We definitely don't want spaces in cases like `` html`  ` ``, so it's consistent to apply the same rule.
   * 2.0 will collapse newlines into a space if the element is not too long: `\n<div>...</div>\n` can potentially be collapsed into ` <div>...</div> `, where we certainly don't want the spaces. This can prevent such fluctuation.
* Does not depend on trimming behavior from a specific library. AFAIK not all libraries do this, e.g. `lit-html` from Google.

saschanaz: There was a similar opinion when I open the prettier PR, though. The conclusion was: let's make this right first and decide further styling later.

sidvishnoi: One reason I was opposing this is that there are too many (avoidable) changes caused by this, which makes it slightly difficult for git blame. Fine with what you decide though.

> Does not depend on trimming behavior from a specific library. 

👍 💯

saschanaz: >which makes it slightly difficult for git blame

True that. But we shouldn't use prettier in that case 😂, like, the big changes it forced to ReSpec when it first introduced HTML template formatting...

sidvishnoi: That's what we get for opinionated code formatters 😄

PR #2803, opened 2020-03-24
fix(core/data-cite): clean up before save only

sidvishnoi:
- We can `data-cite` in later plugins now (will be needed in external dependency list)
- Also, xref to clean up after itself


marcoscaceres:
Hmm... I think we should probably not do the cleanup (and remove the cleanup from `core/data-cite.js`). Otherwise, it makes it really hard to debug things when things go wrong.

sidvishnoi:
We should probably do better cleanup before export though

marcoscaceres:
> We should probably do better cleanup before export though

Good idea! We could run it on save!

marcoscaceres:
What might be better is that each module has its own cleanup listener tho. Otherwise centralizing everything is a bit messy.

marcoscaceres:
(breaking tests aside, lol)

PR #2804, opened 2020-03-24
chore(karma.conf.js): annonate unit test failures in GitHub PR

sidvishnoi:
Show a failure summary in GitHub actions, without scrolling into the logs.
annonated test failures in GitHub PR

~TODO: create a custom karma reporter maybe.~


sidvishnoi:
Phew. This took some effort and lots of learning!
Please see the [files](https://github.com/w3c/respec/pull/2804/files) tab for failing tests :tada: (I've intentionally failed some tests)

package.json
```diff
@@ -51,6 +51,7 @@
     "karma-mocha": "^1.3.0",
     "karma-mocha-reporter": "^2.2.5",
     "karma-opera-launcher": "^1.0.0",
+    "karma-respec-github-action-reporter": "github:sidvishnoi/karma-respec-github-action-reporter",
```
saschanaz: If this is mainly for ReSpec maybe make it a local module?

sidvishnoi: Karma only picks from node_modules :disappointed:

saschanaz: Seems there is a way to `require()` a module though: https://karma-runner.github.io/4.0/config/plugins.html

sidvishnoi: Ouch. Didn't see it.

saschanaz: Make sure that you also add `"karma-*"` to the `plugins` field or it will stop autoloading karma plugins.

karma.conf.js
```diff
@@ -173,5 +173,94 @@ module.exports = function (config) {
     options.detectBrowsers.enabled = false;
     options.browsers = process.env.BROWSERS.split(" ");
   }
+
+  if (process.env.GITHUB_WORKFLOW) {
+    const { readdirSync } = require("fs");
+    const defaultPlugins = readdirSync("node_modules").filter(m =>
+      m.startsWith("karma-")
```
saschanaz: Didn't `"karma-*"` work? 😮

sidvishnoi: Oh. You literally meant `"karma-*"`. :sweat_smile:

karma.conf.js
```diff
@@ -173,5 +173,90 @@ module.exports = function (config) {
...
+      { "reporter:respec-github-action": ["type", GithubActionReporter] },
+    ];
+    options.reporters.push("respec-github-action");
+    options.plugins = ["karma-*"].concat(localPlugins);
+  }
+
   config.set(options);
 };
+
+function GithubActionReporter() {
```
saschanaz: I really want to move this to a separate file, something like `tools/github-action-reporter.js`. Is there a reason to keep it here?

sidvishnoi: Not a good reason. Just wanted to keep it within karma as it's related. Will move.

tools/github-action-reporter.js
```diff
@@ -0,0 +1,79 @@
...
+}
+
+/**
+ * @param {string} command
+ * @param {string} message
+ * @param {Record<string, string>} options
+ */
+function print(command, message = "", options = {}) {
+  let optionsString = Object.entries(options)
+    .map(([k, v]) => `${k}=${escapeProperty(v)}`)
```
saschanaz: Node has `URLSearchParams` too, you can just use it instead of escaping everything manually.

tools/github-action-reporter.js
```diff
@@ -0,0 +1,79 @@
...
+/**
+ * @typedef {{ test: string[], file: string, line: number, col: number, message: string }} Failure
+ * @returns {Failure}
+ */
+function parseFailure(result) {
+  // convert newlines into array and flatten
+  const log = result.log.flatMap(message => message.split("\n"));
+  const { suite, description } = result;
+  const message = log[0].replace("Error: ", "");
+  const location = log[2].split(":9876/base/", 2)[1].replace(/\)$/, "");
```
saschanaz: Could you extract the port number as a const with some comment? Might be easier to understand later.

tools/github-action-reporter.js
sidvishnoi: Let me try... I think this is using a different escaping than URLSearchParams. I copied it from [@actions/core](https://github.com/actions/toolkit/tree/master/packages/core)

sidvishnoi: Didn't work :disappointed: 
Did I do what you expected or you meant something else?

saschanaz: URLSearchParams uses `&` as the separator and here it's `,`, so `.replace(/&/, ",")` is needed.

tools/github-action-reporter.js
```diff
@@ -52,7 +53,9 @@ function printFailure(failure) {
  * @param {Record<string, string>} options
  */
 function print(command, message = "", options = {}) {
-  let optionsString = new URLSearchParams(options).toString();
+  let optionsString = new URLSearchParams(options)
+    .toString()
+    .replace(/&/g, "%2C");
```
saschanaz: I don't think `%2C` will work as a separator 👀

sidvishnoi: I'm trying to follow what @actions/core did :thinking: 
Should I try `,` instead?

sidvishnoi: Wait. Did you mean to not run replace again and again, and instead run it once on final string?

saschanaz: I think `,` is what the previous code did, so yes.

saschanaz: No, just here, because the purpose here is to use `,` as a separator between key-values.

sidvishnoi: I'm gonna try once more... I should probably be sleeping on it :stuck_out_tongue:

sidvishnoi: /me gives up for tonight

saschanaz: Good night!

sidvishnoi: Feel free to edit it while I sleep :smile_cat:

.github/workflows/pr-push.yml
```diff
@@ -7,62 +7,62 @@ on:
...
-        name: Cache node modules
-        uses: actions/cache@v1
-        with:
-          path: ./node_modules
-          key: ${{ runner.os }}-node-10-${{ hashFiles('**/package-lock.json') }}
-      - if: "!contains(steps.log.outputs.message, 'ci skip')"
-        run: npm install
-      - if: "!contains(steps.log.outputs.message, 'ci skip')"
-        run: npm run lint
+  # lint:
```
sidvishnoi: IDK why I commented it out instead of deleting :laughing:

tools/github-action-reporter.js
sidvishnoi: Had an unrelated error. I refactored to use `Array.prototype.flatMap`, which doesn't work on Node v10 (which is used in our GitHub workflow), and I was debugging ~15 wrong commits.
Also, had to unescape `%2F` to `/`.

PR #2809, opened 2020-03-25
chore(ci): fix node_modules caching issue

sidvishnoi:
Unfortunately, Snyk will run unnecessarily, but it's better than randomly failing tests.


sidvishnoi:
Here's what was happening:
We use the cache with linter, and then run `npm install --ignore-scripts` - which removed puppeteer from node_modules, and then save it in the cache, causing subsequent jobs randomly failing.

Issue #2560, opened 2019-10-30
Auto generate list of external dependencies


marcoscaceres:
It would be nice ReSpec outputted a a list of terms cited by reference... and, if we know which what the type of a term is, we labelled the term correctly. For example:

"The HTML spec defines the following concepts: queue a task. It also defines the following elements: `<body>`. And it defines the following IDL fragments: `HTMLWhateverThing` interface, and the `Foo` dictionary."

sidvishnoi:
`xref` adds a `data-type` attribute on all the links, so we just need to collect all terms by `data-cite` and then by `data-type` and render them. 

What about the links which editors added without using xref? Should we show them as "misc" ("xyz also defines a, b, c") or ignore them?

From the current structure of code, this all has to be done before calling `linkInlineCitations` (in `core/link-to-dfn`), as it includes a cleanup step which removes `data-cite*` stuff (which I think should be done in end of all plugins).

Perhaps we should do a refactoring of link-to-dfn.js and data-cite.js first.

marcoscaceres:
> What about the links which editors added without using xref? Should we show them as "misc" ("xyz also defines a, b, c") or ignore them?

We should add them... but we need to look at conventions of how specs do this manually already. For example:
https://fetch.spec.whatwg.org/#index

Basically, the sections:
"Terms defined by this specification"
"Terms defined by reference"

ridhishjain:
@marcoscaceres @sidvishnoi I would like to work on this. Seems challenging, a little help regarding understanding this more thoroughly would be great!

sidvishnoi:
As Marcos said "we need to look at conventions of how specs do this manually already". So, presently, we need to do some research how other documents try to list terms defined externally, and then decide which one we want to support. 
Some examples of what we're trying to generate:
https://w3c.github.io/manifest/#terms-defined-by-reference
https://w3c.github.io/sensors/#index-defined-elsewhere
https://w3c.github.io/payment-method-basic-card/#dependencies
https://fetch.spec.whatwg.org/#index-defined-elsewhere

So, we need to find more examples and figure out what we want to support. Only after that, we'll look into implementation.

sidvishnoi:
Taking following features from examples above:

- https://w3c.github.io/manifest/#terms-defined-by-reference
  - links to the [reference] also 
- https://fetch.spec.whatwg.org/#index-defined-elsewhere, https://w3c.github.io/sensors/#index-defined-elsewhere
  - back references to usage in current document
[...]

PR #2810, opened 2020-03-25
feat(core/dfn-index): generate list of externally referenced terms

sidvishnoi:
Closes https://github.com/w3c/respec/issues/2560
[Demo with [payment-request]](https://respec-preview.netlify.com/?spec=https%3A%2F%2Fw3c.github.io%2Fpayment-request%2F&version=https%3A%2F%2Fdeploy-preview-2810--respec-pr.netlify.com%2Frespec-w3c.js&respecConfig=setTimeout%28%28%29+%3D%3E+document.body.insertAdjacentHTML%28%27beforeend%27%2C+%27%3Csection+id%3D%22index%22%3E%3C%2Fsection%3E%27%29%2C+0%29%3B)

### DIFF
--- a/src/core/dfn-index.js
+++ b/src/core/dfn-index.js
@@ -57,8 +57,8 @@ export async function run(conf) {
  */
 function createExternalTermIndex(toCiteDetails) {
   const data = collectExternalTerms(toCiteDetails);
-  const dataSortedBySpec = [...data.entries()].sort((a, b) =>
-    a[0].localeCompare(b[0])
+  const dataSortedBySpec = [...data.entries()].sort(([specA], [specB]) =>
+    specA.localeCompare(specB)
   );
   return html`<ul class="index">
     ${dataSortedBySpec.map(
--- a/src/core/dfn-index.js
+++ b/src/core/dfn-index.js
@@ -10,6 +10,7 @@ import { fetchAsset } from "./text-loader.js";
 import { getTermFromElement } from "./xref.js";
 import { hyperHTML as html } from "./import-maps.js";
 import { renderInlineCitation } from "./render-biblio.js";
+import { sub } from "./pubsubhub.js";
 
 export const name = "core/dfn-index";
 
@@ -39,6 +40,8 @@ export async function run(conf) {
     ${createExternalTermIndex(toCiteDetails)}
   </section>`;
   index.append(externalTermIndex);
+
+  sub("beforesave", cleanup);
 }
 
 /**
@@ -51,7 +54,7 @@ function createExternalTermIndex(toCiteDetails) {
   );
   return html`<ul class="index">
     ${dataSortedBySpec.map(
-      ([spec, entries]) => html`<li>
+      ([spec, entries]) => html`<li data-spec="${spec}">
         ${renderInlineCitation(spec)} defines the following:
         <ul>
           ${entries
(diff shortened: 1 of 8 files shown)


tests/spec/core/dfn-index-spec.js
```diff
@@ -0,0 +1,244 @@
...
+    /** @type {HTMLElement} */
+    let index;
+    beforeAll(async () => {
+      const ops = makeStandardOps({ xref: "web-platform" }, body);
+      const doc = await makeRSDoc(ops);
+      index = doc.getElementById("index-defined-elsewhere");
+    });
+
+    it("doesn't list local terms", () => {
+      const terms = index.querySelectorAll("ul li li");
```
sidvishnoi:
```suggestion
      const terms = index.querySelectorAll(".index-term");
```

src/core/dfn-index.js
```diff
@@ -0,0 +1,214 @@
...
+    return;
+  }
+
+  const styleEl = document.createElement("style");
+  styleEl.textContent = await loadStyle();
+  document.head.appendChild(styleEl);
+
+  index.classList.add("appendix");
+  if (!index.querySelector("h2")) {
+    index.prepend(html`<h2>Index</h2>`);
```
marcoscaceres: could be localized.

src/core/dfn-index.js
```diff
@@ -0,0 +1,214 @@
...
+  }
+
+  const toCiteDetails = citeDetailsConverter(conf);
+
+  const externalTermIndex = html`<section id="index-defined-elsewhere">
+    <h3>Terms defined by reference</h3>
```
marcoscaceres: as above... I guess all the headings are localization targets.

src/core/dfn-index.js
```diff
@@ -0,0 +1,214 @@
...
+
+  sub("beforesave", cleanup);
+}
+
+/**
+ * @param {ReturnType<typeof citeDetailsConverter>} toCiteDetails
+ */
+function createExternalTermIndex(toCiteDetails) {
+  const data = collectExternalTerms(toCiteDetails);
+  const dataSortedBySpec = [...data.entries()].sort((a, b) =>
```
marcoscaceres: Could this be?: 
```suggestion
  const dataSortedBySpec = [...data.entries()].sort(([a], [b]) => a.localeCompare(b))
```
Can't remember if that's valid destructure or not...

src/core/dfn-index.js
```diff
@@ -0,0 +1,214 @@
...
+  "enum",
+  "exception",
+  "extended-attribute",
+  "interface",
+  "method",
+  "typedef",
+]);
+
+// Terms of these _types_ are suffixed with their type info.
+const TYPED_TYPES = new Map([
```
marcoscaceres: Maybe it could be we only have the special cases here... if it's not in the TYPED_TYPES, then just use what is given?  🤔 doesn't really matter too much tho.

src/core/dfn-index.js
sidvishnoi: How about `[SPEC] defines the following: `?

src/core/dfn-index.js
sidvishnoi: It gets noisy otherwise, so limited what gets type info. For example, "creating an event dfn", "foo enum-value".
Also, it's used to map human-friendly terms, like { x, element-attr } becomes "x attribute (for y element)"

tests/spec/core/dfn-index-spec.js
```diff
@@ -0,0 +1,244 @@
+"use strict";
+import { flushIframes, makeRSDoc, makeStandardOps } from "../SpecHelper.js";
+
+describe("Core — dfn-index", () => {
+  afterAll(flushIframes);
+
+  it("adds default heading to section#index", async () => {
+    const body = `<section id="index">
+      <p>PASS</p>
```
marcoscaceres: this pass is not checked in the test... might be just worth checking that PASS is next sibling of the heading.

Issue #1668, opened 2018-05-13
Separate static and dynamic parts


saschanaz:
For example:

`webidl-clipboard.js` has a part that can be static (adding clipboard copy buttons) and a dynamic UI part (adding event listeners to those added buttons). Currently respec2html removes any dynamic UI features and this causes weird situation that users cannot use the feature in respec2html-generated page.

We may separate those two parts so that we can ultimately:

1. serve static parts as respec2html already does
2. serve dynamic UI parts as separate modules
3. avoid removeOnSave

What do you think?

marcoscaceres:
@plehegar, tl;dr: ReSpec in "live mode" (e.g., Editor's drafts) adds useful things like buttons that allow readers to copy/paste WebIDL with one click. The W3C currently doesn't allow JS on TR (apart from fixup.js), so we are wondering... what would we need to be allowed to ship some additional JS on TR?

saschanaz:
Bikeshed adds inline scripts, maybe we can do the same?

marcoscaceres:
> Bikeshed adds inline scripts, maybe we can do the same?

On documents published on /TR/? If yes, then we can make a case to the W3C.

saschanaz:
>On documents published on /TR/?

Yes, check this out for example: https://www.w3.org/TR/webaudio/

saschanaz:
https://www.w3.org/TR/clreq/ has an arbitrary JS file. Not sure it's explicitly allowed or not.

marcoscaceres:
Ok, we can definitely do the either inline or we can host our own scripts. I'd personally like to have the scripts hosted (same a ReSpec currently is)... we have very few scripts that would need hosting. Primarily the copy/paste IDL one.

plehegar:
Missed this somehow, catching up...

It is no longer true that we don't allow JS except for fixup.js.

W3C TR allows "recognized" JS. The list is at https://www.w3.org/scripts/ (this is not a public page apparently, not sure why). https://www.w3.org/TR/webaudio/ uses MathJAX for example. https://www.w3.org/TR/clreq/ might be a mistake however (but pubrules doesn't seem to flag it).
Keep in mind that similarly to CSS files, we don't allow changes except for bug fixes to the scripts once the document has been published (thus we link to specific library versions). You may publish a new version of the script for every publication but, given that we'd need to review the script for each publication, we won't scale fast if it becomes common practice.

[...]

PR #2811, opened 2020-03-26
feat(core/dfn-panel): run in exported documents

sidvishnoi:
This is a very hacky way. Anyone with better ideas?


tests/spec/core/dfn-panel-spec.js
```diff
@@ -165,4 +165,46 @@ describe("Core — dfnPanel", () => {
...
   });
+
+  it("works in exported document", async () => {
+    const rdoc = await makeRSDoc(ops);
+    const doc = await getExportedDoc(rdoc);
+
+    const dfnOne = doc.querySelector("dfn#dfn-one");
+    dfnOne.click();
+    const panel = doc.getElementById("dfn-panel");
+    expect(panel).toBeTruthy();
```
sidvishnoi: works locally :thinking:

saschanaz:
#1668 is related. I think we should make the whole module to be injected as a `<script>` so that it can run in a same way both in raw and exported cases.

tests/spec/core/dfn-panel-spec.js
saschanaz: That GitHub action is so cool

saschanaz:
My thought: If we have any states that can/must be computed early, the code may expose them as a `<script type="application/json">` so that a module as a `<script>` can use it.

sidvishnoi:
How about create a runtime profile (`profiles/w3c-static.js`)? We can inject it after `fixup.js` (as a `script[src]`). My idea is to export a `main()` as static part of a module and make base-runner run `main` instead of `run`.

saschanaz:
I think creating a new profile can introduce breaking changes too easily. I vote to inject as a inline script.

marcoscaceres:
If it helps, the W3C can host the panel code for us. Inlining is 🤢.

saschanaz:
>If it helps, the W3C can host the panel code for us. Inlining is 🤢.

What if we have to change the structure and it breaks old specs? Inlining can be 🤢 but backward compatibility is 🤢🤢.

sidvishnoi:
We've decided to go with inlining. But we need to figure out a strategy for that. Inlining like in this PR is :nauseated_face:

marcoscaceres:
Difficult tradeoffs here... but seems inline is the way to go.

sidvishnoi:
> My thought: If we have any states that can/must be computed early, the code may expose them as a `<script type="application/json">` so that a module as a `<script>` can use it.

@saschanaz can you expand on this approach? I'm not liking the way I've done in this PR, so I'm open to other ideas.

sidvishnoi:
> The existing dfn-panel module inserts this new runtime module into the document

We would still need to serialize using `fn.toString()`, right?

saschanaz:
>We would still need to serialize using fn.toString(), right?

No, we would get the runtime module as a string and then `<script>${module}</script>`.

src/core/dfn-panel.js
```diff
@@ -67,15 +74,15 @@ function createPanel(dfn) {
   const href = `#${id}`;
   const links = document.querySelectorAll(`a[href="${href}"]`);
 
-  /** @type {HTMLElement} */
-  const panel = hyperHTML`
+  const tempNode = document.createElement("div");
+  tempNode.innerHTML = `
```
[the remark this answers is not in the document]
sidvishnoi: Still need to get `firstElementChild` at least. Can't style a `DocumentFragment` 😅

sidvishnoi:
Closed in favor of https://github.com/w3c/respec/pull/2820

PR #2812, opened 2020-03-27
feat(core/dfn-panel): smarter positioning, better styles

sidvishnoi:
Now the panel has a caret (triangle, whatever it's called). The panel is now placed close to where user has clicked (and always below the click location, never left or right of dfn). If there is not enough space, the panel moves but the caret represents the location of user click, providing a better UX.
Also, changed the colors to make it look less like Windows 95.


assets/dfn-panel.css
```diff
@@ -4,20 +4,36 @@ dfn {
...
 .dfn-panel {
+  --fill: #fff;
   position: absolute;
   left: var(--left); /* set via JS */
   top: var(--top); /* set via JS */
   z-index: 35;
-  height: auto;
-  width: max-content;
-  max-width: 300px;
-  max-height: 500px;
```
sidvishnoi: If we want overflow with max-height, then we would need to add a wrapper to dfn-panel to display the caret. Otherwise, as both caret and panel are `position: absolute`, the panel hides the caret.

sidvishnoi:
[Demo](https://user-images.githubusercontent.com/8426945/77779914-39206700-7079-11ea-957a-d394be7e938b.gif) (GIF, 3.8MB)

marcoscaceres:
Amazing! that's super cool.

assets/dfn-panel.css
marcoscaceres: let's go with this for now. If we need to, we can modify later.

PR #2814, opened 2020-03-27
chore(tools/builder): minify CSS

sidvishnoi:
Reduces W3C bundle size by 8KB

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1295,6 +1295,15 @@
       "resolved": "https://registry.npmjs.org/ci-info/-/ci-info-1.6.0.tgz",
       "integrity": "sha512-vsGdkwSCDpWmP80ncATX7iea5DWQemg1UgCW5J8tqjU3lYw4FBYuj89J0CTVomA7BEfvSZd84GmHko+MxFQU2A=="
     },
+    "clean-css": {
+      "version": "4.2.3",
+      "resolved": "https://registry.npmjs.org/clean-css/-/clean-css-4.2.3.tgz",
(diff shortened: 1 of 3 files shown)


saschanaz:
How about gzipped size? 👀

sidvishnoi:
> What about gzipped size? 👀

No change 👻

saschanaz:
So no practical change. I'm afraid that this just complicates our build script...

sidvishnoi:
No benefit with download, but that's 8kb less JavaScript to parse.

sidvishnoi:
We can merge the functionality of `string` plugin in this. Might simplify (less plugins)

saschanaz:
> No benefit with download, but that's 8kb less JavaScript to parse.

But it's just removing whitespaces, not really any optimization as JS minifiers do, right?

sidvishnoi:
> But it's just removing whitespaces, not really any optimization as JS minifiers do, right?

Yep. But still less JavaScript to parse.

saschanaz:
I think parsing whitespace is lightning fast on every engine...

marcoscaceres:
Hhmm... @saschanaz does have a point. If it’s only white space, then it might not be worth adding the dependency.

sidvishnoi:
It's whitespace, and a little bit of comments - all inside a JavaScript string. Let me confirm if there is no improvement in gzip size. If not, I'll close this.

sidvishnoi:
``` diff
Size comparison (bytes):
- 294719
+ 286512
gzip:
- 92545
+ 91128
```

sidvishnoi:
(I probably tested same gzip file earlier)

sidvishnoi:
@saschanaz should I merge the functionality of `rollup-plugin-string` with this? One less dependency.

saschanaz:
> should I merge the functionality of rollup-plugin-string with this? One less dependency.

I have no idea how it works, but please do!

PR #2815, opened 2020-03-29
fix(tools/github-action-reporter): support multiline failure messages

sidvishnoi:
The failure message got cut in https://github.com/w3c/respec/pull/2717: 
| then | now |
| --- | --- |
| image | image |

### DIFF
--- a/tests/spec/core/examples-spec.js
+++ b/tests/spec/core/examples-spec.js
@@ -51,7 +51,7 @@ describe("Core — Examples", () => {
     expect(marker.querySelector(".example-title").textContent).toBe(": EX");
     expect(example.getAttribute("title")).toBeNull();
     expect(example.textContent.trim()).toBe(
-      "Example 1: EX\n  \n{\n  CONTENT\n}"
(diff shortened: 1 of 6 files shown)


tests/spec/core/examples-spec.js
```diff
@@ -51,7 +51,7 @@ describe("Core — Examples", () => {
     expect(marker.querySelector(".example-title").textContent).toBe(": EX");
     expect(example.getAttribute("title")).toBeNull();
     expect(example.textContent.trim()).toBe(
-      "Example 1: EX\n  \n{\n  CONTENT\n}"
+      "Example 1:  EX\n  \n{\n  CONTENT\n}"
```
saschanaz: Why does Chrome report twice? It doesn't seem to be a bug in this script.

sidvishnoi: Yeah. I am confused too. Probably something temporary. The one below is reported correctly.
We can fix later if we see it in real life again

PR #2816, opened 2020-03-29
fix(w3c/headers): indicate as recommended

saschanaz:
Somehow `return` is omitted 😲

### DIFF
--- a/src/w3c/templates/sotd.js
+++ b/src/w3c/templates/sotd.js
@@ -108,8 +108,8 @@ export default (conf, opts) => {
                   ? renderImplementationReportURI(conf)
                   : ""}
                 ${conf.sotdAfterWGinfo ? opts.additionalContent : ""}
-                ${conf.notRec ? renderNotRec(conf) : ""}
-                ${conf.isRec ? renderIsRec() : ""} ${renderDeliverer(conf)}
+                ${conf.isRec ? renderIsRec() : renderNotRec(conf)}
+                ${renderDeliverer(conf)}
                 <p>
                   This document is governed by the
                   <a
@@ -185,8 +185,7 @@ function renderImplementationReportURI(conf) {
   </p>`;
 }
 
-function renderNotRec(conf) {
-  const { anOrA, textStatus } = conf;
+function renderNotRec({ anOrA, textStatus }) {
   return html`<p>
     Publication as ${anOrA} ${textStatus} does not imply endorsement by the W3C
     Membership. This is a draft document and may be updated, replaced or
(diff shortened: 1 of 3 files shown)


tests/spec/w3c/headers-spec.js
```diff
@@ -44,6 +52,27 @@ describe("W3C — Headers", () => {
       expect(doc.querySelector(".head h2").textContent).toContain(
         "W3C Editor's Draft"
       );
+      expect(
+        normalizedIncludes(
```
saschanaz: `innerText` could be used instead but it requires the document to be visually rendered. The test currently disables rendering by `.display = "none"` and replacing this with `.visibility = "hidden" enables `innerText`, which do you prefer?

src/w3c/templates/sotd.js
```diff
@@ -196,7 +195,7 @@ function renderNotRec(conf) {
 }
 
 function renderIsRec() {
-  html`<p>
+  return html`<p>
```
sidvishnoi: 😱

tests/spec/w3c/headers-spec.js
sidvishnoi: Why would be need to use `innerText` instead?

saschanaz: With `innerText` this can be simpler: `expect(document.getElementById("sotd").innerText).toContain(...)`

tests/spec/w3c/headers-spec.js
```diff
@@ -44,6 +52,27 @@ describe("W3C — Headers", () => {
...
+        specStatus: "REC",
+      };
+      Object.assign(ops.config, newProps);
+      const doc = await makeRSDoc(ops);
+      expect(
+        normalizedIncludes(
+          doc.getElementById("sotd"),
+          "is endorsed by the Director as a W3C Recommendation."
+        )
+      ).toBe(true);
```
sidvishnoi: Can we use something like `expect(norm(el.textContent)).toContain(...)`?

tests/spec/w3c/headers-spec.js
sidvishnoi: `.visibility = "hidden"` seems like a good idea, even if we don't need `innerText` here (if https://github.com/w3c/respec/pull/2816#discussion_r399832452 works)

src/w3c/templates/sotd.js
```diff
@@ -108,8 +108,8 @@ export default (conf, opts) => {
                   ? renderImplementationReportURI(conf)
                   : ""}
                 ${conf.sotdAfterWGinfo ? opts.additionalContent : ""}
-                ${conf.notRec ? renderNotRec(conf) : ""}
-                ${conf.isRec ? renderIsRec() : ""} ${renderDeliverer(conf)}
+                ${conf.isRec ? renderIsRec() : renderNotRec(conf)}
+                ${renderDeliverer(conf)}
```
sidvishnoi: Nice cleanup 💯

tests/spec/w3c/headers-spec.js
saschanaz: Seems we don't need `innerText` anyway!

PR #2820, opened 2020-04-02
feat(core/dfn-panel): run in exported documents

sidvishnoi:
Alternative to https://github.com/w3c/respec/pull/2811
@saschanaz Did you mean something like this?


sidvishnoi:
@saschanaz ICYMI

saschanaz:
Bit busy with other issue, will review soon when it ends 🙏

sidvishnoi:
No problem! Take care

sidvishnoi:
Pros of this approach:
- clean
- can't forget to serialize some function
- live documents run same code as exported documents

Cons:
- Losing sourcemap support in live documents
- Extra build step - need to run terser etc. separately
- another file

saschanaz:
1. IMO `referencesToHTML()` part should be precomputed and inserted as a JSON dictionary so that the runtime code can focus on UI part.
2. Do we really need to run terser in this short file? How many bytes we save when gzipped?

>another file

This should be good though. A module with multiple purposes (getting references and showing UI) is not ideal.

sidvishnoi:
> IMO referencesToHTML() part should be precomputed and inserted as a JSON dictionary so that the runtime code can focus on UI part.

That would be a lot of unused data injected into exported doc. Although Bikeshed does this (creates an `<aside>` for each dfn, I prefer if we do it at runtime.

> Do we really need to run terser in this short file? How many bytes we save when gzipped?

```
5001 normal.js
1963 normal.js.gz 

2154 terser.js   
1106 terser.js.gz 
```

gkellogg:
IIRC, the JSON-LD script element optionally added for SEO contains these definitions (or certainly could). Perhaps it could serve dual purpose.

src/core/dfn-panel.runtime.js
```diff
@@ -0,0 +1,168 @@
...
+  return `<ul>${listItems.join("")}</ul>`;
+}
+
+/** @param {HTMLAnchorElement} link */
+function getReferenceTitle(link) {
+  const section = link.closest("section");
+  if (!section) return null;
+  const heading = section.querySelector("h1, h2, h3, h4, h5, h6");
+  if (!heading) return null;
+  return norm(heading.textContent);
```
saschanaz, line 124: I think we can use `.innerText.trim()` here which does the normalization for us. Or we can just inline it as no other line uses it.

saschanaz:
>That would be a lot of unused data injected into exported doc. Although Bikeshed does this (creates an <aside> for each dfn, I prefer if we do it at runtime.

No idea what Bikeshed exactly does but what if we only extract the link IDs? Those shouldn't be a lot of bytes and we can do the remaining UI thing on runtime.

sidvishnoi:
Comparing the amount of code in `referencesToHTML` vs what we would add by pre-computation, I think it's better to do it in runtime.

sidvishnoi:
The injected could should probably be wrapped as an IIFE.

saschanaz:
>Comparing the amount of code in referencesToHTML vs what we would add by pre-computation, I think it's better to do it in runtime.

I think it also makes debugging easier as we get the whole linking structure in text form, and also potentially SEO as @gkellogg mentioned (although I'm not an expert on that part.)

sidvishnoi:
@gkellogg Can you expand on how we can add definitions and their references via JSON-LD?

sidvishnoi:
> I still hope we precompute the data but I guess it can be done in a separate patch.

I gave it some thought today, and I think we should do this before this patch. Will be sending one soon..

saschanaz:
If we ever want to do JSON-LD thing here, this might be a clue: https://schema.org/DefinedTermSet

src/core/dfn-panel.runtime.js
```diff
@@ -0,0 +1,172 @@
...
+  const { id } = dfn;
+  const href = dfn.dataset.href || `#${id}`;
+  /** @type {NodeListOf<HTMLAnchorElement>} */
+  const links = document.querySelectorAll(`a[href="${href}"]:not(.index-term)`);
+
+  const tempNode = document.createElement("div");
+  tempNode.innerHTML = `
+    <aside class="dfn-panel" id="dfn-panel">
+      <span class="caret"></span>
+      <b><a class="self-link" href="${href}">Permalink</a></b>
```
marcoscaceres, line 69: Nit: do we really need `<b>` ?

src/core/dfn-panel.runtime.js
```diff
@@ -0,0 +1,172 @@
...
+/** @param {HTMLElement} dfn */
+function createPanel(dfn) {
+  const { id } = dfn;
+  const href = dfn.dataset.href || `#${id}`;
+  /** @type {NodeListOf<HTMLAnchorElement>} */
+  const links = document.querySelectorAll(`a[href="${href}"]:not(.index-term)`);
+
+  const tempNode = document.createElement("div");
+  tempNode.innerHTML = `
+    <aside class="dfn-panel" id="dfn-panel">
```
marcoscaceres, line 67: Bonus points if you add in "(Exported)" somewhere ;)

marcoscaceres: (no problem if this is a followup)

src/core/dfn-panel.runtime.js
```diff
@@ -0,0 +1,172 @@
+// @ts-check
+if (document.respecIsReady) {
+  document.respecIsReady.then(() => dfnPanel());
```
marcoscaceres, line 3:
```suggestion
  document.respecIsReady.then(dfnPanel);
```

sidvishnoi:
Closing in favor of https://github.com/w3c/respec/pull/2811

src/core/dfn-panel.runtime.js
sidvishnoi: Looked nice IMO. Also, we're using it as `display: block` :sweat_smile:

src/core/dfn-panel.runtime.js
sidvishnoi: Filed https://github.com/w3c/respec/issues/2935

PR #2826, opened 2020-04-09
refactor(core/data-cite): do not require conf in citeDetailsConverer

sidvishnoi:
We can avoid passing conf to citeDetailsConverter, and have a unique instance which we can reuse. Simplifies code a bit.

Note to reviewer: Review commit by commit and ignore whitespace changes.

### DIFF
--- a/src/core/data-cite.js
+++ b/src/core/data-cite.js
@@ -26,9 +26,6 @@ export const name = "core/data-cite";
 
 const THIS_SPEC = "__SPEC__";
 
-/**
- * @param {HTMLElement} elem
- */
 async function toLookupRequest(elem) {
   const originalKey = elem.dataset.cite;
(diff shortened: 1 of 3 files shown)


src/core/data-cite.js
```diff
@@ -24,78 +24,80 @@ import {
...
+  const originalKey = elem.dataset.cite;
+  const { key, frag, path } = toCiteDetails(elem);
+  let href = "";
+  let title = "";
+  // This is just referring to this document
+  if (key === THIS_SPEC) {
+    console.log(
+      elem,
+      `The reference "${key}" is resolved into the current document per \`conf.shortName\`.`
+    );
```
sidvishnoi, line 42: We probably don't need this message.

marcoscaceres: Yeah, can go

Issue #2830, opened 2020-04-12
Refactor linker modules


sidvishnoi:
Presently, the linker modules, mainly `core/data-cite` and `core/link-to-dfn` are doing a lot of related/unrelated work. Would be helpful to simplify the linking logic. We should run them as 3 modules in following order: 
1. `core/link-to-dfn`: links to local definitions and marks everything else external
2. `core/xref`: links external terms
3. `core/data-cite`: converts data-cite attributes created by above two modules (and user generated ones) to href values - i.e., `linkInlineCitations()` becomes `run()` for core/data-cite

sidvishnoi:
The task has completed already with linked PRs.

PR #2831, opened 2020-04-13
refactor(core/data-cite): run data-cite to href converter as module

sidvishnoi:
This one simplifies `core/data-cite` by requiring it to only convert `data-cite` attributes to `href` attributes (and related markup changes)


src/core/link-to-dfn.js
```diff
@@ -87,11 +92,6 @@ export async function run(conf) {
   } else {
     showLinkingError(possibleExternalLinks);
   }
-
-  await linkInlineCitations();
```
sidvishnoi, line 91: This becomes `core/data-cite` module, which runs after `core/link-to-dfn`.

src/core/link-to-dfn.js
```diff
@@ -76,6 +76,11 @@ export async function run(conf) {
 
   showLinkingError(badLinks);
 
+  // This needs to run before core/xref adds its data-cite and updates
+  // conf.normativeReferences and conf.informativeReferences.
+  await dataCite(conf);
```
sidvishnoi, line 81: Failing to run this causes one test to fail in which an external definition (defined non-normatively) moves to normative references section.

profiles/w3c.js
```diff
@@ -33,10 +33,10 @@ const modules = [
   import("../src/core/best-practices.js"),
   import("../src/core/figures.js"),
   import("../src/core/webidl.js"),
-  import("../src/core/data-cite.js"),
   import("../src/core/biblio.js"),
-  import("../src/core/webidl-index.js"),
   import("../src/core/link-to-dfn.js"),
+  import("../src/core/data-cite.js"),
```
sidvishnoi, line 38: In near future, we'll have this:
```suggestion
  import("../src/core/link-to-dfn.js"),
  import("../src/core/xref.js"),
  import("../src/core/data-cite.js"),
```

src/core/link-to-dfn.js
sidvishnoi: Need a better name for this function. Plus, move its definition to core/link-to-dfn

src/core/data-cite.js
```diff
@@ -2,29 +2,28 @@
...
   showInlineWarning,
   wrapInner,
 } from "./utils.js";
-import { sub } from "./pubsubhub.js";
 export const name = "core/data-cite";
 
-const THIS_SPEC = "__SPEC__";
+/**
+ * An arbitrary constant value used as an alias to current spec's shortname. It
+ * exists to simplify code as passing `conf.shortName` everywhere gets clumsy.
```
saschanaz, line 24: How about migrating to a class as #2752 did?

sidvishnoi: It's not just this module that makes use of `toCiteDetails`, which is why we introduced `THIS_SPEC` in https://github.com/w3c/respec/pull/2826
Migrating to class like in #2752 is helpful when accessing `conf` within same module.

saschanaz: >Migrating to class like in #2752 is helpful when accessing conf within same module.

Well no, it's helpful when needing to pass `conf` again and again and that `conf` can't be kept in the same module. Accessing `conf` within same module should be replaced by class property, not `conf`.

sidvishnoi: Sorry, not following. `THIS_SPEC`  (or earlier `conf.shortName`) is being used in `core/data-cite@toCiteDetails` as well as `core/link-to-dfn@updateReferences`. How would we access this common constant with the class approach?

saschanaz: Each instance can keep `conf` as a class property value so that its methods can access it without passing it everywhere just to access `conf.shortName`.

sidvishnoi: How do we access `toCiteDetails` outside `core/data-cite` when we've moved it into a Plugin class?

saschanaz: In that case we can keep a module function version of `toCiteDetails` and then a new wrapper method with the same name to pass `conf` to the function.

That said, I didn't see this is just an existing code. No need to change it in this PR IMO.

saschanaz: Well not totally, it's introduced yesterday? 😮

src/core/link-to-dfn.js
```diff
@@ -274,3 +274,34 @@ function showLinkingError(elems) {
...
 }
+
+/**
+ * Update references due to `data-cite` attributes.
+ *
+ * Also, make sure self-citing doesn't cause current document getting added to
+ * bibliographic references section.
+ * @param {Conf} conf
+ */
+function updateReferences(conf) {
```
saschanaz, line 285: Much better readability!

src/core/data-cite.js
sidvishnoi: Yes. Should've asked your feedback there

PR #2832, opened 2020-04-13
refactor(core/xref): run as module

sidvishnoi:


src/core/link-to-dfn.js
```diff
@@ -12,11 +12,13 @@ import {
   wrapInner,
 } from "./utils.js";
 import { THIS_SPEC, toCiteDetails } from "./data-cite.js";
-import { run as addExternalReferences } from "./xref.js";
 import { definitionMap } from "./dfn-map.js";
 
 export const name = "core/link-to-dfn";
 
+/** @type {HTMLElement[]} */
+export const possibleExternalLinks = [];
```
saschanaz, line 20: I really don't want more module level state variables 1) as they are major blockers against #2187 as those variables only initialize once and 2) it makes harder to track how modules depends on others. Modules should ideally only export functions and constants, not variables.

I guess 2) is hard to solve in current code structure though, but maybe we can create a new object to save states and pass it instead of current `conf` to make modules more modularized:

```
const respecDoc = {
  configuration, // current conf
  possibleExternalLinks,
  /* ... other states that can change during module runs ... */
};
```

sidvishnoi: I agree with your arguments and understand the pain with #2187.

But having a common mutable state might also be problematic - who modified the state? Having module level exports helps here a bit as we know who imported it and as well as we limit the scope of shared state by exporting a single variable. Shared common state is similar to storing state in DOM.

sidvishnoi: GitHub app cleared my comment twice and also resolved the conversation once. 🙃

saschanaz: >who modified the state?

TypeScript knows who did it 😁

saschanaz: Another argument: storing states in a single object will help us know how many states we have. That will help potential refactoring.

sidvishnoi: > >who modified the state?
> 
> TypeScript knows who did it 😁

Nested object? 

> Storing in DOM is okay as each run should always get a newly initialized DOM root, while storing in module level will be accumulated for each run in #2187. 

It'll get slow though. Might be fine in this case - add data-xref* to possibleExternalLinks

Agree with knowing about state of state. But a lot of it is intermediate state, like between only two modules - rest of codebase need not know of its existence.

Talk tomorrow, sleeping time. I'll use desktop website. 😁

saschanaz: >Objects also? 😮

We can always type it and then find references for each field.

>It'll get slow though. Might be fine in this case - add data-xref* to possibleExternalLinks

I mean it's okay for modularization sake, but not preferable as it will needlessly make DOM dirty.

>But a lot of it is intermediate state, like between only two modules - rest of codebase need not know of its existence.

IMO that's okay, no need to use everything that can be used.

sidvishnoi: Should we begin with the shared object state with this PR, or let this in and then work on that?

saschanaz: Maybe a separate PR. I'll file a separate issue for this 👀

PR #2836, opened 2020-04-14
refactor(core/data-cite): simplify linking

sidvishnoi:
`findLinkTarget` was misleading as it mutated its `anchor` parameter.
Although `useLinkTarget` isn't a great name either - it's very much overloaded. Suggestions welcome.

### DIFF
--- a/src/core/link-to-dfn.js
+++ b/src/core/link-to-dfn.js
@@ -68,7 +68,10 @@ export async function run(conf) {
         titleToDfns.get(target.title).has(target.for)
     );
     if (linkTarget) {
-      useLinkTarget(linkTarget, anchor, titleToDfns);
+      const hasExternalTarget = processAnchor(anchor, linkTarget, titleToDfns);
+      if (hasExternalTarget) {
+        possibleExternalLinks.push(anchor);
+      }
     } else {
       if (anchor.dataset.cite === "") {
         badLinks.push(anchor);
@@ -132,25 +135,26 @@ function collectDfns(title) {
 }
 
 /**
- * @param {import("./utils.js").LinkTarget} target
  * @param {HTMLAnchorElement} anchor
+ * @param {import("./utils.js").LinkTarget} target
  * @param {CaseInsensitiveMap} titleToDfns
  */
-function useLinkTarget(target, anchor, titleToDfns) {
+function processAnchor(anchor, target, titleToDfns) {
+  let hasExternalTarget = false;
   const { linkFor } = anchor.dataset;
   const dfn = titleToDfns.get(target.title).get(target.for);
   if (dfn.dataset.cite) {
     anchor.dataset.cite = dfn.dataset.cite;
   } else if (linkFor && !titleToDfns.get(linkFor)) {
-    possibleExternalLinks.push(anchor);
+    hasExternalTarget = true;
   } else if (dfn.classList.contains("externalDFN")) {
     // data-lt[0] serves as unique id for the dfn which this element references
     const lt = dfn.dataset.lt ? dfn.dataset.lt.split("|") : [];
     anchor.dataset.lt = lt[0] || dfn.textContent;
-    possibleExternalLinks.push(anchor);
+    hasExternalTarget = true;
   } else {
     if (anchor.dataset.idl === "partial") {
-      possibleExternalLinks.push(anchor);
+      hasExternalTarget = true;
     } else {
       anchor.href = `#${dfn.id}`;
       anchor.classList.add("internalDFN");
@@ -162,6 +166,7 @@ function useLinkTarget(target, anchor, titleToDfns) {
   if (isCode(dfn)) {
     wrapAsCode(anchor, dfn);
   }
+  return hasExternalTarget;
 }
 
 /**
--- a/src/core/link-to-dfn.js
+++ b/src/core/link-to-dfn.js
@@ -56,11 +56,11 @@ export async function run(conf) {
   /** @type {HTMLAnchorElement[]} */
   const badLinks = [];
 
-  const localLinkSelector =
-    "a[data-cite=''], a:not([href]):not([data-cite]):not(.logo):not(.externalDFN)";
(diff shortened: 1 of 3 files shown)


src/core/link-to-dfn.js
```diff
@@ -56,23 +56,27 @@ export async function run(conf) {
...
+  for (const anchor of localAnchors) {
     const linkTargets = getLinkTargets(anchor);
-    const foundDfn = linkTargets.some(target => {
-      return findLinkTarget(target, anchor, titleToDfns);
-    });
-    if (!foundDfn && linkTargets.length !== 0) {
+    const linkTarget = linkTargets.find(
+      target =>
+        titleToDfns.has(target.title) &&
+        titleToDfns.get(target.title).has(target.for)
```
sidvishnoi, line 68: Patiently waiting for optional chaining...

src/core/link-to-dfn.js
```diff
@@ -56,23 +56,27 @@ export async function run(conf) {
...
+    );
+    if (linkTarget) {
+      useLinkTarget(linkTarget, anchor, titleToDfns);
+    } else if (linkTargets.length !== 0) {
```
saschanaz, line 72: How about:

```suggestion
    } else if (!linkTargets.length) {
```

I think this pattern is used frequent enough.

src/core/link-to-dfn.js
```diff
@@ -132,14 +136,8 @@ function collectDfns(title) {
  * @param {HTMLAnchorElement} anchor
  * @param {CaseInsensitiveMap} titleToDfns
  */
-function findLinkTarget(target, anchor, titleToDfns) {
+function useLinkTarget(target, anchor, titleToDfns) {
```
saschanaz, line 139: `mutateLinkTarget` if its sole purpose is mutation?

src/core/link-to-dfn.js
sidvishnoi: Just found out `else {` works also - unless we're missing some tests

src/core/link-to-dfn.js
sidvishnoi: It also mutates `possibleExternalLinks`

saschanaz: AFAICT this function doesn't touch `possibleExternalLinks`, though?

saschanaz: Ah okay, GitHub skipped some lines. Well, in that case, maybe `process`? It's also way too general but better than `use` as the latter doesn't imply it'll be mutated.

saschanaz: Or maybe refactor this function further to just return the anchor is a possible external link or not.

sidvishnoi: This refactor is turning out to be more difficult then I expected..

saschanaz: Maybe just define a local boolean variable and return it at the last line?

sidvishnoi: I was trying to:
```js
if (hasLocalTarget(anchor, dfn, titleToDfns)) {
  processAnchor(anchor, dfn);
} else {
  possibleExternalLinks.push(anchor);
}
```

Didn't work out fine. Went with your suggestion in https://github.com/w3c/respec/pull/2836/commits/505c12eee8aeeb59a180c08e5576d71362cf516f.

saschanaz: That's a great refactoring if possible, sad that it's harder than expected 😅

Somehow I prefer `mutate` for the name as now that's the main purpose of the function.

sidvishnoi: This turned out to be difficult as the order of conditions mattered here. We would be essentially repeating all conditionals to check if it has local target, then use same conditionals to mutate the anchor.
Is the present one acceptable?

src/core/link-to-dfn.js
```diff
@@ -56,23 +56,30 @@ export async function run(conf) {
...
-      return findLinkTarget(target, anchor, titleToDfns);
-    });
-    if (!foundDfn && linkTargets.length !== 0) {
+    const linkTarget = linkTargets.find(
+      target =>
+        titleToDfns.has(target.title) &&
+        titleToDfns.get(target.title).has(target.for)
+    );
+    if (linkTarget) {
+      const localMatchFound = processAnchor(anchor, linkTarget, titleToDfns);
```
saschanaz, line 71: Personally prefer `foundLocalMatch` or `hasLocalMatch` which is a bit shorter.

src/core/link-to-dfn.js
```diff
@@ -128,43 +136,35 @@ function collectDfns(title) {
...
-  ) {
-    return false;
-  }
   const dfn = titleToDfns.get(target.title).get(target.for);
   if (dfn.dataset.cite) {
     anchor.dataset.cite = dfn.dataset.cite;
+    localMatchFound = true;
   } else if (linkFor && !titleToDfns.get(linkFor)) {
-    possibleExternalLinks.push(anchor);
+    localMatchFound = false;
```
saschanaz, line 151: This is weird. How about `noLocalMatch` and set it true if not found, and then return `!noLocalMatch`?

src/core/link-to-dfn.js
saschanaz: >Is the present one acceptable?

Yup 👍

src/core/link-to-dfn.js
sidvishnoi: > This is weird.

Yes! It's meant to be a noop here, but if we skip this condition, tests fail :tada: Decided setting it to false was better than a comment? Maybe comment was better option.

`noLocalMatch` is a lot of negations IMO.

saschanaz: >noLocalMatch is a lot of negations IMO.

I think it's no problem if it makes cleaner code and not causes confusion.

sidvishnoi:
Thanks for making sure we only let the good code in! :superhero:

Issue #2152, opened 2019-03-13
Add Turtle format language module


BigBlueHat:
Follow-up from #1063

## Description of problem

Many specs at the W3C (and elsewhere) use the [Turtle](https://www.w3.org/TR/turtle/) format.

There is **[a PR to highlight.js for adding Turtle](https://github.com/highlightjs/highlight.js/pull/1553/commits)** that has apparently been available since shortly after I filed #1063.

The JSON-LD WG is nearing CR for the [JSON-LD 1.1](https://w3c.github.io/json-ld-syntax/) syntax document, and we use a good bit of Turtle, so having this added would help us there particularly.

Also, credit goes to @cygri for finding this highlighter! https://github.com/w3c/json-ld-syntax/pull/67#issuecomment-472408686

Thanks!
🎩 

P.S.: [The PR](https://github.com/highlightjs/highlight.js/pull/1553/commits) also includes a highlighter for SPARQL which W3C specs and notes also use from time to time.

cygri:
There was some further work on top of that PR by @VladimirAlexiev, which led to this branch:

https://github.com/VladimirAlexiev/highlight.js/tree/feature/language-sparql-ttl

I've tried this for SPARQL and Turtle, and it works well. I had to follow the [build instructions](https://highlightjs.readthedocs.io/en/latest/building-testing.html) to create a `highlight.pack.js` from source.

It looks like maintainership of highlight.js is somewhat adrift at the moment (highlightjs/highlight.js#1678), so PRs are not being merged.

marcoscaceres:
We are currently moving await from using the hljs bundle. So, the way to do this is to use `preProcess` to load the highlighter language:

```JS
      async function loadSparqlLang() {
        const action = "highlight-load-lang";
        const langURL = new URL(/*URL to your language file*/).href;
        // this needs to match the name of the property created in highlight js for the language
        const propName = "sparql-ttl";
        const lang = "sparql-ttl"; // check what this should be 
        worker.postMessage({ action, langURL, propName, lang });
        return new Promise(resolve => {
          worker.addEventListener("message", function listener({ data }) {
            const { action: responseAction, lang: responseLang } = data;
            if (responseAction === action && responseLang === lang) {
              worker.removeEventListener("message", listener);
              resolve();
            }
          });
        });
      }
      var respecConfig = {
        preProcess: [loadSparqlLang],
        // other stuff
      }
```

[...]

PR #2838, opened 2020-04-15
chore(profiles/w3c-common): freeze profile to v25.5.0

sidvishnoi:
🚨 We would no longer be updating `w3c-common` profile. 🚨
The profile will continue to be hosted at https://www.w3.org/Tools/respec/respec-w3c-common

---

## Why?
It exposes some of ReSpec's internals, so we can't safely make changes to our internal structure and move forward. [A lot of legacy specs](https://github.com/search?p=4&q=org%3Aw3c+pubsubhub&type=Code) use those internals, so we can't just remove them.
The most recent incident which triggered this decision is the breaking of https://w3c.github.io/json-ld-syntax/, which [makes use of](https://github.com/w3c/json-ld-syntax/blob/a65dd9acf04bce870392afd2c074c44bf86f9607/common/common.js#L60-L61) local-linking portion of core/link-to-dfn. [json-ld-syntax] broke as it wanted to hook between end of local-linking and start of external-linking - which we [do not expose anymore](https://github.com/w3c/respec/issues/2830).

## How can we extend ReSpec now?
Presently, there is no way to do it. Not breaking specs is our priority. We do have plans to provide extensibility in future.
If you've a use case that other specs can also use, consider opening an issue/sending a PR so everyone can benefit from it!

## Want new features and bug fixes?
Use the `respec-w3c` profile. It is nearly 90KB (30KB gzip) smaller than `respec-w3c-common`.
Most of specs need to simply replace:
https://www.w3.org/Tools/respec/respec-w3c-common with
https://www.w3.org/Tools/respec/respec-w3c

TODO:
- [ ] Remove all content from modern profiles that references legacy code, like `expose-module`, `pub("end, *)"`. Blocked by https://github.com/w3c/respec/issues/2152


sidvishnoi:
Maybe we should revisit https://github.com/w3c/respec/pull/2702

PR #2840, opened 2020-04-15
fix(tools/respecDocWriter): prevent hang on never resolving promise

sidvishnoi:
In https://github.com/w3c/spec-generator/issues/292, `document.respecIsReady` never resolved.

Tried it with spec-generator locally, and the error response was like:
```
Sorry, there was an error generating the HTML. Please report this issue!
...
Error: Error: Evaluation failed: Timeout: document.respecIsReady didn't resolve in 22792ms.
[stack trace]
```

Without this patch, there was no response and browser never closed.

As an addon, it would be nice if we can show the ReSpec error/warnings somehow. console.log on server doesn't tell anything to end user.


saschanaz:
Should we let `respecIsReady` itself timeout?

sidvishnoi:
> Should we let respecIsReady itself timeout?

That would mean hardcoding the timeout? Or a new config option (this fails for https://github.com/w3c/spec-generator/issues/292)?

saschanaz:
30sec seems a reasonable hardcoded default timeout, maybe we can also set a config later if needed.

sidvishnoi:
Not sure if we should do this, but got curious how we may do it :sweat_smile: 
``` js
const respecDonePromise = new Promise((resolve, reject) => {
  let shouldReject = true;
  const onEnd = respecConfig => {
    shouldReject = false;
    resolve(respecConfig);
  };
  setTimeout(() => {
    if (shouldReject) {
      unsub({ topic: "end-all", cb: onEnd });
      reject(new Error("Timeout"));
    }
  }, TIMEOUT);
  sub("end-all", onEnd, { once: true });
});
```

sidvishnoi:
There is one difference here though. In this PR, we wait `timeout - alreadySpent` time for document.respecIsReady. `timeout` implies total time - fetching document, verifying and running ReSpec and generating HTML. Here, it's just `timeout` time for respecIsReady.

saschanaz:
I was thinking about a promise that rejects if not resolved before due.

```js
function timeout(promise, ms) {
    promise.then(resolve, reject);
    setTimeout(reject, ms);
  });
};
```

saschanaz:
>In this PR, we wait timeout - alreadySpent time for document.respecIsReady.

IMO that's overkill, the purpose here is not to timeout with precision. It's okay if it doesn't run forever. How about just:

```js
await timeout(document.respecIsReady, 30000);
```

... with the `timeout` defined above?

sidvishnoi:
> the purpose here is not to timeout with precision.

That kinda breaks the promise to timeout in 30s though.

saschanaz:
This might just skip passing the timer:

```js
await timeout(page.evaluate(...));
```

Might be better to prevent potentially being stuck after `respecIsReady`.

saschanaz:
Oh wait, why not just:

```js
try {
  await timeout(fetchAndWriteInternal(browser, ...), ms);
} finally {
  browser.close();
}
```

... and extract the code inside `try` as a separate function?

sidvishnoi:
I prefer adding timeout on internal promises, so we know which promise timed out. Adding it on top-level hides where the actual was.

saschanaz:
Hmmm okay fine. Still prefer `timeout()` than the current inline effort though

sidvishnoi:
Releasing v25.6.1

PR #2844, opened 2020-04-16
tests(core/dfn-index): test existence of each external term

sidvishnoi:
Faced some regressions while working on https://github.com/w3c/respec/pull/2508

In particular, with following IDL:
``` webidl
[Exposed=Window]
interface Employee {
  [Default] object toJSON();
}
```
if we do not define `toJSON()`, it just links to itself. With #2508, we're able to link it to WebIDL spec thus adding another external term to list.


saschanaz:
> if we do not define `toJSON()`, it just links to itself.

Not sure I get it, linking to itself sounds like a bug?

sidvishnoi:
Yes. That sounds like a bug, and apparently, it is getting fixed in #2508
Problem is associated with `data-link-for` attribute.
``` html
<pre class="idl" data-link-for="Employee">
  [Exposed=Window]
  interface Employee {
    [Default] object toJSON();
  };
</pre>
```

saschanaz:
Can the fix be separated from there, as the PR is "feat" not "fix"?

sidvishnoi:
I'll try to figure out exact cause of that bug. #2508 is a bit convulated right now. Will investigate and send a fix.

Issue #2845, opened 2020-04-17
Unable to publish because of CSS validator issues


marcoscaceres:
Echidna is upset :( 

```
✗ Echidna failed to publish https://w3c.github.io/screen-orientation/W3CTRMANIFEST
 for the following reasons:

   The file has been retrieved.
   The metadata have been extracted.
   The document failed Specberus.
 👎 Unknown pseudo-element or pseudo-class “:absolute”
     ⮡ cf https://www.w3.org/pubrules/doc/rules/?profile=WD#cssValide
 👎 Unknown pseudo-element or pseudo-class “:var”
     ⮡ cf https://www.w3.org/pubrules/doc/rules/?profile=WD#cssValide
 👎 Unknown pseudo-element or pseudo-class “:var”
     ⮡ cf https://www.w3.org/pubrules/doc/rules/?profile=WD#cssValide
 👎 Unknown pseudo-element or pseudo-class “:var”
     ⮡ cf https://www.w3.org/pubrules/doc/rules/?profile=WD#cssValide

More details are available here:

 https://labs.w3.org/echidna/api/status?id=a38b65b6-b664-4b0c-9fbd-d77d6d6c2754

If you think you found a bug, please fill in and submit this issue:

 https://github.com/w3c/echidna/issues/new?title=Bug:%20%5BWHAT%5D&body=%5BEXPLANATION%5D%0A%0AFound%20while%20trying%20to%20publish%20%60https%3A%2F%2Fw3c.github.io%2Fscreen-orientation%2FW3CTRMANIFEST%60%20(see%20the%20%5Bpublication%20job%5D(https%3A%2F%2Flabs.w3.org%2Fechidna%2Fapi%2Fstatus%3Fid%3Da38b65b6-b664-4b0c-9fbd-d77d6d6c2754)).&labels=from-template

~ 4/17/2020, 1:47:39 AM
````

sidvishnoi:
CSS Custom properties are not yet supported with CSS validator https://github.com/w3c/css-validator/pull/173 :disappointed:
I'll send a PR to remove our use of custom properties until it is supported. Would need to incorporate it in our frozen w3c-common profile also. :disappointed:

PR #2846, opened 2020-04-17
fix(assets/dfn-panel.css): remove on save

sidvishnoi:
Fixes https://github.com/w3c/respec/issues/2845

---
<details><summary>Outdated</summary>
**Revert this PR as soon as validator is fixed.**

w3c/css-validator doesn't support custom properties yet (https://github.com/w3c/css-validator/issues/111), so echidna failed to publish.
Setting position of panel via `panel.style.left` is straightforward, but setting `left` for `.dfn-panel::before` (the caret/triangle), JS won't allow. We would have to change the markup otherwise ~- which is something I'm not sure I want to do~.

~God forgive me, for this hack :pray:~
Update: Hack removed, changed markup instead.
</summary>
<details?


sidvishnoi:
On a second thought, it might not be a bad idea to change the markup instead, :thinking:

src/core/dfn-panel.js
```diff
@@ -10,7 +10,7 @@ export const name = "core/dfn-panel";
 export async function run() {
   const css = await loadStyle();
   document.head.insertBefore(
-    hyperHTML`<style>${css}</style>`,
+    hyperHTML`<style class="removeOnSave">${css}</style>`,
```
saschanaz, line 13: We want to enable dfn panel on saved documents, no?

sidvishnoi: yes, we want to. but not yet. gotta fix echidna breakage first. will add it back with the fix in https://github.com/w3c/respec/pull/2820

sidvishnoi: Did you get a change to see other two attempts?
https://github.com/w3c/respec/pull/2846/commits/86e8f5a88d0e7be7b00f192c610516a54d9f4e6f is good if we want a workaround on css-validator, but way too hacky.
https://github.com/w3c/respec/pull/2846/commits/fcc0c9a662e7e1a1b0181e2d287c017332c3b298 is good practice, but I don't like adding markup for decorative changes.

sidvishnoi:
> No builds/ update here wink

Will need to send another PR for builds/ then. Because release script no longer generates w3c-common. What do you prefer - here or a new PR?

saschanaz:
>Because release script no longer generates w3c-common. What do you prefer - here or a new PR?

Updating w3c-common here sounds fine.

Issue #2848, opened 2020-04-17
ARIA specs broken - definitionMap not defined


jnurthen:
## Important info

* URL to affected spec: https://w3c.github.io/aria/
* ReSpec version: 25.6.0
* [x] I did a "hard refresh", but it's still busted. 

## Description of problem

ARIA (and a number of specs) have shared terms. We clean these up by removing any which are not referenced in https://github.com/w3c/aria-common/blob/master/script/resolveReferences.js which includes the following where it attempts to remove references from respec

                    $p.getDfnTitles().forEach(function( item ) {
                        if (respecConfig.definitionMap[item]) {
                            delete respecConfig.definitionMap[item];
                        }
                    });

This no longer works

This worked when we published in December so my guess is that this is caused by https://github.com/w3c/respec/pull/2682

Can you help us get this working again. I don't mind making minor code changes to get it working.

jnurthen:
https://github.com/w3c/aria/issues/1244

sidvishnoi:
That's indeed a regression. We [recently stopped](https://github.com/w3c/respec/pull/2838) supporting the w3c-common profile, but I'll get the fix for this bug in.

Alternatively, if you could explain what exactly resolveReferences.js does, we might be able to integrate it in ReSpec. I've seen [similar](https://github.com/w3c/respec/issues/2767#issuecomment-590022201) thing being used in json-ld specs if I'm not wrong.

@saschanaz I'm thinking of creating a side-channel for updates to w3c-common profile. We won't actively support it, but may add fixes for severe regressions. My plan is to branch off v25.5.0, build manually when needed (should be rare) and copy builds to /builds folder when we release.

saschanaz:
Sounds good, we should still maintain it even when we don't add features.

marcoscaceres:
Thanks for reporting this, @jnurthen. The `w3c-common` profile that exposed `definitionMap` as a global was dropped in 2020 (PR #2838), and the module was refactored into `src/core/dfn-map.js`. Since then, the ARIA specs have moved to their own build process with `aria-common/script/resolveReferences.js`, which no longer depends on ReSpec's internals.

As the `legacy` label indicates, this was specific to the deprecated `w3c-common` profile. Closing as resolved by the ARIA team's migration.

PR #2849, opened 2020-04-17
fix(core/link-to-dfn): add definitionMap back to conf

sidvishnoi:
Closes https://github.com/w3c/respec/issues/2848

### DIFF
--- a/src/core/link-to-dfn.js
+++ b/src/core/link-to-dfn.js
@@ -89,6 +89,14 @@ export async function run(conf) {
   }
 
   await linkInlineCitations(document, conf);
+
+  // It's here for legacy support of aria and json-ld specs.
+  // See https://github.com/w3c/respec/issues/2019
+  // See https://github.com/w3c/respec/issues/2848
+  Object.assign(conf, {
+    definitionMap: { ...Object.fromEntries(definitionMap) },
+  });
+
   // Added message for legacy compat with Aria specs
   // See https://github.com/w3c/respec/issues/793
   pub("end", "core/link-to-dfn");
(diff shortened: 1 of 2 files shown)


sidvishnoi:
@jnurthen Can you validate the behavior by 
replacing https://www.w3.org/Tools/respec/respec-w3c-common
with https://deploy-preview-2849--respec-pr.netlify.app/respec-w3c-common.js?

It won't throw error, but please verify links work as before.

jnurthen:
It runs now but it isn't removing the elements from definitionMap. I think this is because it used to be case sensitive but no longer is and is all lower-case. Is this correct? Happy to modify our code for this change.

sidvishnoi:
> I think this is because it used to be case sensitive but no longer is and is all lower-case. Is this correct?

Yes. It's all lowercase now. Hopefully doesn't break your code much. Let me know how it goes.

jnurthen:
So this lets me update definitionMap but the terms are still appearing in the respec List of Definitions and any normative or infomormative references which the definition pulls into the document is still appearing in the references list.

sidvishnoi:
Can you try using https://unpkg.com/respec@24.38.1/builds/respec-w3c-common.js and share what terms should not appear?
I tried to use v24.38.1 (this is what https://www.w3.org/TR/wai-aria-1.2/ used) on https://github.com/w3c/aria/commit/f519fca855c11004f35b22ed127077414a3af0ea (as this was the last commit before December update), but I found differences between the TR version and my local one even with same ReSpec version.

sidvishnoi:
Closing as w3c-common profile is no longer maintained. Read [migration guide](https://github.com/w3c/respec/wiki/respec-w3c-common-migration-guide) in wiki.

Issue #2860, opened 2020-05-11
MDN configuration should allow defining different key


marcoscaceres:
The mdn config option currently takes a single argument `true`, and derives which data to pull down from MDN based on the `shortName`... however, there are cases where the `data` is not at the `shortName` (e.g., `netinfo` on MDN, while the spec's short name was `netinfo-api`). 

It would be great if `mdn` allowed string override (e.g., `mdn: "netinfo"`), in addition to `true`... and then true can just use the short name, otherwise, use the provided string override.

PR #2864, opened 2020-05-12
feat(core/mdn-annotation): allow providing custom key

sidvishnoi:
Closes https://github.com/w3c/respec/issues/2860

### DIFF
--- a/tests/spec/core/mdn-annotation-spec.js
+++ b/tests/spec/core/mdn-annotation-spec.js
@@ -2,15 +2,44 @@ import { flushIframes, makeRSDoc, makeStandardOps } from "../SpecHelper.js";
 
 describe("Core - MDN Annotation", () => {
   afterAll(flushIframes);
+
+  const baseJsonPath = `${window.location.origin}/tests/data/mdn-annotation/`;
   let doc;
   beforeAll(async () => {
     const ops = makeStandardOps({
-      mdn: {
-        baseJsonPath: `${window.location.origin}/tests/data/mdn-annotation/`,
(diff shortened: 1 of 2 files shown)


tests/spec/core/mdn-annotation-spec.js
```diff
@@ -2,17 +2,46 @@ import { flushIframes, makeRSDoc, makeStandardOps } from "../SpecHelper.js";
...
+    it("prefers `mdn.key` over shortName", async () => {
+      const ops = makeStandardOps({
+        mdn: { baseJsonPath, key: "payment-request" },
+        shortName: "whatever",
+      });
+      const doc = await makeRSDoc(ops, "spec/core/mdn-annotation.html");
+      expect(getMdnBox(doc).matches("aside.mdn")).toBeTrue();
+    });
+
+    it("supports mdn to be a string", async () => {
```
sidvishnoi, line 35: This test sadly fetches over network, as we cannot set `mdn.baseJsonPath` (used for mock JSON source) on a string.
I will add a util for to know if we're in test env for these things in a separate PR.

sidvishnoi: Well, all these tests are going over network right now anyway as tests don't add `mdn.specMapUrl` :man_facepalming:

marcoscaceres: We can live with that...

sidvishnoi:
Wiki updated: https://github.com/w3c/respec/wiki/mdn

Issue #2866, opened 2020-05-12
<var> doesn't match if it includes a newline


saschanaz:
## Important info

* URL to affected spec: 
* ReSpec version: 25.7.0
* [x] I did a "hard refresh", but it's still busted. 

## Description of problem

What happened (e.g., it crashed)?:

Clicking a `var` does not highlights others.

But what I was expecting?: Both should be highlighted

Optional, steps to reproduce:
1. Insert

    ```html
    <var>my previous</var>
    <var>my
    previous</var>
    ```

2. Check clicking either one highlights the other.

PR #2867, opened 2020-05-12
fix(core/highlightVars): allow multiline variables

sidvishnoi:
Closes https://github.com/w3c/respec/issues/2866


sidvishnoi:
> That was blazing fast

So was the review! 🏎️

PR #2871, opened 2020-05-15
chore(ci): update checkout to v2

sidvishnoi:
v2 is [claimed](https://github.com/actions/checkout/#whats-new) to be faster.

It is indeed faster! :shipit: 
image
image


marcoscaceres:
Oh, 10x faster! Must be a 10x developer!

PR #2872, opened 2020-05-15
chore(ci): use npm ci instead of npm i

sidvishnoi:
As we have a package-lock, we can use it, and it's faster than `npm install`.


sidvishnoi:
Thanks to puppeteer being hostile to npm cache, we don't get benefits of this change.

marcoscaceres:
I wonder if this will change in the future. I guess we can always revisit.

sidvishnoi:
We longer install puppeteer in CI, so this can come back in :tada:

sidvishnoi:
Oops. `npm ci` clears `node_modules` so our present cache is useless :laughing: 
Will add cache for `~/.npm` instead, to see if we get improvements.

marcoscaceres:
sounds good

PR #2876, opened 2020-05-18
fix(core/structure): support large levels in TOC

thijsbrentjens:
See discussion in Geonovum repo: https://github.com/Geonovum/respec/pull/185

Calculating the level for the TOC failed when large (>=10) number of sections where available.


sidvishnoi:
Not sure if we should test this or it's fine. cc @marcoscaceres

marcoscaceres:
Wouldn't hurt to add a test... catch regressions in the future.

tests/spec/core/structure-spec.js
```diff
@@ -172,6 +172,25 @@ describe("Core - Structure", () => {
     );
   });
 
+  it("should correctly put all headings until maxTocLevel in ToC", async () => {
+    const times = (n, fn) =>
+      Array.from({ length: n }, (_, i) => fn(i)).join("\n");
+    const body = `
+      ${makeDefaultBody()}
+      ${times(9, () => `<section><h2>pass</h2></section>`)}
```
marcoscaceres, line 180: Might be good to add comment here why the `9` and `10` different sections are needed. 

Also, as we are checking maxTocLevel, would we want to go to level 3 and check level 3 stuff is not actually there?

sidvishnoi: > would we want to go to level 3 and check level 3 stuff is not actually there?

That's tested under `should limit ToC depth with maxTocLevel`

> Might be good to add comment here why the 9 and 10 different sections are needed.

:+1:

PR #2878, opened 2020-05-18
chore(ci): skip puppeteer install

sidvishnoi:
GitHub actions [already have](https://github.com/actions/virtual-environments/blob/master/images/linux/Ubuntu1804-README.md) Chrome and Firefox installed, so we can skip our own install..

Cons: A rare risk of puppeteer API getting out of sync with installed Chrome/Firefox version.
Pros: Much faster CI (4m 26s to 3m 15s), smaller cache (157MB compressed to 35MB) (need cache purging to see - will be automatic in future, I checked with temporary manual cache purges)

What's next?
- We can now use `npm ci` :tada: 
- We can also remove `actions/setup-node`, as Node v12 is also installed.

### DIFF
--- a/.github/workflows/pr-push.yml
+++ b/.github/workflows/pr-push.yml
@@ -6,6 +6,10 @@ on:
       - gh-pages
   pull_request: {}
 
+env:
+  PUPPETEER_SKIP_CHROMIUM_DOWNLOAD: 1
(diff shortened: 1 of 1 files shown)


marcoscaceres:
very nice! specially the speed increases and no needless downloads!

PR #2888, opened 2020-05-20
fix(core/dfn-panel): avoid using CSS variables

sidvishnoi:
Reverts hot-fix from https://github.com/w3c/respec/pull/2846 (See https://github.com/w3c/respec/issues/2845)
Precursor to https://github.com/w3c/respec/pull/2820

This changes the markup for dfnPanel, mainly converts the caret to a `<span>` instead of `::before` as we cannot programmatically set position of `::before` without CSS variables - which aren't supported by nuHTML validator.

### DIFF
--- a/assets/dfn-panel.css
+++ b/assets/dfn-panel.css
@@ -1,39 +1,42 @@
 /* dfn popup panel that list all local references to a dfn */
+/**
+ * TODO: Revert changes due to https://github.com/w3c/respec/pull/2888 when
+ * https://github.com/w3c/css-validator/pull/111 is fixed.
+ */
 dfn {
   cursor: pointer;
 }
 
 .dfn-panel {
-  --fill: #fff;
   position: absolute;
-  left: var(--left); /* set via JS */
-  top: var(--top); /* set via JS */
   z-index: 35;
   min-width: 300px;
   max-width: 500px;
   padding: 0.5em 0.75em;
   margin-top: 0.6em;
   font: small Helvetica Neue, sans-serif, Droid Sans Fallback;
-  background: var(--fill);
+  background: #fff;
   color: black;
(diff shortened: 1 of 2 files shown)


assets/dfn-panel.css
```diff
@@ -1,39 +1,42 @@
...
+}
+.dfn-panel:not(.docked) > .caret::before,
+.dfn-panel:not(.docked) > .caret::after {
   content: "";
   position: absolute;
   border: 10px solid transparent;
   border-top: 0;
-  border-bottom: 9px solid #a2a9b1; /* triangle outline */
-  top: -9px;
-  left: calc(var(--caret-offset, 0px) + 0.75em); /* set via JS */
```
saschanaz, line 33: Might be better to use `attr()` in the future rather than just reverting this. Google just posted an intent to implement for the feature. https://groups.google.com/a/chromium.org/forum/#!msg/blink-dev/FGCgsKmylhw/BSUrk2roCQAJ

Since the validator is quite inactive, the browser implementation might be quicker.

saschanaz: Looking at the code, the validator already supports `attr()` so we probably can just use it when the browser support lands.

sidvishnoi: Even though nuHTML allows advanced `attr()`, [CSS validator](http://jigsaw.w3.org/css-validator/validator) doesn't.

saschanaz: Ah right, it can parse the syntax but the validation is broken. 🤔

saschanaz: Hey, actually it does validate it!

```
.abc {
  width: calc(attr(data-foo px) / 10);
}
```

Try this code.

sidvishnoi: Oh right! I missed the `px` suffix

PR #2892, opened 2020-05-21
chore(package): convert some dependencies to devDependencies

sidvishnoi:
Faster installs!


sidvishnoi:
Was transient. Probably some network issue. Worked on CI restart.

Issue #1050, opened 2017-01-20
Magic sections for defining WebIDL


marcoscaceres:
I would like it if ReSpec could do more of the manual labor for me when it comes to IDL sections. 

I want to just write:

```HTML
<section>
   <pre class="idl">
    interface WhatEver{};
   </pre>
</section>
```

And for that to become:

```HTML
<section data-dfn-for="WhatEver" data-link-for="WhatEver">
   <h2><dfn>WhatEver</dfn> interface</h2>
   <pre class="idl">
    interface WhatEver{};
   </pre>
</section>
```

pradeepgangwar:
Working on it.

marcoscaceres:
Too much magic.

PR #2893, opened 2020-05-21
feat(core/webidl): add data-dfn-for to IDL sections

sidvishnoi:
Part of #1050


src/core/webidl.js
```diff
@@ -358,6 +359,38 @@ function renderWebIDL(idlElement, index) {
...
+/**
+ * Add data-dfn-for to the closest section if not present already.
+ * @param {HTMLPreElement} idlElement
+ * @param {ReturnType<typeof webidl2.parse>} parse
+ */
+function addDataDfnFor(idlElement, parse) {
+  const closestSection = idlElement.closest("section");
+  if (closestSection.hasAttribute("data-dfn-for")) return;
+
+  const topLevelEntities = ["dictionary", "interface", "callback interface"];
```
sidvishnoi, line 372: Need "interface mixin" also, right? Or do we want to support only "dictionary" and "interface"?
https://github.com/w3c/respec/blob/f62ee6b95d79c6008d01da35a95464b5854b870b/src/core/webidl.js#L240-L244

src/core/webidl.js
```diff
@@ -358,6 +359,38 @@ function renderWebIDL(idlElement, index) {
...
+      dfnFors.push(dfnFor);
+    }
+  }
+  if (dfnFors.length === 1) {
+    closestSection.dataset.dfnFor = dfnFors[0];
+  } else if (!dfnFors.length) {
+    return;
+  } else {
+    closestSection.dataset.dfnFor = "";
+    const options = dfnFors.map(dfnFor => `"${dfnFor}"`).join(", ");
```
sidvishnoi, line 387: Maybe we should not suggest these options? User might want to add something else?

src/core/webidl.js
marcoscaceres: yeah, it would be all the top level ones...

src/core/webidl.js
marcoscaceres: yeah, don't suggest these yet... remember that this currently works today:

```
<section data-dfn-for="Whatever">
   
    <section>
    <pre class="idl">
        interface Whatever{};
        interface SomethingElse{};
    <pre>
    </section>
</section>
```

marcoscaceres: actually, the above is probably bad practice and doesn't matter...

Issue #2874, opened 2020-05-15
Use custom ReSpec version while generating via respec2html


sidvishnoi:
There was a discussion earlier about pinning ReSpec version while exporting document - bypassing the version spec was originally using. (https://github.com/w3c/respec/pull/2702)

I've found another use case for doing so. I was thinking of creating a GitHub Action for monitoring regressions in ReSpec, by using actual specs. Idea is similar to using https://respec-preview.netlify.app, but automate it in the CI process. With each commit in `develop`, we'll build a few selected specs against the version spec is using (latest release) and the one with changes of current commit, and find any new errors/warnings or changes to links.

PR #2896, opened 2020-05-21
chore(ci): split into PR and push workflows

sidvishnoi:
Idea is:
- We run all tests for PRs (pr.yml). Good work.
- When merge the PR, all tests have already passed. Plus, we always sync with develop before merge, so things are unlikely to have changed.
- But to be sure, we still run fast tests when PR is merged, or when we release (push: gh-pages). To not waste CPU cycles, we:
  - ignore lint -> not high priority.
  - run karma only in Chrome, as Firefox is slower and there is unlikely any change since PR tests. We can also do the opposite: test more browsers on merge.
- We can run other jobs on merge, like https://github.com/w3c/respec/pull/2895/ or regression tests (See https://github.com/w3c/respec/issues/2874). These need not run on every PR synchronize event.

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -1,32 +1,11 @@
-name: Node CI
+name: Node CI (push)
 on:
   push:
     branches:
       - develop
       - gh-pages
-  pull_request: {}
-
-env:
-  PUPPETEER_SKIP_CHROMIUM_DOWNLOAD: 1
-  PUPPETEER_EXECUTABLE_PATH: /usr/bin/google-chrome
 
 jobs:
-  lint:
-    name: Check linting issues
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-node@v1
-        with:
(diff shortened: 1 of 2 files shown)


.github/workflows/push.yml
```diff
@@ -0,0 +1,48 @@
...
+          node-version: 12
+      - name: Cache node modules
+        uses: actions/cache@v1
+        with:
+          path: ~/.npm
+          key: ${{ runner.os }}-node-12-${{ hashFiles('**/package-lock.json') }}
+      - name: install & build
+        run: |
+          npm ci
+          npm run build:w3c & npm run build:geonovum
```
marcoscaceres, line 45: do we need to build:dini also?

sidvishnoi: There are no tests for dini profile yet.

marcoscaceres: True...

sidvishnoi:
Oh dear! Forgot skipping puppeteer install in push.yml 😞

PR #2899, opened 2020-05-23
chore: make tests and CI run for w3c-common profile

sidvishnoi:
- Sync the GitHub action improvements from develop (remove redundant commands, avoid Chromium install etc.)
- Run tests with respect to w3c-common profile (in CI as well as `npm run test:karma`)
- Don't run headless tests, as they use `respec-w3c` profile anyway.

Another chore!


package.json
```diff
@@ -68,6 +68,7 @@
     "build:w3c-debug": "npm run build:w3c -- --debug",
     "build:w3c-with-jquery": "node ./tools/builder.js --profile=w3c-common",
     "build:w3c": "node ./tools/builder.js --profile=w3c",
+    "build:w3c-common": "node ./tools/builder.js --profile=w3c-common",
```
saschanaz, line 71: We already have `build:w3c-with-jquery` 😁

sidvishnoi: Ahhhh :facepalm:

PR #2901, opened 2020-05-23
fix(core/dfn-panel): precompute all dfnPanels

sidvishnoi:
This pre-computes all dfn panels and appends them to body. Then click listener only needs to hide/show and position already existing panels.

Precursor to https://github.com/w3c/respec/pull/2820

Tests have changed a lot because earlier we relied on existence of panel in DOM, but now we've to check if it's visible (`.hidden`) or not.


src/core/dfn-panel.js
```diff
@@ -14,6 +14,13 @@ export async function run() {
     document.querySelector("link")
   );
 
+  /** @type {HTMLElement} */
+  const elems = document.querySelectorAll("dfn, .index-term");
+  for (const el of elems) {
+    const panel = createPanel(el);
+    document.body.append(panel);
```
sidvishnoi, line 21: Would it be better to add all panels to a fragment and then append all at once?

src/core/dfn-panel.js
```diff
@@ -69,9 +75,10 @@ function createPanel(dfn) {
   const href = dfn.dataset.href || `#${id}`;
   const links = document.querySelectorAll(`a[href="${href}"]:not(.index-term)`);
 
+  const panelId = `dfn-panel-for-${dfn.id}`;
   /** @type {HTMLElement} */
   const panel = hyperHTML`
-    <aside class="dfn-panel" id="dfn-panel">
+    <aside class="dfn-panel hidden removeOnSave" id="${panelId}">
```
sidvishnoi, line 81: `removeOnSave` until https://github.com/w3c/respec/pull/2820

src/core/dfn-panel.js
```diff
@@ -142,7 +149,10 @@ function getReferenceTitle(link) {
  * @param {{ x: number, y: number }} clickPosition
  */
 function displayPanel(dfn, panel, { x, y }) {
-  document.body.appendChild(panel);
+  for (const el of document.querySelectorAll(".dfn-panel")) {
```
sidvishnoi, line 152: Maybe?
```suggestion
  for (const el of document.querySelectorAll(".dfn-panel:not(.hidden)")) {
```

src/ui/dfn-list.js
```diff
@@ -78,7 +78,9 @@ function labelDfnIfExported(dfn) {
  * @param {HTMLElement} dfn a definition
  */
 function labelDfnIfUnused(dfn) {
-  const isUsed = document.querySelector(`a[href^="#${dfn.id}"]`);
+  const isUsed = document.querySelector(
+    `:not(.dfn-panel) > :not(li) > a[href^="#${dfn.id}"]`
```
sidvishnoi, line 82: I'm not sure how this works, but it works :sweat_smile:

assets/dfn-panel.css
```diff
@@ -84,3 +84,7 @@ dfn {
   max-height: 30vh;
   overflow: auto;
 }
+
+.dfn-panel.hidden {
```
saschanaz, line 88: There is an attribute named [`hidden`](https://html.spec.whatwg.org/multipage/interaction.html#the-hidden-attribute) which does the same 👀

src/core/dfn-panel.js
saschanaz: That might be cleaner.

assets/dfn-panel.css
sidvishnoi: How can I forget!

sidvishnoi:
> I personally prefer to only compute dfn maps and do anything else on runtime

Is this store references as JSON idea? If so, I'm not convinced how it may be used. A parser would need as much effort to parse `<aside>` than to find and parse a `<script type="application/json">`

src/core/dfn-panel.js
```diff
@@ -49,29 +57,36 @@ function deriveAction(clickTarget) {
...
 function createPanel(dfn) {
   const { id } = dfn;
   const href = dfn.dataset.href || `#${id}`;
   const links = document.querySelectorAll(`a[href="${href}"]:not(.index-term)`);
 
+  const panelId = `dfn-panel-for-${dfn.id}`;
   /** @type {HTMLElement} */
   const panel = hyperHTML`
-    <aside class="dfn-panel" id="dfn-panel">
+    <aside class="dfn-panel removeOnSave" id="${panelId}" hidden>
```
marcoscaceres, line 89: why remove?

sidvishnoi: https://github.com/w3c/respec/pull/2901#discussion_r429569305

src/ui/dfn-list.js
marcoscaceres: Maybe add this 👆 as a comment? The query is indeed complex.

saschanaz:
> A parser would need as much effort to parse `<aside>` than to find and parse a `<script type="application/json">`

It's not about efficiency, it's more about splitting - creating/opening panels at runtime while complex mapping at build time.

That said, there is no blocker here, let's :shipit: and see how it goes.

Issue #2914, opened 2020-05-26
Be nice to link "coded" things


marcoscaceres:
Consider: 

```HTML
<a data-cite=
               "!HTML#multipart/form-data-encoding-algorithm"><code>multipart/form-data</code>
               encoding algorithm</a>.
```

It would be super nice to just write:
```HTML 
[=`multipart/form-data` encoding algorithm=]
```

PR #2916, opened 2020-05-27
fix(core/inlines): allow escaping forward-slash in terms

sidvishnoi:
Closes https://github.com/w3c/respec/issues/2914 as backticks are already supported but `/` is not

### DIFF
--- a/src/core/inlines.js
+++ b/src/core/inlines.js
@@ -80,7 +80,7 @@ function inlineElementMatches(matched) {
   const [element, attribute] = splitBySlash(value, 2);
   const [xrefType, xrefFor, textContent] = attribute
     ? ["element-attr", element, attribute]
-    : ["element|element-attr", null, element];
+    : ["element", null, element];
   const code = html`<code
     ><a data-xref-type="${xrefType}" data-xref-for="${xrefFor}"
       >${textContent}</a
--- a/src/core/inlines.js
+++ b/src/core/inlines.js
@@ -81,11 +81,17 @@ function inlineElementMatches(matched) {
     .replace("\\/", "%%")
     .split("/", 2)
     .map(s => s && s.trim().replace("%%", "/"));
-  const [xrefType, xrefFor, textContent] = attribute
-    ? ["element-attr", element, attribute]
-    : ["element|element-attr", null, element];
+  const [xrefFor, textContent] = attribute
+    ? [element, attribute]
+    : [null, element];
+  const xrefType = [];
+  if (attribute || value.includes("\\/")) {
+    xrefType.push("element-attr");
+  } else if (!attribute) {
+    xrefType.push("element");
+  }
   const code = html`<code
-    ><a data-xref-type="${xrefType}" data-xref-for="${xrefFor}"
+    ><a data-xref-type="${xrefType.join("|")}" data-xref-for="${xrefFor}"
       >${textContent}</a
     ></code
   >`;
--- a/src/core/inlines.js
+++ b/src/core/inlines.js
@@ -77,10 +77,13 @@ const inlineElement = /(?:\[\^[^^]+\^\])/; // Inline [^element^]
  */
 function inlineElementMatches(matched) {
   const value = matched.slice(2, -2).trim();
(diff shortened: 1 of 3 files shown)


src/core/inlines.js
```diff
@@ -77,10 +77,13 @@ const inlineElement = /(?:\[\^[^^]+\^\])/; // Inline [^element^]
  */
 function inlineElementMatches(matched) {
   const value = matched.slice(2, -2).trim();
-  const [element, attribute] = value.split("/", 2).map(s => s && s.trim());
+  const [element, attribute] = value
+    .replace("\\/", "%%")
+    .split("/", 2)
+    .map(s => s && s.trim().replace("%%", "/"));
```
sidvishnoi, line 83: In future, we can do:
``` js
const [element, attribute] = value.split(/(?<!\\)\//, 2).map(s => s && s.trim());
const textContent = element.replace("\\/", "/")
```

``` js
// example
const fn = s => s.split(/(?<!\\)\//, 2).map(s => s && s.trim().replace("\\/", "/"));
fn("application/x-www-form-urlencoded");
// Array [ "application", "x-www-form-urlencoded" ]
fn("application\\/x-www-form-urlencoded")
// Array [ "application/x-www-form-urlencoded" ]
```

Open to better ideas.

saschanaz: What about just `.split(/\\?\//, 2)`?

src/core/inlines.js
```diff
@@ -77,10 +77,13 @@ const inlineElement = /(?:\[\^[^^]+\^\])/; // Inline [^element^]
...
   const [xrefType, xrefFor, textContent] = attribute
     ? ["element-attr", element, attribute]
-    : ["element", null, element];
+    : ["element|element-attr", null, element];
```
saschanaz, line 86: Does this mean there is no element-only syntax anymore?

src/core/inlines.js
saschanaz: LGTM for now, maybe better to define this as a separate function as this is also used elsewhere. A comment about the future change will be great.

src/core/inlines.js
sidvishnoi: Yes. There isn't. 
Maybe we can add `element-attr` only if `value.includes("\\/")`, because elements don't have slashes. Ideally, this should be `attr-value`, but Bikeshed/specs have data messed up.

saschanaz: Could you give me an example for me to understand better? 👀

sidvishnoi: The example in the referenced issue is one: `application/x-www-form-urlencoded`. It's not an attribute (`element-attr`), but an attribute value `attr-value` for `enctype` attribute.

> Maybe we can add element-attr only if value.includes("\\/"), because elements don't have slashes.

This is partially wrong assumption, as it can be `element-attr` without `/`

sidvishnoi: How about we support this:
```
[^element^], [^element/element-attr^] // general, good usage
[^/element-attr^] // for terms like above
(we don't support attr-value in ReSpec yet, might in near future)
```

saschanaz: > but HTML exported it as (or Bikeshed understood it to be) `element-attr`.

Shouldn't we fix HTML instead of doing `element|element-attr`? This type of mess shouldn't be frequent.

sidvishnoi: Good point. I wonder why I jumped on the fix :flushed:.
I just checked, and there are only 3 instances near https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#attr-fs-enctype:

```
    The "application/x-www-form-urlencoded" keyword and corresponding state.
    The "multipart/form-data" keyword and corresponding state.
    The "text/plain" keyword and corresponding state.
```

saschanaz: > Good point. I wonder why I jumped on the fix 😳.

This is why a peer review is essential even for obvious changes 😁

sidvishnoi: Filed https://github.com/whatwg/html/issues/5586

PR #2917, opened 2020-05-27
perf(core/mdn-annotation): avoid request to specmap.json

sidvishnoi:
... and added typescript annotations

There is a scope for a large number of refactor/cleanups in this module. Should I do them here or separate PRs?

### DIFF
--- a/src/core/mdn-annotation.js
+++ b/src/core/mdn-annotation.js
@@ -42,11 +42,6 @@ async function loadStyle() {
   }
 }
 
-function fetchAndCacheJson(url, maxAge) {
-  if (!url) return {};
-  return fetchAndCache(url, maxAge).then(r => r.json());
-}
-
 function insertMDNBox(node) {
   const targetAncestor = node.closest("section");
   const { previousElementSibling: targetSibling, parentNode } = targetAncestor;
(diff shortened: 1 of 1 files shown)


src/core/mdn-annotation.js
```diff
@@ -189,3 +171,33 @@ function getMdnKey(conf) {
...
+/**
+ * @param {string} key MDN key
+ * @param {object} mdnConf
+ * @param {string} [mdnConf.specMapUrl]
+ * @param {string} [mdnConf.baseJsonPath]
+ * @param {number} [mdnConf.maxAge]
+ *
+ * @typedef {{ [browser in keyof MDN_BROWSERS]: { version_added: string } }} MdnSupportEntry
+ * @typedef {{ name: string, title: string, summary: string, support: MdnSupportEntry }} MdnEntry
+ * @typedef {{ [id: string]: MdnEntry[] }} MdnData
```
saschanaz, line 184:
```suggestion
 * @typedef {Record<string, MdnEntry[]>} MdnData
```

src/core/mdn-annotation.js
```diff
@@ -189,3 +171,33 @@ function getMdnKey(conf) {
...
 }
+
+/**
+ * @param {string} key MDN key
+ * @param {object} mdnConf
+ * @param {string} [mdnConf.specMapUrl]
+ * @param {string} [mdnConf.baseJsonPath]
+ * @param {number} [mdnConf.maxAge]
+ *
+ * @typedef {{ [browser in keyof MDN_BROWSERS]: { version_added: string } }} MdnSupportEntry
```
saschanaz, line 182:
```suggestion
 * @typedef {Record<keyof MDN_BROWSERS, { version_added: string }>} MdnSupportEntry
```

sidvishnoi: I kinda prefer this to `Record<>`, as we can specify what the key is (`browser` as a string here).

saschanaz: Well, I think `keyof MDN_BROWSERS` already is a useful hint in this case 😉

sidvishnoi: Good point!

src/core/mdn-annotation.js
```diff
@@ -189,3 +170,31 @@ function getMdnKey(conf) {
...
+/**
+ * @param {string} key MDN key
+ * @param {object} mdnConf
+ * @param {string} [mdnConf.specMapUrl]
+ * @param {string} [mdnConf.baseJsonPath]
+ * @param {number} [mdnConf.maxAge]
+ *
+ * @typedef {Record<keyof MDN_BROWSERS, { version_added: string }>} MdnSupportEntry
+ * @typedef {{ name: string, title: string, summary: string, support: MdnSupportEntry }} MdnEntry
+ * @typedef { Record<string, MdnEntry[]>} MdnData
```
saschanaz, line 183:
```suggestion
 * @typedef {Record<string, MdnEntry[]>} MdnData
```

PR #2918, opened 2020-05-27
refactor(core/mdn-annotation): cleanup, better type definitions

sidvishnoi:


src/core/mdn-annotation.js
```diff
@@ -43,42 +41,44 @@ async function loadStyle() {
 
 function insertMDNBox(node) {
   const targetAncestor = node.closest("section");
```
marcoscaceres, line 43: I wonder if we should check `targetAncestor` not being null too? It could be an authoring mistake.

PR #2921, opened 2020-05-29
fix(core/xref): give inline spec context least priority

sidvishnoi:
See https://github.com/w3c/manifest/pull/883#issue-424889006


marcoscaceres:
Test?

PR #2922, opened 2020-05-29
fix(core/xref): add data-cite as authored in spec context

sidvishnoi:
This avoids the need to add `aliasOf` entries in `localBiblio`
It allows ReSpec to search and link using the same authored `data-cite`. Earlier, we could search using authored data-cite, but link was always based on `shortname`.
See: https://github.com/w3c/manifest/pull/878#discussion_r432331678

### DIFF
--- a/src/core/xref.js
+++ b/src/core/xref.js
@@ -389,8 +389,12 @@ function addDataCiteToTerms(elems, queryKeys, data, conf) {
  */
 function addDataCite(elem, query, result, conf) {
   const { term } = query;
-  const { uri, shortname: cite, normative, type, for: forContext } = result;
-
+  const { uri, shortname, spec, normative, type, for: forContext } = result;
+  let cite = shortname;
+  // if authored spec context had `result.spec`, use it instead of shortname
+  if (query.specs && query.specs.some(specs => specs.includes(spec))) {
+    cite = spec;
+  }
   const path = uri.includes("/") ? uri.split("/", 1)[1] : uri;
   const [citePath, citeFrag] = path.split("#");
   const dataset = { cite, citePath, citeFrag, type };
--- a/src/core/xref.js
+++ b/src/core/xref.js
@@ -390,11 +390,8 @@ function addDataCiteToTerms(elems, queryKeys, data, conf) {
 function addDataCite(elem, query, result, conf) {
   const { term, specs = [] } = query;
   const { uri, shortname, spec, normative, type, for: forContext } = result;
-  let cite = shortname;
   // if authored spec context had `result.spec`, use it instead of shortname
-  if (specs.flat().includes(spec)) {
-    cite = spec;
-  }
+  const cite = specs.flat().includes(spec) ? spec : shortname;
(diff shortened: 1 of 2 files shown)


src/core/xref.js
```diff
@@ -389,8 +389,12 @@ function addDataCiteToTerms(elems, queryKeys, data, conf) {
  */
 function addDataCite(elem, query, result, conf) {
   const { term } = query;
-  const { uri, shortname: cite, normative, type, for: forContext } = result;
-
+  const { uri, shortname, spec, normative, type, for: forContext } = result;
+  let cite = shortname;
+  // if authored spec context had `result.spec`, use it instead of shortname
+  if (query.specs && query.specs.some(specs => specs.includes(spec))) {
```
sidvishnoi, line 395:
```suggestion
  if (query.specs && query.specs.flat().includes(spec)) {
```

src/core/xref.js
```diff
@@ -388,9 +388,10 @@ function addDataCiteToTerms(elems, queryKeys, data, conf) {
  * @param {any} conf
  */
 function addDataCite(elem, query, result, conf) {
-  const { term } = query;
-  const { uri, shortname: cite, normative, type, for: forContext } = result;
-
+  const { term, specs = [] } = query;
+  const { uri, shortname, spec, normative, type, for: forContext } = result;
+  // if authored spec context had `result.spec`, use it instead of shortname
+  const cite = specs.flat().includes(spec) ? spec : shortname;
```
saschanaz, line 394: It seems `specs.flat()` happens frequently. Can it be flattened inside `getSpecContext`?

sidvishnoi: The server relies on specContext being a nested array - first sub-array has highest priority and so on.

saschanaz: What happens if everything is in the first subarray?

sidvishnoi: This was done to avoid ambiguities with specContext. Consider:

``` html
<script>
  respecConfig.xref = ["fetch", "html"];
</script>
<body>
 <section data-cite="fetch dom"> 
   <a>body</a>
  <a>EventHandler</a> ...
 </section>
</body>
```

If we have everything in first specContext array, we have an ambiguity (body is defined in both fetch and html). So, we use a fallback chain, which goes like: search in [fetch, dom], if not found, search in [fetch, html]

saschanaz: Understood, thanks. Although this should still be able to flattened first as this function is called after the query, but not sure how much it would improve the code.

sidvishnoi: Would `query.specs.some(specs => specs.includes(spec)` be better than `flat()`?

saschanaz: Not sure as it's significantly longer 🤔

sidvishnoi: Though it avoids creating a new array..

saschanaz: Well, shorter code is better than micro-optimization 👀

Issue #2861, opened 2020-05-11
Better support indicators for MDN boxes


marcoscaceres:
Elsewhere, @sideshowbarker wrote:

@marcoscaceres This reminds me that I’ve been meaning to suggest an enhancement to the MDN annotations generated by Respec: We can add an implementation-status indicator to the collapsed MDN buttons. Take a look at https://wicg.github.io/webusb/#device-usage

The (Bikeshed-generated) MDN annotations there shout pretty loudly (maybe too loudly…) that `USBDevice` is only implemented in one engine.

Then for comparison, go to https://drafts.csswg.org/css-logical-1/#caption-side and scroll down. The implementation-status indicators in the MDN annotations there are pretty useful. The CSS Logical Properties and Values spec is a mix of features at varying levels of implementation maturity: some implemented in all engines, some in two engines, and some only in one engine. A reader can get a pretty quick sense of which are which just with quick scroll through the spec.

Due to the baroque complexity of BCD status data for some features, programatically identifying whether a particular browser currently supports a particular feature requires some work. But I’ve already done that work — three times: Once in Free Pascal (for the HTML spec) and once in Python (for Bikeshed) and once also in JavaScript (for https://github.com/w3c/mdn-spec-links).

We could re-use the same algorithm for Bikeshed — and probably also copy over the JavaScript code for it that I already have. I’ve been meaning to write a Bikeshed patch for it, but haven’t managed to get around to it quite yet.

marcoscaceres:
Sample of not supported: 

![Screenshot 2020-05-11 18 30 45](https://user-images.githubusercontent.com/870154/81540832-949e7e00-93b5-11ea-89c5-aa243feab73e.png)

Sample of supported: 

![Screenshot 2020-05-11 18 30 59](https://user-images.githubusercontent.com/870154/81540838-9700d800-93b5-11ea-8922-2b1163108a6d.png)

sideshowbarker:
I realize now I misspoke a bit in what’s in the issue description here. It’s not actually necessary to re-write an algorithm for generating the implementation-status indicatators.

That’s because all the https://w3c.github.io/mdn-spec-links/ JSON data already now includes an `engines` array that lists the shortname of each engine that supports a particular feature; like this:

```json
    "dom-mediadeviceinfo-deviceid": [
        {
            "engines": [
                "blink",
                "gecko"
            ], 
```
So you can just check the length of that `engines` array.

PR #2923, opened 2020-05-29
feat(core/mdn-annotation): show engine support indicators 

sidvishnoi:
Closes https://github.com/w3c/respec/issues/2861


src/core/mdn-annotation.js
```diff
@@ -31,6 +31,14 @@ const MDN_BROWSERS = {
   webview_android: "WebView Android",
 };
 
+const localizationStrings = {
+  en: {
+    inAllEngines: "This feature is in all current engines.",
```
saschanaz, line 36: "major" engines, maybe. Servo is an engine too 😄

src/core/mdn-annotation.js
```diff
@@ -31,6 +31,14 @@ const MDN_BROWSERS = {
...
+    inSomeEngines: "This feature is in less than two current engines.",
```
saschanaz, line 37: "This feature has poor support"? The number should be shown next anyway.

src/core/mdn-annotation.js
```diff
@@ -188,3 +198,29 @@ function findElements(data) {
...
+    return html`<span title="${l10n.inSomeEngines}">🚫</span>`;
+  }
+}
+
+/**
+ * @param {MdnEntry['engines']} engines
+ * @returns {HTMLParagraphElement|undefined}
+ */
+function getEngineSupport(engines) {
+  if (engines.length === 3) {
```
saschanaz, line 220: Why 3 when `MDN_BROWSERS` includes many browsers?

src/core/mdn-annotation.js
sidvishnoi: Was copying Bikeshed

src/core/mdn-annotation.js
sidvishnoi: "blink", "gecko" and "webkit"

saschanaz: Ah okay, it's from https://github.com/w3c/mdn-spec-links/blob/346c5fac134f99f4bd52920a1be1542f661aede6/.browser-compat-data-process.js#L365.

src/core/mdn-annotation.js
saschanaz: Do we need to sync the messages?

sidvishnoi: Nope. We need something for starters though..

src/core/mdn-annotation.js
marcoscaceres: s/poor/limited maybe.

src/core/mdn-annotation.js
marcoscaceres: Agree... we should start with this set. It's nice that both ReSpec and BikeShed are showing the same boxes/info for now.

saschanaz: I think adding "major" should still be fine, though. EdgeHTML is still in a lot of machines and there are also other minor ones.

saschanaz: Is there a reason you don't want it? 😕

sidvishnoi: Oops. Missed your newer comments before merge @saschanaz. Will improve messages in a followup

src/core/mdn-annotation.js
sidvishnoi: > "This feature has poor support"? The number should be shown next anyway.

"This feature has limited support (n engines)"?

src/core/mdn-annotation.js
sidvishnoi: "This feature is in all major engines"

sidvishnoi: @saschanaz do you think this message is fine? Open to suggestions.

saschanaz: That looks fine 😁

src/core/mdn-annotation.js
saschanaz: The support table should show how its limited, so maybe no need for "n engines" part.

Issue #792, opened 2016-05-26
Plug echidna in test suite


dontcallmedom:
To catch bugs that appear in the export to HTML phase of ReSpec (such as #786 and #791) before releases, it would be great if the ReSpec test suite could run one of its reference documents through the pipeline of:
respec → spec-generator → echidna
 (using the underlying libraries rather than the on-line services)

deniak:
We are about to release a new version of specberus that will provide a [REST API](https://github.com/w3c/specberus#5-rest-api). As I'm writing this, it's not deployed yet but this is only a matter of days.
The REST API can also help if you don't want to add a dependency to specberus.

marcoscaceres:
oh! that would be great!

marcoscaceres:
@deniak, it's probably easier for me to hook up the JS API

marcoscaceres:
We should make this our highest priority going forward. It would solve/expose a lot of issues and would actually make the generation tests that we do useful.

deniak:
@marcoscaceres in case it helps, see [how we do it in Echidna](https://github.com/w3c/echidna/blob/master/lib/specberus-wrapper.js#L16)

marcoscaceres:
@deniak thanks, btw! working from your example!

marcoscaceres:
Hoping to find time this week to finish this off.

marcoscaceres:
Notes from meeting with @dontcallmedom: 
 - [ ] On Travis, import specberus into travis, then run it against a spec with all the statuses.

marcoscaceres:
We figured out the end point :D https://www.w3.org/pubrules/api/ 

For example:
https://www.w3.org/pubrules/api/validate?url=https://www.w3.org/TR/payment-request/&profile=auto

In theory, can also run locally... but not tried yet.

sidvishnoi:
Plan is to support Pubrules in spec-prod (https://github.com/w3c/spec-prod/pull/100) and enable spec-prod in ReSpec CI.

marcoscaceres:
Closing. Spec validation has moved to external CI tooling (spec-prod, Echidna) rather than being integrated into ReSpec's test suite.

Issue #639, opened 2016-03-10
Tests should test spec generation against pubrules


halindrome:
Since a major customer for respec is w3c publications, we should ensure that generated specs (continue to) pass pubrules.

xfq:
Related: https://github.com/w3c/respec/issues/792

sidvishnoi:
We should do this in [respec-validator](https://github.com/marcoscaceres/respec-validator) probably.

sidvishnoi:
Closing in favor of https://github.com/w3c/respec/issues/792

PR #2945, opened 2020-06-04
fix(core/dfn-panel): fix heading markup due to pubrules id rule

sidvishnoi:
[Pubrules wanted](https://github.com/w3c/web-share/runs/738231021#step:3:53) an ID on each for `<h2>Referenced in:</h2>`. Falling back to less semantic `<b>` as generating ID on these headings doesn't make sense.

Need to prioritize on https://github.com/w3c/respec/issues/792 / https://github.com/w3c/respec/issues/639 now.

### DIFF
--- a/src/core/dfn-panel.js
+++ b/src/core/dfn-panel.js
@@ -49,7 +49,7 @@ function createPanel(dfn) {
         <a class="self-link" href="${href}">Permalink</a>
         ${dfnExportedMarker(dfn)}
       </div>
-      <h2>Referenced in:</h2>
+      <b>Referenced in:</b>
       ${referencesToHTML(id, links)}
     </aside>
   `;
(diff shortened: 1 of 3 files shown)


src/core/dfn-panel.js
```diff
@@ -49,7 +49,7 @@ function createPanel(dfn) {
         <a class="self-link" href="${href}">Permalink</a>
         ${dfnExportedMarker(dfn)}
       </div>
-      <h2>Referenced in:</h2>
+      <b>Referenced in:</b>
```
marcoscaceres, line 52: Should it be marked as an aria heading? It kinda serves as a landmark.

sidvishnoi: Ideally yes, but I think it's a matter of time pubrules checker starts treating them same as h2.
Note to self: `<b role="heading" aria-level="2">Referenced in:</b>`

PR #2955, opened 2020-06-08
feat(linter-rules/no-unused-vars): declared var must be used

sidvishnoi:
Based on feedback from https://github.com/tabatkins/bikeshed/issues/655#issuecomment-638139916

Enabled by default during development only.


src/core/linter-rules/no-unused-vars.js
```diff
@@ -0,0 +1,66 @@
...
+ * Checks that an variable is used if declared (the first use is treated as
+ * declaration).
+ */
+import LinterRule from "../LinterRule.js";
+import { lang as defaultLang } from "../l10n.js";
+import { norm } from "../utils.js";
+
+const name = "no-unused-vars";
+
+const meta = {
```
marcoscaceres, line 14: Make sure the message here looks like one you would get from ESLint... that is, it should be clear that a "variable X was defined but never used".

sidvishnoi: We won't have access to "X" in linter message, so maybe:
"Variable was defined, but never used. Add a `data-ignore-unused` attribute to the `<var>`, or consider using `<em>`/ `<code>`"

src/core/linter-rules/no-unused-vars.js
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * @param {*} _
+ * @param {Document} doc
+ * @return {import("../LinterRule").LinterResult}
+ */
+function linterFunction(_, doc) {
+  const offendingElements = [];
+
+  for (const section of doc.querySelectorAll("section")) {
```
sidvishnoi, line 33: Should we pick only those sections that contain algorithms?
See: https://github.com/w3c/payment-request/blob/e9515622/index.html#L280

src/core/linter-rules/no-unused-vars.js
marcoscaceres: >  "Add a data-ignore-unused attribute to the <var>, or consider using <em>/ <code>"

Only the "data-ignore-unused" seems like an good suggestion... the other not so much as they are working around the problem, not addressing the problem explicitly.

marcoscaceres: No idea what's going on with my comment's formatting above... 😂

src/core/linter-rules/no-unused-vars.js
marcoscaceres: I think so... we might need to test out a few specs.

src/core/linter-rules/no-unused-vars.js
sidvishnoi: > Only the "data-ignore-unused" seems like an good suggestion... the other not so much as they are working around the problem, not addressing the problem explicitly.

Maybe author used `<var>` for its style when they meant `<em>` or `<code>` (See https://github.com/w3c/webdriver/pull/1515/files)?

> No idea what's going on with my comment's formatting above... :joy:

Missed backticks on `<em>` :balloon:

sidvishnoi: Also, we can't add a `data-ignore-unused` to the var micro-syntax.

marcoscaceres: Using `<var>` for style is an author error, so let’s not worry about it (or encourage that).

src/core/linter-rules/no-unused-vars.js
```diff
@@ -0,0 +1,84 @@
...
+  /**
+   * Check if a <section> contains a `".algorithm"`
+   *
+   * The selector matches:
+   * ``` html
+   * <section><ul class="algorithm"></ul></section>
+   * ```
+   * The selector does not match:
+   * ``` html
+   * <section><div><ul class="algorithm"></ul></div></section>
```
sidvishnoi, line 41: Should this be supported? Seems unlikely, but possible

src/core/linter-rules/no-unused-vars.js
```diff
@@ -0,0 +1,85 @@
...
```
marcoscaceres, line 41: What’s the rationale for this one?

src/core/linter-rules/no-unused-vars.js
marcoscaceres: Heh, commented at the same time 😂 yes, I think we should support it

src/core/linter-rules/no-unused-vars.js
sidvishnoi: The selector gets complex (match section only if section contains algorithm, but not any subsection). Looking at specs, most `.algorithm` were in directly section, not wrapped in other elements. Might go the imperative way if selector gets too complex and we want to support it.

PR #2957, opened 2020-06-12
feat(respecDocWriter): use locally installed respec version

sidvishnoi:
Closes https://github.com/w3c/respec/issues/2874 

``` bash
# use the same version as in document
respec2html --src examples/basic.built.html -o /dev/null

# use the local version instead 🎉 
respec2html --src examples/basic.built.html -o /dev/null --use-local
```

### DIFF
--- a/tests/headless.js
+++ b/tests/headless.js
@@ -27,11 +27,22 @@ describe("Headless (examples)", () => {
     );
     await expectAsync(exe.run()).toBeResolved();
   });
+
+  it("uses local respec version with --use-local", async () => {
+    const exe = toExecutable(
+      toCommand(`http://localhost:${port}/examples/basic.built.html`, {
+        useLocal: true,
+      })
+    );
+    const logs = await exe.run();
+    expect(logs).toContain("Intercepted");
+  });
 });
 
-function toCommand(src) {
+function toCommand(src, { useLocal = false } = {}) {
   const command = `node ./tools/respec2html.js ${src}`;
-  const options = ["-e", `--timeout ${timeout}`];
+  const options = ["-e", `--timeout ${timeout}`, "--verbose"];
+  if (useLocal) options.push("--use-local");
   return `${command} ${options.join(" ")}`;
 }
 
@@ -41,15 +52,14 @@ function toExecutable(cmd) {
       return cmd;
     },
     run() {
+      const env = { ...process.env, FORCE_COLOR: "0" };
       return new Promise((resolve, reject) => {
-        const childProcess = exec(cmd, (err, data) => {
+        exec(cmd, { env }, (err, _stdout, stderr) => {
           if (err) {
             return reject(err);
           }
-          resolve(data);
+          resolve(stderr);
         });
-        childProcess.stdout.pipe(process.stdout);
-        childProcess.stderr.pipe(process.stderr);
       });
     },
   };
--- a/tools/respecDocWriter.js
+++ b/tools/respecDocWriter.js
@@ -3,7 +3,7 @@
  */
 const puppeteer = require("puppeteer");
 const path = require("path");
-const { mkdtemp, writeFile } = require("fs").promises;
+const { mkdtemp, readFile, writeFile } = require("fs").promises;
 const { tmpdir } = require("os");
 
 const noop = () => {};
@@ -13,6 +13,7 @@ const noop = () => {};
  * @param {string} src A URL or filepath that is the ReSpec source.
  * @param {object} [options]
  * @param {number} [options.timeout] Milliseconds before processing should timeout.
+ * @param {boolean} [options.useLocal] Use locally installed ReSpec instead of the one in document.
  * @param {(error: RsError) => void} [options.onError] What to do if a ReSpec processing has an error. Does nothing by default.
  * @param {(warning: RsError) => void} [options.onWarning] What to do if a ReSpec processing has a warning. Does nothing by default.
  * @param {(msg: string, timeRemaining: number) => void} [options.onProgress]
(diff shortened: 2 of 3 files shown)


tools/respecDocWriter.js
```diff
@@ -60,7 +60,12 @@ async function fetchAndWrite(
   src,
   out,
   whenToHalt,
-  { timeout = 300000, disableSandbox = false, debug = false } = {}
+  {
+    timeout = 300000,
+    disableSandbox = false,
+    debug = false,
+    useLocal = false,
```
sidvishnoi, line 67: Should this be a string/boolean instead of just boolean? Having a string could allow using a custom URL (say, from unpkg) instead of using local one.

tests/headless.js
```diff
@@ -33,23 +31,30 @@ function toExecutable(cmd) {
...
-  const executables = testURLs.map(url => {
-    const nullDevice =
-      process.platform === "win32" ? "\\\\.\\NUL" : "/dev/null";
-    const disableSandbox = process.env.TRAVIS ? " --disable-sandbox" : "";
-    const cmd = `node ./tools/respec2html.js -e${disableSandbox} --timeout 30 --src ${url} --out ${nullDevice}`;
-    return toExecutable(cmd);
-  });
+
   let testCount = 1;
   for (const exe of executables) {
```
saschanaz, line 55:
```suggestion
  for (const [i, exe] of executables.entries()) {
```

This may remove `testCount++`.

tools/respecDocWriter.js
```diff
@@ -116,6 +125,67 @@ async function fetchAndWrite(
...
+  switch (host) {
+    case "www.w3.org":
+      return (
+        path.startsWith("/Tools/respec/") && !path.includes("respec-highlight")
+      );
+    case "w3c.github.io":
+      return path.startsWith("/respec/builds/");
+    default:
+      // localhost, file://, and everything else
+      return /\/builds\/respec-[\w-]+\.js$/.test(path);
```
saschanaz, line 185: Why not just `respec-[\w-]+\.js`?

sidvishnoi: Could also match respec-worker, respec-highlight

saschanaz: Yes, I mean why `builds/`?

sidvishnoi: We're only replacing `respec-[profile]` here, which is in `builds/`. Let me try if we can do without `/builds/`

sidvishnoi: Adding `/builds/` reduces risk of false-positive though. A mismatch here means crash. Say, some adds their config as `respec-config.js` (haven't seen cases, but possible). `builds/respec-config.js` is less likely.

saschanaz: Is this story possible?: We catch `respec-config.js` or whatever we don't support, but it's replacement is not available, so we just passthrough in this case.

sidvishnoi: Due to https://github.com/puppeteer/puppeteer/issues/4208, we need to disable request interception as soon as possible (otherrwise worker script crashes). So, I'm hesitant to try to match more scripts. Ideally, I would like to match respec-worker and respec-highlight also (and probably support an entirely offline experience)

saschanaz: Hmm. Will migration to playwright help? 🤔

tools/respecDocWriter.js
marcoscaceres: I don't have a strong opinion. Whatever you think is best... or leave it for now, and add the custom URL when we need it.

tools/respecDocWriter.js
sidvishnoi: So I played with playwright a bit. Although I didn't check if this problem exists there, playwright has its own problems. 

Playwright not supporting using existing browser binaries is a major blocker in my opinion. One can [install all/some binaries in a folder and then re-use them](https://github.com/microsoft/playwright/blob/master/docs/installation.md), but it doesn't let me reuse the pre-installed browsers. Puppeteer allows it, and we do use it in github-actions (I also use it locally).

saschanaz: It may not be a blocker though:

1. The download already happens for Firefox run, and we already are caching the dependencies via `actions/cache`.
1. Looking at https://github.com/microsoft/playwright/issues/2623#issuecomment-646051077, it seems there is `executablePath` to use a custom path, although I didn't find a way to skip the download.

sidvishnoi: Download from cache is kinda slow though. Given we already have a browser installed, it'll definitely feel slower to not use it.

PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD=1 should let us skip download though.

Nice find on passing executable path. Can set it with env variables.

I'll play more with it, if we do want to go with playwright. Though, I'm okay with puppeteer also.

sidvishnoi: It's https://github.com/microsoft/playwright/issues/5952 vs https://github.com/puppeteer/puppeteer/issues/4208 now.

sidvishnoi: @saschanaz I think intercepting only `builds/respec-{profile}` should be good enough for now?

saschanaz: Yeah, it seem it won't be fixed anytime soon.

marcoscaceres (approved):

tools/respecDocWriter.js
rwoll: @saschanaz importScripts from workers should now be fixed on vNext/master of playwright. When v1.11.0 comes out, feel free to give it a try and raise an issue if you encounter problems. 😄 

NB: This fix should simply prevent the importScripts from hanging; I'm not sure of the general story for request interception of SW traffic.

PR #2965, opened 2020-06-15
fix(ui/save-html): use new epub service

sidvishnoi:
Thanks to @iherman at https://github.com/iherman/respec2epub/issues/30#issuecomment-644169851!


iherman:
Thank you guys.

Issue #1562, opened 2018-03-07
Some way to drop "Latest published version" link for specs not published yet


tidoust:
ReSpec automatically generates the "Latest Published Version" link, which is obviously great. That said, that link does not exist before a spec gets published as First Public Working Draft, leading to a broken link in first versions of the Editor's Draft. In some cases, groups may work on Editor's Drafts that are initial proposals without concrete plans to publish the spec in the "/TR/" space.

That is not a major issue, for sure, but I haven't seen any way to prevent ReSpec from generating that link (save by using the "unofficial" spec status) and I wonder if something like ```"publishedURI": null``` could be useful. At the same time, if there is a config setting, it's possible that people may forget to remove it after publication...

sandhawke:
This is a major issue for me.   Suggested workaround?

marcoscaceres:
@sandhawke, can you help me understand a little more which of your specs it's affecting? I'm assuming you also have a "ED" before a FPWD?

sandhawke:
Yeah, specifically it's in a CG.   So even after FPWD it'll never have a /TR URL.

Example: https://credweb.org/report/snapshot

So, really there are two orthogonal problems here: ED before FPWD, and CG-ED, neither of which are supported by respec.  I've dealt with ED-before-FPWD enough over the years (dozens of times before respec came along) that I don't want to just ignore that one, but it's true the hack you see me using on that page (shortname = "not-published-as-TR") probably does suffice for that situation.  It's less good for the CG-ED, which might run a year or more before perhaps eventually becoming a WG's FPWD and getting a real shortname.

marcoscaceres:
>  ED before FPWD, 

Ok, we should come up with something clever here.... like "ED" without a `previousStatus` drops the "Latest published version". 

I agree we need a CG-ED status. That makes sense. But we might need to take that TR-Design as it might require additional image assets.  

> It's less good for the CG-ED, which might run a year or more before perhaps eventually becoming a WG's FPWD and getting a real shortname.

Agree.

marcoscaceres:
> I agree we need a CG-ED status.

Actually, we have CG-DRAFT.

sandhawke:
That's different. That's like a WD.  It's a publication approved by the CG as a non-final Report.  For example. the CG-DRAFT form of the CG-ED I linked above is: https://credweb.org/report/20181011

marcoscaceres:
[...]

PR #2968, opened 2020-06-16
fix(w3c/headers): allow skipping latest published version in ED

sidvishnoi:
Fixes https://github.com/w3c/respec/issues/1562

To skip adding a "Latest published version" link:
``` js
var respecConfig = {
  specStatus: "ED",
  latestVersion: null,
};
```

### DIFF
--- a/src/w3c/templates/headers.js
+++ b/src/w3c/templates/headers.js
@@ -129,14 +129,12 @@ export default (conf, options) => {
                 >${conf.thisVersion}</a
               >
             </dd>
-            <dt>${l10n.latest_published_version}</dt>
-            <dd>
-              ${conf.latestVersion
-                ? html`<a href="${conf.latestVersion}"
-                    >${conf.latestVersion}</a
-                  >`
-                : "none"}
-            </dd>
+            ${conf.latestVersion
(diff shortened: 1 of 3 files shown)


marcoscaceres:
hmm... looking at this... maybe we should just allow `conf.latestVersion` to be set to `null`?

sidvishnoi:
Agree with above. Though note that `conf.latestVersion` is our internal thing

src/w3c/templates/headers.js
```diff
@@ -129,14 +129,12 @@ export default (conf, options) => {
...
-                    >${conf.latestVersion}</a
-                  >`
-                : "none"}
-            </dd>
+            ${conf.latestVersion
+              ? html`<dt>${l10n.latest_published_version}</dt>
+                  <dd>
+                    <a href="${conf.latestVersion}">${conf.latestVersion}</a>
+                  </dd>`
+              : ""}
```
sidvishnoi, line 137: Should we keep the old behavior of showing "none" as latest version or skip it altogether?

marcoscaceres: Oh, good point! Yeah, “none” is good.

PR #2989, opened 2020-06-30
feat(core/markdown): treat markdown code snippets as 'example'

sidvishnoi:
``` markdown
    ``` js
    console.log("hey!!");
    ```
```
becomes:
``` html
<pre class="example js">
console.log("hey!!");
</pre>
```

This might be a breaking change (nested `.example` if people already added their own `pre.example`).
A safer way might be to use DOM manipulation to convert `<pre><code class="lang">` to `<pre class="example lang"><code>`

### DIFF
--- a/src/core/markdown.js
+++ b/src/core/markdown.js
@@ -62,8 +62,10 @@ class Renderer extends marked.Renderer {
     }
 
     const html = super.code(code, language, isEscaped);
+
     const { example, illegalExample } = metaData;
     if (!example && !illegalExample) return html;
+
     const title = example || illegalExample;
     const className = `${language} ${example ? "example" : "illegal-example"}`;
     return html.replace(
--- a/src/core/markdown.js
+++ b/src/core/markdown.js
@@ -68,10 +68,7 @@ class Renderer extends marked.Renderer {
 
     const title = example || illegalExample;
     const className = `${language} ${example ? "example" : "illegal-example"}`;
-    return html.replace(
-      /^<pre[^>]*>/,
-      `<pre title="${title}" class="${className}">`
-    );
+    return html.replace("<pre>", `<pre title="${title}" class="${className}">`);
   }
 
   /**
--- a/src/core/markdown.js
+++ b/src/core/markdown.js
@@ -54,7 +54,7 @@ const ampEntity = /&amp;/gm;
 
 class Renderer extends marked.Renderer {
   code(code, infoString, isEscaped) {
-    const { language, example } = Renderer.parseInfoString(infoString);
+    const { language, ...metaData } = Renderer.parseInfoString(infoString);
 
     // regex to check whether the language is webidl
     if (/(^webidl$)/i.test(language)) {
@@ -62,10 +62,13 @@ class Renderer extends marked.Renderer {
     }
 
     const html = super.code(code, language, isEscaped);
-    if (!example) return html;
+    const { example, illegalExample } = metaData;
+    if (!example && !illegalExample) return html;
+    const title = example || illegalExample;
+    const className = `${language} ${example ? "example" : "illegal-example"}`;
     return html.replace(
       /^<pre[^>]*>/,
-      `<pre title="${example}" class="${language} example">`
+      `<pre title="${title}" class="${className}">`
     );
   }
 
--- a/src/core/markdown.js
+++ b/src/core/markdown.js
@@ -53,35 +53,43 @@ const gtEntity = /&gt;/gm;
 const ampEntity = /&amp;/gm;
 
 class Renderer extends marked.Renderer {
-  code(code, language, isEscaped) {
+  code(code, infoString, isEscaped) {
+    const { language, example } = Renderer.parseInfoString(infoString);
+
     // regex to check whether the language is webidl
     if (/(^webidl$)/i.test(language)) {
       return `<pre class="idl">${code}</pre>`;
     }
 
-    // Extract the first comment containing "@example" as title, and treat this
-    // code block as an "example" if such comment exists.
-    const firstNonEmptyLine = code.split("\n", 3).find(s => s.trim());
-    const commentRegex = /^(\/\/|\/\*|<!--|#)\s*@example:?/;
-    if (firstNonEmptyLine && commentRegex.test(firstNonEmptyLine)) {
-      code = code.replace(`${firstNonEmptyLine}\n`, "");
+    const html = super.code(code, language, isEscaped);
+    if (!example) return html;
+    return html.replace(
+      /^<pre[^>]*>/,
+      `<pre title="${example}" class="${language} example">`
+    );
+  }
 
-      const titleValue = firstNonEmptyLine
-        .replace(commentRegex, "")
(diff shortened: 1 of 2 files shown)


sidvishnoi:
The second commit makes:
``` markdown
    ``` js
    // i'm da title
    console.log("hey!!");
    ```
```
render as:
![image](https://user-images.githubusercontent.com/8426945/86172402-e3f26680-bb3b-11ea-8816-0302a91ff7e8.png)

marcoscaceres:
hmmm, I don't think we should do the title based on the first comment tbh.

sidvishnoi:
How about 
```markdown
    ``` js
    // @example: i'm example title
    console.log("hey!!");
    ```
```

This way, we can have: `<pre class="lang">`, `<pre class="example lang">`, and `<pre class="example lang" title="title">`

marcoscaceres:
Special key might be good. Good idea!

marcoscaceres:
@saschanaz would like your thoughts on this, if you have time.

saschanaz:
Hmm. Can we do this:

~~~
``` js example
hi();
```
~~~

... rather than adding it implicitly?

sidvishnoi:
Last time I checked this, marked didn't let change the lexer behavior on that. I'll try again.

Not sure about title part...
~~~
```js example title here
hello()
```
~~~
...maybe?

saschanaz:
> 
> 
> Last time I checked this, marked didn't let change the lexer behavior on that. I'll try again.
> 
> Not sure about title part...
> 
> ```
> ```js example title here
> hello()
> ```
> ```
> 
> ...maybe?

If the API allows accessing each "information text" maybe we can even do `js example title="title here"`.

sidvishnoi:
Just checked, marked now includes entire first line in "information text" (earlier it only picked first word). So, we can utilize that easily now.
Are you sure we want this way though? I kinda preferred having an `@example` comment as it'll render fine on GitHub and other places as a comment, while adding it to the backtick line doesn't render.

saschanaz:
I think it's more syntactic and less hacky, as CommonMark says [the "information text" can be used freely](https://spec.commonmark.org/0.29/#fenced-code-blocks).

>The first word of the info string is typically used to specify the language of the code sample, and rendered in the class attribute of the code tag. However, this spec does not mandate any particular treatment of the info string.

----

> it'll render fine on GitHub and other places as a comment, while adding it to the backtick line doesn't render.

If it's important to support non-ReSpec then yes I like `@example` too. I'm not sure though?

sidvishnoi:
Agree with above.

> If it's important to support non-ReSpec then yes I like @example too. I'm not sure though?

Lets call @marcoscaceres for a tie breaker!

sidvishnoi:
Note to future self: here's a regex if we go the info string route: https://regex101.com/r/vksZMb/1

saschanaz:
We could even allow inline JSON and have fun with its extensibility 😄

sidvishnoi:
> We could even allow inline JSON and have fun with its extensibility :smile:

Can't imagine the possibilities lol.

saschanaz:
Random thought:

~~~
```js example: "Example Title"
function myCode() {}
```
~~~

with the syntax `langname JSON_INNER` where `JSON_INNER` is parsed via ``JSON.parse(`{${json_inner}}`)``.

This way the syntax looks simple, the parsing code is straightforward, and we can keep backward compatibility when we ever want to add more features here.

marcoscaceres:
Let's go with `js "example": "title"`

sidvishnoi:
| input  | output | 
| -- | -- |
| image |  image |

src/core/markdown.js
```diff
@@ -53,12 +53,43 @@ const gtEntity = /&gt;/gm;
...
+    const { language, example } = Renderer.parseInfoString(infoString);
+
     // regex to check whether the language is webidl
     if (/(^webidl$)/i.test(language)) {
       return `<pre class="idl">${code}</pre>`;
     }
-    return super.code(code, language, isEscaped);
+
+    const html = super.code(code, language, isEscaped);
+    if (!example) return html;
```
sidvishnoi, line 65: TODO: also support `.illegal-example`

sidvishnoi:
@saschanaz this would be helpful with https://github.com/marcoscaceres/respec.org/pull/94. Got cycles for review?

src/core/markdown.js
```diff
@@ -53,12 +53,48 @@ const gtEntity = /&gt;/gm;
...
+
+    const html = super.code(code, language, isEscaped);
+
+    const { example, illegalExample } = metaData;
+    if (!example && !illegalExample) return html;
+
+    const title = example || illegalExample;
+    const className = `${language} ${example ? "example" : "illegal-example"}`;
+    return html.replace(
+      /^<pre[^>]*>/,
```
saschanaz, line 72: Is this always safe? In other words, is it guaranteed that markdown will never emit `>` inside pre? We know pre can include another HTML elements, so I want to make sure here.

sidvishnoi: The regex will break on cases like `<pre title="foo>">`, but work fine with `<pre title="foo"><code>` or `<pre title="foo">>` (stops at first `>`). As we're replacing content coming from marked, it should be mostly safe.

saschanaz: Could you add tests where pre includes:
1. > as a plain text
2. HTML tags

That would ensure this will be safe.

tests/spec/core/markdown-spec.js
```diff
@@ -497,6 +497,44 @@ function getAnswer() {
...
+
+      const example = pre.closest(".example");
+      expect(example).toBeFalsy();
+    });
+
+    it("treats as example if example metadata exists", async () => {
+      const body = `
+        <section data-format="markdown" id="test">
+        \`\`\`js "example": "the title"
+        console.log("hey!!");
```
sidvishnoi, line 521:
```suggestion
        \`\`\`html "example": "the title"
        <div>&lt;soup</div>
```

src/core/markdown.js
sidvishnoi: Added test. 
Though, even replacing `"<pre>"` would be fine instead of regex, but regex is safer for future changes.

tests/spec/core/markdown-spec.js
```diff
@@ -497,6 +497,44 @@ function getAnswer() {
...
+      `;
+      const ops = makeStandardOps(null, body);
+      const doc = await makeRSDoc(ops);
+
+      const example = doc.querySelector("#test .example");
+      expect(example).toBeTruthy();
+      const title = example.querySelector(".example-title");
+      expect(title).toBeTruthy();
+      expect(title.textContent).toContain("the title");
+      expect(example.querySelector("pre").classList).toContain("js");
```
sidvishnoi, line 533:
```suggestion
      expect(example.querySelector("pre").classList).toContain("html");
```

src/core/markdown.js
saschanaz: Hmm, how about just `<pre>` for now and do regex when we know there is a need? I don't think triple ticks are expected to have any attributes.

Issue #3005, opened 2020-07-09
HTML in markdown code snippets doesn't work


sidvishnoi:
<details>
<summary>Unescaped Input</summary>

```html
<!DOCTYPE html>
<html>
  <head>
    <meta charset='utf-8'>
    <title>Replace me with a real title</title>
    <script src='../profiles/w3c.js' async class='remove'></script>
    <script class='remove'>
      var respecConfig = {
        specStatus: "ED",
        shortName: "shortname-api",
        editors: [{ name: "Your Name" }],
        github: "https://github.com/w3c/some-API/",
        format: "markdown",
      };
    </script>
  </head>
  <body>

    ## test

    ```html
    <!DOCTYPE html>
    <html lang="en">
    <head>
      <meta charset="UTF-8">
      <meta name="viewport" content="width=device-width, initial-scale=1.0">
      <title>Document</title>
    </head>
    <body>
      <div class="issue" data-number="280"></div>
      <script>
        alert("whoa");
      </script>
    </body>
    </html>
    ```

    <section id='abstract'>
      <p>This is required.</p>
    </section>
    <section id='sotd'>
      <p>This is required.</p>
    </section>
  </body>
</html>
```

</details>

<details>
<summary>Output (HTML in snippet evaluated, including the script!)</summary>

![image](https://user-images.githubusercontent.com/8426945/87032305-08b4a100-c202-11ea-8d9d-f6c0ab1425a9.png)

</details>

And if I try to escape:

<details>
<summary>Escaped Input</summary>

```html
    ## test

    ```html
    &lt;!DOCTYPE html&gt;
    &lt;html lang="en"&gt;
    &lt;head&gt;
      &lt;meta charset="UTF-8"&gt;
      &lt;meta name="viewport" content="width=device-width, initial-scale=1.0"&gt;
      &lt;title&gt;Document&lt;/title&gt;
    &lt;/head&gt;
    &lt;body&gt;
      &lt;div class="issue" data-number="280"&gt;&lt;/div&gt;
      &lt;script&gt;
        alert("whoa");
      &lt;/script&gt;
    &lt;/body&gt;
    &lt;/html&gt;
    ```
```

</details>

<details>
<summary>Output</summary>

![image](https://user-images.githubusercontent.com/8426945/87032378-28e46000-c202-11ea-8081-1ebcbd28eb79.png)

</details>

Also, unindenting everything to left evaluates some part of input:
<details>

[...]

PR #3007, opened 2020-07-09
fix(core/data-include): support markdown code blocks with unescaped html

sidvishnoi:
Closes https://github.com/w3c/respec/issues/3005

https://respec.org/docs/src.html finally looks like good!

### DIFF
--- a/src/core/data-include.js
+++ b/src/core/data-include.js
@@ -9,20 +9,11 @@
 //  This module only really works when you are in an HTTP context, and will most likely
 //  fail if you are editing your documents on your local drive. That is due to security
 //  restrictions in the browser.
-import { getElementIndentation, runTransforms } from "./utils.js";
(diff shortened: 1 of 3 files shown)


sidvishnoi:
One issue now is, we can either use `data-include-format=markdown`, or `format: "markdown"`. Not sure why. Markdown processor breaks code snippets with multiple lines into paragraphs. For example:

~~~markdown
```bash
respec2html --src source.html --out index.html

# "respec2html" is an alias for "respec". You can also run:
respec --src source.html --out index.html
```
~~~
becomes:
```html
<pre><code class="bash hljs" aria-busy="false">respec2html --src source.html --out index.html


"respec2html" is an alias for "respec". You can also run:

</code><p><code class="bash">respec --src source.html --out index.html</code></p></pre>
```

PR #3019, opened 2020-07-16
fix(core/contrib): missing full stop at end of contrib element

marcoscaceres:


marcoscaceres:
Actually, no... we can use `<span id="gh-contributors"></span>.` That's much more flexible.

Issue #2889, opened 2020-05-20
Remove legacy code related to w3c-common profile


sidvishnoi:
- [ ] Remove code due to legacy support
- [ ] Fix highlightjs dynamic language loading - uses accessing worker with `require`
- [ ]  Fix tools/respecDocWriter.js - uses exporter with `require`
- [ ] Remove `require` support - do not `expose` anymore - BREAKING CHANGE

sidvishnoi:
Before we start deprecation warning, we should add a migration guide.
As we plan to remove `require(["some-respec-internal-thing"])`, we should figure out a strategy for extensions also (See https://github.com/w3c/respec/issues/1976), otherwise migration is like "check back later".

JohnRiv:
I'm not sure if https://github.com/w3c/respec/wiki/respec-w3c-common-migration-guide was ever created, but that page does not exist anymore

sidvishnoi:
Damn bots deleted page. Thanks for reporting. It's available again.
![image](https://user-images.githubusercontent.com/8426945/94987904-49b11d00-0587-11eb-99d6-34aef04a91a3.png)

carmacleod:
The migration guide is gone again, fyi.

sidvishnoi:
Thanks for reporting @carmacleod. Added it back again, and restricted wiki edit access for some time.

marcoscaceres:
The w3c-common deprecation tracked here is effectively done: the profile is frozen, the deprecation warning is in place, the migration guide exists, jQuery is removed, and legacy code was cleaned up (6 of 9 items checked).

The remaining unchecked items are tracked separately: extensions API (#1976), hljs dynamic loading (#4934), and removing `require` support (would need its own issue as a breaking change). Closing this umbrella tracker. Thanks @sidvishnoi for driving this.

PR #3023, opened 2020-07-16
chore: no longer build jQuery dependency

sidvishnoi:
Part of https://github.com/w3c/respec/issues/2889

- Well, one less dependency build.
- Keeping the jQuery in package.json only to keep w3c-common branch in sync, so we don't need to `npm install` when switching branch locally. Also helps in using same CI cache :smile:. Can remove if you prefer.


sidvishnoi:
Bonus: It reduced build time in GitHub Action and Netlify build by 1 second :smile: :shipit:

saschanaz:
I'm okay with it. Sad that we can't have a comment in package.json, though.

PR #3042, opened 2020-07-31
chore: add Chinese translation

xfq:


src/core/figures.js
```diff
@@ -37,8 +37,8 @@ const localizationStrings = {
     list_of_figures: "Lista de Figuras",
   },
   zh: {
-    fig: "圖 ",
-    list_of_figures: "List of Figures",
+    fig: "图 ",
```
xfq, line 40: Note that I changed this because the original text "圖" was `zh-hant` instead of `zh-hans`. In https://github.com/w3c/respec/blob/7a8e5f295020fd076cd84b84124624519aee28b8/src/core/utils.js#L322-L323 `zh-hans` is an alias of `zh`.

Issue #3035, opened 2020-07-26
Proposal: improving error reporting


sidvishnoi:
Create a custom error class:
``` js
// extending error gives us stack trace
class ReSpecError extends Error {
  // which module had error. can help in filtering errors.
  module: string;
  message: string;
  details?: string;

  // how to solve the error, link to docs etc.
  hint?: string;

  // offending elements
  elements?: Element[],
  // ... or offending respecConfig property
  config?: string,
  // and maybe more or less members if needed
}
```

Instead of `pub("error", message)`, do `pub("error", new ReSpecError(...))` to get a more structured and consistent error message. Then we can selectively show parts of error in UI and console.

Also, by `document.dispatchEvent(new CustomEvent('respecerror' | 'respecwarning', { detail: respecError }))` in `sub("error" | "warn")`, we can attach event listeners in respecDocWriter instead of watching for console messages. This way, we can better display errors/warnings with CI or spec-generator service.

Thoughts @saschanaz @marcoscaceres?

I played a bit with this earlier in [better-error-doc-writer](https://github.com/w3c/respec/compare/better-error-doc-writer) branch (also includes some nice-ish API changes to respecDocWriter).

marcoscaceres:
Looks like a good change.

sidvishnoi:
Moving `dispatchEvent` part to https://github.com/w3c/respec/issues/1084

PR #3062, opened 2020-08-13
refactor(*): make errors consistent

sidvishnoi:
Closes https://github.com/w3c/respec/issues/3035

Adds a `RespecError extends Error` class to make all errors consistent. `showError` and `showWarning` are improved versions of `showInlineError` and `showInlineWarning`.

**Edit:** following should help in review:
- Most relevant changes are in: `core/utils`, `core/ui` and `core/pubsubhub`.
- Then `respecDocWriter` adapts to new error format, while being backward compatible with old string-only format.
- Everything else then is replacing pub(error, msg) and showInlineError with showError (and showWarning), so viewing any one of other files should be enough.
- Side-by-side diff would be more helpful.


sidvishnoi:
## Demo (first run):

In console:
image

In UI:
image

src/core/utils.js
```diff
@@ -949,3 +950,54 @@ export class CaseInsensitiveMap extends Map {
...
   }
 }
+
+export class Err extends Error {
+  /**
+   * @param {string} message
+   * @param {string} plugin Name of plugin that caused the error.
+   * @param {object} [options]
+   * @param {boolean} [options.isWarning]
+   * @param {string} [options.title]
```
marcoscaceres, line 960: "title" is Kida weird... 

```suggestion
   * @param {string} [options.message]
```

sidvishnoi: `title` is what gets applied to offending elements. similar to what we had in `showInlineError`. We already have message as a required constructor parameter.

src/core/utils.js
```diff
@@ -949,3 +898,54 @@ export class CaseInsensitiveMap extends Map {
     return super.delete(key.toLowerCase());
   }
 }
+
+export class Err extends Error {
```
sidvishnoi, line 902: Consider other names: 
- `ReSpecError`: too long
- `RsError`?

src/core/issues-notes.js
```diff
@@ -154,7 +154,8 @@ function handleIssues(ins, ghIssues, conf) {
           title.classList.add("issue-number");
           ghIssue = ghIssues.get(dataNum);
           if (!ghIssue) {
-            pub("warning", `Failed to fetch issue number ${dataNum}`);
```
sidvishnoi, line 157: Yay!! A bug caught :partying_face:

src/core/pubsubhub.js
```diff
@@ -75,11 +75,12 @@ export function unsub({ topic, cb }) {
 }
 
 sub("error", err => {
-  console.error(err, err.stack);
+  console.error(err, Object.fromEntries(Object.entries(err)));
```
sidvishnoi, line 78: This is similar to doing a `console.error` followed by a `console.dir`, but within the same `console.error`.

src/core/utils.js
marcoscaceres: ok, makes sense... maybe worth a comment so I remember next time :)

src/core/utils.js
marcoscaceres: RsError would be better IMO

src/core/utils.js
```diff
@@ -949,3 +890,56 @@ export class CaseInsensitiveMap extends Map {
...
+  }
+
+  toHTML() {
+    const plugin = this.plugin ? `(${this.plugin}): ` : "";
+    const hint = this.hint ? ` ${this.hint}.` : "";
+    const elements = Array.isArray(this.elements)
+      ? ` Occurred at: ${joinAnd(this.elements.map(generateMarkdownLink))}.`
+      : "";
+    const details = this.details
+      ? `\n\n<details>\n${this.details}\n</details>\n`
```
marcoscaceres, line 931: Any reason why we don't html`<details>? here?

marcoscaceres: ah, never mind... we markdown it below

src/core/webidl.js
```diff
@@ -184,7 +180,7 @@ function defineIdlName(escaped, data, parent) {
     const styledName = data.type === "operation" ? `${name}()` : name;
     const ofParent = parentName ? ` \`${parentName}\`'s` : "";
     const msg = `Missing \`<dfn>\` for${ofParent} \`${styledName}\` ${data.type}. [More info](https://github.com/w3c/respec/wiki/WebIDL-thing-is-not-defined).`;
-    showInlineWarning(unlinkedAnchor, msg, "");
+    pub("warn", new RsError(msg, name, { elements: [unlinkedAnchor] }));
```
marcoscaceres, line 183: yay!

src/core/anchor-expander.js
```diff
@@ -45,7 +47,8 @@ export function run() {
       default: {
         a.textContent = a.getAttribute("href");
         const msg = "ReSpec doesn't support expanding this kind of reference.";
-        showInlineError(a, msg, `Can't expand "#${id}".`);
+        const title = `Can't expand "#${id}".`;
+        pub("error", new RsError(msg, name, { title }));
```
marcoscaceres, line 51: can we pass the `a`?

sidvishnoi: Can't believe I missed this one after 3 self-reviews!

src/core/webidl.js
sidvishnoi: There is a subtle bug here :smile: 
`name` (plugin name) is shadowed by `name` in this `defineIdlName` function

sidvishnoi:
This is a breaking change for respecDocWriter. Putting on hold with further breaking changes.

sidvishnoi:
This is how information is available to the Node.js API now:
image

And to the browser UI:
image

tools/respecDocWriter.js
```diff
@@ -271,13 +271,23 @@ async function evaluateHTML(version, timer) {
...
+        return JSON.stringify(err);
+      }
+    }, handle);
   }
 
   page.on("console", async message => {
     const args = await Promise.all(message.args().map(stringifyJSHandle));
     const msgText = message.text();
-    const text = args.filter(msg => msg !== "undefined").join(" ");
+    const text = args.filter(msg => msg !== "undefined")[0] || "";
```
sidvishnoi, line 290: Note to self: this needs some attention.

tools/respecDocWriter.js
```diff
@@ -271,13 +271,21 @@ async function evaluateHTML(version, timer) {
...
   /** @param {import('puppeteer').JSHandle<any>} handle */
   async function stringifyJSHandle(handle) {
-    return await handle.executionContext().evaluate(o => String(o), handle);
+    return await handle.executionContext().evaluate(obj => {
+      if (typeof obj === "string") {
+        // Old ReSpec versions might report errors as strings.
+        return JSON.stringify({ message: String(obj) });
+      } else {
+        // Ideally: `obj instanceof RsError` and `RsError instanceof Error`.
+        return JSON.stringify(obj);
```
sidvishnoi, line 280: Note to self: A possible serialization of `RsError.elements`:
```js
if (obj.elements && Array.isArray(obj.elements)) {
  err.elements = obj.elements.map(
    /** @param {HTMLElement} el */ el => {
      el.classList.remove("respec-offending-element");
      if (el.id.includes("respec-offender")) el.id = "";
      const className = el.classList.length ? `.${el.classList}` : "";
      const id = el.id ? `#${el.id}` : "";
      const text = `[text="${el.textContent.replace(/"/g, '\\"')}"]`;
      return `${el.localName}${className}${id}${text}`;
    }
  );
}
```

saschanaz:
This is huuuuge, could this be split somehow? 👀

sidvishnoi:
Not sure if possible to split, as it'll leave repo in inconsistent state, but following should help in review:

- Most relevant changes are in: `core/utils`, `core/ui` and `core/pubsubhub`.
- Then `respecDocWriter` adapts to new error format, while being backward compatible with old string-only format.
- Everything else then is replacing pub(error, msg) and showInlineError with showError (and showWarning), so viewing any one of other files should be enough.

(Should've added this note first, sorry!)

marcoscaceres (approved):

PR #3065, opened 2020-08-14
chore(ui/dfn-list): remove ui/dfn-list in favor of id=index

sidvishnoi:
As [`<section id="index">`](https://respec.org/docs/#id-index) provides same (or even better) functionality.


sidvishnoi:
It's a breaking change. Lets put it on hold to release with other upcoming breaking changes, if any.

marcoscaceres:
This seems pretty harmless, to be honest. I'd be ok with shipping this whenever. We might not even need to mark it as a breaking change, it's... kind of a feature (that we removed :P).

PR #3067, opened 2020-08-16
chore(ci): use problemMatcher for karma jasmine

sidvishnoi:
Came across https://github.com/actions/toolkit/blob/master/docs/problem-matchers.md
We might not need to maintain a hack.


sidvishnoi:
It's matching output twice - first in "live run", then in failure summary. We can change the reporter maybe. Or maybe, not worth it :thinking:

sidvishnoi:
This works in VS Code though :smile:

sidvishnoi:
Closing it GitHub actions don't have good enough `problemMatcher` support yet.

PR #3082, opened 2020-08-19
feat(core/respec): document.respec gives respecIsReady, ReSpec version

sidvishnoi:
Part of https://github.com/w3c/respec/issues/3089


src/core/respec.js
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * This module adds a `respec` object to the `document` with the following
+ * readonly properties:
+ *  - version: returns version of ReSpec Script.
+ *  - ready: an alias to `document.respecIsReady`.
+ */
+
+export const name = "core/respec";
+
+class ReSpec {
```
sidvishnoi, line 11: Creating a class here definitely looks weird, but it's due to next PR (`class ReSpec extends EventTarget`)

src/core/respec.js
```diff
@@ -0,0 +1,24 @@
...
+  }
+
+  get ready() {
+    return document.respecIsReady;
+  }
+}
+
+export function init() {
+  const respec = new ReSpec();
+  Object.defineProperty(document, "respec", { get: () => respec });
```
saschanaz, line 23: How about:

```suggestion
  Object.defineProperty(document, "respec", { value: respec });
```

This way it will be a readonly property.

src/type-helper.d.ts
```diff
@@ -55,6 +55,10 @@ interface Window {
 
 interface Document {
   respecIsReady: Promise<void>;
```
saschanaz, line 57: I think this returns the config object, so maybe this should be also changed.

src/core/respec.js
sidvishnoi: TIL!

src/type-helper.d.ts
sidvishnoi: `respecIsReady` does resolve with config. I was thinking if we should resolve `respec.ready` with nothing. Any use cases to get final config from `ready`?
cc @marcoscaceres

src/core/respec.js
```diff
@@ -0,0 +1,24 @@
+// @ts-check
+/**
+ * This module adds a `respec` object to the `document` with the following
+ * readonly properties:
+ *  - version: returns version of ReSpec Script.
+ *  - ready: an alias to `document.respecIsReady`.
+ */
+
+export const name = "core/respec";
```
sidvishnoi, line 9: Should we call it something else? Like "core/respec-global"?

sidvishnoi: Or maybe even `core/document.respec`

marcoscaceres: "core/respec-global" sounds good

src/type-helper.d.ts
marcoscaceres: I would keep them consistent, tbh. There is no real use case... but we should choose one to deprecate. Having two is confusing... not sure why that happened.

marcoscaceres (approved):
Looks fine... leave the rename and promise thing to you.

src/type-helper.d.ts
sidvishnoi: [Searched GitHub for `document.respecIsReady`](https://github.com/search?q=document.respecIsReady&type=Code) and found no use of resolved `respecConfig`. So, will be removing it (i.e., `Promise<respecConfig> -> Promise<void>`) in another PR (will still mark it as breaking change).
Will also be deprecating `document.respecIsReady` next, as mentioned in PR description.

PR #3091, opened 2020-08-26
feat(core/exporter): prettify HTML output

sidvishnoi:
Though it was something nice to have while working with diffing part of an [auto-deploy GitHub action](https://github.com/sidvishnoi/w3c-deploy) ([this precisely](https://github.com/sidvishnoi/w3c-deploy-test/runs/1030986115?check_suite_focus=true#step:3:82), or [as image](https://user-images.githubusercontent.com/8426945/91339568-48f0d280-e7f4-11ea-9f40-086354595821.png))

| Before | After |
| -- | -- |
| image | image |


tools/builder.js
```diff
@@ -19,7 +19,7 @@ colors.setTheme({
  * @param {RegExp[]} opts.include
  */
 function string(opts) {
-  const minifier = new CleanCSS();
+  const minifier = new CleanCSS({ format: "keep-breaks" });
```
sidvishnoi, line 22: This will increase the bundle size slightly (more newlines), but might be worth it.

src/core/exporter.js
```diff
@@ -92,4 +93,17 @@ function cleanup(cloneDoc) {
...
+    el.innerHTML = `\n${el.innerHTML}\n`;
+  });
+  cloneDoc.querySelectorAll("head > *").forEach(el => {
+    el.outerHTML = `\n${el.outerHTML}`;
+  });
+  cloneDoc.querySelectorAll(":root > *").forEach(el => {
+    el.outerHTML = `\n${el.outerHTML}\n`;
+  });
+}
+
```
sidvishnoi, line 108: This might be delegated to `tidy` instead and make it opt-in, or keep it pretty by default. WDYT?

marcoscaceres: Pretty by default is good IMO.

marcoscaceres (approved):
I wouldn't object if the tree was outputted with nesting of 2 spaces for the DOM tree. It would make the source a little bigger of course, but will be nicer if anyone wants to actually view source on these documents.

sidvishnoi:
> I wouldn't object if the tree was outputted with nesting of 2 spaces for the DOM tree.

Maybe some other time. Need to take block/inline elements into account, otherwise it becomes a balloon.
I wouldn't like to integrate prettier in ReSpec :laughing:

PR #3096, opened 2020-08-31
feat(w3c):  2020 process + patent policy

dontcallmedom:
This pull request prepares ReSpec for transition to Process 2020, with:
* support for the two new types of CR (CR Snapshot and CR Draft) with their respective SOTD updates; this introduces a new boolean configuration option `lsMode` (for "living standard mode") which sets how the request for input is formulated in the SOTD
* support for updateable Rec (with a new boolean config option `updateableRec`) which adapts the SOTD to highlight whether a given Rec is allowed to be updated or not; it also opens the way for Recs with proposed additions and corrections (set with the array config option `revisionTypes` which can take "addition" and/or "correction" as values), in which case a `revisedRecEnd` date is expected to be configured as well
* other updates to SOTD / doc head (in particular, requiring github issues to be linked from the head of the document)

I think this includes all the changes that I know are needed. I'm assuming it is not ready yet for merging though so am looking for feedback on what would be needed to do so.

In any case, this should not be merged before September 15 (the date at which the new process is coming in operation).


sidvishnoi:
@dontcallmedom Thanks for the PR!
I can't do a detailed review right now, but you can run `npm run lint -- --fix` to autofix (most of) the linting issues.

src/w3c/headers.js
```diff
@@ -301,6 +305,26 @@ export function run(conf) {
...
+    !(
+      conf.otherLinks &&
+      conf.otherLinks.find(linkGroup =>
+        linkGroup.data.find(
+          l =>
+            l.href &&
+            l.href.toString().match(/^https:\/\/github\.com\/.*\/issues/)
+        )
+      )
+    )
```
sidvishnoi, line 321: Can we extract this into a separate function `hasGitHubIssuesLink(conf)` or something similar?

src/w3c/headers.js
```diff
@@ -301,6 +305,13 @@ export function run(conf) {
   conf.anOrA = precededByAn.includes(conf.specStatus) ? "an" : "a";
   conf.isTagFinding =
     conf.specStatus === "finding" || conf.specStatus === "draft-finding";
+
+  if (conf.isRecTrack && !hasGitHubIssuesLink(conf)) {
+    pub(
+      "error",
+      "Rec-track documents needs to link to github issues from their head; consider setting config option `github`"
```
marcoscaceres, line 312:
```suggestion
      "Rec-track documents must link to Github issues from their head. Please use the [`github`](https://respec.org/docs/#github) configuration option."
```

src/w3c/headers.js
```diff
@@ -721,3 +779,17 @@ function normalizeOrcid(orcid) {
...
+    conf.github ||
+    (conf.otherLinks &&
+      conf.otherLinks.find(linkGroup =>
+        linkGroup.data.find(
+          l =>
+            l.href &&
+            l.href.toString().match(/^https:\/\/github\.com\/.*\/issues/)
+        )
+      ))
+  );
```
marcoscaceres, line 794: I think we should exclusively key off `conf.github`. Using other links this way should be considered deprecated.

src/w3c/templates/sotd.js
```diff
@@ -177,33 +119,110 @@ function renderIsNoTrack(conf, opts) {
...
+    statusExplanation = html`A Candidate Recommendation Snapshot has received
+      <a href="https://www.w3.org/2020/Process-20200915/#dfn-wide-review"
+        >wide review</a
+      >
+      and is intended to gather
+      <a href="${conf.implementationReportURI}">implementation experience</a>.`;
+    updatePolicy = "";
+    if (conf.lsMode) {
+      reviewPolicy = `Comments are welcome at any time but most especially before ${conf.humanCREnd}.`;
+    } else {
```
marcoscaceres, line 144: Maybe want to check `conf.humanCREnd` here? 

```suggestion
    } else if(conf.humanCREnd) {
```

src/w3c/templates/sotd.js
```diff
@@ -177,33 +119,110 @@ function renderIsNoTrack(conf, opts) {
...
+    <p>${reviewPolicy}</p>`;
 }
 
-function renderIsRec() {
+function renderIsRec({
+  updateableRec,
+  revisionTypes = [],
+  humanRevisedRecEnd,
+}) {
+  let reviewTarget;
```
marcoscaceres, line 185:
```suggestion
  let reviewTarget = "";
```

src/w3c/templates/sotd.js
```diff
@@ -338,9 +362,31 @@ function linkToWorkingGroup(conf) {
   if (!conf.wg) {
     return;
   }
+  let proposedChanges = "";
```
marcoscaceres, line 365: I think `null` should be ok here... hyperHTML should then remove it.  

```suggestion
  let proposedChanges = null;
```

marcoscaceres (commented):
Some initial feedback. Thanks for doing this @dontcallmedom!

src/w3c/headers.js
dontcallmedom: I think deprecating using other links for github issues sounds like a good idea, but I don't think it needs to be done at the same time as the already potentially disruptive change to process 2020.

In other words, if a spec already complies with that part of pubrules using `otherLinks` rather than `github`, I don't see any reason this should be treated as an error. Having a deprecation plan for `otherLinks` referring to github issues should be handled separately (and likely through a warning rather than an error)

src/w3c/templates/sotd.js
dontcallmedom: I'm not sure I understand - `humanCREnd` is needed in both conditions, and its absence is reported as an error in `w3c/headers.js`. What would this additional if brings?

dontcallmedom:
How should we approach the potential release of this for September 15 (next Tuesday)?

sidvishnoi:
@marcoscaceres Should we backport this to w3c-common profile, or deprecate it for real this time?

src/w3c/headers.js
marcoscaceres: Yeah, I agree. Will do that as a followup.

src/w3c/templates/sotd.js
marcoscaceres: Ah, I might have misunderstood.

marcoscaceres:
> How should we approach the potential release of this for September 15 (next Tuesday)?

We can release as soon as it's merged. Seems like this should be good to go.

marcoscaceres (approved):
Looks good to me.

src/w3c/headers.js
```diff
@@ -571,6 +592,43 @@ export function run(conf) {
...
+    conf.revisionTypes &&
+    conf.revisionTypes.length > 0
+  ) {
+    let unknown;
+    if ((unknown = conf.revisionTypes.find(x => !revisionTypes.includes(x)))) {
+      pub(
+        "error",
+        `\`specStatus\` is "REC" with unknown revision type '${unknown}'`
+      );
+    }
```
sidvishnoi, line 608:
```suggestion
    const unknownRevisionType = conf.revisionTypes.find(x => !revisionTypes.includes(x));
    if (unknownRevisionType) {
      pub(
        "error",
        `\`specStatus\` is "REC" with unknown revision type '${unknownRevisionType}'`
      );
    }
```

sidvishnoi:
We might need docs for some of the new features. Adding a checklist here:

- [ ] wgPatentPolicy
- [ ] isRecTrack
- [ ] revisionTypes
- [ ] updateableRec
- [ ] revisedRecEnd
- [ ] lsMode

sidvishnoi (approved):

marcoscaceres:
We should decide which of the above are private/internal and which WGs will use. I need to check.

PR #3115, opened 2020-09-15
fix(w3c/sotd): replace lsMode with pubMode

sidvishnoi:
In an afterthought of https://github.com/w3c/respec/pull/3096, we decided to replace `conf.lsMode` with `conf.pubMode === "LS"`, as the latter may allow more options in future without a separate configuration type.


dontcallmedom (approved):

Caglar1881:
👍

Issue #3129, opened 2020-09-22
ORCID icon not showing up when switching to respect-w3c


andrea-perego:
## Important info

- URL to affected spec: https://raw.githack.com/w3c/dxwg/andrea-perego-fix-respec-w3c-common-warning/dcat/index.html
- ReSpec version: 25.16.4 
- [x] I did a "hard refresh", but it's still busted.

## Description of problem

**What happened?**

We recently tried to switch the DCAT specification to `respec-w3c`, as per https://github.com/w3c/respec/issues/2889

After that, we saw that the ORCID SVG icon is not showing up for some of the editors.

You can check the differences by comparing these two versions:
- Version with `respec-w3c` - ORCID icon missing from some editors: https://raw.githack.com/w3c/dxwg/andrea-perego-fix-respec-w3c-common-warning/dcat/index.html
- Version with `respec-w3c-common` - ORCID icon shown correcly: https://w3c.github.io/dxwg/dcat/

The editor's data in `config.js` are all specified consistently, so this should not be the problem:

https://raw.githack.com/w3c/dxwg/andrea-perego-fix-respec-w3c-common-warning/dcat/config.js

The only differences between the editors is that the ones without the ORCID icon have fields `url` and `companyURL` set.

**Expected behavior (e.g., it shouldn't crash):**

ORCID icon showing up for all editors having an ORCID.

**Optional, steps to reproduce:**

1.  ...

sidvishnoi:
Regression from https://github.com/w3c/respec/pull/3063. Not sure why hyperHTML only rendered last instance. Filed https://github.com/WebReflection/hyperHTML/issues/401. Sent https://github.com/w3c/respec/pull/3131 for ReSpec

andrea-perego:
Thanks for addressing this issue, @sidvishnoi .

However, the ORCID icons are not showing up yet:

https://raw.githack.com/w3c/dxwg/andrea-perego-fix-respec-w3c-common-warning/dcat/index.html

I guess this is because no new built was created (the current one in GH is still v25.16.4).

Do you know when this will happen?

sidvishnoi:
Hey, I just released with this fix. Should be fine now :)

andrea-perego:
Yep. Confirmed. Thanks a lot, @sidvishnoi !

PR #3063, opened 2020-08-13
refactor(core/templates): extract common templates from profiles

sidvishnoi:
Part of https://github.com/w3c/respec/issues/2743

Moves `show-logo`, `show-people` and `show-link` templates from DINI and W3C profiles to core/templates. 
Also improves JSDoc definitions.

Less files to look at!! :tada:

PR #3131, opened 2020-09-22
fix(templates/show-people): ensure ORCID icon can render multiple times

sidvishnoi:
Fixes https://github.com/w3c/respec/issues/3129

Regression from https://github.com/w3c/respec/pull/3063. Not sure why hyperHTML only rendered last instance.
Filed https://github.com/WebReflection/hyperHTML/issues/401.


src/core/templates/show-people.js
```diff
@@ -114,7 +93,23 @@ export default function showPeople(persons = []) {
     }
     if (p.orcid) {
       contents.push(
-        html`<a class="p-name orcid" href="${p.orcid}">${orcidIcon} </a>`
+        html`<a class="p-name orcid" href="${p.orcid}"
```
saschanaz, line 96: I think we want to do `orcidIcon.cloneNode(true)` instead, as the previous code was moving `orcidIcon` around the parents.

sidvishnoi: Oh my! What would I do without you!

saschanaz (approved):
LGTM but do we need a test?

saschanaz (approved):
👍

PR #3133, opened 2020-09-23
chore(ci): run action only if relevant files change

sidvishnoi:
https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths

It'll run here, as the PR contains files matching the relevant patterns (`.github/workflows/**`). But when only changelog or other such files are modified in a PR or push to main branches, actions won't be triggered.


sidvishnoi:
Would've liked it more if it allowed running specific jobs based on pattern. On a second thought, might not be worth adding given extra lines of configuration.
At least it doesn't run tests when CHANGELOG or README are modified though :laughing:

marcoscaceres (approved):
Really useful!

Issue #3135, opened 2020-09-23
Issue titles in issue summary section are split between columns


andrea-perego:
## Important info

- URL to affected spec: https://raw.githack.com/w3c/dxwg/dcat-versioning/dcat/
- ReSpec version: 25.16.5
- [x] I did a "hard refresh", but it's still busted.

## Description of problem

**What happened?:**

In the spec above, part of the issue title in the issue summary section has been moved to the 2nd column, with a resulting effect which is quite confusing for readers, and looks like a rendering error.

To see it, open the link below and scroll up to the "Issue summary" section:

https://raw.githack.com/w3c/dxwg/dcat-versioning/dcat/#introduction

**Expected behavior (e.g., it shouldn't crash):**

Issue titles are not split between columns.

**Optional, steps to reproduce:**

1.  ...

sidvishnoi:
The problem seems to be gone for now.
IIRC, the problem was caused when there is a single issue in `issue-summary` (in a way, it's a rendering error: CSS multi-column is buggy).

sidvishnoi:
Okay. I see it. Adding `display: inline-block` to `#issue-summary li` should fix that. Will send a PR soon.

PR #3152, opened 2020-10-05
fix(assets/respec.css): prevent splitting title in issue-summary with single item

sidvishnoi:
Fixes https://github.com/w3c/respec/issues/3135

| Before | After |
| -- | -- |
| image | image |

Works fine with multiple issues also: image


saschanaz:
So are we workarounding a CSS implementation error? If then I prefer adding a comment possibly with links to browser bugs.

sidvishnoi:
Not sure if it's an implementation error. I couldn't find any relevant browser bugs.
Looks like this is how multi-column works. I'll read the spec to confirm.

marcoscaceres:
I think @sidvishnoi fix is correct, in as far as each issue should be treated as a self contained unit. What multi-col would be correct for a run of text. Having said that, the proper fix here would be for us to use CSS grid (for the same reason I mentioned... each "issue" is actually self contained block of information).

marcoscaceres (approved):
I can live with this... or better would be CSS Grid to solve this.

saschanaz:
If every browser is showing the same behavior I think we can just proceed.

sidvishnoi:
One difference with grid comes when titles have uneven length (multi-column behaves as masonry layout, grid keeps equal height). Not sure which one we prefer.
| Grid | Multicol |
| -- | -- |
| image | image |

saschanaz:
I think multicol is better here, as we don't need the empty space.

andrea-perego:
Thanks, @sidvishnoi

Issue #3156, opened 2020-10-08
Provide shortcut to IDL fragment


dontcallmedom:
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when I'm on the definition of an interface (attribute, methods, etc) (e.g. having followed a link from another spec) and I can't easily move to the IDL fragment where the said item is defined.

**Describe the solution you'd like**
An easily identifiable link from the said item back to where it is defined as an IDL fragment.

**Describe alternatives you've considered**
N/A

**Can you or your organization fund the work or help with development?**
I'd be happy to help implementing or reviewing if there is interest

**Additional context**
cc @tidoust since he expressed a similar interest in such a feature

sidvishnoi:
To confirm, are you looking for a shorter way of doing this:
![image](https://user-images.githubusercontent.com/8426945/95437749-008c0f00-0973-11eb-8d2a-c8d6f9e344e2.gif)

dontcallmedom:
not necessarily shorter (in terms of clicks), but much clearer. See e.g. https://w3c.github.io/webrtc-pc/webrtc.html#rtcrtpsender-interface - finding where to click to get to the IDL fragment takes way to long.

sidvishnoi:
Something like this might be good:
![image](https://user-images.githubusercontent.com/8426945/95441070-4f3ba800-0977-11eb-90a2-dd56b1505795.png)

PR #3157, opened 2020-10-08
feat(core/dfn-panel): add link to jump to IDL fragment

sidvishnoi:
Closes https://github.com/w3c/respec/issues/3156
Demo: https://deploy-preview-3157--respec-pr.netlify.app/

### DIFF
--- a/assets/dfn-panel.css
+++ b/assets/dfn-panel.css
@@ -55,7 +55,6 @@ dfn {
 
 .dfn-panel > div {
   display: flex;
-  gap: 0.5em;
 }
 
 .dfn-panel a.self-link {
@@ -65,6 +64,7 @@ dfn {
 
 .dfn-panel .marker {
   padding: 0.1em;
+  margin-left: 0.5em;
   border-radius: 0.2em;
   text-align: center;
   white-space: nowrap;
--- a/assets/dfn-panel.css
+++ b/assets/dfn-panel.css
@@ -77,8 +77,8 @@ dfn {
   box-shadow: 0 0 0 0.125em #1ca5f940;
 }
 .dfn-panel .marker.idl-block {
-  background: #f8ddaa;
-  box-shadow: 0 0 0 0.125em #c8870b4d;
+  background: #8ccbf2;
+  box-shadow: 0 0 0 0.125em #0670b161;
 }
 
 .dfn-panel a:not(:hover) {
--- a/assets/dfn-panel.css
+++ b/assets/dfn-panel.css
@@ -53,20 +53,32 @@ dfn {
   color: #333;
 }
 
-.dfn-panel a.self-link {
-  font-weight: bold;
+.dfn-panel > div {
+  display: flex;
+  gap: 0.5em;
 }
 
-.dfn-panel .dfn-exported {
-  float: right;
+.dfn-panel a.self-link {
+  font-weight: bold;
+  margin-right: auto;
+}
+
+.dfn-panel .marker {
   padding: 0.1em;
   border-radius: 0.2em;
   text-align: center;
   white-space: nowrap;
   font-size: 90%;
-  background: #d1edfd;
   color: #040b1c;
+}
+
+.dfn-panel .marker.dfn-exported {
+  background: #d1edfd;
   box-shadow: 0 0 0 0.125em #1ca5f940;
+}
+.dfn-panel .marker.idl-block {
+  background: #f8ddaa;
+  box-shadow: 0 0 0 0.125em #c8870b4d;
 }
 
 .dfn-panel a:not(:hover) {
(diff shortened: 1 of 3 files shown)


src/core/dfn-panel.js
```diff
@@ -60,12 +60,38 @@ function createPanel(dfn) {
...
+      idlFragmentId = parentIdlBlock.id;
+      break;
+    }
+  }
+
+  if (!idlFragmentId) return null;
+  return html`<a
+    href="#${idlFragmentId}"
+    class="marker idl-block"
+    title="Jump to defining IDL block"
```
sidvishnoi, line 90: "Defining IDL block" may not be exactly correct here.

sidvishnoi: Declaration?

dontcallmedom:
```suggestion
    title="Jump to IDL declaration"
```

src/core/dfn-panel.js
```diff
@@ -60,12 +60,34 @@ function createPanel(dfn) {
...
 
+/**
+ * @param {HTMLElement} dfn
+ * @param {NodeListOf<HTMLAnchorElement>} links
+ */
+function idlMarker(dfn, links) {
+  if (!dfn.hasAttribute("data-idl")) return null;
+
+  for (const anchor of links) {
+    const parentIdlBlock = anchor.closest("pre.idl");
```
dontcallmedom, line 77: I'm not sure exactly what `links` are, but this doesn't seem quite right; at least applying this to the WebRTC spec doesn't yield the right result for the `RTCRtpSender` interface for instance.

assets/dfn-panel.css
```diff
@@ -53,21 +53,33 @@ dfn {
...
+}
+
 .dfn-panel a.self-link {
   font-weight: bold;
+  margin-right: auto;
 }
 
-.dfn-panel .dfn-exported {
-  float: right;
+.dfn-panel .marker {
```
dontcallmedom, line 66: when the panel is at the top of the viewport, the IDL button is not clickable as the "ReSpec" command menu overlaps it (at least in my arguably quick & specific test); I'm not sure how to address this cleanly

src/core/dfn-panel.js
sidvishnoi: Maybe the `respec-pr` demo didn't replace the ReSpec script in WebRTC spec. Can you try replacing the ReSpec script with https://deploy-preview-3157--respec-pr.netlify.app/respec-w3c.js locally?

`links` here are anchor elements that reference this dfn BTW.

dontcallmedom: > doesn't yield the right result
more specifically it links to the `RTCPeerConnection` IDL declaration rather than `RTCRtpSender`

dontcallmedom: > Maybe the `respec-pr` demo didn't replace the ReSpec script in WebRTC spec. Can you try replacing the ReSpec script with https://deploy-preview-3157--respec-pr.netlify.app/respec-w3c.js locally?

For clarification, I'm using a local build of the PR and integrating it locally in the WebRTC spec - it's not a deployment problem as far as I can tell.

> links here are anchor elements that reference this dfn BTW.

Thanks - the problem is that any IDL fragment that uses the said interface as a type will be in that list of links and will match `closest("pre.idl")`; there is no reason that the first fragment will be the right one.

sidvishnoi: @dontcallmedom can you check again?

dontcallmedom: working great, now, thanks!

dontcallmedom (approved):

src/core/dfn-panel.js
```diff
@@ -60,12 +60,35 @@ function createPanel(dfn) {
...
+  for (const anchor of links) {
+    if (anchor.dataset.linkType !== dfn.dataset.dfnType) continue;
+    const parentIdlBlock = anchor.closest("pre.idl");
+    if (parentIdlBlock && parentIdlBlock.id) {
+      const href = `#${parentIdlBlock.id}`;
+      return html`<a
+        href="${href}"
+        class="marker idl-block"
+        title="Jump to IDL declaration"
+        >IDL{}</a
```
marcoscaceres, line 85: We probably don't need the "{}".

marcoscaceres (approved):
Kinda squishy... we might want to give some padding: 
Screen Shot 2020-10-09 at 3 26 12 pm

Otherwise, looks great.

marcoscaceres:
Something like: 
Screen Shot 2020-10-09 at 3 30 59 pm

That's with a .dfn-panel .marker `margin-left: 0.15cm`

marcoscaceres:
One more nit... the "IDL" background should probably match the background color of the IDL box heading (#8ccbf2). So: 

Screen Shot 2020-10-09 at 3 34 24 pm

sidvishnoi:
> Kinda squishy... we might want to give some margin: 
> Screen Shot 2020-10-09 at 3 26 12 pm
> 
> Otherwise, looks great. 
> 
> 

Does Safari not support gap property yet?

Issue #3089, opened 2020-08-24
Add a global ReSpec API


sidvishnoi:
```
document.respec = {
  // same as window.respecVersion
  version: string,
  errors: ReSpecError[],
  warnings: ReSpecError[],
  // same as document.respecIsReady
  ready: Promise<void>,
  // catch ReSpec errors/warnings (extends EventTarget)
  addEventListener: ("error" | "warning") => void,
  // export static HTML
  toHTML: () => Promise<string>,
}
```
- [x] `document.respec` basic (https://github.com/w3c/respec/pull/3082)
- [x] `document.respec.ready` (https://github.com/w3c/respec/pull/3082)
- [x] deprecate `document.respecIsReady` (https://github.com/w3c/respec/pull/3090)
- [x] `document.respec.toHTML()` - Part of https://github.com/w3c/respec/issues/2889 (https://github.com/w3c/respec/pull/3403)
- [x] `document.respec.errors[]`, `document.respec.warnings[]` - Closes https://github.com/w3c/respec/issues/1084 (https://github.com/w3c/respec/pull/3348)
- [ ] `document.respec.addEventListener("error" | "warning")` - Part of https://github.com/w3c/respec/issues/3035 (https://github.com/w3c/respec/pull/3338)

dontcallmedom:
I'm not sure if it makes sense in this context, but one feature [I needed recently]( which could be hosted in this global API: in a postProcess script in ReSpec, I [wanted to use data from the WebIDL parser](https://github.com/w3c/webrtc-stats/pull/580/files#diff-98a5b017f8b861a70f52aee2a9a2bb55R19), but there is no way to access the parser or the parsed IDL tree right

sidvishnoi:
I'm not sure if `document.respec` would be the right place to expose this as I would like to keep the API surface as small as possible here to avoid any breaking changes.
FWIW, we do have plans for a new plugins API (See https://github.com/w3c/respec/issues/1976#issuecomment-659673129), where it might be easier to expose this functionality.

PR #3158, opened 2020-10-09
feat(tools/respecDocWriter): add improved toHTML() method

sidvishnoi:
## DEPRECATED:

- `fetchAndWrite` is deprecated (will be removed in future).

## BREAKING CHANGES:

- `--out=""` CLI option no longer writes to stdout.

## Node.js API Changes:

- `toHTML` is more structured and requires fewer arguments:
  - `toHTML` does not write output unlike `fetchAndWrite`.
  - `toHTML` returns processed HTML as well as errors & warnings.
  - `toHTML` does not have `haltOnError` and `haltOnWarning` object parameter.
  - `options.debug` in `toHTML` is renamed to `options.devtools`.
- Removed colors from errors and warnings returned from `toHTML` to make them easier to parse.
- ReSpec version being used is printed during processing.

### Example usage:

```js
const { toHTML } = require('respec');
try {
  const { html, errors, warnings } = await toHTML(src, options);
} catch (fatalError) {
  console.error(fatalError);
}
```

## CLI Improvements:

- Use `--out stdout` to write to stdout.
- Ignore `--out` option to ignore output. No need to pass `/dev/null` anymore.

## Internal:

- lots of refactoring and cleanup.
- functions are defined where used, and follow more consistent order.

### DIFF
--- a/tools/respec2html.js
+++ b/tools/respec2html.js
@@ -13,9 +13,7 @@ colors.setTheme({
   verbose: "cyan",
   warn: "yellow",
 });
-const convertToHTML = require("./respecDocWriter");
-const { writeFile } = require("fs").promises;
-const path = require("path");
+const { convertToHTML, write } = require("./respecDocWriter");
 
 const commandLineArgs = require("command-line-args");
 const getUsage = require("command-line-usage");
@@ -170,25 +168,3 @@ const usageSections = [
   }
   process.exit(0);
 })();
-
-/**
- * @param {string | "stdout" | null | "" | undefined} destination
- * @param {string} html
- */
-async function write(destination, html) {
-  switch (destination) {
-    case "":
-    case null:
-    case undefined:
-      break;
-    case "stdout":
-      process.stdout.write(html);
-      break;
-    default: {
-      const newFilePath = path.isAbsolute(destination)
-        ? destination
-        : path.resolve(process.cwd(), destination);
-      await writeFile(newFilePath, html, "utf-8");
-    }
-  }
-}
--- a/tools/respec2html.js
+++ b/tools/respec2html.js
@@ -1,9 +1,6 @@
 #!/usr/bin/env node
-
-/* jshint node: true, browser: false */
 "use strict";
 const colors = require("colors");
-const fetchAndWrite = require("./respecDocWriter").fetchAndWrite;
 colors.setTheme({
   data: "grey",
   debug: "cyan",
@@ -16,6 +13,9 @@ colors.setTheme({
   verbose: "cyan",
   warn: "yellow",
 });
+const convertToHTML = require("./respecDocWriter");
+const { writeFile } = require("fs").promises;
+const path = require("path");
 
 const commandLineArgs = require("command-line-args");
 const getUsage = require("command-line-usage");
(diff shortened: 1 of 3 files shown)


sidvishnoi:
This combined with https://github.com/w3c/respec/issues/3035 and `document.respec.toHTML` from https://github.com/w3c/respec/issues/3089 should give us a lot better Node.js setup, specially as CI usage increases.

tools/respecDocWriter.js
```diff
@@ -293,4 +338,6 @@ async function write(destination, html) {
   }
 }
 
+exports.convertToHTML = convertToHTML;
```
sidvishnoi, line 341: Happy to name it something better.

tools/respecDocWriter.js
```diff
@@ -1,135 +1,173 @@
...
+  /** @type {RsError[]} */
+  const warnings = [];
+  const onError = error => {
+    errors.push(error);
+    options.onError(error);
+  };
+  const onWarning = warning => {
+    warnings.push(warning);
+    options.onWarning(warning);
+  };
```
marcoscaceres, line 54: I wonder if we should curry these two? 

```JS
function errorReporter(errors, handler) {
    return error => {
        errors.push(error);
        handler(error)
    }
}
```

tools/respecDocWriter.js
```diff
@@ -1,135 +1,173 @@
...
+ */
+async function fetchAndWrite(src, out, whenToHalt = {}, options = {}) {
+  const colors = require("colors");
+  colors.setTheme({ debug: "cyan", error: "red", warn: "yellow" });
+
+  console.warn(
+    colors.warn(
+      "DEPRECATION WARNING: `fetchAndWrite` is deprecated and will be removed in a future version. Please use `convertToHTML` instead."
+    )
+  );
```
marcoscaceres, line 125: I wonder if we should should define generic `showWarn(msg)` and `showError(msg)`. Would avoid the ugly looking:

```
console.error(colors.error(
```

tools/respecDocWriter.js
marcoscaceres: Name seems ok to me.

marcoscaceres (approved):
Mostly just refactor/nits/suggestions for your consideration - but overall looks fine.

tools/respecDocWriter.js
sidvishnoi: Maybe it's fine here. Can avoid duplication as just two nearby instances. Reduces indirection

tools/respec2html.js
```diff
@@ -133,18 +131,35 @@ const usageSections = [
...
+      onError(error) {
+        console.error(
+          colors.error(`💥 ReSpec error: ${colors.debug(error.message)}`)
+        );
+      },
+      onWarning(warning) {
+        console.warn(
+          colors.warn(`⚠️ ReSpec warning: ${colors.debug(warning.message)}`)
+        );
+      },
```
sidvishnoi, line 148: Keeping these as is, as they'll be changed when https://github.com/w3c/respec/pull/3062 comes in anyway

tools/respecDocWriter.js
marcoscaceres: Yeah, agree, it's fine.

tools/respecDocWriter.js
sidvishnoi: How about `respecToHTML` though?

tools/respecDocWriter.js
```diff
@@ -1,135 +1,173 @@
...
+ * @param {boolean} [options.devtools] Show the Chromium window with devtools open for debugging.
+ * @return {Promise<{ html: string, errors: RsError[], warnings: RsError[] }>}
+ * @throws {Error} If failed to process.
  */
-async function fetchAndWrite(
-  src,
-  out,
-  whenToHalt = {},
-  {
+async function convertToHTML(src, options = {}) {
```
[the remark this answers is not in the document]
sidvishnoi: Oops. We can't do that. That would mean if user passes some option, then they must pass onError, onWarning also.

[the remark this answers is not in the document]
sidvishnoi: Fixed above in https://github.com/w3c/respec/pull/3158/commits/246f1d32883d9ba4543868d3b3f224b798e6874f

PR #3225, opened 2020-11-23
chore(ci/changelog): update create-pull-request to v3

sidvishnoi:
This will get rid of `set-env` and `add-path` deprecation errors in GitHub pages (example: https://github.com/w3c/respec/actions/runs/378730123)

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -18,7 +18,7 @@ jobs:
       - run: github_changelog_generator -u w3c -p respec --no-unreleased
         env:
           CHANGELOG_GITHUB_TOKEN: ${{ secrets.CHANGELOG_GITHUB_TOKEN }}
-      - uses: peter-evans/create-pull-request@v2
+      - uses: peter-evans/create-pull-request@v3
         with:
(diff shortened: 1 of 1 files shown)


saschanaz (approved):
TIL this exists, I might also consider using it in webidl-updater to autoupdate the spec list.

.github/workflows/changelog.yml
```diff
@@ -18,7 +18,7 @@ jobs:
       - run: github_changelog_generator -u w3c -p respec --no-unreleased
         env:
           CHANGELOG_GITHUB_TOKEN: ${{ secrets.CHANGELOG_GITHUB_TOKEN }}
```
saschanaz, line 20: The docs say [`secrets.GITHUB_TOKEN` is autogenerated inside an Action](https://docs.github.com/en/free-pro-team@latest/actions/reference/authentication-in-a-workflow#about-the-github_token-secret), is there a reason not to use it?

sidvishnoi: The default has 1000 API request limit per hour (it used to be 10k), so we run out of quota. Using my personal token here.

(This changelog generator needs to be rewritten by someone)

PR #3229, opened 2020-11-26
chore(bundle): mangle object properties to get smaller bundle

sidvishnoi:
Mangles long keys names in objects to get smaller bundle. Keys like [`no_issues_in_spec`](https://github.com/w3c/respec/blob/33532d9f29b5188614ca16d75098eaa9df86875a/src/core/issues-notes.js#L21-L29) get renamed to have smaller names.

|              | Before |  After | Difference |
| ------------ | -----: | -----: | ---------: |
| Uncompressed | 306548 | 282476 |      24072 |
| Brotli       |  84039 |  81095 |       2944 |
| Gzip         |  97392 |  94140 |       3252 |

### DIFF
--- a/tools/builder.js
+++ b/tools/builder.js
@@ -6,6 +6,7 @@ const { promises: fsp } = require("fs");
 const path = require("path");
 const { rollup } = require("rollup");
 const alias = require("rollup-plugin-alias");
+const { terser } = require("rollup-plugin-terser");
 const CleanCSS = require("clean-css");
 const commandLineArgs = require("command-line-args");
 const getUsage = require("command-line-usage");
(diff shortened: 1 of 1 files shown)


sidvishnoi:
Oops. I ran tests on debug version.. Wait for me..

sidvishnoi:
This didn't work out as expected. I'll try this again in future.
One alternative is to use regex to allow/block certain properties/patterns. Maybe `localizationStrings` can have `_` in key names. This way, `respecConfig` ones won't get mangled. Creating a full list of reserved keys is cumbersome for sure.

sidvishnoi:
Some other time...

PR #3244, opened 2020-12-07
fix(core/xref): clarify msg for normative/informative ref mismatch

tidoust:
ReSpec renders a warning when a spec makes a normative reference to a term that is defined as an informative term by another spec. This is great but the warning message was:
`Adding an informative reference to "${term}" from "${cite} in a normative section`.

This message is misleading. First, the use of "Adding" only makes sense from a code perspective (the code detects the problem when it populates the list of references). More importantly, this makes the reader feel that the spec is making an informative reference, whereas the actual problem is precisely the opposite.

This updates the message to a hopefully clearer:
`Normative reference to "${term}" found but term is defined informatively in "${cite}"`.


sidvishnoi (approved):

marcoscaceres (approved):
Agree, this was confusing. We might need to wordsmith this further.

PR #3306, opened 2021-02-15
fix(core/list-sorter, dfn-index): use HTML lang attribute for localized sort

sidvishnoi:
Found when debugging Firefox CI failures. Apparently, GitHub Actions didn't set up an expected LANGUAGE ([sample run](https://github.com/w3c/respec/runs/1902727344#step:6:8)), so Firefox's `Array.sort` didn't work as expected.
This is probably a Firefox bug - it should've used `lang` attribute while sorting if LANGUAGE is not available. cc: @saschanaz

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -19,37 +19,14 @@ jobs:
     name: Check linting issues
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-node@v1
-        with:
-          node-version: 12
-      - name: Cache node modules
-        uses: actions/cache@v1
-        with:
-          path: ~/.npm
-          key: ${{ runner.os }}-node-12-${{ hashFiles('**/package-lock.json') }}
-      - run: npm ci
-      - run: npm run lint
+      - run: exit 0
 
   test-headless:
     name: Headless Tests
     runs-on: ubuntu-latest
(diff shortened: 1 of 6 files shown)


saschanaz:
Hmm, tests still fail.

>  This is probably a Firefox bug - it should've used `lang` attribute while sorting if LOCALE is not available. cc: @saschanaz

👀 Could you file a bug?

sidvishnoi:
> Hmm, tests still fail.

I didn't update other cases yet. See the action run for https://github.com/w3c/respec/pull/3306/commits/224d845589229219f5b98a5dd2b69f775b027ada as a confirmation.

> :eyes: Could you file a bug?

Will do. Was confirming if you think it's a valid bug.

saschanaz:
> Was confirming if you think it's a valid bug.

Frankly I'm not entirely sure how it should decide the locale. MDN says it's implementation defined, but the SpiderMonkey team should know better.

sidvishnoi:
Found https://bugzilla.mozilla.org/show_bug.cgi?id=1685075

sidvishnoi:
Not sure if we should "fix" our impl, or just set proper locale in CI. Can't find any reason for a user computer will have C.utf8 set as locale.

sidvishnoi:
Closing in favor of https://github.com/w3c/respec/pull/3307

PR #3310, opened 2021-02-16
fix(w3c): improve license + unofficial status support

### LINKED ISSUE #3064: Unofficial specs using CC-BY
In a spec with `specStatus: "unofficial", license: "cc-by"`, ReSpec emits an error:

> You cannot use license "`cc-by`" with W3C Specs. Please set `respecConfig.license: "w3c-software-doc"` instead.

In the documentation for [specStatus](https://github.com/w3c/respec/wiki/specStatus), `unofficial` is:

> An unofficial draft. Use this if you're producing a document outside of W3C but want to use styles that match those of W3C specifications, for instance to propose a new document while hosting it on your own server. Note that this automatically licenses the content under CC-BY v3.0. If you want a different copyright, you will need to set the various copyright configuration options.	

(BTW, CC-BY v3.0 should probably be CC-BY v4.0 here. See [the code here](https://github.com/w3c/respec/blob/b623ed094e11b251c9b6c03bf347afb3359ffd23/src/w3c/templates/headers.js#L288).)

Since "unofficial" is considered a document *outside* of W3C and the default license is CC-BY, maybe ReSpec shouldn't emit this [error](https://github.com/w3c/respec/blob/2bcffb680c784b0fe65c606bc08c5e22fc797872/src/w3c/headers.js#L238-L241)?

marcoscaceres:
Closes #3064 

It no longer hard codes cc-by into the template, which was overriding license selection.

### DIFF
--- a/src/w3c/headers.js
+++ b/src/w3c/headers.js
@@ -186,29 +186,42 @@ const noTrackStatus = [
 ];
 const cgbg = ["CG-DRAFT", "CG-FINAL", "BG-DRAFT", "BG-FINAL"];
 const precededByAn = ["ED", "IG-NOTE"];
-const licenses = {
-  cc0: {
-    name: "Creative Commons 0 Public Domain Dedication",
-    short: "CC0",
-    url: "https://creativecommons.org/publicdomain/zero/1.0/",
-  },
-  "w3c-software": {
-    name: "W3C Software Notice and License",
-    short: "W3C Software",
-    url: "https://www.w3.org/Consortium/Legal/2002/copyright-software-20021231",
-  },
-  "w3c-software-doc": {
-    name: "W3C Software and Document Notice and License",
-    short: "W3C Software and Document",
-    url:
-      "https://www.w3.org/Consortium/Legal/2015/copyright-software-and-document",
-  },
-  "cc-by": {
-    name: "Creative Commons Attribution 4.0 International Public License",
-    short: "CC-BY",
-    url: "https://creativecommons.org/licenses/by/4.0/legalcode",
-  },
-};
+const licenses = new Map([
+  [
+    "cc0",
+    {
+      name: "Creative Commons 0 Public Domain Dedication",
+      short: "CC0",
+      url: "https://creativecommons.org/publicdomain/zero/1.0/",
+    },
+  ],
+  [
+    "w3c-software",
+    {
+      name: "W3C Software Notice and License",
+      short: "W3C Software",
+      url:
+        "https://www.w3.org/Consortium/Legal/2002/copyright-software-20021231",
+    },
+  ],
+  [
+    "w3c-software-doc",
+    {
+      name: "W3C Software and Document Notice and License",
+      short: "W3C Software and Document",
+      url:
+        "https://www.w3.org/Consortium/Legal/2015/copyright-software-and-document",
+    },
+  ],
+  [
+    "cc-by",
+    {
+      name: "Creative Commons Attribution 4.0 International Public License",
+      short: "CC-BY",
+      url: "https://creativecommons.org/licenses/by/4.0/legalcode",
+    },
+  ],
+]);
 
 const baseLogo = Object.freeze({
   id: "",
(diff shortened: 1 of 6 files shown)


tests/spec/w3c/headers-spec.js
```diff
@@ -1001,22 +1001,103 @@ describe("W3C — Headers", () => {
...
+        "https://creativecommons.org/licenses/by/4.0/legalcode"
+      );
+    });
+
+    it("falls back to cc-by when license is unknown and spec status is unofficial", async () => {
+      const ops = makeStandardOps();
+      ops.config = {
+        shortName: "whatever",
+        specStatus: "unofficial",
+        licenses: "not a thing",
```
marcoscaceres, line 1026:
```suggestion
        license: "not a thing",
```

src/w3c/headers.js
```diff
@@ -243,14 +256,29 @@ export function run(conf) {
   if (conf.isUnofficial && !Array.isArray(conf.logos)) {
     conf.logos = [];
   }
+  if (conf.isUnofficial) {
+    if (conf.license && !licenses.has(conf.license)) {
```
sidvishnoi, line 260: Should we check this regardless of `conf.isUnofficial`?

src/w3c/headers.js
```diff
@@ -385,7 +413,7 @@ export function run(conf) {
       !conf.isNoTrack &&
       !conf.isSubmission
     ) {
-      const msg = "Document on track but no previous version.";
```
sidvishnoi, line 388: This one is probably not desired. All other `msg` end with full-stop.

src/w3c/templates/headers.js
```diff
@@ -282,25 +282,23 @@ function renderCopyright(conf) {
...
-          )}.
-        </p>`
-    : conf.overrideCopyright
-    ? [conf.overrideCopyright]
-    : renderOfficialCopyright(conf);
+  if (conf.isUnofficial && conf.licenseInfo) {
+    return html`<p class="copyright">
+      This document is licensed under a
+      ${linkLicense(
+        `${conf.licenseInfo.name}`,
```
sidvishnoi, line 294:
```suggestion
        conf.licenseInfo.name,
```

tests/spec/w3c/headers-spec.js
```diff
@@ -1001,22 +1001,103 @@ describe("W3C — Headers", () => {
...
 
-  describe("license - w3c-software-doc", () => {
+  describe("license configuration", () => {
+    it("defaults to cc-by when spec status is unofficial", async () => {
+      const ops = makeStandardOps();
+      ops.config = {
+        shortName: "whatever",
+        specStatus: "unofficial",
+        editors: [{ name: "foo" }],
+      };
```
sidvishnoi, line 1011:
```suggestion
      const ops = makeStandardOps({
        shortName: "whatever",
        specStatus: "unofficial",
      });

```

tests/spec/w3c/headers-spec.js
```diff
@@ -1241,18 +1322,6 @@ describe("W3C — Headers", () => {
         /XXX\s+&\s+the\s+Contributors\s+to\s+the/
       );
     });
-    it("takes additionalCopyrightHolders into account when spec is unofficial", async () => {
```
sidvishnoi, line 1244: Could this be a breaking change? I'm not sure I follow the rationale behind this.

marcoscaceres: `additionalCopyrightHolders` was being used incorrectly here. It was being used as equivalent of the  `replaceCopyright` option. I think that was a oversight - specially for "unofficial" docs. I checked specs using `additionalCopyrightHolders` and none are "unofficial".

src/w3c/headers.js
marcoscaceres: This appears to be a bug in core UI:

https://github.com/w3c/respec/blob/49ddeac2e88a635032d65f21a7244350512c1c14/src/core/ui.js#L278

It adds a "." to the hint. We probably don't want that regardless in i18n'zed contexts. I'll remove it.

tests/spec/w3c/headers-spec.js
marcoscaceres: forgot about this :P

src/w3c/headers.js
marcoscaceres: yes, but I want to do that as a followup. I'm just interested in dealing with unofficial for now... eventually we need to rip out all the unofficial stuff from the W3C profile. None of that stuff should be here.

tests/spec/w3c/headers-spec.js
sidvishnoi: Also the ones below

src/w3c/headers.js
marcoscaceres: Just in case, fix is now here: https://github.com/w3c/respec/pull/3310/files#diff-b1ee7df5d4991f7198879953fab78023c1ed2dcb5e6d45b2f8994e50d3459c24

tests/spec/w3c/headers-spec.js
sidvishnoi: I'll fix the rest..

sidvishnoi (approved):

Issue #2981, opened 2020-06-24
Proposal: add a `prepare()` pass before `run()` pass


sidvishnoi:
ReSpec has "prepare" code like [this](https://github.com/w3c/respec/blob/b9dab188f0eb460980fcabac718e85cb421fed47/src/core/worker.js#L10-L21) or [this](https://github.com/w3c/respec/blob/7faae154e57d4af864a83b50adf744029fbd7496/src/core/style.js#L15-L16) in many modules. My idea is to add a `async prepare(conf)` in such modules, and do a pass where all modules run their `prepare()`, followed by our existing `run()` pass.

Why?
- I was trying to reorder some plugins to make overall ReSpec appear to load faster (mainly create ToC before content below the fold). But some plugins add headings dynamically (not available until they `run()`), but ToC requires those headings (so we've a dependency chain making ToC render late). If we can generate those headings in a prepare, then we can easily generate ToC earlier (See this demo of impact: https://respec-run-toc-early.netlify.app).
- This would also help with https://github.com/w3c/respec/pull/2187 where we can keep mutations in prepare, and not make them top-level.

Thoughts? @saschanaz @marcoscaceres

saschanaz:
This sounds good! I see no immediate negative point.

marcoscaceres:
Like it! /me presses 👍 button.

PR #3311, opened 2021-02-16
feat(core/base-runner): add a prepare pass

sidvishnoi:
Closes https://github.com/w3c/respec/issues/2981
(Review commit by commit)

### DIFF
--- a/src/core/base-runner.js
+++ b/src/core/base-runner.js
@@ -24,7 +24,10 @@ export async function runAll(plugs) {
   runnables.forEach(
     plug => !plug.name && console.warn("Plugin lacks name:", plug)
   );
+  respecConfig.state = {};
+  await executePreparePass(runnables, respecConfig);
   await executeRunPass(runnables, respecConfig);
+  respecConfig.state = {};
   pub("plugins-done", respecConfig);
 
   await postProcess(respecConfig);
@@ -36,6 +39,16 @@ export async function runAll(plugs) {
 
 function isRunnableModule(plug) {
   return plug && (plug.run || plug.Plugin);
+}
+
+async function executePreparePass(runnables, config) {
+  for (const plug of runnables.filter(p => p.prepare)) {
+    try {
+      await plug.prepare(config);
+    } catch (err) {
+      console.error(err);
+    }
+  }
 }
 
 async function executeRunPass(runnables, config) {
--- a/src/core/base-runner.js
+++ b/src/core/base-runner.js
@@ -11,43 +11,6 @@ import { removeReSpec } from "./utils.js";
 
 export const name = "core/base-runner";
 
-function toRunnable(plug) {
-  const name = plug.name || "";
-  if (!name) {
-    console.warn("Plugin lacks name:", plug);
-  }
-  return config => {
-    // eslint-disable-next-line no-async-promise-executor
-    return new Promise(async (resolve, reject) => {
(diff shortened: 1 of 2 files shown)


sidvishnoi:
Impact on caniuse:
(I've high latency with respec.org, so screenshots are perfect, but important thing is start/end time for caniuse so it appears in header faster)

Before:
image
After:
image


Before:
image
After: 
image

Before:
image
After:
image

src/core/base-runner.js
```diff
@@ -11,61 +11,63 @@ import { removeReSpec } from "./utils.js";
...
+        } catch (err) {
+          reject(err);
+        } finally {
+          clearTimeout(timerId);
+          performance.mark(`${name}-end`);
+          performance.measure(name, `${name}-start`, `${name}-end`);
+        }
+      });
+    } catch (error) {
+      console.error(error);
```
saschanaz, line 83: It seems we don't need the outer try-catch. Seemingly it can only happen when `performance` doesn't exist.

src/core/base-runner.js
```diff
@@ -22,7 +22,10 @@ export async function runAll(plugs) {
   runnables.forEach(
     plug => !plug.name && console.warn("Plugin lacks name:", plug)
   );
+  respecConfig.state = {};
+  await executePreparePass(runnables, respecConfig);
   await executeRunPass(runnables, respecConfig);
+  delete respecConfig.state;
```
saschanaz, line 28: Maybe `respecConfig.state = {}` again?

src/core/caniuse.js
```diff
@@ -60,18 +48,29 @@ export async function run(conf) {
...
+  const apiUrl = options.apiURL || API_URL;
+  // Initiate a fetch, but do not wait. Try to fill the cache early instead.
+  conf.state[name] = {
+    fetchPromise: fetchStats(apiUrl, options),
+  };
+}
+
+export async function run(conf) {
+  const options = conf.caniuse;
+  if (!options || !options.feature) return;
```
saschanaz, line 66:
```suggestion
  if (!options?.feature) return;
```

src/core/base-runner.js
sidvishnoi: The `new Promise()` can still `reject(err)`. Also, `setTimeout` can reject.
I agree this function can be made better.

saschanaz: Oops, true! Okay then.

src/core/caniuse.js
sidvishnoi: OMG are we ready for optional chaining now?! :star_struck:

sidvishnoi: Optional chain all the things!! https://caniuse.com/mdn-javascript_operators_optional_chaining

sidvishnoi: Wait. Should we wait for Samsung Internet?

saschanaz: That's a good question... I think the data is just outdated and it's actually supported. Filed https://github.com/mdn/browser-compat-data/issues/9146

marcoscaceres: Yeah, let's start using optional chaining.

marcoscaceres (approved):

sidvishnoi:
(Just fixed commit dates in last force-push)

saschanaz (approved):
🚀

PR #3312, opened 2021-02-16
refactor(core): run global plugins like regular plugins

sidvishnoi:
We can further refactor these to be imported directly into profiles, not sure if that's preferable.

### DIFF
--- a/src/core/pre-process.js
+++ b/src/core/pre-process.js
@@ -9,43 +9,30 @@
  *      want to be using a new module with your own profile
  */
 import { showError } from "./utils.js";
-import { sub } from "./pubsubhub.js";
 
 export const name = "core/pre-process";
 
-let doneResolver;
-export const done = new Promise(resolve => {
-  doneResolver = resolve;
-});
-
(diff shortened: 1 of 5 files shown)


src/core/pre-process.js
```diff
@@ -9,43 +9,30 @@
...
-            const msg = `Function ${f.name} threw an error during \`preProcess\`.`;
-            const hint = "See developer console.";
-            showError(msg, name, { hint });
-            console.error(err);
-          }
-        });
-      const values = await Promise.all(promises);
-      result.push(...values);
-    }
-    doneResolver(result);
```
sidvishnoi, line 48: Unless we had a reason to resolve with `result`...

saschanaz (approved):
Cool, everything is now more consistent and less implicit 😍

src/core/pre-process.js
marcoscaceres: Probably not.

marcoscaceres (approved):

PR #3317, opened 2021-02-18
refactor: use sade for CLI

sidvishnoi:
## respec2html

```diff
- respec -s SRC -o OUT
+ respec SRC OUT
```
Above works while still supporting older format (~with a deprecation warning~).
image

## builder

```diff
- tools/builder.js --profile PROFILE
+ tools/builder.js PROFILE
```

Breaking change, but fine as it's internal.

## dev-server

Same as before.

## Notes to reviewer:

- `tools/dev-server.js`: Wrapped all functions under a `run(opts)` function (args is no longer a global), so mainly whitespace changes.
- Review/merge commit by commit.


tools/respec2html.js
```diff
@@ -13,156 +15,90 @@ colors.setTheme({
...
-    console.info(getUsage(usageSections));
-    return process.exit(2);
+cli.action((source, destination, opts) => {
+  if (!source) {
+    if (!opts.src) {
+      cli.help();
+      process.exit(1);
+    }
+    const msg =
+      `-s / --src option is deprecated in favor of [source] positional argument.` +
```
marcoscaceres, line 55: I'm not sure we should bother warning. Using either positional or the explicit means should be fine.

tools/respec2html.js
```diff
@@ -13,156 +15,90 @@ colors.setTheme({
...
+      ` See --help for new syntax.`;
+    console.warn("⚠", colors.yellow(msg));
+    source = opts.src;
   }
-  if (parsedArgs.help) {
-    console.info(getUsage(usageSections));
-    return process.exit(0);
+  if (!destination && opts.out) {
+    const msg =
+      `-o / --out option is deprecated in favor of [destination] positional argument.` +
```
marcoscaceres, line 62: as above... supporting the explicit means seems fine. Not something we would remove in the long term.

marcoscaceres (approved):
Just nit about the deprecation comments, as we discussed.

PR #3320, opened 2021-02-19
feat(core/extensions): add extensions API

sidvishnoi:
Closes https://github.com/w3c/respec/issues/1976

Support two extension hooks: before-markdown and after-markdown.
Support two extension utils: showError and showWarning.

### DIFF
--- a/examples/basic.html
+++ b/examples/basic.html
@@ -15,6 +15,15 @@
           name: "Your Name",
           url: "https://your-site.com",
         }],
+        extensions: [
+          {
+            name: "my-extension",
+            on: "before-markdown",
+            async run(conf, utils) {
+              utils.showWarning("oops from extension", this.name);
+            }
+          }
+        ],
         github: "https://github.com/w3c/some-API/",
         testSuiteURI: "https://w3c-test.org/some-API/",
         implementationReportURI: "https://w3c.github.io/test-results/some-API",
--- a/src/core/base-runner.js
+++ b/src/core/base-runner.js
@@ -1,6 +1,7 @@
 // @ts-check
 // Module core/base-runner
 // The module in charge of running the whole processing pipeline.
+import { utils as extensionUtils, setupExtensions } from "./extensions.js";
 import { run as includeConfig } from "./include-config.js";
 import { init as initReSpecGlobal } from "./respec-global.js";
 import { run as overrideConfig } from "./override-configuration.js";
@@ -51,7 +52,9 @@ async function executePreparePass(runnables, config) {
   }
 }
 
-async function executeRunPass(runnables, config) {
+async function executeRunPass(corePlugins, config) {
+  const runnables = setupExtensions(corePlugins, respecConfig);
+
   for (const plug of runnables) {
     const name = plug.name || "";
 
@@ -68,6 +71,10 @@ async function executeRunPass(runnables, config) {
         try {
           if (plug.Plugin) {
             await new plug.Plugin(config).run();
+            resolve();
+          } else if (plug.on && plug.run) {
+            // is extension
+            await plug.run(config, extensionUtils);
             resolve();
           } else if (plug.run) {
             await plug.run(config);
--- a/src/core/extensions.js
+++ b/src/core/extensions.js
@@ -1,0 +1,62 @@
+// @ts-check
+import { showError, showWarning } from "./utils.js";
+
+export const name = "core/extensions";
+
+/**
+ * @param {any[]} corePlugins
+ * @param {Conf} conf
+ */
+export function setupExtensions(corePlugins, conf) {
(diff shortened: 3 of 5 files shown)


marcoscaceres:
I think this is still risky... I'd prefer people just user preProcess and postProcess. Hooking into anything makes things fragile, as we've seen.

sidvishnoi:
> Hooking into anything makes things fragile, as we've seen.

I understand the concern, but I think the previous hooks were fragile as we exposed them everywhere without much thought. If we limit what we expose, we can have a much more powerful API. 

As an example, if you look at [respec.org/docs/src.html#L29-L36](https://github.com/marcoscaceres/respec.org/blob/9229a5484ef2dfaeebad2711b1e9c7b46100c22f/static/docs/src.html#L29-L36) (and [respec.org/docs/src.html#L98-L128](https://github.com/marcoscaceres/respec.org/blob/9229a5484ef2dfaeebad2711b1e9c7b46100c22f/static/docs/src.html#L98-L128) in particular), you'll see what hacks I had to make to overcome the limitations of preProcess and postProcess :sweat_smile:

marcoscaceres:
> As an example, if you look at respec.org/docs/src.html#L29-L36 (and respec.org/docs/src.html#L98-L128 in particular), you'll see what hacks I had to make to overcome the limitations of preProcess and postProcess 😅

I see... what's a rough sketch of what that would look like with the new api?

saschanaz:
Looks cool. Which specs would use the markdown hooks?

sidvishnoi:
Haven't tested real specs that would like to use `after-markdown` hook, but [respec.org/docs/src.html#L115-L125](https://github.com/marcoscaceres/respec.org/blob/9229a5484ef2dfaeebad2711b1e9c7b46100c22f/static/docs/src.html#L115-L125) could use it like this:

<details>
<summary>diff: docs/src.html</summary>

```diff
diff --git a/static/docs/src.html b/static/docs/src.html
index 9c5375d..03418a3 100644
--- a/static/docs/src.html
+++ b/static/docs/src.html
@@ -34,6 +34,26 @@
           postProcessEnhance,
           cleanup,
         ],
+        extensions: [
+          {
+            name: "note-advisement",
+            on: "after-markdown",
+            run() {
+              for (const p of document.querySelectorAll("p")) {
+                if (p.firstElementChild?.localName === "strong") {
+                  const text = p.firstElementChild.textContent.trimStart();
+                  if (text.startsWith("Note:")) {
+                    p.classList.add("note");
+                    p.firstElementChild.remove();
+                  } else if (text.startsWith("Warning:")) {
+                    p.classList.add("advisement");
+                    p.firstElementChild.remove();
+                  }
+                }
+              }
+            },
+          },
+        ],
         logos: [
           {
             src: "/respec-logo.png",
@@ -112,18 +132,6 @@
           .replace(/\`\|(\w+)/g, `\`\\|${ZERO_WIDTH_SPACE}$1`)
           .replace(/(\w+)\|\`/g, `$1${ZERO_WIDTH_SPACE}\\|\``);
 
-        // Add .note and .advisement classes based on line prefix
-        content = content
-          .split("\n")
-          .map(line => {
-            if (/^.{0,5}\s*Warning:/.test(line))
-              return `<div class="advisement">\n\n${line}</div>`;
-            if (/^.{0,5}\s*Note:/.test(line))
-              return `<div class="note">\n\n${line}</div>`;
-            return line;
-          })
-          .join("\n");
-
         return content;
       }
```

</details>

With a potential `after-inline-processing` hook (which runs before linker), [docs/src.html#L99-L113](https://github.com/marcoscaceres/respec.org/blob/9229a5484ef2dfaeebad2711b1e9c7b46100c22f/static/docs/src.html#L99-L113) will simplify a lot.

Similarly, if we could run [`fixLinks()`](https://github.com/marcoscaceres/respec.org/blob/9229a5484ef2dfaeebad2711b1e9c7b46100c22f/static/docs/src.html#L147), say, `on: "after-linker"`, we would get access to `local-refs-exist` linker in a better way. Currently linters run after `doc.respec.ready`, which means we can miss linter warnings in respec2html (to workaround this, we [wait for 1 second after `respec.ready`](https://github.com/w3c/respec/blob/22bd2782c9ea1fb226f0344131b0f29a071a5a51/tools/respecDocWriter.js#L89-L90) which is problematic in itself). 

Will test with some other real specs with potential hooks and update here.

(PS: Missed notifications as I accidentally unsubscribed from this PR).

marcoscaceres:
I think think this is going to be extremely footgunny. 

> Currently linters run after doc.respec.ready, which means we can miss linter warnings in respec2html (to workaround this, we wait for 1 second after respec.ready which is problematic in itself).

Yeah, we should change the linters back into normal plugins as you suggested elsewhere.

PR #3321, opened 2021-02-21
refactor(core/utils): remove IDBKeyVal and use raw idb with xref-db

sidvishnoi:
`IDBKeyVal` sole user was xref-db. This PR makes xref-db use raw idb store and adds some other improvements (including less code).

refactor(core/xref-db): to make xref DB easier to inspect/debug, we now store both result as well as the query. There's also a secondary index (`byTerm` = `query.term`) which makes the inspection easier.

Also moved `__LAST_VERSION_CHECK__` from the IDB store to localStorage, in order to make the IDB store more consistent.

Improved TypeScript support while we're at it (no "any" types now).

Also, removed a stupid test that tried to test the caching behavior, but didn't actually test the caching behavior.

| Before | After |
| --- | --- |
| image  | image |
| | image |


src/core/xref-db.js
```diff
@@ -73,27 +74,30 @@ async function shouldBustCache(cache) {
...
-export async function cacheXrefData(data) {
+export async function cacheXrefData(queries, results) {
   try {
     const cache = await getIdbCache();
-    // add data to cache
-    await cache.addMany(data);
+    const tx = cache.transaction(STORE_NAME, "readwrite");
+    for (const query of queries) {
+      tx.objectStore(STORE_NAME).add({ query, result: results.get(query.id) });
+    }
```
marcoscaceres, line 91: Bit easier on the eye... 

```suggestion
    for (const query of queries) {
      const result = results.get(query.id);
      tx.objectStore(STORE_NAME).add({ query, result });
    }
```

marcoscaceres (approved):
Nice one!

PR #3322, opened 2021-02-22
fix(w3c): prevent CG/BGs from using wrong status

### LINKED ISSUE #3017: CG/BG drafts should use CG-DRAFT, not ED
When the group is W3C's WICG, ReSpec error if the status is not CG-DRAFT or CG-FINAL(i.e., "ED" is not allowed).

marcoscaceres:
Closes #3017

Prevents CG/BG drafts from using "ED" as the status. It also checks that WGs are not using a status intended for CG/BGs.


src/type-helper.d.ts
```diff
@@ -183,3 +183,11 @@ module "core/xref" {
 
   export type XrefDatabase = IDBPDatabase<XrefDBScheme>;
 }
+
+enum W3CGroupType {
+  "bg",
+  "cg",
+  "ig",
+  "wg",
+  "",
```
sidvishnoi, line 192: I don't think this case is possible.

src/w3c/defaults.js
```diff
@@ -52,4 +54,53 @@ export function run(conf) {
...
+function validateStatusForGroup(conf) {
+  const { specStatus, groupType } = conf;
+  switch (groupType) {
+    case "cg": {
+      if (![...cgbgStatus, "unofficial"].includes(specStatus)) {
+        const msg = `W3C Community Group documents can't use \`"${specStatus}"\` for the ${docLink(
+          "specStatus"
+        )} configuration option. Please use one of: ${toMDCode(
+          cgStatus
+        )}. Automatically falling back to \`"CG-DRAFT"\`.`;
```
sidvishnoi, line 72: Split into `msg` and `hint`?

src/w3c/headers.js
```diff
@@ -257,12 +258,22 @@ export function run(conf) {
     conf.logos = [];
   }
   if (conf.isUnofficial) {
+    if (!conf.specStatus) {
```
sidvishnoi, line 261: This combination is not possible, due to following code abpve:
```js
  conf.isUnofficial = conf.specStatus === "unofficial";
````

**Review by sidvishnoi — APPROVED** [review]

File: src/w3c/headers.js (hunk 3 above)
Comment (marcoscaceres) [reply src/w3c/headers.js]: Good catch 🙇‍♂️

File: src/w3c/defaults.js [hunk 4]
```diff
@@ -52,4 +54,55 @@ export function run(conf) {
...
+        showError(msg, name, { hint });
+        conf.specStatus = "BG-DRAFT";
+      }
+      break;
+    }
+    case "wg": {
+      if (!cgbgStatus.includes(specStatus)) {
+        const msg = `Invalid value for ${docLink(
+          "specStatus"
+        )} a W3C Working Group document. Please see ${docLink(
```
sidvishnoi, line 96: Missed one for `hint`. Can update later..

PR #3323, opened 2021-02-22
tests(core/markdown): workaround highlight.js lang detection bug

sidvishnoi:
hljs was detecting the test block as "CSS".

Follow up: As we've access to highlight languages, we should probably use `highlight(lang, code)` instead of `highlightAuto(code, langSuggestions)`. Should also move respec-hljs into respec repo, to make such changes easier.

### DIFF
--- a/tests/spec/core/markdown-spec.js
+++ b/tests/spec/core/markdown-spec.js
@@ -301,7 +301,7 @@ describe("Core - Markdown", () => {
       \`\`\`
 
       \`\`\`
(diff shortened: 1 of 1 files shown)


marcoscaceres (approved):
if it helps..

marcoscaceres:
>  we should probably use highlight(lang, code) instead of highlightAuto(code, langSuggestions). 

Agree... and maybe just fall back to js if not specified. Language detection is bogus.

PR #3329, opened 2021-02-23
chore(package): update deps

marcoscaceres:
closes #3327
closes #3326 
closes #3325 
closes #3324


sidvishnoi (approved):
I wonder if there are instances where we can use https://github.com/w3c/respec/pull/3325..

marcoscaceres:
> ?boolean=${value}

Maybe... I guess we need to look at some cases. I don't recall any cases where what we do is not clear, but might be worth having a look.

PR #3330, opened 2021-02-23
refactor(tests/test-build): use fs.promises instead of promisify

### LINKED ISSUE #3188: Drop promisify() from tests
[fs.promises](https://nodejs.org/api/fs.html#fs_fs_promises_api) API has been there for a while. We should be able to use it without breaking changes.

marcoscaceres:
closes #3188


saschanaz (approved):
👍

marcoscaceres:
thanks for the speedy review! ❤️

PR #3333, opened 2021-02-23
fix(tools/dev-server): regression, browsers arg must be an array

marcoscaceres:
Looks like a small regression from #3317 ... was wondering how Sade determined the types. Now I know 🤣

### DIFF
--- a/tools/dev-server.js
+++ b/tools/dev-server.js
@@ -17,8 +17,8 @@ sade("./tools/dev-server.js", true)
   .option("-p, --profile", "Name of profile to build.", "w3c")
   .option("-i, --interactive", "Run in interactive mode.", false)
   .option(
-    "--browsers",
-    'Browsers for Karma unit tests, space seperated (e.g., "Chrome Firefox")'
+    "--browser",
+    'Browser for Karma unit tests (e.g., "Chrome"). Multiple allowed.'
   )
   .option("--grep", "Run specific tests using karma --grep")
   .action(opts => run(opts))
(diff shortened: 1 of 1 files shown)


tools/dev-server.js
```diff
@@ -16,18 +16,20 @@ const SERVE_PORT = 5000;
...
   .parse(process.argv);
 
 function run(args) {
   let isActive = false;
-
   const karmaConfig = karma.config.parseConfig(
     path.join(__dirname, "../karma.conf.js"),
     {
-      browsers: args.browsers,
+      browsers: args.browsers?.split(/\s/) || [],
```
sidvishnoi, line 32: I think we also want something like:
```js
if (typeof args.browsers === "string")
  args.browsers = args.browsers.split(/\s/);
```
As we can call it like:
```sh
npm start -- --browsers Chrome --browsers Firefox
# args.browsers is now an array of strings
```

sidvishnoi: Maybe rename the option from browsers to browser to make it clearer

tools/dev-server.js
```diff
@@ -16,18 +16,21 @@ const SERVE_PORT = 5000;
...
 
 function run(args) {
   let isActive = false;
-
   const karmaConfig = karma.config.parseConfig(
     path.join(__dirname, "../karma.conf.js"),
     {
-      browsers: args.browsers,
+      browsers:
+        typeof args.browser === "string" ? [args.browser] : args.browser,
```
sidvishnoi, line 33:
```suggestion
        typeof args.browser === "string" ? [args.browser] : args.browser || [],
```

sidvishnoi (approved):

tools/dev-server.js
sidvishnoi: Unless it can accept undefined?

PR #3334, opened 2021-02-23
fix(templates/show-link): localize colons usage

### LINKED ISSUE #3041: Colons in the specification metadata block
ReSpec can automatically generate spec metadata like "This version:", "Latest published version:", or "Participate:".

In https://github.com/w3c/respec/blob/f564b3aa141a02f267866f36e51e238bf1ed94c4/src/core/github.js#L23-L56 , there is no colon after the strings in `en`. Some languages use `U+FF1A` instead of `U+003A` for colon, so I think a better way to deal with the colon in a spec metadata entry is to either add colons to the strings (`participate: "Participate:"` instead of `participate: "Participate"`), or let ReSpec automatically add a colon based on the language tag.

(Not sure this is a feature request or a bug, so tagging "feature request" for now.)

marcoscaceres:
closes #3041


src/dini/templates/headers.js
```diff
@@ -133,7 +133,7 @@ export default conf => {
             ${showPeople(conf.authors)}
           `
         : ""}
-      ${conf.otherLinks ? conf.otherLinks.map(showLink) : ""}
+      ${conf.otherLinks ? conf.otherLinks.map(link => showLink(link)) : ""}
```
marcoscaceres, line 136: I changed these because `showLink()`'s function signature doesn't match what `.map()` expects.

sidvishnoi: I think this is probably not needed here, as we have control over the signature of `showLink` (it expects only one param). Issues arise when there's more than one parameter.

marcoscaceres: yeah, true... I'll switch 'em back.

tests/spec/w3c/headers-spec.js
```diff
@@ -1141,7 +1141,7 @@ describe("W3C — Headers", () => {
       const otherLinks = [
         {
           class: "key-other-link",
-          key: "KEY",
+          key: "KEY:",
```
sidvishnoi, line 1144: It's a little "breaking change".

marcoscaceres: It's a gray area (doesn't "break" anything) :) I searched around and not many specs that are affected. I'll send PRs for any I see, as most are not using this for any valid reason (e.g., they should use `"github"` instead).

sidvishnoi (approved):

src/core/github.js
```diff
@@ -21,9 +21,9 @@ export const github = new Promise((resolve, reject) => {
...
   en: {
-    file_a_bug: "File a bug",
-    participate: "Participate",
-    commit_history: "Commit history",
+    file_a_bug: "File a bug:",
+    participate: "Participate:",
+    commit_history: "Commit history:",
   },
   ko: {
     participate: "참여",
```
saschanaz, line 29: Shouldn't these also be updated? And should there be `colon` field instead of repeating it everywhere? 🤔

marcoscaceres: Ah, sorry. I wasn't sure if ko also used ":". I'll update those.

PR #3335, opened 2021-02-23
feat(core/webidl): support linking special operations

### LINKED ISSUE #982: Support linking to stringifier
Given:

```JS
interface Foo {
   stringifier DOMString();
};
```

One should be able to link to the stringifier somehow.

marcoscaceres:
Support linking and auto-definition of special operations: setters, getters, deleters, etc.

closes #982 

Now supports
```HTML
<section data-dfn-for="Bar">
  <h2>Annonymous special operations</h2>
  <pre class="idl">
    [Exposed=Window]
    interface Bar {
      getter DOMString ();
      setter DOMString ();
      stringifier DOMString ();
      deleter DOMString ();
    };
  </pre>
  <p>
    <dfn>getter</dfn> <dfn>setter()</dfn> <dfn>stringifier</dfn>
  </p>
</section>
```


src/core/webidl.js
```diff
@@ -231,7 +232,9 @@ function getNameAndId(defn, parent = "") {
 
 function resolveNameAndId(defn, parent) {
   let name = getDefnName(defn);
-  let idlId = getIdlId(name, parent);
+  // For getters, setters, etc. "annonymous-getter",
+  const prefix = defn.special && defn.name === "" ? "annonymous-" : "";
```
saschanaz, line 236: I wonder this could just be `-` so that the id would be `dom-foo--getter`. Because 1) the "named" version also has the behavior of "anonymous" one as [it's just a syntax sugar on top of it](https://heycam.github.io/webidl/#example-50d7ab88) 2) and it's just shorter 😄

(typo: a"n"onymous)

src/core/webidl.js
```diff
@@ -298,12 +301,17 @@ function getDefnName(defn) {
     case "enum-value":
       return defn.value;
     case "operation":
-      return defn.name;
+      return defn.special ? getSpecialName(defn) : defn.name;
```
saschanaz, line 304: How about:

```suggestion
      return defn.name || defn.special;
```

src/core/webidl.js
marcoscaceres: the "--" looks weird to me... I kinda want it to be explicitly named "anonymous". Having said that, the ids are just for us so not super concerned about length.

src/core/webidl.js
marcoscaceres: Ah, lol! yeah :D

saschanaz (approved):
Okay then, no blocker.

marcoscaceres:
yay! thanks @saschanaz 🎉

marcoscaceres:
And, TIL how to spell anonymous properly lol! 🤣

PR #3336, opened 2021-02-23
fix(webidl): wrap idl in a code element

### LINKED ISSUE #3189: Wrap IDL code with <pre><code>
Currently ReSpec inserts `.idlHeader` into `<pre>`, which forces crawlers to use ReSpec-specific code to remove it.

Doing `<pre><span class="idlHeader">...</span><code>...</code></pre>` could be better, as crawlers can query `pre code` as a standard way. It's also semantically better, as `<pre>` does not automatically say the content is a code. [HTML](https://html.spec.whatwg.org/multipage/grouping-content.html#the-pre-element) says:

>To represent a block of computer code, the pre element can be used with a code element

It also can replace a hacky code inside ReSpec by querying `pre code`:

https://github.com/w3c/respec/blob/7faae154e57d4af864a83b50adf744029fbd7496/src/core/webidl-index.js#L67-L70

See also https://github.com/whatwg/html/pull/3768.

marcoscaceres:
Closes #3189


saschanaz (approved):
Looks cool 👍

PR #3338, opened 2021-02-23
feat(core/respec-global): dispatch errors and warnings 

sidvishnoi:
Closes https://github.com/w3c/respec/issues/1084

- Added an event ~`"respec-start"`~ `"__private_respec_start__"` on `document` that signals `document.respec` is available. Can change its target/name etc.
- `document.respec` is now an `EventTarget` that can emit "error" and "warning" `CustomEvent`s

```js
document.addEventListener("__private_respec_start__", (ev) => {
  document.respec.addEventListener("error", ({ detail: rsError }) => console.error(rsError));
  document.respec.addEventListener("warning", ({ detail: rsError }) => console.warn(rsError));
  document.respec.ready.then(() => {
    console.log(document.respec.errors);
    console.log(document.respec.warnings);
  })
}, { once: true })
```

As a follow up, I intend to use this in respecDocWriter, instead of relying on console messages.

### DIFF
--- a/src/core/respec-global.js
+++ b/src/core/respec-global.js
@@ -47,7 +47,8 @@ class ReSpec extends EventTarget {
 export function init() {
   const respec = new ReSpec();
   Object.defineProperty(document, "respec", { value: respec });
-  document.dispatchEvent(new CustomEvent("respec-start"));
+  // Internal. Do not use.
+  document.dispatchEvent(new CustomEvent("__private_respec_start__"));
 
   let respecIsReadyWarningShown = false;
   Object.defineProperty(document, "respecIsReady", {
--- a/src/core/respec-global.js
+++ b/src/core/respec-global.js
@@ -13,11 +13,25 @@ import { sub } from "./pubsubhub.js";
 
 export const name = "core/respec-global";
 
-class ReSpec {
+class ReSpec extends EventTarget {
   constructor() {
+    super();
     /** @type {Promise<void>} */
     this._respecDonePromise = new Promise(resolve => {
       sub("end-all", resolve, { once: true });
+    });
+    this.errors = [];
+    this.warnings = [];
+
+    sub("error", rsError => {
+      console.error(rsError, rsError.toJSON());
+      this.dispatchEvent(new CustomEvent("error", { detail: rsError }));
+      this.errors.push(rsError);
+    });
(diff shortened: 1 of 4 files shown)


marcoscaceres (commented):
I like it!

marcoscaceres (approved):

marcoscaceres:
Will wait for @saschanaz before merging, as this is a small but important addition.

src/core/respec-global.js
```diff
@@ -33,6 +47,7 @@ class ReSpec {
 export function init() {
   const respec = new ReSpec();
   Object.defineProperty(document, "respec", { value: respec });
+  document.dispatchEvent(new CustomEvent("respec-start"));
```
saschanaz, line 50: Can we use the existing `<script onload>` for this?

marcoscaceres: Should we fire this at all... who is expected to use this?

sidvishnoi: `onload` can get fired before the code that adds `document.respec` is run.

> who is expected to use this

`respecDocWriter` can use it to start listening to error/warning events, similar to how we add a console message listener. Tests are also using it.

It's a guarantee that respec has started processing and is emitted as soon as respec starts (a little bit after DOMReady).

marcoscaceres: maybe we should call it "private-respec-start" or something - just in case people find it and are tempted to use it.

sidvishnoi: Let me add it when I use it in respec2html. Might figure out a way to not need it.

sidvishnoi: What's wrong if people use it though?

marcoscaceres: We have multiple mechanism in place already for people to use, so I'm worried about the maintainability of having yet another that could cause issues down the road. Want to avoid another `w3c-common` situation.

sidvishnoi: Renamed to `__private_respec_start__`

sidvishnoi:
Should I split this PR into two: dispatch and collect (errors/warnings array)? i.e. do dispatch one when we need to use it.

sidvishnoi:
Closing for now in favor of https://github.com/w3c/respec/pull/3348

PR #3339, opened 2021-02-24
fix(assets/respec.css): darken pre background in examples

marcoscaceres:
Makes the `<pre>`s inside `.example` blocks a little darker, and matches what Bikeshed does: 

Screen Shot 2021-02-24 at 12 24 54 pm

This is nice in the it better splits up code and text:

Screen Shot 2021-02-24 at 12 29 51 pm

### DIFF
--- a/assets/respec.css
+++ b/assets/respec.css
@@ -234,10 +234,6 @@ aside.example .marker > a.self-link {
   color: inherit;
 }
 
-.example pre {
-  background-color: rgba(0, 0, 0, .03);
-}
-
 h2 > a.self-link,
 h3 > a.self-link,
 h4 > a.self-link,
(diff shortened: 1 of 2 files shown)


sidvishnoi (approved):

assets/respec.css
```diff
@@ -234,6 +234,10 @@ aside.example .marker > a.self-link {
   color: inherit;
 }
 
+.example pre {
```
sidvishnoi, line 237: We have an examples.css. Might make sense to move it there.m

sidvishnoi (approved):

assets/respec.css
marcoscaceres: ah! good point.

Issue #3292, opened 2021-02-03
respec2html does not load local data-include


gkellogg:
## Important info

- URL to affected spec: https://github.com/json-ld/json-ld-star/ (https://json-ld.github.io/json-ld-star/)
- ReSpec version: 26.0.0
- [x] I did a "hard refresh", but it's still busted. (N/A)

## Description of problem

In attempting to use the [spec-pub GitHub action](https://github.com/w3c/spec-prod/blob/main/docs/options.md) to validate PRs and Push actions, I found that it failed to validate the file due to what seems to be failure to load files from the local directory using `data-include`. I reproduced this using `respec2html` using the following:

    respec2html  --src index.html --out /tmp/foo.html

The relevant lines from index.html are:

```html
  <div class="informative hidden">
    <!-- Terms included for referencability, but not displayed -->
    <div data-include="common/terms.html"></div>
    <div data-include="common/algorithm-terms.html"></div>
  </div>
```

(By the way, it would be useful if --help or --version gave some ReSpec version info).

**What happened (e.g., it crashed)?:**

It failed with the following (abbreviated) output:

```text
💥 ReSpec error: data-include failed for element:  [object HTMLDivElement] TypeError: Failed to fetch
💥 ReSpec error: `data-include` failed: `common/terms.html` (Failed to fetch). See console for details.
💥 ReSpec error: data-include failed for element:  [object HTMLDivElement] TypeError: Failed to fetch
💥 ReSpec error: `data-include` failed: `common/algorithm-terms.html` (Failed to fetch). See console for details.
```
I suspect that this is for the same reason that I need to run a local HTTP server for local development in the browser. However, the sand-box reasons for a browser to not load off of the local filesystem seem different than those for an NPM script.

**Expected behavior (e.g., it shouldn't crash):**

Validates document including loading resources.

**Optional, steps to reproduce:**

1.  Run `respec2html  --src index.html --out /tmp/foo.html` or similar.

sidvishnoi:
@marcoscaceres should we include a simple static server in respec2html? 
I plan to run the server in spec-prod GitHub action, but might be worth adding it in ReSpec itself?

sidvishnoi:
Either way, need to fix docs for `d ata-include`

dontcallmedom:
> services like spec-generator that call toHTML() often will end up creating lots of servers

Also, it may be that once we get enough adoption of the github action, we don't need the spec-generator anymore.

marcoscaceres:
[...]

PR #3340, opened 2021-02-24
feat(tools/respec2html): add --localhost option for http server

marcoscaceres:
Quick straw-person for #3292 for discussion.


package.json
```diff
@@ -82,6 +82,7 @@
   },
   "dependencies": {
     "colors": "^1.4.0",
+    "express": "^4.17.1",
```
sidvishnoi, line 85: First thought before anything else: overkill for a static server! 😆

tools/respec2html.js
```diff
@@ -43,7 +48,9 @@ cli
   .option("-w, --haltonwarn", "Abort if ReSpec generates warnings.", false)
   .option("--disable-sandbox", "Disable Chromium sandboxing if needed.", false)
   .option("--devtools", "Enable debugging and show Chrome's DevTools.", false)
-  .option("--verbose", "Log processing status to stdout.", false);
+  .option("--verbose", "Log processing status to stdout.", false)
+  .option("--localhost", "Spin up a local server to peform processing.", false)
```
sidvishnoi, line 52: Maybe we can support something like `--localhost=optionalPort` instead of two options.

package.json
marcoscaceres: Yeah, was thinking the same... at the same time, it's nice to use. Do you have a suggestion for a good alternative. Generally, I use https://www.npmjs.com/package/http-server but it doesn't look like it has an API.

tools/respec2html.js
marcoscaceres: Could do... need to check how to set up that kind of option.

package.json
sidvishnoi: I used `serve-handler` in spec-prod

sidvishnoi:
Maybe we can use local server as default behavior?

marcoscaceres:
> Maybe we can use local server as default behavior?

I'm not comfortable with spinning up a web sever on a user's machine by default, tbh. Even if it's just for a few seconds. "The Web" is scary.

tools/respec2html.js
marcoscaceres: Concluded to keeping option separate.

tools/respec2html.js
```diff
@@ -43,7 +64,9 @@ cli
...
```
sidvishnoi, line 68: This message some a room for improvements. It looks a bit misleading right now.

tools/respec2html.js
```diff
@@ -2,6 +2,22 @@
...
+    req,
+    res,
+    finalhandler(req, res, {
+      onerror(err) {
+        console.error(err.stack);
+      },
+    })
+  );
+});
+const path = require("path");
```
sidvishnoi, line 20: Can we wrap these in a `StaticServer` class?

sidvishnoi: Done with 2158b46

tools/respec2html.js
marcoscaceres: Feel free to change.... the current wording makes sense to me so I couldn't come up with anything better 😢.

marcoscaceres (commented):
The new class for the server looks good to me. You will need to self approve to merge it, but happy to review any new changes you make.

tools/respec2html.js
sidvishnoi: Can't think of any better either. Right now, it sounds like the server will perform processing..

tools/respec2html.js
```diff
@@ -38,6 +42,49 @@ class Logger {
   }
 }
 
+class StaticServer {
+  /**
+   * @param {number} port
+   * @param {string} source
+   */
+  constructor(port, source) {
```
marcoscaceres, line 50: I wonder if we should split construction from source validation... they seem to be different concerns.

marcoscaceres (commented):
Just a minor thing for the constructor

tools/respec2html.js
marcoscaceres: That's not entirely untrue... if you squint just right.

tools/respec2html.js
sidvishnoi: It encapsulates related logic well enough though..

marcoscaceres: Up do you. I don't mind either way.

sidvishnoi (approved):

PR #3372, opened 2021-03-09
perf(core/a11y): start loading axe earlier

sidvishnoi:


src/core/base-runner.js
```diff
@@ -43,6 +43,7 @@ function isRunnableModule(plug) {
 
 async function executePreparePass(runnables, config) {
   for (const plug of runnables.filter(p => p.prepare)) {
+    config.state[plug.name] = {};
```
marcoscaceres, line 46: Hmm... not a fan of attaching more stuff to `config`. Why not have this as an internal state inside the plugin itself.

marcoscaceres: I'd suggest doing that now... look at what happened with my terrible linter API :( I kept saying, "I'll fix it next week... I'll fix it next week..." and years later, still there 😢

PR #3374, opened 2021-03-11
feat(core/inlines): support attr-value type for elements

sidvishnoi:


src/core/inlines.js
```diff
@@ -77,16 +77,28 @@ const inlineElement = /(?:\[\^[^^]+\^\])/; // Inline [^element^]
...
+  const [xrefType, xrefFor, textContent, linkingText] = (() => {
+    if (attrValue) {
+      const text = `${element} ${attribute} "${attrValue}"`;
+      return ["attr-value", `${element}/${attribute}`, text, attrValue];
+    } else if (attribute) {
+      return ["element-attr", element, attribute];
+    } else {
+      return ["element", null, element];
+    }
+  })();
```
sidvishnoi, line 93: If only JS supported switch-case as expressions...

tests/spec/core/inlines-spec.js
```diff
@@ -334,6 +335,9 @@ describe("Core - Inlines", () => {
     const iframeAllowAnchor = doc.querySelector("#test2 a");
     expect(iframeAllowAnchor.textContent).toBe("allow");
     expect(iframeAllowAnchor.hash).toBe("#attr-iframe-allow");
+    const inputModeAnchor = doc.querySelector("#test3 a");
+    expect(inputModeAnchor.textContent).toBe(`html-global inputmode "text"`);
```
sidvishnoi, line 339: For `[^ link/rel/manifest ^]` will render as `link rel "manifest"`.
Any better textContent?
Should we support renaming like  `[^ link/rel/manifest | link[rel="manifest"] ^]`

marcoscaceres (approved):
Looks good.

tests/spec/core/inlines-spec.js
marcoscaceres: as discussed, should just be manifest

PR #3376, opened 2021-03-11
Fix including markdown

SteveALee:
I've been exploring `include` and `markdown` for [w3c/coga](https://github.com/w3c/coga) and found the combination was only working in 2 very limited cases.

After lots of userland permutation testing I finally debugged the code and found a broken destructuring assignment. This PR fixes it.

There seems to be some test code coverage issues as well as this didn't break the tests but my fix does. I can look at those too if you like?

Ideally this would be released ASAP as it is holding me back.

Thanks.

### DIFF
--- a/src/core/data-include.js
+++ b/src/core/data-include.js
@@ -21,19 +21,19 @@ export const name = "core/data-include";
  * @param {boolean} options.replace
  */
 function fillWithText(el, data, { replace }) {
-  const { includeFormat } = el.dataset;
+  const { format } = el.dataset;
   let fill = data;
-  if (includeFormat === "markdown") {
+  if (format === "markdown") {
     fill = markdownToHtml(fill);
   }
 
-  if (includeFormat === "text") {
+  if (format === "text") {
     el.textContent = fill;
   } else {
     el.innerHTML = fill;
   }
 
-  if (includeFormat === "markdown") {
+  if (format === "markdown") {
     restructure(el);
   }
 


marcoscaceres:
Hi @SteveALee, it seems per the implementation that `data-include-format` is correct, and not `data-format`.  So I think the bug is actually here:

https://github.com/w3c/respec/blob/38ef244b8e9099f59cf5b036c7d5c98e8fa3f98d/src/core/markdown.js#L304-L306

Those should also check `data-include-format` and then I think it will work.

SteveALee:
I can't believe this is the case and im not on my pc to test but I've been using Firefox and it appears the tests use headless Chromium. Surly not a browser variation?

SteveALee:
@marco I can update the PR but why do you say Also? Is that for legacy? If I used data-format and not data-format-include it will be as I copied it from an example. So the error may be quite widespread.

What would you like me to do?

saschanaz:
`data-format=markdown` is to treat the texts in the block as markdown, whereas `data-include-format=markdown` is to treat texts included by `data-include` as markdown. Those are two different cases.

sidvishnoi:
@SteveALee can you share link to the document where you're having issues?

SteveALee:
Oh gosh, I completely missed that distinction!! I assumed one covered both cases. Didn't realise the behaviour varied as it seems the transformation would happen after the include expansion.

Might need pointing out in the docs?

There's no published doc yet, I am experimenting with creating a  TR note and WAI web pages from a single content source for the design guide in  W3C/coga repo. The idea is a file per section which is in markdown for easy editing and rendering via respec includes and jekyll SSG.

I really need nested includes as therec nested sections with introductions but hoping can work around by hoisting headings to including respec and pasingvtgat for jekyll.

I'll go back to exploring with the new knowledge about two different options.

Then I'll publish a branch for this issue if required.

Thanks.

SteveALee:
Given my missunderstanding about the 2 distinct attributes for markdown I'll close this and raise an issue with more details.

Thanks

PR #3342, opened 2021-02-24
Improved logs

sidvishnoi:
Before:
image

After:
image

After (deluxe edition (markdown formatting in terminal)):
image

Less verbose, hyperlink style and a warning:
image

And if colors are too much:
image

Fatal errors:
image
image
image
image

Detailed errors when available:
image

PR #3387, opened 2021-03-16
feat(tools/respec2html): provide more detailed logs

sidvishnoi:
Remaining part of https://github.com/w3c/respec/pull/3342

image
image

Didn't provide `ReSpecError.details` yet, as it gets noisy. Will add sometime later as a followup (maybe behind a --log-level arg?).


marcoscaceres (approved):
So sad no private methods yet...

PR #3388, opened 2021-03-16
fix(conformance): use datatracker for IETF urls

dontcallmedom:
tools.ietf.org urls are being deprecated per https://mailarchive.ietf.org/arch/msg/tools-discuss/oYrAxb3KayPzZ4SNB1DVZTDPPNo/


sidvishnoi:
@dontcallmedom can you also update `src/dini/conformance.js`?

dontcallmedom:
I'm happy to, I wasn't sure if this was appropriate for me to do so or if the DIN maintainers would want to take a different stab at it.

sidvishnoi (approved):
Thanks!

PR #3390, opened 2021-03-20
refactor: build respec-highlight alongside respec-worker

sidvishnoi:
This brings in `respec-highlight` from https://github.com/w3c/respec-hljs into this repo, so we don't need to maintain it separately.
- [ ] Ask someone at W3C to host `respec-highlight` from this repo instead of other.


marcoscaceres:
I can set up the proxy.

marcoscaceres (approved):

worker/respec-highlight.js
```diff
@@ -0,0 +1,22 @@
...
+  languages: ["abnf", "css", "http", "javascript", "json", "xml"],
+});
+
+highlight.registerLanguage("abnf", abnf);
+highlight.registerLanguage("css", css);
+highlight.registerLanguage("http", http);
+highlight.registerLanguage("javascript", javascript);
+highlight.registerLanguage("json", json);
+highlight.registerLanguage("xml", xml);
+
```
marcoscaceres, line 21:
```suggestion
const langMap = new Map([
  ["abnf", abnf], 
  ["css", css],
  ["http", http],
  ["javascript", javascript],
  ["json", json],
  [ "xml", xml],
])

highlight.configure({
  tabReplace: "  ", // 2 spaces
  languages:  [...langMap.keys()],
});

for(const [key, highlighter] of langMap.entries()) {
  highlight.registerLanguage(key, highlighter);
}

```

sidvishnoi: I think the former is straight-forward. Other requires more brain processing.
Opinions:tm:

sidvishnoi:
Afterthought: why do we even build/ship `respec-worker` when we inline it in `src/core/worker.js`?
cc: @marcoscaceres

PR #3398, opened 2021-03-22
chore(ci/pr): print size of respec-w3c

sidvishnoi:


.github/workflows/pr.yml
```diff
@@ -50,6 +50,11 @@ jobs:
         run: |
           npm ci
           npm run build:w3c
+      - name: Size of respec-w3c
+        run: |
+          npx -q brrr builds/respec-w3c.js -o builds
```
marcoscaceres, line 55: This is very cute, but it would be more cute 🐨 if it showed in comparison to the old build... otherwise, it doesn't mean much, right?

sidvishnoi: That would require building twice.. 
There's https://github.com/preactjs/compressed-size-action to do the heavy lifting, but only tells of compressed size. I'm interested in both compressed and uncompressed. I might create an action for that maybe :cat:

sidvishnoi:
Will build an action later based on: https://github.com/w3c/respec/blob/e464b16f180b16b245d7141d74969e9d7c785596/.github/workflows/pr.yml#L32-L50

.github/workflows/pr.yml
marcoscaceres: > That would require building twice...

Hmm... could we look at sizes before "npm run build:w3c" runs, then store them somewhere?

sidvishnoi: Building adds like 3-4 seconds, so I think might be okay to be build twice.
Storing adds its own maintenance issues.

marcoscaceres: > Building adds like 3-4 seconds, so I think might be okay to be build twice.

Yeah, that's fine. It's only for us, after all. 

> Storing adds its own maintenance issues.

Oh, I was thinking just a simple environment variable, like:  

```
ORIGINAL_BUILD_SIZE=$(stat -f%z builds/respec-w3c.js)
```

Just for quick comparison.

sidvishnoi: That would be comparing to previous release. I'm interested in comparing since last commit to base branch.

sidvishnoi: Unless you mean comparing size with the one we just built in previous step, which is something I'll do when I create that action.

marcoscaceres: Ok cool.

PR #3401, opened 2021-03-23
refactor(core/linter/check-charset): run as regular plugin

sidvishnoi:


tests/spec/SpecHelper.js
```diff
@@ -57,6 +57,60 @@ export function makeRSDoc(opts, src, style = "") {
...
+          var respecConfig = ${JSON.stringify(config || {}, null, 2)};
+        </script>
+        <script type="module">
+          import * as baseRunner from "/base/src/core/base-runner.js";
+          const modules = ${JSON.stringify(modules)};
+          const importedModules = await Promise.all(
+            modules.map(m => import("/base" + m))
+          );
+          await baseRunner.runAll(importedModules);
+        </script>
```
sidvishnoi, line 84: Experimenting with "unit testing" modules as "units"...

sidvishnoi:
```diff
START:
  Core — data-abbr
    ✔ processes definition abbreviations
    ✔ correctly parses and abbreviates different Abbreviation syntax
    ✔ warns when used with unsupported elements
    ✔ allows different abbreviation combinations to link to dfn
    ✔ normalizes title of added abbr for unwanted spaces

- Finished in 1.335 secs / 1.315 secs
+ Finished in 0.307 secs / 0.294 secs
```

tests/spec/SpecHelper.js
```diff
@@ -57,6 +57,78 @@ export function makeRSDoc(opts, src, style = "") {
   });
 }
 
+/**
+ * Create a doc for unit tests.
```
marcoscaceres, line 61: This is neat! I like the idea of running things in isolation. However, we should use this sparingly, as running all the interdependencies always catches many bugs in other code.

sidvishnoi: I'm thinking of creating a separate folder for unit tests, which would run before current (integration) tests. Might move some of the integration tests into unit tests, where it makes sense.
Agree on using unit tests sparingly (keeping in mind unit tests are unit tests!)

tests/spec/SpecHelper.js
marcoscaceres: Sure, but speed is not a big concern on the testing side, right? `fit(` and `fdescribe(` solve mostly for it during development. But making sure that we catch one plugin breaking/regressing another is more critical, which is usually what happens in practice (when I touch the code 😅).

tests/spec/SpecHelper.js
marcoscaceres: I like the division of labor. The utils module does something similar, in as far as we test exported methods on their own.... I like where this is going!

tests/unit/SpecHelper.js
```diff
@@ -6,19 +6,19 @@ const iframes = [];
...
 export function makePluginDoc(
   plugins,
-  { config = {}, head = "", body = "" } = {}
+  { config = {}, head = `<meta charset="UTF-8" />`, body = "" } = {}
 ) {
   return getDoc(`
     <!DOCTYPE html>
     <html lang="en">
       <head>
-        ${head !== null ? `<meta charset="UTF-8" />${head}` : ""}
```
sidvishnoi, line 21: Ok. I want to add default charset in all tests except check-charset :smile: 
Lets hope we don't need to edit `<head>` often in other tests.

src/core/linter-rules/check-charset.js
```diff
@@ -1,35 +1,31 @@
...
+    msg: `文档只能包含一个 charset 属性为 utf-8 的 \`<meta>\` 标签`,
+    hint: `将此行添加到文档的 \`<head>\` 部分—— \`<meta charset="utf-8">\` 或将 charset 设置为 utf-8（如果尚未设置）。`,
   },
 };
+const l10n = getIntlData(localizationStrings);
 
-// Fall back to english, if language is missing
-const lang = defaultLang in meta ? defaultLang : "en";
+export function run(conf) {
+  if (!conf.lint?.[ruleName]) {
```
sidvishnoi, line 23: As linter names have hyphens, we need to configure them as:
```js
var respecConfig = {
 lint: {
   "check-charset": true,
  },
};
```

maybe we can alias the config options so we can use:
```js
var respecConfig = {
 lint: {
   checkCharset: true,
  },
};
```

marcoscaceres: yeah, argh... really screwed that up too.

marcoscaceres (approved):

src/core/linter-rules/check-charset.js
sidvishnoi: Will do that when all other linters have migrated to `run()` API.

sidvishnoi: Filed https://github.com/w3c/respec/issues/3438 as a TODO

PR #3402, opened 2021-03-24
refactor(styles): convert css files to js

sidvishnoi:
This PR converts all `fetchAsset(file.css)` and `import(text!file.css)` to regular JS imports. 
The benefit being, cleaner code and less async-await, and slightly smaller bundle size. Worth doing?

<details>
<summary>Size changes</summary>

| File             |  After | Before | Diff      |
| ---------------- | -----: | -----: | --------: |
| respec-w3c.js    | 305747 | 308252 | -2.5Kb 🟢 |
| respec-w3c.js.br |  84910 |  84815 | +0.1Kb 🔴 |
| respec-w3c.js.gz |  99233 |  98219 | +1.0Kb 🔴 |

</details>

### DIFF
--- a/src/core/dfn-index.js
+++ b/src/core/dfn-index.js
@@ -46,14 +46,14 @@ const CODE_TYPES = new Set([
  * @typedef {{ term: string, type: string, linkFor: string, elem: HTMLAnchorElement }} Entry
  */
 
-export async function run() {
+export function run() {
   const index = document.querySelector("section#index");
   if (!index) {
     return;
   }
 
   const styleEl = document.createElement("style");
-  styleEl.textContent = await css;
+  styleEl.textContent = css;
   document.head.appendChild(styleEl);
 
   index.classList.add("appendix");
--- a/src/core/dfn-index.js
+++ b/src/core/dfn-index.js
@@ -6,7 +6,7 @@
  */
 
 import { addId, getIntlData, norm } from "./utils.js";
-import { fetchAsset } from "./text-loader.js";
+import css from "../styles/dfn-index.css.js";
 import { getTermFromElement } from "./xref.js";
 import { html } from "./import-maps.js";
 import { renderInlineCitation } from "./render-biblio.js";
@@ -53,7 +53,7 @@ export async function run() {
   }
 
   const styleEl = document.createElement("style");
-  styleEl.textContent = await loadStyle();
+  styleEl.textContent = await css;
   document.head.appendChild(styleEl);
 
   index.classList.add("appendix");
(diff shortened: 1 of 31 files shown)


marcoscaceres (commented):
If we are doing this, then we might as well just get rid of the *.css.js files and fold them into the modules. Then we can do proper "CSS in JS" stuff.

saschanaz:
Yeah, I agree that it would be better, assuming that every CSS file is small enough and has no visible performance degradation from this.

sidvishnoi:
Wouldn't regular css-in-js require a build step? `.css.js` files don't change our development workflow (directly using src/ without building).

src/core/dfn-index.js
```diff
@@ -46,14 +46,14 @@ const CODE_TYPES = new Set([
  * @typedef {{ term: string, type: string, linkFor: string, elem: HTMLAnchorElement }} Entry
  */
 
-export async function run() {
+export function run() {
```
sidvishnoi, line 49: Should I keep the async here for consistency? Or, not add async where not needed?

saschanaz: I would just remove it.

saschanaz:
> Wouldn't regular css-in-js require a build step? `.css.js` files don't change our development workflow (directly using src/ without building).

Well, I mean, why import it when we can just put it right in the modules? That shouldn't require any build step.

sidvishnoi:
> why import it when we can just put it right in the modules

I think they're large enough to look bad in plugin files:
```bash
# ls assets/* | xargs wc -l | sort -n
    7 assets/algorithms.css
   36 assets/dfn-index.css
   43 assets/examples.css
   45 assets/datatype.css
   49 assets/var.css
   54 assets/issues-notes.css
   94 assets/highlight.css
  111 assets/dfn-panel.css
  127 assets/caniuse.css
  136 assets/mdn-annotation.css
  140 assets/webidl.css
  285 assets/respec.css
  356 assets/ui.css
```

saschanaz (approved):
Ooops, 100+ line files are definitely too big.

Okay, I support this, assuming that the minifier works as expected.

sidvishnoi:
Minification works as expected:
| Non-minified (now) | Minified (now) | Minified (before) |
| -- | -- | -- |
| image | image | image |

saschanaz:
Cool! Interesting that it keeps newlines after each rule 👀

sidvishnoi:
The newlines were there before, but as `\n` in `""`, now they're in backticks.

marcoscaceres (approved):
Ok, let's start here.

PR #3409, opened 2021-03-27
experiment: try to get line number of errors

sidvishnoi:
By adding a special comment at the end of each line as a pre-process step (in Node, not respec), we can estimate the location of error in source, which can be very useful in CI. For every element, we can find the "nearest" (with some heuristics) HTML comment matching that pattern and we'll get access to line number.

- Does anyone know of a way to add such comments? We want to skip `<style>`, `<script>` and `<title>` elements for sure.
- `showError` can have error location hints, e.g., "at the end of element", "at the start of element", "at line N in this element" (WebIDL.js already gives us last one).

https://user-images.githubusercontent.com/8426945/112732752-d1d61500-8f61-11eb-8d72-6c6d0318d3f9.mp4

### DIFF
--- a/examples/basic.commented.html
+++ b/examples/basic.commented.html
@@ -1,0 +1,117 @@
+<!DOCTYPE html><!-- respec-sourcemap#examples/basic.html:1 -->
+<html><!-- respec-sourcemap#examples/basic.html:2 -->
+  <head><!-- respec-sourcemap#examples/basic.html:3 -->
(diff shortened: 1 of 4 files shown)


marcoscaceres:
>  The solution probably involves using a parser like parse5 or htmlparser2, does anyone have some experience with them?

Not me, unfortunately.

lastthyme (approved):

PR #3410, opened 2021-03-28
refactor(tools/dev-server): add promisified KarmaServer

sidvishnoi:
Karma, as it should've been.
In preparation of unit test support...

### DIFF
--- a/tools/dev-server.js
+++ b/tools/dev-server.js
@@ -17,26 +17,23 @@ class KarmaServer {
   /**
    * @param {string} configFile
    * @param {string} [browser]
-   * @param {string} [grep]
    */
-  constructor(configFile, browser, grep) {
-    this._configFile = configFile;
-    this._browsers = browser ? [browser] : undefined;
-    this._grep = grep;
+  constructor(configFile, browser, grep = "") {
+    const browsers = browser ? [browser] : [];
+    this._karmaConfig = karma.config.parseConfig(configFile, {
+      browsers,
+      autoWatch: false,
+      port: KARMA_PORT,
+      logLevel: karma.constants.LOG_INFO,
+      client: {
+        args: ["--grep", grep || ""],
+      },
+      mochaReporter: { ignoreSkipped: true },
+    });
     this._isActive = null;
   }
 
   start() {
-    this._karmaConfig = karma.config.parseConfig(this._configFile, {
-      browsers: this._browsers,
-      autoWatch: false,
-      port: KARMA_PORT,
-      logLevel: karma.constants.LOG_WARN,
-      client: {
-        args: ["--grep", this._grep || ""],
-      },
-      mochaReporter: { ignoreSkipped: true },
-    });
     this.karmaServer = new karma.Server(this._karmaConfig);
     this.karmaServer.start();
     return new Promise(resolve =>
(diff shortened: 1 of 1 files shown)


tools/dev-server.js
```diff
@@ -13,34 +13,69 @@ const { Builder } = require("./builder");
...
 
+class KarmaServer {
+  /**
+   * @param {string} configFile
+   * @param {string} [browser]
+   * @param {string} [grep]
+   */
+  constructor(configFile, browser, grep) {
+    this._configFile = configFile;
+    this._browsers = browser ? [browser] : undefined;
```
marcoscaceres, line 24: Unless it conflicts with Karma, it might be better to use an empty array here instead of `undefined`.

tools/dev-server.js
```diff
@@ -13,34 +13,69 @@ const { Builder } = require("./builder");
...
+    this._grep = grep;
+    this._isActive = null;
+  }
+
+  start() {
+    this._karmaConfig = karma.config.parseConfig(this._configFile, {
```
marcoscaceres, line 30: is `parseConfig()` has no side effects, it might better to do this in the ctor.

tools/dev-server.js
```diff
@@ -13,34 +13,69 @@ const { Builder } = require("./builder");
...
+      mochaReporter: { ignoreSkipped: true },
+    });
+    this.karmaServer = new karma.Server(this._karmaConfig);
+    this.karmaServer.start();
+    return new Promise(resolve =>
+      this.karmaServer.once("browsers_ready", resolve)
+    );
+  }
+
+  async stop() {
```
marcoscaceres, line 47:
```suggestion
  stop() {
```

tools/dev-server.js
```diff
@@ -13,34 +13,69 @@ const { Builder } = require("./builder");
...
+    return new Promise(resolve =>
+      karma.stopper.stop(this._karmaConfig, code =>
```
marcoscaceres, line 49: Might want to add a comment here as to why we need the timeout (if known).

tools/dev-server.js
```diff
@@ -13,34 +13,69 @@ const { Builder } = require("./builder");
...
+        setTimeout(() => resolve(code), 0)
+      )
+    );
+  }
+
+  async run() {
+    if (this._isActive) return;
+    this._isActive = true;
+    karma.runner.run(this._karmaConfig, () => {});
+    await new Promise(res => this.karmaServer.once("run_complete", res));
```
marcoscaceres, line 59: Just for consistency...
```suggestion
    await new Promise(resolve => this.karmaServer.once("run_complete", resolve));
```

marcoscaceres (approved):
Few suggestions, but otherwise looks good 👍

tools/dev-server.js
```diff
@@ -13,34 +13,69 @@ const { Builder } = require("./builder");
 const KARMA_PORT = 9876;
 const SERVE_PORT = 5000;
 
+class KarmaServer {
+  /**
+   * @param {string} configFile
+   * @param {string} [browser]
+   * @param {string} [grep]
+   */
+  constructor(configFile, browser, grep) {
```
marcoscaceres, line 22:
```suggestion
  constructor(configFile, browser, grep = []) {
```

tools/dev-server.js
sidvishnoi: Good idea!

tools/dev-server.js
sidvishnoi: Realized late it was a mistake :sweat_smile: 
Undefined is important.

PR #3418, opened 2021-03-29
tests: support testing individual modules

sidvishnoi:
- Run unit tests before (current) integration tests.
- Convert `highlight-vars` test to unit test as an example.
- Update `npm start` to run both unit tests and integration tests.
- Update CI to run both tests.
- Extracted from https://github.com/w3c/respec/pull/3401


tests/unit/core/highlight-vars-spec.js
```diff
@@ -1,13 +1,14 @@
...
-
-// TODO: add tests:
-// - right colors are applied
+import { flushIframes, makePluginDoc } from "../SpecHelper.js";
 
 describe("Core - highlightVars", () => {
   afterAll(flushIframes);
-  const testBody = `
+
+  const plugins = ["/src/core/highlight-vars.js"];
```
sidvishnoi, line 8: Using absolute paths from /src as it gets a little better editor support (not including renaming)

tests/unit/SpecHelper.js
```diff
@@ -0,0 +1,88 @@
...
+  const html = `
+    <!DOCTYPE html>
+    <html lang="en">
+      <head>
+        ${head}
+        <script>
+          var respecConfig = ${JSON.stringify(config || {}, null, 2)};
+        </script>
+        <script type="module">
+          const plugins = ${JSON.stringify(plugins)};
```
marcoscaceres, line 24: Might be better to wrap this in a self-invoking  `async` and the sprinkle some `await`s around... it's hard to follow the logic around error handling as it currently is - and try/catch will be easier on the eyes.

tests/unit/SpecHelper.js
```diff
@@ -0,0 +1,88 @@
...
+        </script>
+      </head>
+      <body>${body}</body>
+    </html>
+  `;
+
+  return new Promise((resolve, reject) => {
+    const ifr = document.createElement("iframe");
+    const timeoutId = setTimeout(() => {
+      reject(new Error(`Timed out waiting.`));
```
marcoscaceres, line 51: We should probably make this error more descriptive.

tests/unit/SpecHelper.js
```diff
@@ -0,0 +1,88 @@
...
+    }, jasmine.DEFAULT_TIMEOUT_INTERVAL);
+    ifr.addEventListener("load", async () => {
```
marcoscaceres, line 53: It might be a good opportunity to refactor out the async function here to make this whole block a little smaller. Maybe curry it to take the `timeoutId`.

marcoscaceres (changes requested):

tests/unit/SpecHelper.js
```diff
@@ -0,0 +1,105 @@
...
+/**
+ * Create a doc for unit tests.
+ * @param {string[]} plugins Paths of plugins to load and run. Example: `/src/core/algorithms.js`
+ * @param {object} config JSON-serializable respecConfig object.
+ * @return {Promise<Document>}
+ */
+export function makePluginDoc(
+  plugins,
+  config = {},
+  { head = `<meta charset="UTF-8" />`, body = "" } = {}
```
marcoscaceres, line 13: Is there any case where meta charset should be overriden? We should probably keep it inline below and just ="" this default.

marcoscaceres (approved):
MUUUUUUCH nicer. Just a nit.

tests/unit/SpecHelper.js
sidvishnoi: the check-charset test is the only reason for its existence. Maybe can use null to override it, and undefined to use default?

PR #3425, opened 2021-03-30
refactor(tests/webidl): update legacy constructors + cleanup

marcoscaceres:
* Adds missing `[Exposed=*]` to interfaces, 
 * Updates to use inline constructors.
 * s/named constructors/LegacyFactoryFunction 
 * A bunch of random cleanup


tests/spec/core/webidl-spec.js
```diff
@@ -338,33 +344,46 @@ describe("Core - WebIDL", () => {
...
+    `;
+    const ops = makeStandardOps(null, body);
+    const doc = await makeRSDoc(ops);
+
+    const text = `
+[Exposed=Window]
+interface SuperStar {
+  constructor();
+  constructor(boolean bar, sequence<double> foo);
+};
```
saschanaz, line 365: I wonder we could dedupe this, currently `body` and `text` are quite duplicated.

marcoscaceres: It's a bit messy, because they are testing whitespace.

tests/spec/core/webidl-spec.js
```diff
@@ -69,11 +69,11 @@ describe("Core - WebIDL", () => {
     });
   });
 
-  describe("records", () => {
-    it("handles record types", async () => {
+  describe("dictionaries", () => {
```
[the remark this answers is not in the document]
marcoscaceres: Good point... the structure a whole for this file needs a cleanup. But will do that some other day.

tests/spec/core/webidl-spec.js
```diff
@@ -271,7 +273,12 @@ describe("Core - WebIDL", () => {
   it("should handle interfaces", () => {
     let target = doc.getElementById("if-basic");
     target.querySelector(".idlHeader").remove();
```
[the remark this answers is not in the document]
marcoscaceres: Good idea! done globally.

tests/spec/core/webidl-spec.js
```diff
@@ -69,26 +69,104 @@ describe("Core - WebIDL", () => {
...
   });
 
-  describe("records", () => {
-    it("handles record types", async () => {
-      const body = `
-        <section id="records">
-          <h2>Testing records</h2>
-          <pre class="idl">
-            dictionary Foo {
-              record&lt;DOMString, USVString> pass;
```
saschanaz, line 79: Now I see no test for records 👀

It would be easier to review if the constructor part and the cleanup part is splitted.

marcoscaceres: Yeah, I probably should have kept records - I thought those had been removed from IDL for some reason (wrong assumption). I'll add it back. 

As for the rest, I'm trying to get rid of WebIDL.html. It's full of errors and because of it's size (and because we never really look at it), it's not feasible to maintain.

tests/spec/core/webidl-spec.js
```diff
@@ -92,14 +92,116 @@ describe("Core - WebIDL", () => {
     });
   });
 
+  describe("dictionaries", () => {
```
saschanaz, line 95: Could this be moved back? Seems like an unneeded change.

marcoscaceres:
It might be nice in the future to structure the tests in the same way the IDL spec is structured. Will make it easier to keep things in sync and also gives a good indication of any missing things.

marcoscaceres:
going ahead and merging this, as it's been sitting for a while. Will try to do further cleanup when I find some time.

PR #3428, opened 2021-03-31
chore(ci/regressions): test respec-w3c differences across branches 

sidvishnoi:
This workflow finds the impact of changes in ReSpec on real-world specs. It's run on manual requests, to avoid wasting resources. It works as follows:
1. Build the W3C profile, and cache it (`actions/upload-artifact`) .
2. For each `matrix.source`:
   1. Process ReSpec document using the ReSpec script in the document, and write `before.html`.
   2. Process ReSpec document using the ReSpec script generated in Step 1 (thanks to https://github.com/w3c/respec/pull/2957), and write `after.html`.
   3. Compare before.html and after.html. If there is a difference, exit with error code to draw manual attention.

A sample workflow run is available at https://github.com/w3c/respec/actions/runs/704622451 (corresponding to https://github.com/w3c/respec/pull/3428/commits/4aff5e79edd1bcb3445b076f6e1fa09b7cde17f1)


.github/workflows/regressions.yml
```diff
@@ -0,0 +1,72 @@
...
+
+jobs:
+  build:
+    name: Build W3C profile
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 12
```
marcoscaceres, line 17: so olde?

.github/workflows/regressions.yml
```diff
@@ -0,0 +1,72 @@
...
+  test:
+    name: Diff ${{ matrix.source }}
+    runs-on: ubuntu-latest
+    needs: build
+    strategy:
+      matrix:
+        source:
+          - https://w3c.github.io/manifest/
+          - https://w3c.github.io/payment-request/
+          - https://w3c.github.io/trace-context/
```
marcoscaceres, line 42: maybe https://w3c.github.io/screen-wake-lock/ instead... at it's one that we control?

marcoscaceres: You can probably add geolocation too and maybe push API... they might be good as they don't change very much.

.github/workflows/regressions.yml
```diff
@@ -0,0 +1,72 @@
...
+        with:
+          name: builds
+          path: ./builds/
+      - run: git status
+      - run: node ./tools/respec2html ${{ matrix.source }} before.html -t 30 --verbose
+      - run: node ./tools/respec2html ${{ matrix.source }} after.html -t 30 --verbose --use-local
+      - name: Install git-delta # prettier git diff
+        run: |
+          set -v
+          wget -q -O delta.tar.gz https://github.com/dandavison/delta/releases/download/0.7.1/delta-0.7.1-x86_64-unknown-linux-gnu.tar.gz
```
marcoscaceres, line 64: wonder if this versioned software will become a problem to maintain? Probably not... just a thought.

marcoscaceres (approved):
Few little questions mostly... but looks good. 

I'm a bit worried that we are following a moving target though... maybe we want a fake "kitchen sink" spec that tests a whole bunch of things... that might allow us to catch more errors in practice.

.github/workflows/regressions.yml
sidvishnoi: I was thinking of converting it into a separate GitHub action, which I can update as needed.
`cargo install` took ~4min, so had to use a pre-compiled binary.

.github/workflows/regressions.yml
sidvishnoi: Using v12 everywhere else and I think spec-generator is on v12 also. So, should move all of them together. Breaking change?
Node v12 goes into maintenance 2020-11-30 FWIW.

.github/workflows/regressions.yml
sidvishnoi: Not changing much is not a concern, as we're diffing the live spec with different respec versions.

sidvishnoi:
> I'm a bit worried that we are following a moving target though

Do want to clarify that we'll run regression test on PR branches manually, when we feel like something in the spec might break.
Also, as tests will run on live spec using two different ReSpec versions, at almost same time, it's not much of a moving target.

.github/workflows/regressions.yml
sidvishnoi: `trace-context` is good for testing as it makes extensive use of `data-include="file.md" data-include-format='markdown'`.

.github/workflows/regressions.yml
marcoscaceres: We should be tied to LTS instead (everywhere)... 12 seems way too limiting.

.github/workflows/regressions.yml
marcoscaceres: true... though we really can't assure the quality of that document.

marcoscaceres:
> Might be better to test the actual practice instead of trying to copy practice in kitchen sink test?

Tradeoffs... happy to try this out for now and see how we go.

.github/workflows/regressions.yml
sidvishnoi: It's important to observe usage in the wild than in controlled environment.

marcoscaceres: Can't argue with that :)

.github/workflows/regressions.yml
marcoscaceres: Ok, but time taken on Ci shouldn't be an issue right? That's the "cost of doing business".

sidvishnoi: No issues now as I've moved this into a separate GitHub action. It takes 1s to install rather than 4min.

image

.github/workflows/regressions.yml
sidvishnoi: Started https://github.com/w3c/respec/pull/3450

.github/workflows/regressions.yml
marcoscaceres: Your meme game is strong, Padawan.

PR #3450, opened 2021-04-12
fix(package): use node v14, npm v6

sidvishnoi:


package.json
```diff
@@ -4,8 +4,8 @@
   "license": "W3C",
   "description": "A technical specification pre-processor.",
   "engines": {
-    "node": ">=10",
-    "npm": ">=5"
+    "node": ">=14",
+    "npm": ">=6"
```
sidvishnoi, line 8: Should we use npm 7 now? Seems mainstream now.

package.json
```diff
@@ -4,8 +4,8 @@
   "license": "W3C",
   "description": "A technical specification pre-processor.",
   "engines": {
-    "node": ">=10",
-    "npm": ">=5"
+    "node": ">=14",
```
marcoscaceres, line 7: I don't think the code would have run on 10... so this is a bug  fix.... but!!!!

package.json
marcoscaceres: Yeah, and this would be the breaking change.

marcoscaceres (approved):

marcoscaceres (approved):

PR #3451, opened 2021-04-12
refactor(core/linter/check-internal-slots): run as regular plugin

sidvishnoi:


profiles/w3c.js
```diff
@@ -66,6 +66,7 @@ const modules = [
   import("../src/core/custom-elements/index.js"),
   /* Linters must be the last thing to run */
   import("../src/core/linter-rules/check-charset.js"),
+  import("../src/core/linter-rules/check-internal-slots.js"),
```
sidvishnoi, line 69: Enabling on in w3c profile. Not sure if it's used anywhere though (disabled by default).

marcoscaceres: Agree.

marcoscaceres (approved):

PR #3458, opened 2021-04-15
fix(core/dfn-panel): make dfn panel keyboard accessible

### LINKED ISSUE #2825: Open dfnPanel using keyboard
Presently, [core/dfn-panel](https://github.com/w3c/respec/blob/58b9c4442d3dcc7d671a3270341d7f0da8fe4f25/src/core/dfn-panel.js) can be opened only with a mouse click. It would be good for accessibility if we can open it using keyboard. For that, `<dfn>` would need to be keyboard focusable first, and we should probably open panel on `Enter` - opening just on focus might be annoying.

marcoscaceres:
closes #2825 

cc @jnurthen. Would appreciate your input if you have time.


src/core/dfn-panel.runtime.js
```diff
@@ -97,10 +140,67 @@ function displayPanel(dfn, panel, { x, y }) {
     const caret = panel.querySelector(".caret");
     caret.style.left = `${newCaretOffset}px`;
   }
+
+  // If we are displaying because of a keydown,
+  // we want to trap focus
+  if (eventType === "keydown") {
```
jnurthen, line 146: This doesn't work for screen readers. Screen readers fire a synthetic click event when activating links so none of this logic will work. You can't branch on event type for this logic. Why not move focus for a mouse user too? You can dismiss the dialog when they click anywhere else on the screen.

jnurthen (changes requested):
appears to work nicely for keyboard users but fails for screen reader users. I think this is due to assuming that screen readers fire key events (which they normally don't) when activating links.

marcoscaceres:
@jnurthen any screen reader in particular? It works really nicely with VoiceOver at least.

jnurthen:
@marcoscaceres - not for me.... also tried with NVDA and didn't work there either.

marcoscaceres:
Had a chat with @jnurthen and he schooled me on a whole bunch of things. Updated code to reflect what we talked about... obvious on reflection, but differentiating between key and click events was obviously a bad thing. Now fixed.

marcoscaceres:
I did some investigation into testing keyboard interaction via Karma and Jasmine.... unfortunately, I couldn't find any useful resources. I did some experiments locally, but it turned into a giant hack. 

We might need to eventually move to a framework that supports keyboard events. 

Having said that, the clicks() still work fine, and give us similar enough expected behavior, so I think we are ok with what we have when it comes to tests.

src/core/dfn.js
```diff
@@ -17,6 +17,12 @@ export function run() {
       dfn.dataset.dfnType = "idl";
     }
 
+    // Make it possible to reach dfns by tabbing, allowing keyboard actions as needed.
+    dfn.tabIndex = 0;
```
sidvishnoi, line 21: Let's add tabIndex and title etc. in dfn-panel.js. Will help if we want to add an option to disable the feature in future, and keeps related code together.

sidvishnoi (commented):
Will review in detail when I get on computer. Looking good though.

tests/spec/core/dfn-spec.js
```diff
@@ -21,6 +21,22 @@ describe("Core — Definitions", () => {
...
+      <dfn>dfn 1</dfn>
+      <dfn>dfn 2</dfn>
+      <dfn>dfn 3</dfn>
+      <dfn>dfn 4</dfn>
+    </section>`;
+    const ops = makeStandardOps(null, body);
+    const doc = await makeRSDoc(ops);
+    const sec = doc.getElementById("dfns");
+    const dfns = sec.querySelectorAll("dfn");
+    expect(dfns.length).toHaveSize(4);
```
sidvishnoi, line 36:
```suggestion
    expect(dfns).toHaveSize(4);
```

src/core/dfn.js
marcoscaceres: Agree about moving the title attribute, but I'm conflicted about tabindex. I see tabbing the dfn's as independent of the dfn panel. The dfn-panel just happens to respond to the keyboard activation of this element type.

sidvishnoi: I don't think there are any other reasons why `<dfn>` should be focusable.

src/core/dfn-panel.runtime.js
```diff
@@ -1,27 +1,38 @@
...
-    /** @type {HTMLElement} */
-    const el = event.target;
+    if (!(target instanceof HTMLElement)) return;
+
+    // For keys, we only care about Enter key to activate the panel
+    // otherwise it's activated via a click.
+    if (type === "keydown" && event.key !== "Enter") return;
+
+    const action = deriveAction(event);
+    const coords = deriveCoordinates(event);
```
sidvishnoi, line 27: We only need coordinates during "show", so lets move this call there.

src/core/dfn-panel.runtime.js
```diff
@@ -32,30 +43,50 @@ function dfnPanel() {
...
+
+/**
+ * @param {MouseEvent|KeyboardEvent} event
+ */
+function deriveCoordinates(event) {
+  if (event instanceof MouseEvent) {
+    return { x: event.clientX, y: event.clientY };
+  }
+  // Otherwise, KeyboardEvent...
+  const target = /** @type HTMLElement */ (event.target);
```
sidvishnoi, line 61: Following should work fine also (cleaner IMO):

```suggestion
  /** @type HTMLElement */
  const target = event.target;
```

src/core/dfn-panel.runtime.js
```diff
@@ -32,30 +43,50 @@ function dfnPanel() {
...
+  return { x, y };
 }
 
-/** @param {HTMLElement} clickTarget */
-function deriveAction(clickTarget) {
-  const hitALink = !!clickTarget.closest("a");
-  if (clickTarget.closest("dfn, .index-term")) {
-    return hitALink ? null : "show";
+/**
+ * @param {Event} event
```
sidvishnoi, line 71:
```suggestion
 * @param {MouseEvent|KeyboardEvent} event
```

src/core/dfn-panel.runtime.js
```diff
@@ -97,10 +128,90 @@ function displayPanel(dfn, panel, { x, y }) {
...
+    switch (event.key) {
+      // Hitting "Tab" traps us in a nice loop around elements.
+      case "Tab": {
+        event.preventDefault();
+        let nextIndex = currentIndex + (event.shiftKey ? -1 : +1);
+        if (nextIndex < 0) {
+          nextIndex = lastIndex;
+        } else if (nextIndex > lastIndex) {
+          nextIndex = 0;
+        }
```
sidvishnoi, line 188: Hopefully not too clever:

```suggestion
        nextIndex = nextIndex - lastIndex * Math.floor(nextIndex / lastIndex);
```
(With `const lastIndex = anchors.length;`)

src/core/dfn-panel.runtime.js
```diff
@@ -97,10 +128,90 @@ function displayPanel(dfn, panel, { x, y }) {
...
+}
+
+/**
+ *
+ * @param {NodeListOf<HTMLAnchorElement>} anchors
+ * @param {HTMLElement} panel
+ * @param {HTMLElement} dfn
+ * @returns
+ */
+function createTrapListener(anchors, panel, dfn) {
```
sidvishnoi, line 175: Maybe we can use the trap we've in `core/ui`:
https://github.com/w3c/respec/blob/77ab0c7507d9960c50745dde954a87e199e2be81/src/core/ui.js#L94-L120

src/core/dfn-panel.runtime.js
```diff
@@ -97,10 +128,90 @@ function displayPanel(dfn, panel, { x, y }) {
...
+        }
+        break;
+
+      // Hitting "Escape" returns focus to dfn.
+      case "Escape":
+        hidePanel(panel);
+        dfn.focus();
+        break;
+    }
+    anchors.item(currentIndex).focus();
```
sidvishnoi, line 208: Otherwise it moves focus to current item again..

```suggestion
        dfn.focus();
        return;
    }
    anchors.item(currentIndex).focus();
```

sidvishnoi (changes requested):
Almost there...

src/core/dfn-panel.runtime.js
marcoscaceres: Tried that originally, but doesn't seem to like it.

src/core/dfn.js
marcoscaceres: I'd still like to keep significant modifications to dfn elements in dfn.js. Feels more natural having it here.

src/core/dfn-panel.runtime.js
marcoscaceres: This one only deals with `.target`, which is only relevant for `Event`.

src/core/dfn-panel.runtime.js
marcoscaceres: Screen Shot 2021-04-19 at 1 25 00 pm

src/core/dfn-panel.runtime.js
marcoscaceres: good point.

src/core/dfn-panel.runtime.js
marcoscaceres: I think that's bit too clever tbh (at least, I would find that to read/understand).

src/core/dfn-panel.runtime.js
marcoscaceres: Turns out the ui.js code is actually busted in Safari :( (but works fine in Chrome!) 

In Safari, we need the explicit anchor list, otherwise Safari just moves the cursor to wherever it thinks it's best... sometime to the browser's address bar... otherwise, to other links.

marcoscaceres: Filed https://github.com/w3c/respec/issues/3462

src/core/dfn.js
marcoscaceres: Ah, forgot that Indexed terms also have popup menu (and they are `<span>`!)... so relocating this code to dfn-panel.

marcoscaceres: Screen Shot 2021-04-19 at 4 25 13 pm

Now:

Screen Shot 2021-04-19 at 4 27 26 pm

sidvishnoi: Why role=link?

src/core/dfn-panel.js
```diff
@@ -24,6 +24,11 @@ export async function run() {
   const panels = document.createDocumentFragment();
   for (const el of elems) {
     panels.append(createPanel(el));
+    // Make it possible to reach el by tabbing,
+    // allowing keyboard action as needed.
+    el.tabIndex = 0;
+    el.setAttribute("role", "link");
+    el.setAttribute("aria-haspopup", "dialog");
```
jnurthen, line 31: I like having role=link - this resolves my 1 big concern that this was adding widget semantics to a structural role. Links are widgets so it is natural to focus them. 
1 improvement would be to add an accessible description to indicate that this will "show usages of foo" - the easiest way to do this would be using the title attribute.

src/core/dfn-panel.runtime.js
```diff
@@ -32,30 +43,50 @@ function dfnPanel() {
...
-  });
+  };
+}
+
+/**
+ * @param {MouseEvent|KeyboardEvent} event
+ */
+function deriveCoordinates(event) {
+  if (event instanceof MouseEvent) {
+    return { x: event.clientX, y: event.clientY };
```
jnurthen, line 58: I did notice the dialog being in the incorrect position when running with NVDA. I think this is because NVDA uses top left of the bounding box as clientX and clientY when clicking. Not sure what we can do about it (and whether it is an issue that needs resolving)

jnurthen (approved):
I'm happy to go forward as-is with the comments being addressed in subsequent issues if you want as this is a huge improvement over the current code.

src/core/dfn-finder.js
```diff
@@ -182,7 +182,10 @@ export function decorateDfn(dfnElem, idlAst, parent, name) {
     const lCaseParent = parent.toLowerCase();
     const middle = lCaseParent ? `${lCaseParent}-` : "";
     let last = name.toLowerCase().replace(/[()]/g, "").replace(/\s/g, "-");
-    if (last === "") last = "the-empty-string";
+    if (last === "") {
+      last = "the-empty-string";
+      dfnElem.setAttribute("aria-label", "the empty string");
```
sidvishnoi, line 187: Nit: gotta reduce that bundle size 😄
```suggestion
      dfnElem.setAttribute("aria-label", last);
```

src/core/dfn-panel.runtime.js
```diff
@@ -97,10 +128,90 @@ function displayPanel(dfn, panel, { x, y }) {
...
+        }
+        break;
+
+      // Hitting "Escape" returns focus to dfn.
+      case "Escape":
+        hidePanel(panel);
+        dfn.focus();
+        return;
+    }
+    anchors.item(currentIndex).focus();
```
sidvishnoi, line 208: currentIndex is updated only on Tab, so maybe move this line there.

src/core/dfn-finder.js
marcoscaceres: "the-empty-string" is different to "the empty string", so will keep this one.

sidvishnoi: Ah. Missed that.

src/core/dfn-panel.runtime.js
marcoscaceres: Yeah, I noticed that Chrome + VO sometimes does the same thing... but sometimes it makes the event's `.clientX = 0`. I'd be hesitant to fix that unless it becomes a problem, as it seems to be a browser bug.

jnurthen: I'm wondering if we get a zero (highly unlikely) we simulate a click in the middle of the bounding box

marcoscaceres: Chrome indeed sends clientX/clientY zeros... this kinda bad because it makes AT detectable. Anyway, added some code to handle that.

src/core/dfn.js
marcoscaceres: See first point in https://github.com/w3c/respec/issues/2825#issuecomment-811516805

src/core/dfn-panel.js
```diff
@@ -41,15 +41,28 @@ function createPanel(dfn) {
...
+      aria-modal="true"
+      aria-label="Links in this document to definition: ${definition}"
+    >
       <span class="caret"></span>
       <div>
-        <a class="self-link" href="${href}">Permalink</a>
+        <a
+          class="self-link"
+          href="${href}"
+          aria-label="Permalink for definition: ${definition}"
```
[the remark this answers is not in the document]
marcoscaceres: Expanded description to say the permalink closes the dialog.

marcoscaceres:
Thanks again for your help @jnurthen! 

@sidvishnoi, should be all good to go 🤞

src/core/dfn-panel.runtime.js
```diff
@@ -97,10 +137,86 @@ function displayPanel(dfn, panel, { x, y }) {
...
+      // Hitting "Tab" traps us in a nice loop around elements.
+      case "Tab": {
+        event.preventDefault();
+        let nextIndex = currentIndex + (event.shiftKey ? -1 : +1);
+        if (nextIndex < 0) {
+          nextIndex = lastIndex;
+        } else if (nextIndex > lastIndex) {
+          nextIndex = 0;
+        }
+        currentIndex = nextIndex;
```
sidvishnoi, line 198:
```suggestion
        currentIndex += event.shiftKey ? -1 : +1;
        if (currentIndex < 0) {
          currentIndex = lastIndex;
        } else if (currentIndex > lastIndex) {
          currentIndex = 0;
        }
```

src/core/dfn-panel.runtime.js
```diff
@@ -97,10 +137,86 @@ function displayPanel(dfn, panel, { x, y }) {
...
+        anchors.item(currentIndex).focus();
+        break;
+      }
+
+      // Hitting "Enter" on an anchor releases the trap.
+      case "Enter":
+        hidePanel(panel);
```
sidvishnoi, line 205: The click behavior is to "dock" the panel on anchor click. Is this due to focus issues?

sidvishnoi (approved):
Great improvement over the current version!

src/core/dfn-panel.runtime.js
marcoscaceres: just feels more natural to close and move the cursor... and yes, it's about focus too.

sidvishnoi: Let's stick with hide for now. Docking will require to move the focus to permalink as well as add ability to jump back into panel, which complicates things.

marcoscaceres:
Great work everyone! Bouncing around specs with keyboard is now quite nice. 

There are a few browser bugs:

  * Chrome seems to  .focus(), but doesn't visually show the focus.  
  * [Firefox fires two events](https://bugzilla.mozilla.org/show_bug.cgi?id=1695636) (keyup + click) - hi @saschanaz!! ❤️

but we will need to fix those in browsers.

Issue #3442, opened 2021-04-10
WebIDL CSS isn't injected


sidvishnoi:
## Important info

- URL to affected spec: https://wicg.github.io/uuid/
- ReSpec version: 26.5.0
- [x] I did a "hard refresh", but it's still busted.

PR #3460, opened 2021-04-17
fix(core/webidl): ensure loading of webidl.css

sidvishnoi:
Fixes https://github.com/w3c/respec/issues/3442

### DIFF
--- a/src/core/webidl.js
+++ b/src/core/webidl.js
@@ -382,14 +382,9 @@ export async function run() {
   if (!idls.length) {
     return;
   }
-  if (!document.querySelector(".idl:not(pre), .webidl:not(pre)")) {
-    const link = document.querySelector("head link");
-    if (link) {
-      const style = document.createElement("style");
-      style.textContent = css;
-      link.before(style);
-    }
-  }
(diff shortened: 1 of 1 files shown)


src/core/webidl.js
```diff
@@ -382,14 +382,9 @@ export async function run() {
...
   }
-  if (!document.querySelector(".idl:not(pre), .webidl:not(pre)")) {
-    const link = document.querySelector("head link");
-    if (link) {
-      const style = document.createElement("style");
-      style.textContent = css;
-      link.before(style);
-    }
-  }
+  const style = document.createElement("style");
```
marcoscaceres, line 385: But shouldn't there be some IDL defined before we add the stylesheet? I wonder what went wrong in that other spec?

sidvishnoi: The check in previous block ensures that. (early return if no `pre.idl, pre.webidl`)

marcoscaceres: Oh, lol. I see now. Yeah, this was a silly bug.

marcoscaceres (approved):

PR #3475, opened 2021-04-21
chore(tools/release): validate HTML before release

marcoscaceres:


sidvishnoi:
I think https://github.com/w3c/respec/pull/3473 should be enough.

marcoscaceres:
Not sure, the validator also gets updated periodically - it could happen between a merge and a release. I'd rather have this check in to be as sure as we can be.

tools/release.js
```diff
@@ -7,6 +7,8 @@ const { exec } = require("child_process");
 const loading = require("loading-indicator");
 const MAIN_BRANCH = "develop";
 const DEBUG = false;
+const vnu = require("vnu-jar");
+const tmp = require("tmp");
```
sidvishnoi, line 11: Can we avoid adding this dependency (and its dependency) for something Node.js allows easily?

sidvishnoi:
```js
const tempFile = path.join(os.tmpdir(), "index.html");
```

sidvishnoi (approved):

PR #3491, opened 2021-04-27
fix(core/inlines): don't link IDL inside dfn

### LINKED ISSUE #3482: dfn.role=link containing a link
## Important info

- URL to affected spec: https://w3c.github.io/webrtc-pc/ (if it were updated to ReSpec 26.6.0)
- ReSpec version: 26.6.0

## Description of problem
`<dfn data-cite=url>` generates markup à la `<dfn role=link><a href=url></a></dfn>` - I presume the `role=link` is for the definition panel, but it is not valid to have an element with `role=link` encompassing a link element.

marcoscaceres:
closes #3482


src/core/inlines.js
```diff
@@ -131,13 +131,20 @@ function inlineRefMatches(matched) {
...
+ * @param {Text} text
  */
-function inlineXrefMatches(matched) {
+function inlineXrefMatches(matched, text) {
   // slices "{{" at the beginning and "}}" at the end
-  const ref = matched.slice(2, -2).trim();
-  return ref.startsWith("\\")
-    ? matched.replace("\\", "")
-    : idlStringToHtml(norm(ref));
+  const ref = norm(matched.slice(2, -2).trim());
```
sidvishnoi, line 138: `norm` does the trim:
```suggestion
  const ref = norm(matched.slice(2, -2));
```

src/core/inlines.js
```diff
@@ -131,13 +131,20 @@ function inlineRefMatches(matched) {
...
+  const ref = norm(matched.slice(2, -2));
+  if (ref.startsWith("\\")) {
+    return matched.replace("\\", "");
+  }
+
+  const node = idlStringToHtml(ref);
+  // If it's inside a dfn, it should just be coded, not linked.
+  // This is because dfn elements are treated as links by ReSpec via role=link.
+  const renderAsCode = !!text.parentElement.closest("dfn");
+  return renderAsCode ? inlineCodeMatches(`\`${node.textContent}\``) : node;
```
sidvishnoi, line 147: Minor optimization:
```suggestion
  // If it's inside a dfn, it should just be coded, not linked.
  // This is because dfn elements are treated as links by ReSpec via role=link.
  const renderAsCode = !!text.parentElement.closest("dfn");
  return renderAsCode ? inlineCodeMatches(`\`${node.textContent}\``) : idlStringToHtml(ref);
```

src/core/inlines.js
```diff
@@ -131,13 +131,20 @@ function inlineRefMatches(matched) {
...
-    : idlStringToHtml(norm(ref));
+  const ref = norm(matched.slice(2, -2));
+  if (ref.startsWith("\\")) {
+    return matched.replace("\\", "");
+  }
+
+  const node = idlStringToHtml(ref);
+  // If it's inside a dfn, it should just be coded, not linked.
+  // This is because dfn elements are treated as links by ReSpec via role=link.
+  const renderAsCode = !!text.parentElement.closest("dfn");
```
sidvishnoi, line 146: Would it make sense if we do `parentElement.localName === "dfn"`? If `text` is not inside a `<dfn>`, it'll end up traversing till root, which is the general case.

As an aside, `.closest()` needs a maxDepth option.

src/core/inlines.js
```diff
@@ -287,38 +294,36 @@ export function run(conf) {
...
-        const node = inlineRFC2119Matches(t);
-        df.append(node);
-      } else {
-        // FAIL -- not sure that this can really happen
-        throw new Error(
-          `Found token '${t}' but it does not correspond to anything`
-        );
+        continue;
+      }
+      switch (true) {
```
sidvishnoi, line 299: Let's keep this refactor out of this PR.

src/core/inlines.js
marcoscaceres: I'm worried that it won't catch all cases. Like if it's a span or whatever.

sidvishnoi (approved):

PR #3492, opened 2021-04-27
refactor(core/inlines): use switch instead of ifs

marcoscaceres:
Because it's beautiful 🤺


src/core/inlines.js
```diff
@@ -294,38 +294,36 @@ export function run(conf) {
...
-        const node = inlineAbbrMatches(t, txt, abbrMap);
-        df.append(node);
-      } else if (keywords.test(t)) {
-        const node = inlineRFC2119Matches(t);
-        df.append(node);
-      } else {
-        // FAIL -- not sure that this can really happen
-        throw new Error(
-          `Found token '${t}' but it does not correspond to anything`
-        );
```
sidvishnoi, line 328: Might be useful to keep this..

sidvishnoi (approved):
‮🤺‮‮

src/core/inlines.js
marcoscaceres: As we've never hit this in the last N years, I figured we are safe to just remove it. Something has to match so in *theory* this won't ever trigger... I'm feeling confident in removing it.

PR #3500, opened 2021-04-30
fix(core/utils): removes attributes from dfn when calling makeSafeCopy

### LINKED ISSUE #3486: Automatic link to sub-sections duplicate definition markup
## Important info

- URL to affected spec: https://w3c.github.io/webrtc-pc/webrtc.html
- ReSpec version: 26.6.1

## Description of problem
The WebRTC spec uses automatic links to one of its subsection with `<a href="#rtcicetransport"></a>`.
This gets expanded with the full markup of the title of that section, including the `<dfn>` of the `RTCIceTransport` name (although as a `span dfn-type=interface` rather than a `dfn` element). That definition then gets wrapped into the dfn panel popup trigger - the panel never actually gets shown because the link takes precedence, but it doesn't make sense for the definition to be copied over, and with the latest accessibility improvements, it creates a situation where an element with `tabindex` gets embedded in a link, which the markup validator flags as an error.

marcoscaceres:
closes #3486


tests/spec/core/anchor-expander-spec.js
```diff
@@ -78,7 +78,8 @@ describe("Core - anchor-expander", () => {
     expect(doc.querySelector("#expansion *[id]")).toBeNull();
 
     const [firstSpan, secondSpan] = doc.querySelectorAll("#expansion a > span");
-    expect(firstSpan.title).toBe("pass");
+    // title get dropped from dfns
```
marcoscaceres, line 81:
```suggestion
    // title gets dropped from dfns
```

sidvishnoi (approved):

PR #3507, opened 2021-05-04
fix(core/anchor-expander): don't copy attributes when expanding

### LINKED ISSUE #3486: Automatic link to sub-sections duplicate definition markup

marcoscaceres:
Should close #3486

### DIFF
--- a/src/core/utils.js
+++ b/src/core/utils.js
@@ -552,14 +552,23 @@ export function getLinkTargets(elem) {
  * Changes name of a DOM Element
  * @param {Element} elem element to rename
  * @param {String} newName new element name
+ * @param {Object} options
+ * @param {boolean} options.copyAttributes
+ *
  * @returns {Element} new renamed element
  */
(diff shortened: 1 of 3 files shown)


sidvishnoi (approved):
🤞

PR #3508, opened 2021-05-05
tests(w3c/group-spec): ipr disclousures have new URLs

marcoscaceres:
This is for upcoming changes to the ipr disclosure URLs.


marcoscaceres:
Blocked on https://github.com/w3c/w3c-api/issues/103

sidvishnoi:
Is there any use of `wgId` after this change?

marcoscaceres:
> Is there any use of wgId after this change?

Hopefully not... if we get short wg names from the w3c, then those IDs can go away. 

I'll find out.

PR #3509, opened 2021-05-05
fix(lint/wpt-tests-exist): make separate warnings

marcoscaceres:
Easier to work with: 

Screen Shot 2021-05-05 at 3 08 01 pm

### DIFF
--- a/src/core/linter-rules/wpt-tests-exist.js
+++ b/src/core/linter-rules/wpt-tests-exist.js
@@ -10,8 +10,9 @@ export const name = "core/linter-rules/wpt-tests-exist";
 
 const localizationStrings = {
   en: {
-    msg: "Non-existent Web Platform Tests",
-    hint: "Please fix the tests mentioned.",
+    msg: "The following tests could not be found in Web Platform Tests:",
+    hint:
+      "Check [wpt.live](https://wpt.live) to see if they've been deleted or renamed.",
   },
 };
 const l10n = getIntlData(localizationStrings);
--- a/src/core/linter-rules/wpt-tests-exist.js
+++ b/src/core/linter-rules/wpt-tests-exist.js
@@ -10,9 +10,9 @@ export const name = "core/linter-rules/wpt-tests-exist";
 
 const localizationStrings = {
   en: {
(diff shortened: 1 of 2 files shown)


sidvishnoi (approved):
Do check how it looks from command line.

marcoscaceres:
Good point. Will do.

marcoscaceres:
Not the worse... not the best...: 

Screen Shot 2021-05-05 at 5 00 03 pm

marcoscaceres:
Is the above fixable to remove the HTML?

sidvishnoi:
I think this will require updating the markdown renderer in respec2html.

sidvishnoi:
I was also wondering if it would make sense to showWarning for each element with missing tests, instead of displaying unique tests across the document. Might help in hunting down bad tests.

marcoscaceres:
> I was also wondering if it would make sense to showWarning for each element with missing tests, instead of displaying unique tests across the document. Might help in hunting down bad tests.

Yeah, that might be better... having a single warning for a bunch of broken things is not amazing.

marcoscaceres:
ok, better... 

Screen Shot 2021-05-06 at 4 03 42 pm

sidvishnoi (approved):
Don't forget to update commit msg before merge.

src/core/linter-rules/wpt-tests-exist.js
```diff
@@ -26,35 +27,22 @@ export async function run(conf) {
...
       .map(test => test.trim().split("#")[0])
-      .filter(test => test);
-
-    const missingTests = tests.filter(test => !filesInWPT.has(test));
-    if (missingTests.length) {
-      offendingElements.push(elem);
-      missingTests.forEach(test => offendingTests.add(test));
-    }
+      .filter(test => test && !filesInWPT.has(test))
+      .map(missingTest => {
```
sidvishnoi, line 39: Maybe group by element?

marcoscaceres: I prefer them as they are for now. But will update them after I play around bit.

PR #3539, opened 2021-05-18
Changed toBe(null) to toBeNull()

### LINKED ISSUE #3538: Tests `toBe(null)`, should use `toBeNull()`
We have a lot of tests that use `toBe(null)`... but should use `toBeNull()`.

```
git grep "toBe(null)" tests
```

himanshu007-creator:
resolves #3538


marcoscaceres (approved):
Awesome! thanks @himanshu007-creator

PR #3548, opened 2021-05-20
fix(core/ui): display plugin name after error msg

marcoscaceres:
The plugin name itself is not very useful for regular users, but useful for debugging... so putting it last. 

Screen Shot 2021-05-20 at 12 08 11 pm

### DIFF
--- a/src/core/ui.js
+++ b/src/core/ui.js
@@ -264,7 +264,7 @@ function rsErrorToHTML(err) {
     return err;
   }
 
-  const plugin = err.plugin ? ` <small>Plugin: "${err.plugin}"</small>` : "";
+  const plugin = err.plugin ? ` <small>(Plugin: "${err.plugin}")</small>.` : "";
   const hint = err.hint ? ` ${err.hint}` : "";
   const elements = Array.isArray(err.elements)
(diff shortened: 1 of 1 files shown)


src/core/ui.js
```diff
@@ -273,7 +273,7 @@ function rsErrorToHTML(err) {
     ? `\n\n<details>\n${err.details}\n</details>\n`
     : "";
 
-  const text = `${plugin}${err.message}${hint}${elements}${details}`;
+  const text = `${err.message}${hint}${elements}${details}${plugin}.`;
```
sidvishnoi, line 276: Maybe before details?

src/core/ui.js
```diff
@@ -264,7 +264,7 @@ function rsErrorToHTML(err) {
     return err;
   }
 
-  const plugin = err.plugin ? `(${err.plugin}): ` : "";
+  const plugin = err.plugin ? ` <small>Plugin: "${err.plugin}"</small>` : "";
```
sidvishnoi, line 267: Keep the parenthesis too?

sidvishnoi (approved):

Issue #3569, opened 2021-05-31
Deprecate tocIntroductory- make default behave as tocIntroductory = true


marcoscaceres:
**Is your feature request related to a problem? Please describe.**
The `tocIntroductory` should be deprecated and all documents should have all section headers added to the ToC. This makes documents more accessible be including quick links to the abstract, status of this document, etc.  

**Describe the solution you'd like**
For `tocIntroductory` to go away, and, if Editors don't want something to appear in the ToC, they can just explicitly "no-toc" a section.

PR #3579, opened 2021-05-31
fix(core/structure): include introductory sections in ToC by default

sidvishnoi:
Fixes https://github.com/w3c/respec/issues/3569
- Added a deprecation msg on use of `tocIntroductory`.
- `tocIntroductory: true` by default.

TODO:
- [x] Remove/deprecate from docs:
  - [x] https://respec.org/docs/#introductory
  - [x] https://respec.org/docs/#tocIntroductory
- [ ] Remove from ReSpec: `class="introductory"`, unless it has some special meaning outside ReSpec. It can lead of change in header IDs though.


sidvishnoi:
Not as simple as I imagined it to be 😄

src/core/structure.js
```diff
@@ -194,9 +192,12 @@ function filterHeader(h) {
 }
 
 export function run(conf) {
-  if ("tocIntroductory" in conf === false) {
-    conf.tocIntroductory = false;
+  if ("tocIntroductory" in conf) {
+    const msg = "Configuration option `tocIntroductory` is deprecated.";
+    const hint = `Add a 'notoc' class to remove a section from Table of Contents.`;
```
marcoscaceres, line 197:
```suggestion
    const hint = `Add a 'notoc' class to remove a section from the Table of Contents.`;
```

marcoscaceres (approved):
Not sure about that test fail, but otherwise looks good!

src/core/structure.js
```diff
@@ -194,9 +192,12 @@ function filterHeader(h) {
 }
 
 export function run(conf) {
-  if ("tocIntroductory" in conf === false) {
-    conf.tocIntroductory = false;
+  if ("tocIntroductory" in conf) {
```
marcoscaceres, line 195: I think we should only warn when `tocIntroductory` is false... people doing `tocIntroductory: true` are already doing the right thing.  

```suggestion
  if ("tocIntroductory" in conf && tocIntroductory === "false") {
```

marcoscaceres:
Updated docs: https://github.com/w3c/respec/wiki/tocIntroductory + https://github.com/w3c/respec-web-services/pull/211

marcoscaceres:
@sidvishnoi, any timeframe on this one? Want me to take it over?

sidvishnoi:
Feel free to take over. Need to identify why some tests are failing and update them (hopefully with minimal changes).

marcoscaceres:
Firefox failures are very odd... can't reproduce locally :(

PR #3580, opened 2021-05-31
fix(core/linter/local-refs-exist): check links not starting with hash

sidvishnoi:
Had a URL like `http://localhost:3000/docs/#/docs%23examples` when serving a directory having a `docs` folder (not serving from docs as root). Found via spec-prod.

### DIFF
--- a/src/core/linter-rules/local-refs-exist.js
+++ b/src/core/linter-rules/local-refs-exist.js
@@ -22,7 +22,7 @@ export function run(conf) {
   }
 
   /** @type {NodeListOf<HTMLAnchorElement>} */
-  const elems = document.querySelectorAll("a[href^='#']");
+  const elems = document.querySelectorAll("a[href]");
   const offendingElements = [...elems].filter(isBrokenHyperlink);
   if (offendingElements.length) {
     showWarning(l10n.msg, name, {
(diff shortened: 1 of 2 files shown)


src/core/linter-rules/local-refs-exist.js
```diff
@@ -21,9 +21,22 @@ export function run(conf) {
...
+  /**
+   * Skip links without hashes; and those that point to outside current page.
+   * @returns
+   */
+  const shouldTest = (() => {
+    const { origin, pathname } = new URL(document.baseURI);
+    /** @type {(a: HTMLAnchorElement) => boolean} */
+    const fn = a => a.hash && a.origin === origin && a.pathname === pathname;
+    return fn;
+  })();
```
marcoscaceres, line 33: Might be more consistent just to make is a regular function above `isBrokenHyperlink()` ... can still curried , but just maybe just:

 ```suggestion
  const shouldTest = checkHashSameOrigin();
```

Or even:
```
   .filter(checkHashSameOrigin())
```

marcoscaceres (approved):
Just a style nit. Good find!

sidvishnoi:
I might have created an impossible case... will investigate. Lets not merge until then.

sidvishnoi:
Will fix if issue appears in real use case.

PR #3588, opened 2021-06-02
refactor(worker/respec-worker): use highlight() instead of highlightAuto()

### LINKED ISSUE #3581: Investigate new stuff in hljs 11
There is some new stuff in hljs https://github.com/w3c/respec/pull/3576 ... we should investigate what Changes to make. In particular, it would be nice to drop `hightlightAuto()` and move to just `hightlight()`.

Patil2099:
Fixes  #3581


marcoscaceres:
Thanks @Patil2099! Would you mind reverting the changes to all the build files?

builds/respec-worker.js
```diff
@@ -18,7 +18,7 @@ self.addEventListener("message", ({ data: originalData }) => {
       const { code } = data;
       const langs = data.languages.length ? data.languages : undefined;
       try {
-        const { value, language } = self.hljs.highlightAuto(code, langs);
+        const { value, language } = self.hljs.highlight(code, langs);
```
marcoscaceres, line 21: I don't think this is right, the signature for highlight is: ` highlight(code, {language, ignoreIllegals})`.

Please see https://github.com/highlightjs/highlight.js/blob/main/VERSION_11_UPGRADE.md 

We need to decide if `ignoreIllegals` should be true or not.

builds/respec-worker.js
```diff
@@ -18,7 +18,7 @@ self.addEventListener("message", ({ data: originalData }) => {
       const { code } = data;
       const langs = data.languages.length ? data.languages : undefined;
```
marcoscaceres, line 19: We need to probably want something like: 

```suggestion
      const langs = data.languages.length ? data.languages : ["text"];
```

builds/respec-worker.js
marcoscaceres: There should probably only be one language coming in too... you will need to check why we are sending in multiple languages. If we are sending in unknown languages, probably want to filter those out and we should only end up with 1 language. 

Additionally, you need to find a suitable default (something like "plaintext") - and we want "plaintext" to be included in the the highlighter build steps.

marcoscaceres (changes requested):
Few things to fix here.

marcoscaceres:
Oops, all the comments on builds/respec-worker.js are for worker/respec-worker.js 

All the build things are built automatically, so shouldn't be committed.

marcoscaceres:
Closing due to inactivity.

PR #3597, opened 2021-06-06
refactor(styles/webidl): use CSS :is()

marcoscaceres:


sidvishnoi:
Good thing we've HTML validator integrated in CI 😄

marcoscaceres:
Yeah... 😭

src/styles/webidl.css.js
```diff
@@ -97,24 +97,18 @@ a.idlEnumItem {
   box-shadow: 0 0 5px rgba(81, 167, 232, 0.5);
 }
 
-.respec-button-copy-paste:focus:hover,
-.respec-button-copy-paste.selected:focus {
+.respec-button-copy-paste:is(:focus:hover,.selected:focus) {
```
sidvishnoi, line 100: As a nit, can you remove the `prettier-ignore` on top, run prettier, and de-indent code again?
(We should remove that prettier-ignore, but I don't like commits with only whitespace changes.)

marcoscaceres: let's do that as a followup. It will mess up the other PRs, which need to sit probably for a few months until the CSS validator gets fixed.

PR #3598, opened 2021-06-06
refactor(styles/examples): use CSS is()

marcoscaceres:
Blocked on https://github.com/w3c/css-validator/issues/304

### DIFF
--- a/src/styles/examples.css.js
+++ b/src/styles/examples.css.js
@@ -8,8 +8,7 @@ span.example-title {
   text-transform: none;
 }
 
-aside.example,
-div.example,
+:is(aside,div).example,
 div.illegal-example {
   padding: 0.5em;
   margin: 1em 0;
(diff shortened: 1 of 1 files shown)


src/styles/examples.css.js
```diff
@@ -39,6 +36,7 @@ aside.example div.example {
   border-color: #999;
   background: #fff;
 }
+
 aside.example div.example span.example-title {
```
sidvishnoi, line 40: I feel like there should've been some commas here.. 🤔

marcoscaceres: yeah, that's weird.... maybe we can remove this entirely.

src/styles/examples.css.js
```diff
@@ -39,6 +36,7 @@ aside.example div.example {
...
   color: #999;
 }
```
marcoscaceres, line 42:
```suggestion
```

PR #3605, opened 2021-06-06
fix(styles/caniuse): use cursor: default

marcoscaceres:
The cursor is confusing, as you can't click on the versions.

### DIFF
--- a/src/styles/caniuse.css.js
+++ b/src/styles/caniuse.css.js
@@ -9,7 +9,7 @@ export default css`
   flex-wrap: wrap;
   justify-content: flex-start;
   align-items: baseline;
-  cursor: default;
+  cursor: auto;
 }
 
 button.caniuse-cell {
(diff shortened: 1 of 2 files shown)


src/styles/caniuse.css.js
```diff
@@ -9,7 +9,6 @@ export default css`
   flex-wrap: wrap;
   justify-content: flex-start;
   align-items: baseline;
-  cursor: pointer;
```
sidvishnoi, line 12: Should probably use `cursor: default` instead of pointer or the default `text`?

marcoscaceres: checking...

marcoscaceres: `text` is no good and it's the text selection cursor. And `cursor: default` is the same as none.

sidvishnoi (approved):

Issue #3566, opened 2021-05-28
Provide a flag to disable unitTestServer or integrationTestServer 


marcoscaceres:
**Is your feature request related to a problem? Please describe.**
When testing, the dev server spins up multiple Karma instances. This is frustrating because they stack ontop of each other, making it difficult to debug tests.  

**Describe the solution you'd like**

It would be great to disable integration test or the unitTest server.

sidvishnoi:
I wonder if we can modify karmaConfig to run both at once in devServer.

marcoscaceres:
possibly, tough I don't mind them split up.

PR #3621, opened 2021-06-10
chore(tools/dev-server): run unit tests, integration tests together

sidvishnoi:
Fixes https://github.com/w3c/respec/issues/3566
@marcoscaceres can you test your workflow locally? CI is OK.


marcoscaceres (approved):
This is good in that it solves the issue. However, as a follow up, we should probably merge all these into a single config again, or find some better name for the files, like: 

```
files["unit-tests"];
files["specs"];
files["supportFiles"];
```

PR #3624, opened 2021-06-12
refactor: use nullish coalescing

marcoscaceres:


marcoscaceres:
Test failing because of #3626

src/core/xref.js
```diff
@@ -132,7 +132,7 @@ function normalizeConfig(xref) {
       if (xref.profile) {
         const profile = xref.profile.toLowerCase();
         if (profile in profiles) {
-          const specs = (xref.specs || []).concat(profiles[profile]);
+          const specs = xref.specs?.concat(profiles[profile]) ?? [];
```
sidvishnoi, line 135: The result is not the same here.

sidvishnoi, line 135: The result is not the same here.

sidvishnoi (changes requested):

src/core/xref.js
marcoscaceres: ah, right.

sidvishnoi (approved):
Would love to find more potential uses and create a larger PR 😄

PR #3625, opened 2021-06-12
fix(w3c/headers): allow overriding latestVersion:

### LINKED ISSUE #3618: Allow latestVersion to be overridden by the author
For latestVersion, sometimes it's good to point directly to another URL. (e.g., for level specs ("/TR/foo-3/), sometimes you want latestVersion to be "TR/foo/" )

latestVersion to take a URL... like

https://w3c.github.io/pointerevents/?latestVersion="pointerevents" (resolved against w3c/TR/!) 

**Describe alternatives you've considered**
Using `level` doesn't help. 

cc @plehegar @caribouW3

marcoscaceres:
Closes #3618

For latestVersion, sometimes it's good to point directly to another URL. (e.g., for level specs ("/TR/foo-3/), sometimes you want latestVersion to be "TR/foo/" )


src/w3c/headers.js
```diff
@@ -63,7 +63,7 @@
 //      PAG on the document.
 //  - thisVersion: the URI to the dated current version of the specification. ONLY ever use this for CG/BG
 //      documents, for all others it is autogenerated.
-//  - latestVersion: the URI to the latest (undated) version of the specification. ONLY ever use this for CG/BG
+//  - latestVersion: the URI to the latest version of the specification.
 //      documents, for all others it is autogenerated.
```
saschanaz, line 67: Should the next line be fixed too?

src/w3c/headers.js
```diff
@@ -108,6 +108,17 @@ const W3CDate = new Intl.DateTimeFormat(["en-AU"], {
   day: "2-digit",
 });
 
+//
+/**
+ * Tagged template TR_URL`path`. Resolves against W3C's /TR/
+ * */
+function W3URL(strings, ...values) {
```
saschanaz, line 115: I think typically template string functions have camel cases 👀

src/w3c/headers.js
```diff
@@ -108,6 +108,17 @@ const W3CDate = new Intl.DateTimeFormat(["en-AU"], {
   day: "2-digit",
 });
 
+//
+/**
+ * Tagged template TR_URL`path`. Resolves against W3C's /TR/
```
saschanaz, line 113: (Name doesn't match)

src/w3c/headers.js
```diff
@@ -401,21 +417,20 @@ export function run(conf) {
...
-    const pmat = status2maturity[conf.previousMaturity]
-      ? status2maturity[conf.previousMaturity]
-      : conf.previousMaturity;
-    if (conf.isTagFinding) {
-      conf.prevVersion = `${conf.latestVersion}-${ISODate.format(
-        conf.previousPublishDate
-      )}`;
-    } else if (conf.isCGBG) {
+    const prevMaturity =
+      status2maturity[conf.previousMaturity] || conf.previousMaturity;
```
saschanaz, line 421: Can we use `??` operator here?

src/w3c/headers.js
```diff
@@ -108,6 +108,17 @@ const W3CDate = new Intl.DateTimeFormat(["en-AU"], {
...
 });
 
+//
+/**
+ * Tagged template TR_URL`path`. Resolves against W3C's /TR/
+ * */
+function W3URL(strings, ...values) {
+  const str = strings.reduce((str, part, i) => {
+    return str + part + (values[i] || "");
+  }, "");
```
saschanaz, line 118: I wonder whether map+join is more readable here.

saschanaz (commented):
BTW, is this only to skip `level` from the URL? I wonder whether we should add an option to skip it instead of overriding the whole path 🤔

marcoscaceres (commented):
Thanks for the great feedback @saschanaz. I think I'd like to keep allowing `latestVersion` to be overridden instead of introducing a new pref. We were already allowing it for BG/CG, so this just makes it consistent (+ it's one less special case to think about).

src/w3c/headers.js
```diff
@@ -108,6 +107,14 @@ const W3CDate = new Intl.DateTimeFormat(["en-AU"], {
   day: "2-digit",
 });
 
+/**
+ * Tagged template. Resolves against https://www.w3c.org.
```
sidvishnoi, line 111: Why not use a regular function instead?

saschanaz: There are heavy uses of template string so I think it's better to be a template string function.

saschanaz: Hmm no, it can just be a wrapper without actual need for TSF...

marcoscaceres: yeah, a function is probably fine. Doesn't look as cool tho 😎🤪.

marcoscaceres:
Added docs https://github.com/w3c/respec/wiki/latestVersion

marcoscaceres:
Sent https://github.com/w3c/respec-web-services/pull/210

sidvishnoi (approved):

saschanaz (approved):
👍

src/w3c/headers.js
```diff
@@ -108,6 +107,15 @@ const W3CDate = new Intl.DateTimeFormat(["en-AU"], {
   day: "2-digit",
 });
 
+/**
+ * Resolves against https://www.w3.org.
+ * @param {string} href
+ * @returns
```
saschanaz, line 113: Oh BTW, `{URL}` or just remove this?

marcoscaceres: Will remove. Returns a string.

marcoscaceres, line 113:
```suggestion
```

PR #3638, opened 2021-06-16
refactor(core/utils): use Intl.ListFormat for joining

marcoscaceres:


src/type-helper.d.ts
```diff
@@ -81,6 +81,10 @@ declare function fetch(input: URL, init?: RequestInit): Promise<Response>;
 
 declare namespace Intl {
   class ListFormat {
+    formatToParts(items: string): {
+      type: "element" | "literal";
+      value: string;
+    };
```
sidvishnoi, line 87:
```suggestion
    }[];
```

sidvishnoi, line 87: This should remove the need for ts-ignore in `makeListJoiner`
```suggestion
    }[];
```

sidvishnoi (commented):
I like where this is going! Let's add TS-types to all the new utils though..

src/core/utils.js
```diff
@@ -104,41 +92,54 @@ function markAsOffending(elem, msg, title) {
...
+   * @param {(value: string, index: number, array: string[]) => any} [mapper]
+   */
+  return (items, mapper) => {
+    let elemCount = 0;
+    if (!mapper) {
+      return [formatter.format(items)];
+    }
+    return formatter.formatToParts(items).map(({ type, value }) => {
+      if (type === "element") {
+        return mapper(value, elemCount++, items);
```
sidvishnoi, line 112: Looks like we don't use it.
```suggestion
        return mapper(value, elemCount++);
```

src/core/utils.js
```diff
@@ -104,41 +92,54 @@ function markAsOffending(elem, msg, title) {
...
+ * @param {"conjunction"|"disjunction"} type
+ * @param {"long"|"narrow"} style
+ */
+function joinFactory(type, style = "long") {
+  const formatter = new Intl.ListFormat(docLang, { style, type });
+  /**
+   * @param {string[]} items
+   * @param {(value: string, index: number, array: string[]) => any} [mapper]
+   */
+  return (items, mapper) => {
```
sidvishnoi, line 105: How about something like:
```ts
return (items: T[], mapper?: (item: T) => any), getter?: (item: T) => string) => {}
```

src/core/utils.js
marcoscaceres: I'd rather sent the interface compliant map.

marcoscaceres: (even though we don't use it)

src/core/utils.js
marcoscaceres: Added.

marcoscaceres:
hopefully all good now.

src/core/utils.js
sidvishnoi: I meant:
```diff
-   return (items, mapper) => { 
+   return (items: T[], mapper?: (item: T) => any), getter?: (item: T) => string) => {
+   // above, but in JS (not TS)
```

src/core/utils.js
```diff
@@ -104,41 +92,56 @@ function markAsOffending(elem, msg, title) {
...
+ * @param {"conjunction"|"disjunction"} type
+ * @param {"long"|"narrow"} style
+ */
+function joinFactory(type, style = "long") {
+  const formatter = new Intl.ListFormat(docLang, { style, type });
+  /**
+   * @template T
+   * @param {string[]} items
+   * @param {(value: string, index: number, array: string[]) => any} [mapper]
+   * @return (items: T[], mapper?: (item: T) => any), getter?: (item: T) => string) => {}
```
sidvishnoi, line 105: Let's ignore this for now. We can make it more generic next time we need it.

```suggestion
```

src/core/utils.js
sidvishnoi: Let's ignore this for now. We can make it more generic next time we need it.

src/core/utils.js
```diff
@@ -104,41 +92,55 @@ function markAsOffending(elem, msg, title) {
...
+    let elemCount = 0;
+    if (!mapper) {
+      return [formatter.format(items)];
+    }
+    return formatter.formatToParts(items).map(({ type, value }) => {
+      if (type === "element") {
+        return mapper(value, elemCount++, items);
+      }
+      return value;
+    });
```
sidvishnoi, line 116:
```suggestion
    return formatter.formatToParts(items).map(({ type, value }) => {
      if (type === "element" && mapper) {
        return mapper(value, elemCount++, items);
      }
      return value;
    });
```

src/core/utils.js
sidvishnoi: Actually maybe we can?

sidvishnoi (approved):
Looks good.
Would like to remove the "must pass string[] to `joinAnd`" requirement in future, and make the interface more like `Array.map`.

Issue #3644, opened 2021-06-18
`<dfn data-dfn-type=idl>` is incorrectly allowed without error


tabatkins:
## Important info

- URL to affected spec: <https://www.w3.org/TR/payment-request/#dfn-addressline> (or <https://rawgit.com/w3c/payment-request/79e23d46c704141915005913829e21d13c6cfb44/index.html#dfn-addressline> for the last "live" version that contained the relevant dfn)
- ReSpec version: 26.8.14
- [x] I did a "hard refresh", but it's still busted.

## Description of problem

**What happened (e.g., it crashed)?:** The `[[addressLine]]` slot definition is marked as `data-dfn-type=idl`, and no error shows up in the console.

**Expected behavior (e.g., it shouldn't crash):** Error should be emitted, as that's not a valid [definition type](https://tabatkins.github.io/bikeshed/#dfn-types). (It's a [linking supertype](https://tabatkins.github.io/bikeshed/#link-types) that encompasses all the IDL-related types.)

You can see this dfn show up in webref with the invalid type at <https://github.com/w3c/webref/blob/master/tr/dfns/payment-request.json#L1992>.

(This particular definition should actually be `type=attribute`, but the markup error isn't the significant issue here.)

tidoust:
Is `attribute` the right type to use for internal slots? Would it be worth having a specific type such as `slot`?

FYI, here is the list of dfns in webref that have an `idl` type. They are all for internal slots:

- `[[CurrentPosture]]` in https://w3c.github.io/device-posture/
- `[[cachedPosition]]` in https://w3c.github.io/geolocation-api/
- `[[isHighAccuracy]]` in https://w3c.github.io/geolocation-api/
- `[[watchTasks]]` in https://w3c.github.io/geolocation-api/
- `[[frameCaptureRequested]]` in https://w3c.github.io/mediacapture-fromelement/
- `[[canExposeCameraInfo]]` in https://w3c.github.io/mediacapture-main/
- `[[canExposeMicrophoneInfo]]` in https://w3c.github.io/mediacapture-main/
- `[[Capabilities]]` in https://w3c.github.io/mediacapture-main/
- `[[Constraints]]` in https://w3c.github.io/mediacapture-main/
- `[[devicesAccessibleMap]]` in https://w3c.github.io/mediacapture-main/
- `[[devicesLiveMap]]` in https://w3c.github.io/mediacapture-main/
[...]

PR #3646, opened 2021-06-21
fix(core/dfn): treat internal slots as IDL attributes

marcoscaceres:
Closes https://github.com/w3c/respec/issues/3644 

Require that all internal slots be "for" something.


src/core/dfn.js
```diff
@@ -26,9 +25,36 @@ export function run() {
...
+  const parent = dfn.closest("[data-dfn-for]");
+  if (parent && dfn !== parent && parent.dataset.dfnFor) {
+    dfn.dataset.dfnFor = parent.dataset.dfnFor;
+  }
+
+  // Assure that it's data-dfn-for= something.
+  if (!dfn.dataset.dfnFor) {
+    const msg = `Internal slot "${title}" must be associated with a WebIDL interface.`;
+    const hint =
+      "Use a `data-dfn-for` attribute to associate this dfn with an WebIDL interface.";
```
marcoscaceres, line 56:
```suggestion
      "Use a `data-dfn-for` attribute to associate this dfn with a WebIDL interface.";
```

src/core/dfn.js
```diff
@@ -26,9 +25,36 @@ export function run() {
...
     }
     dfn.dataset.lt = titles.join("|");
   });
 }
+/**
+ *
+ * @param {string} title
+ * @param {HTMLElement} dfn
+ */
+function processInternalSlotsAsIDL(title, dfn) {
```
saschanaz, line 39: Although webref gives IDL attribute and internal slots the same `type: "attribute"`, an internal slot is not an IDL attribute. So I would rename this to something like `associateInternalSlotWithIDL`.

saschanaz (approved):

src/core/dfn.js
dontcallmedom: see also my remark at https://github.com/w3c/respec/issues/3644#issuecomment-864894696

dontcallmedom:
to close #3644 ReSpec would also need to flag as error any definition that doesn't match the established taxonomy

marcoscaceres:
> to close #3644 ReSpec would also need to flag as error any definition that doesn't match the established taxonomy

Will add... I'll also do automatic checking for `[[\method]](that, people, like)`, and set the type to "method". 

Won't export by default.

src/core/dfn.js
```diff
@@ -26,9 +25,36 @@ export function run() {
...
+  const parent = dfn.closest("[data-dfn-for]");
+  if (parent && dfn !== parent && parent.dataset.dfnFor) {
+    dfn.dataset.dfnFor = parent.dataset.dfnFor;
+  }
+
+  // Assure that it's data-dfn-for= something.
+  if (!dfn.dataset.dfnFor) {
+    const msg = `Internal slot "${title}" must be associated with a WebIDL interface.`;
+    const hint =
+      "Use a `data-dfn-for` attribute to associate this dfn with a WebIDL interface.";
```
sidvishnoi, line 56: Can use `docLink` here now 🎉

src/core/dfn.js
```diff
@@ -26,9 +25,36 @@ export function run() {
...
+    showError(msg, name, { hint, elements: [dfn] });
+  }
+  dfn.dataset.dfnType = "attribute";
```
sidvishnoi, line 59: This is something that will be fixed in follow-up, right? i.e., `method` or a new `internalSlot` type?

src/core/dfn.js
```diff
@@ -12,10 +12,9 @@ export function run() {
     const titles = getDfnTitles(dfn);
     registerDefinition(dfn, titles);
 
-    // Treat Internal Slots as IDL.
-    if (!dfn.dataset.dfnType && /^\[\[\w+\]\]$/.test(titles[0])) {
-      dfn.dataset.dfnType = "idl";
-    }
+    const [firstTitle] = titles;
```
sidvishnoi, line 15: Should we call it `linkingText` instead?

sidvishnoi (approved):

src/core/dfn.js
marcoscaceres: yeah, firstTitle is 🤢... why these are called "title" at all is beyond me.

src/core/dfn.js
marcoscaceres: I don't think so.. See, for example:
https://respec.org/xref/?term=%5B%5BCollectFromCredentialStore%5D%5D%28origin%2C+options%2C+sameOriginWithAncestors%29

src/core/dfn.js
```diff
@@ -60,8 +60,7 @@ function processAsInternalSlot(title, dfn) {
   // Assure that it's data-dfn-for= something.
   if (!dfn.dataset.dfnFor) {
     const msg = `Internal slot "${title}" must be associated with a WebIDL interface.`;
-    const hint =
-      "Use a `data-dfn-for` attribute to associate this dfn with a WebIDL interface.";
+    const hint = docLink`Use a ${"[data-dfn-for]"} attribute to associate this dfn with a WebIDL interface.`;
```
saschanaz, line 63: Wow, interesting!

src/core/dfn.js
saschanaz: Will the existing tools throw if we introduce a new `dfnType` here?

dontcallmedom: re new type, discussion started in https://github.com/speced/spec-dfn-contract/issues/1

src/core/inline-idl-parser.js
```diff
@@ -7,16 +7,18 @@ import { html } from "./import-maps.js";
...
-const attributeRegex = /^((?:\[\[)?(?:\w+)(?:\]\])?)$/;
+const attributeRegex = /^((?:\[\[)?(?:\w+(?: +\w+)*)(?:\]\])?)$/;
 const baseRegex = /^(?:\w+)\??$/;
 const enumRegex = /^(\w+)\["([\w- ]*)"\]$/;
 // TODO: const splitRegex = /(?<=\]\]|\b)\./
 // https://github.com/w3c/respec/pull/1848/files#r225087385
 const methodSplitRegex = /\.?(\w+\(.*\)$)/;
-
+const slotSplitRegex = /\.?\/(.+)/;
+const isProbablySlotRegex = /\[\[.+\]\]/;
```
sidvishnoi, line 21:
```suggestion
const isProbablySlotRegex = /^\[\[.+\]\]/;
```

src/core/inline-idl-parser.js
```diff
@@ -7,16 +7,18 @@ import { html } from "./import-maps.js";
...
 // NOTE: [[value]] is actually a slot, but database has this as type="attribute"
-const attributeRegex = /^((?:\[\[)?(?:\w+)(?:\]\])?)$/;
+const attributeRegex = /^((?:\[\[)?(?:\w+(?: +\w+)*)(?:\]\])?)$/;
 const baseRegex = /^(?:\w+)\??$/;
 const enumRegex = /^(\w+)\["([\w- ]*)"\]$/;
 // TODO: const splitRegex = /(?<=\]\]|\b)\./
 // https://github.com/w3c/respec/pull/1848/files#r225087385
 const methodSplitRegex = /\.?(\w+\(.*\)$)/;
-
+const slotSplitRegex = /\.?\/(.+)/;
```
sidvishnoi, line 20: I'm not following this one. Can you write some examples?

src/core/inline-idl-parser.js
```diff
@@ -167,15 +186,20 @@ function renderBase(details) {
...
   const element = html`${parent && renderParent ? "." : ""}<a
-      data-xref-type="attribute"
-      data-link-for=${linkFor}
-      data-xref-for=${linkFor}
+      data-xref-type="${slotType}"
+      data-link-for="${linkFor}"
+      data-xref-for="${linkFor}"
       data-lt="${lt}"
-      ><code>[[${identifier}]]</code></a
+      ><code>[[${identifier}]]${isMethod ? html`(${argsHtml})` : null}</code></a
```
sidvishnoi, line 202: Can probably merge this into `argsHtml` above.

tests/spec/core/dfn-spec.js
```diff
@@ -153,4 +153,75 @@ describe("Core — Definitions", () => {
...
+        {{Test/[[internal slot]]}}
+        {{Test/[[I am_a method]](I, really, ...am)}}
+        {{Window/[[internal slot]]}}
+        </p>
+      </section>
+    `;
+
+    it("sets the data-dfn-type as an attribute", async () => {
+      const ops = makeStandardOps(null, body);
+      const doc = await makeRSDoc(ops);
```
sidvishnoi, line 191: This is repeated (with same args) in each test here. Can probably move it into a `beforeAll`?

tests/spec/core/dfn-spec.js
```diff
@@ -153,4 +153,75 @@ describe("Core — Definitions", () => {
...
+      expect(dfn.dataset.dfnType).toBe("attribute");
+      expect(dfn.dataset.idl).toBe("");
+    });
+
+    it("sets the data-dfn-type as a method, when it's a method", async () => {
+      const ops = makeStandardOps(null, body);
+      const doc = await makeRSDoc(ops);
+      const dfn = doc.getElementById("method");
+      expect(dfn.textContent.trim()).toBe(
+        "[[I am_a method]](I, really, ...am)"
```
sidvishnoi, line 203: Are spaces acceptable in internal slot names? Any real world examples?

dontcallmedom: there are no rules, which is another reason why I'm not very comfortable with labeling them as attributes; there are real-world examples of spaces in internal slots, like [`[[unique id]]`](https://drafts.css-houdini.org/css-layout-api-1/#dom-layoutchild-unique-id-slot) (`term.startsWith('[[') && term.match(/ .+\]/)` lists them all on https://respec.org/xref/filter/)

sidvishnoi: (I should've checked given I wrote the filter tool 😆)

sidvishnoi: Though, at least no internalSlot-method with spaces.

src/core/dfn.js
marcoscaceres: We will eventually need a way to actually verify this... but pretty cool if we can do that!

src/core/inline-idl-parser.js
marcoscaceres: Just `{{Foo/[[bar]](what, ever)}},  gets split into "Foo/" and `[[bar]](what, ever)` below.

src/core/dfn.js
```diff
@@ -26,9 +34,49 @@ export function run() {
...
+ */
+function processAsInternalSlot(title, dfn) {
+  if (!dfn.dataset.hasOwnProperty("idl")) {
+    dfn.dataset.idl = "";
+  }
+
+  // Automatically use the closest data-dfn-for as the parent.
+  /** @type HTMLElement */
+  const parent = dfn.closest("[data-dfn-for]");
+  if (parent && dfn !== parent && parent.dataset.dfnFor) {
```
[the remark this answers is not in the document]
marcoscaceres: I don't think it would be expensive. But I don't find it as clear. Maybe we could do a compromise... how about?:

```JS
if (parent?.dataset.dfnFor && dfn !== parent) {
```

src/core/inline-idl-parser.js
```diff
@@ -7,16 +7,18 @@ import { html } from "./import-maps.js";
 const idlPrimitiveRegex = /^[a-z]+(\s+[a-z]+)+\??$/; // {{unrestricted double?}} {{ double }}
 const exceptionRegex = /\B"([^"]*)"\B/; // {{ "SomeException" }}
 const methodRegex = /(\w+)\((.*)\)$/;
-const slotRegex = /^\[\[(\w+)\]\]$/;
+
+export const slotRegex = /\[\[(\w+(?: +\w+)*)\]\](\([^)]*\)(\.[^)]*\))?)*$/;
```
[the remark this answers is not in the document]
marcoscaceres: You 100% it works the same? I spent 1 hour crafting that beast and it's very temperamental! 🐻

src/core/inline-idl-parser.js
marcoscaceres: to be clear: 
```
nonMethodPart = "Foo"
methodPart = "[[bar]](what, ever)"
```

src/core/dfn.js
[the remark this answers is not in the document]
sidvishnoi: 👍🏼  Though, `dfn !== parent` test before `dfnFor` one.

src/core/inline-idl-parser.js
sidvishnoi: I'm confused about the `\.?` part..

sidvishnoi: Why not `split("/")`?

sidvishnoi: Just to be clear, we're talking about `slotSplitRegex`

src/core/inline-idl-parser.js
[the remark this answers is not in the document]
sidvishnoi: Let's try it and hope we're testing enough. Or, try some cases at regex101.com?
We should add some fast unit tests for our regexes or inline idl parsing as a follow-up.

src/core/inline-idl-parser.js
marcoscaceres: Oh, you are right... that should have been just /\/(.+)/

marcoscaceres: > Why not split("/")?

Wanted to only split at first... although having other "/"s would be weird, but I'd rather keeps the parts separate. It makes the whole thing more sane when it outputs and errors.

tests/spec/core/dfn-spec.js
marcoscaceres: It's specific to this particular `describe()` group, so beforeAll is probably not ideal... wish they had a "before you run these tests".

sidvishnoi: I think `beforeAll` is scoped to `desribe()` blocks

src/core/dfn.js
```diff
@@ -65,18 +66,21 @@ function processAsInternalSlot(title, dfn) {
...
   }
 
-  // Perform validation on the dfn's type type.
+  // Perform validation on the dfn's type.
   const allowedSlotTypes = ["attribute", "method"];
-  if (!allowedSlotTypes.includes(dfn.dataset.dfnType)) {
-    const msg = docLink`Invalid ${"[data-dfn-type]"} for defining an internal slot.`;
-    const hint = `The only allowed types are: ${codedJoinOr(allowedSlotTypes, {
+  const { dfnType } = dfn.dataset;
+  if (!allowedSlotTypes.includes(dfnType) || derivedType !== dfnType) {
```
sidvishnoi, line 78: This check is duplicated from above, given `derivedType` can only be one of `allowedSlotTypes`.
```suggestion
  if (!allowedSlotTypes.includes(dfnType)) {
```

tests/spec/core/dfn-spec.js
marcoscaceres: Ok, will try. The docs said they beforeAll was global.

src/core/dfn.js
marcoscaceres: the `dfnType` could be "method", but the `derivedType` could be attribute.

dontcallmedom:
> Won't export by default.

it looks like this didn't make it in the pull request - i.e. internal slots seems to be exported by default - @marcoscaceres ?

marcoscaceres:
Weird... checking...

marcoscaceres:
Ah, found it... it's following "the contract" (export everything except "dfn")... I'll fix this as part of https://github.com/w3c/respec/pull/3667/

Cc you there!

PR #3657, opened 2021-06-28
fix(core/inlines): link global element attributes

marcoscaceres:
Fixes linking for `[^ /role ^]` attribute.

### DIFF
--- a/src/core/inlines.js
+++ b/src/core/inlines.js
@@ -86,8 +86,7 @@ function inlineElementMatches(matched) {
     const isGlobalAttr = value.startsWith("/");
     if (isGlobalAttr) {
       return ["element-attr", null, forPart];
-    }
-    if (attrValue) {
+    } else if (attrValue) {
       return ["attr-value", `${forPart}/${attribute}`, attrValue];
     } else if (attribute) {
       return ["element-attr", forPart, attribute];
--- a/src/core/inlines.js
+++ b/src/core/inlines.js
@@ -76,17 +76,23 @@ const inlineElement = /(?:\[\^[^^]+\^\])/; // Inline [^element^]
  */
 function inlineElementMatches(matched) {
   const value = matched.slice(2, -2).trim();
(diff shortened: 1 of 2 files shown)


src/core/inlines.js
```diff
@@ -76,17 +76,23 @@ const inlineElement = /(?:\[\^[^^]+\^\])/; // Inline [^element^]
...
     .map(s => s && s.trim())
     .filter(s => !!s);
+
   const [xrefType, xrefFor, textContent] = (() => {
+    // [^ /role ^], for example
+    const isGlobalAttr = value.startsWith("/");
+    if (isGlobalAttr) {
+      return ["element-attr", null, forPart];
+    }
     if (attrValue) {
```
sidvishnoi, line 90: (for consistency)
```suggestion
    } else if (attrValue) {
```

src/core/inlines.js
```diff
@@ -76,17 +76,23 @@ const inlineElement = /(?:\[\^[^^]+\^\])/; // Inline [^element^]
  */
 function inlineElementMatches(matched) {
   const value = matched.slice(2, -2).trim();
-  const [element, attribute, attrValue] = value
+  const [forPart, attribute, attrValue] = value
```
sidvishnoi, line 79: Might be misleading to call it `forPart`? Not sure.

sidvishnoi (approved):

src/core/inlines.js
marcoscaceres: Don't tempt me to `switch` 😂

src/core/inlines.js
marcoscaceres: element is more misleading for the global case, which is why I renamed it. It was weird otherwise. 

Eventually, we will refractor all this... we should have a single parser for all these types.

src/core/inlines.js
sidvishnoi: `switch (true)` goes brrr

Issue #3664, opened 2021-07-01
Refactor duplicate initialization code in profiles


marcoscaceres:
Every profile has:

```JS
async function domReady() {
  if (document.readyState === "loading") {
    await new Promise(resolve =>
      document.addEventListener("DOMContentLoaded", resolve)
    );
  }
}

(async () => {
  const [runner, { ui }, ...plugins] = await Promise.all(modules);
  try {
    ui.show();
    await domReady();
    await runner.runAll(plugins);
  } finally {
    ui.enable();
  }
})().catch(err => {
  console.error(err);
});
```

What we should do instead is just have a `await ReSpec.run(plugins);` that does all of the above. In particular, the UI should always be loaded, etc.

sidvishnoi:
Had a branch locally for this few weeks ago 😄

PR #3666, opened 2021-07-01
refactor: remove duplicate initialization code in profiles

sidvishnoi:
Closes https://github.com/w3c/respec/issues/3664


profiles/aom.js
```diff
@@ -59,23 +53,6 @@ const modules = [
...
-    await runner.runAll(plugins);
-  } finally {
-    ui.enable();
-  }
-})().catch(err => {
-  console.error(err);
-});
+Promise.all(modules)
+  .then(plugins => ReSpec.run(plugins))
+  .catch(err => console.error(err));
```
sidvishnoi, line 58: I would've liked `await ReSpec.run(modules)`, but that would mean passing `moduleImportPromises` to `ReSpec.run`, which feels awkward.
Also, top-level await isn't common enough yet I think?

marcoscaceres: Yeah, once Safari 15 is out we can switch 🥳

src/respec.js
```diff
@@ -0,0 +1,25 @@
...
+export async function run(plugins) {
+  try {
+    ui.show();
+    await domReady();
+    await runAll(plugins);
+  } finally {
+    ui.enable();
+  }
+}
+
```
marcoscaceres, line 18: I'm not a huge fan of mixing together the running the plugins and showing the UI together. I think they should be separate concerns. Also, we should only allow plugins to be run once. 

Could we do something like: 

```suggestion
import { runAll } from "./core/base-runner.js";
import { ui } from "./core/ui.js";

// In case everything else fails, we want the error
window.addEventListener("error", ev => {
  console.error(ev.error, ev.message, ev);
});

let initialized = false;
let pluginsDone; 
const pluginsPromise = new Promise((res) => {
   pluginsDone = res;
});
domReady().then(async () => {
   ui.show();
   await pluginsPromise;
   ui.enable();
});

export async function run(plugins) {
  if (initialized) throw new Error("ReSpec's was already initialized!");
  await domReady();
  await runAll(plugins);
  initialized = true;
  pluginsDone();
}
```

marcoscaceres: Any thoughts on the above @sidvishnoi?

marcoscaceres (changes requested):
Suggested a change...

src/respec.js
sidvishnoi: > Also, we should only allow plugins to be run once.

The way initialization code runs, it's unlikely that plugins can run multiple times. I'm not sure what the concern is here.

> I'm not a huge fan of mixing together the running the plugins and showing the UI together.

I think the mixing is not bad here. The tasks are related and have a sound flow. Unless we add more things like ui, I think the current approach is fine.

marcoscaceres: >  I'm not sure what the concern is here.

I'm worried that because this is the most "public" part of the API, random documents/folks will try to run their own plugins through it more than once. We've seen the "creative" things people do in their own documents when interacting with ReSpec, so I'd like to ensure that never happens.

sidvishnoi: IMO, if they're calling `respec.run(plugins)` multiple times in their profile, nothing stops them from writing their own runners. The fault is on their end and we shouldn't mess with our API to avoid that.

Though, if we're to provide a safer API, we should _probably_ wrap it in a class:
```js
// file: respec.js
class ReSpec {
  constructor(plugins) {
    this.#initialized = false;
  }

  async run() {
    // ...
  }

  // maybe more methods like getPluginRunDurations()?
}

// file: [profile].js
import ReSpec from "./respec.js";
const respec = new ReSpec(plugins);
await respec.run();
```

sidvishnoi: Though, if we're to provide a safer API, we should _probably_ wrap it in a class:
```js
// file: respec.js
class ReSpec {
  constructor(plugins) {
    this.#initialized = false;
  }

  async run() {
    // ...
  }

  // maybe more methods like getPluginRunDurations()?
}

// file: [profile].js
import ReSpec from "./respec.js";
const respec = new ReSpec(plugins);
await respec.run();
```

Edit: class isn't that useful, as a module with global state is similar.

sidvishnoi:
On another note, do we want to wait for `domReady`? If the script is included with `defer` (which we currently recommend), we can let go of `domReady`.

src/respec.js
marcoscaceres: > IMO, if they're calling respec.run(plugins) multiple times in their profile, nothing stops them from writing their own runners. 

It's not in the profile creators I'm worried about (as we review those, so we are protected). I'm worried about random spec editors fining it and using it. Get what I mean?

What about the separation of concerns about UI and plugins part? Can we separate them or are they inherently tied together?

marcoscaceres:
> On another note, do we want a domReady? If the script is included with defer (which we currently recommend), we can let go of domReady.

I again don't trust editors will do the right thing :(

src/respec.js
sidvishnoi: The `respec.run` isn't exposed to webpage. It's part of the bundling process. So, that should be fine?

> What about the separation of concerns about UI and plugins part? Can we separate them or are they inherently tied together?

They look tied together to me, as in: show loader -> run plugins -> show UI.

marcoscaceres:
(note that by Editors, I mean me... I'm pretty sure none of the specs I edit have "defer")

marcoscaceres (approved):
Ok, let's go with this as it's already a vast improvement.

marcoscaceres (commented):
Actually, the only thing I would ask is for the: 

```JS
if (initialized) throw new Error("ReSpec's was already initialized!");
```

Check. I just don't want random people doing unexpected things in user land.

sidvishnoi:
`respec.run` isn't exposed to user land, so I'm not sure what you're concerned about (it's different from `document.respec`). It's just an internal wrapper over baseRunner (also, it probably belongs to baseRunner and not in separate file maybe).

marcoscaceres:
ah, if no one can do:
```
import * as ReSpec from "../src/respec.js";
```

From user land code then yeah. We are golden. I was worried about the above.

marcoscaceres (approved):

PR #3667, opened 2021-07-05
fix(core/dfn): always default data-dfn-type to 'dfn'

marcoscaceres:
The [dfn contract](https://github.com/speced/spec-dfn-contract) requires `data-dfn-type` to default to "dfn".


src/core/dfn.js
```diff
@@ -86,3 +80,26 @@ function processAsInternalSlot(title, dfn) {
     showError(msg, name, { hint, elements: [dfn] });
   }
 }
+
+function addContractDefaults() {
+  // find all dfns that don't have a type and default them to "dfn".
+  /** @type NodeListOf<HTMLElement> */
+  const dfnsWithNoType = document.querySelectorAll(
+    "dfn:is([data-dfn-type=''],:not([data-dfn-type])"
```
sidvishnoi, line 88:
```suggestion
    "dfn:is([data-dfn-type=''],:not([data-dfn-type]))"
```

src/core/dfn.js
```diff
@@ -86,3 +80,26 @@ function processAsInternalSlot(title, dfn) {
...
+    dfn.dataset.dfnType = "dfn";
+  }
+
+  // Per "the contract", export all definitions, except where:
+  //  - Explicitly marked with data-noexport.
+  //  - The type is "dfn" and not explicitly marked for export (i.e., just a regular definition).
+  //  - definitions was included via (legacy) data-cite="foo#bar".
+  /** @type NodeListOf<HTMLElement> */
+  const exportableDfns = document.querySelectorAll(
+    "dfn:not([data-noexport]):not([data-export]):not([data-dfn-type='dfn']):not(data-cite)"
```
sidvishnoi, line 100: Can we use `:is()` here instead of repeating `:not()`?

src/core/dfn.js
```diff
@@ -86,3 +83,26 @@ function processAsInternalSlot(title, dfn) {
...
+    dfn.dataset.dfnType = "dfn";
+  }
+
+  // Per "the contract", export all definitions, except where:
+  //  - Explicitly marked with data-noexport.
+  //  - The type is "dfn" and not explicitly marked for export (i.e., just a regular definition).
+  //  - definitions was included via (legacy) data-cite="foo#bar".
+  /** @type NodeListOf<HTMLElement> */
+  const exportableDfns = document.querySelectorAll(
+    "dfn:not([data-noexport], [data-export], [data-dfn-type='dfn'], [data-cite])"
```
marcoscaceres, line 103: @sidvishnoi, this is better :) Didn't realize I could just combine :not() like this instead of stacking them

src/core/dfn.js
```diff
@@ -65,6 +59,9 @@ function processAsInternalSlot(title, dfn) {
     showError(msg, name, { hint, elements: [dfn] });
   }
 
+  // Don't export internal slots by default, as they are not supposed to be public.
+  if (!dfn.dataset.hasOwnProperty("export")) dfn.dataset.noexport = "";
```
marcoscaceres, line 63: @dontcallmedom, `noexport` is explicit now 🙏

dontcallmedom: thanks :) could you add a test for that case?

marcoscaceres: Yep, see:
https://github.com/w3c/respec/pull/3667/files#diff-67b69965a7e8e8a580e896f9a2101a8fc9490cee82a5357d77c79d66f107b4f2R270-R284

dontcallmedom (approved):

src/core/dfn.js
marcoscaceres: Actually, I won't be lazy... I'll add an actual test independent test.

PR #3677, opened 2021-07-06
fix(core/templates/show-logo): improve attribute handling and errors

marcoscaceres:


src/core/templates/show-logo.js
```diff
@@ -1,35 +1,45 @@
...
-  const a = html`<a href="${logo.url || ""}" class="logo"></a>`;
-  if (!logo.alt) {
-    const msg = "Found spec logo without an `alt` attribute.";
-    showWarning(msg, name, { elements: [a] });
+  const a = html`<a href="${logo.url ?? null}" class="logo"
+    ><img
+      id="${logo.id ?? null}"
+      alt="${logo.alt ?? null}"
+      width="${logo.width ?? null}"
+      height="${logo.height ?? null}"
```
sidvishnoi, line 26: The `?? null` are unnecessary here. If value is null or undefined (which is what `??` will check for), hyperHTML will not set the attribute (unless you're being explicit here).

src/core/templates/show-logo.js
```diff
@@ -1,35 +1,45 @@
...
+      crossorigin
+    />
+  </a>`;
+  const img = a.querySelector("img");
+  if (!img.alt) {
+    const msg = `Logo at index ${index} is missing required "\`alt\`" property.`;
+    const hint = docLink`Add the missing "\`alt\`" property describing the logo. See ${"[logos]"} for more information.`;
+    showError(msg, name, { hint, elements: [img] });
+  }
+  // avoid triggering 404 requests from dynamically generated hyperHTML attribute values
```
sidvishnoi, line 36: Is this still required? What was the original issue here?

src/core/templates/show-logo.js
```diff
@@ -1,35 +1,45 @@
...
-  if (!logo.alt) {
-    const msg = "Found spec logo without an `alt` attribute.";
-    showWarning(msg, name, { elements: [a] });
+  const a = html`<a href="${logo.url ?? null}" class="logo"
+    ><img
+      id="${logo.id ?? null}"
+      alt="${logo.alt ?? null}"
+      width="${logo.width ?? null}"
+      height="${logo.height ?? null}"
+      crossorigin
```
sidvishnoi, line 27: Why do we want `crossorigin` here?

src/core/templates/show-logo.js
marcoscaceres: Remember a few years ago it was messing up our tests by creating 404s. When the image is created with a src in the markup, it would always trigger a download.

src/core/templates/show-logo.js
marcoscaceres: ah, I wanted to assure that empty strings also got treated as `null`.

src/core/templates/show-logo.js
marcoscaceres: Just for "best current practice". I noticed a lot the injected were cross origin.

src/core/templates/show-logo.js
marcoscaceres: Current code produces with empty strings, for example:

```
<img id="" alt="" width="0" height="0" src="">
```

With the `null`s, the attributes won't be there, which is more what I expect.

marcoscaceres: Actually, you are right... here I really want ` || null`,  not `?? null`.

src/core/templates/show-logo.js
marcoscaceres: Good news! Seems this got fixed!

src/core/templates/show-logo.js
```diff
@@ -1,35 +1,43 @@
...
-  if (!logo.alt) {
-    const msg = "Found spec logo without an `alt` attribute.";
-    showWarning(msg, name, { elements: [a] });
+  const a = html`<a href="${logo.url || null}" class="logo"
+    ><img
+      alt="${logo.alt || null}"
+      height="${logo.height || null}"
+      id="${logo.id || null}"
+      width="${logo.width || null}"
+      src="${logo.src || null}"
```
sidvishnoi, line 27: Nit: let's keep attributes sorted: src, alt, id, width, height, crossorigin

src/core/templates/show-logo.js
```diff
@@ -1,35 +1,43 @@
...
+      crossorigin
+    />
+  </a>`;
+  const img = a.querySelector("img");
+  if (!img.alt) {
+    const msg = `Logo at index ${index} is missing required "\`alt\`" property.`;
```
sidvishnoi, line 33: If we've src available here, maybe we can show that in error msg instead of index?

src/core/templates/show-logo.js
```diff
@@ -1,35 +1,43 @@
...
+      alt="${logo.alt || null}"
+      height="${logo.height || null}"
+      id="${logo.id || null}"
+      width="${logo.width || null}"
+      src="${logo.src || null}"
+      crossorigin
+    />
+  </a>`;
+  const img = a.querySelector("img");
+  if (!img.alt) {
```
sidvishnoi, line 32: Why not `logo.alt`?

sidvishnoi (approved):
Looking good. Few nits.

src/core/templates/show-logo.js
marcoscaceres: Yes, we can probably switch back to just checking the object. I was using the image because we already had a reference to it for the old 404 issue.

src/core/templates/show-logo.js
marcoscaceres: I'll show both... why compromise! :D

PR #3681, opened 2021-07-09
fix(core/dfn): ignore legacy slot dfns

marcoscaceres:
There were specs that were posting to internal slots in the ES6 spec, but using data-cite="foo#bar".


src/core/dfn.js
```diff
@@ -23,7 +23,8 @@ export function run() {
 
     const [linkingText] = titles;
     // Matches attributes and methods, like [[some words]](with, optional, arguments)
-    if (slotRegex.test(linkingText)) {
+    // but ignores legacy data-cite="foo#bar" (e.g., a link to a slot in the ES6 spec)
+    if (slotRegex.test(linkingText) && !dfn.dataset.cite) {
```
sidvishnoi, line 27: Might be faster:
```suggestion
    if (!dfn.dataset.cite && slotRegex.test(linkingText)) {
```

sidvishnoi (approved):

marcoscaceres:
Thanks for the review! 🥪

PR #3691, opened 2021-07-14
fix(core/xref): order matters in tree

marcoscaceres:
@sidvishnoi, I think this might have been a mistake (but I'm a bit unsure)? The order at every level matters, so not sure this should be sorted?

### DIFF
--- a/src/core/xref.js
+++ b/src/core/xref.js
@@ -242,7 +242,7 @@ function dedupeSpecContext(specs) {
     const uniqueSpecs = [...new Set(level)].filter(
       spec => !higherPriority.includes(spec)
     );
-    unique.push(uniqueSpecs.sort());
+    unique.push(uniqueSpecs);
   }
   return unique;
 }


sidvishnoi:
In search backend, order matters only level-by-level, not within a level: https://github.com/w3c/respec-web-services/blob/933362f2b9a940192357bc379973f5bf3cb54cc5/routes/xref/lib/search.ts#L151-L161

marcoscaceres:
Sure, but it's still not an "unordered list" - the order is going to matter there too, thanks to "javer-escripts".

sidvishnoi:
The order won't matter as it's a filter there, not a find. We return all matches in a level. The order is not defined (it's usually sorted by term, or by a checksums of entire term result).

marcoscaceres:
Ok, got it.. but the original question holds... should we bother sorting if it doesn't matter?

sidvishnoi:
> should we bother sorting if it doesn't matter?

sorting was added to dedupe and reduce query size. So, I think it's ok to keep as is.

marcoscaceres:
Ok cool. As long as it doesn't mess with order of citations, then it's all good.

I need to check elsewhere tho... I feel like there are weird instances where things are getting reordered where order should win. 

Where "y" in the following case should win over "x". 

```
<p data-cite="x y">
```

PR #3692, opened 2021-07-17
chore(ci): use cache option of setup-node

sidvishnoi:
Instead of actions/cache, use recently added cache input of setup-node.
And also use shorter YAML syntax, for clarity.


marcoscaceres (approved):
Much clearer.

PR #3700, opened 2021-07-21
feat(core/dfn-validators): add dfn type validators

marcoscaceres:
Part 1 of adding support for definition types. 

This introduces the type validators.

### DIFF
--- a/src/core/dfn-validators.js
+++ b/src/core/dfn-validators.js
@@ -12,14 +12,14 @@ export function validateMimeType(text, type, elem, pluginName) {
     if (type.toString() !== text) {
       throw new Error(`Input doesn't match its canonical form: "${type}".`);
     }
-    return true;
   } catch (error) {
     const msg = `Invalid ${type} "${text}": ${error.message}.`;
     const hint =
       "Check that the MIME type has both a type and a sub-type, and that it's in a canonical form (e.g., `text/plain`).";
     showError(msg, pluginName, { hint, elements: [elem] });
+    return false;
   }
-  return false;
+  return true;
 }
 
 /**
@@ -45,13 +45,12 @@ export function validateDOMName(text, type, elem, pluginName) {
 }
 
 /**
- * Validates common variable or other named thing in a spec, like event names.
+ * Used to validates common variable or other named thing in a spec, like event names.
  *
- * @param {"event"} type
  * @type {DefinitionValidator}
  */
 export function validateCommonName(text, type, elem, pluginName) {
-  // Check a-z, maybe a dash followed by words, case insensitive.
+  // Check a-z, maybe a dash and letters, case insensitive.
   // Also, no spaces.
   if (/^[a-z]+(-[a-z]+)*$/i.test(text)) {
     return true; // all good
--- a/src/core/dfn-validators.js
+++ b/src/core/dfn-validators.js
@@ -1,0 +1,62 @@
+import { MIMEType } from "./import-maps.js";
+import { showError } from "./utils.js";
+
+/**
+ * Validates MIME types strings.
+ *
+ * @type {DefinitionValidator} */
+export function validateMimeType(text, type, elem, pluginName) {
+  try {
+    const type = new MIMEType(text);
+    if (type.toString() !== text) {
+      throw new Error(`Input doesn't match it's canonical '${type}')`);
+    }
+    return true;
+  } catch (error) {
+    const msg = `Invalid ${type} "${text}": ${error.message}.`;
+    const hint =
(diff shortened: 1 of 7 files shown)


sidvishnoi (changes requested):
Looking good, but maybe we should first add the user of this validation in a separate PR? I would be able to review better once I know how it's being used. Then we can iteratively add new validations.

marcoscaceres:
I can send the other PR in parallel, but it's gonna be quite large. I'd rather land this in parts, with this being the smallest one.

tests/spec/core/validators-spec.js
```diff
@@ -0,0 +1,96 @@
...
+      const elements = [
+        "h1",
+        "feSpecularLighting",
+        "feGaussianBlur",
+        "body",
+        "html",
+      ];
+      for (const element of elements) {
+        const dfn = document.createElement("dfn");
+        expect(validateDOMName(element, "element", dfn, "foo/bar")).toBeTrue();
```
sidvishnoi, line 19: I came across jasmine's `withContext`, which lets us add some context to `expect` failure messages. Could be useful when testing in a loop.

```suggestion
        expect(validateDOMName(element, "element", dfn, "foo/bar"))
          .withContext(`element name: ${element}`)
          .toBeTrue();
```

image

marcoscaceres: neat!

src/core/dfn-validators.js
```diff
@@ -0,0 +1,63 @@
...
+    const hint =
+      "Check that the MIME type has both a type and a sub-type, and that it's in a canonical form (e.g., `text/plain`).";
+    showError(msg, pluginName, { hint, elements: [elem] });
+  }
+  return false;
+}
+
+/**
+ * Validates the names of DOM attribute and elements.
+ * @param {"attribute" | "element"} type
```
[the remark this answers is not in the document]
sidvishnoi: Isn't `@type {DefinitionValidator}` below overriding this?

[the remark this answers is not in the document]
marcoscaceres: Nope. 
Screen Shot 2021-07-21 at 7 39 17 pm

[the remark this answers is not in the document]
marcoscaceres: oh wait, it trashes the other ones 😤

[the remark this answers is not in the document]
marcoscaceres: Actually, no, it's fine: 

Screen Shot 2021-07-21 at 7 44 00 pm

tests/spec/core/validators-spec.js
sidvishnoi: Let's do it on all new tests?

marcoscaceres: yep. Done.

sidvishnoi (approved):

tests/karma.conf.base.js
```diff
@@ -17,7 +17,8 @@ const files = [
     included: false,
   },
   {
-    pattern: "node_modules/@(idb|hyperhtml|marked|webidl2)/**/*.js",
+    pattern:
+      "node_modules/@(idb|hyperhtml|marked|webidl2|sniffy-mimetype)/**/*.js",
```
sidvishnoi, line 21: Nit: use backticks to make prettier move it into same line as the key?

src/core/import-maps.js
```diff
@@ -20,3 +21,7 @@ export const marked = _marked;
 /** @type {import("pluralize")} */
 // @ts-ignore
 export const pluralize = _pluralize;
+
+/** @type {import("sniffy-mimetype")} */
+// @ts-ignore
```
sidvishnoi, line 26: Do we need this ts-ignore?

sidvishnoi (approved):
Some nits

src/core/import-maps.js
marcoscaceres: dunno, will try without it. Was cargo cutting based on the others.

tests/karma.conf.base.js
marcoscaceres: I imagine prettier will eventually fix this.

src/core/import-maps.js
marcoscaceres: Yeah, it gets upset.

marcoscaceres: I wonder if this is because the dependency doesn't bundle a typescript file? Or because it's lacking the types. 

@saschanaz, I see that the WebIDL2.js import doesn't need the `tsignore`... what magic are you doing there? 🪄

saschanaz: dunno, maybe because it has no `/** @type {import()} */`?

PR #3704, opened 2021-07-23
feat(core/dfn): use class name for definition type

marcoscaceres:
Supports the following definition classes, to align with Bikeshed (per the contract):

  * abstract-op
  * attribute
  * attr-value
  * element
  * element-state
  * event
  * http-header
  * scheme

Adds one new one:

  * media-type

### DIFF
--- a/src/core/dfn.js
+++ b/src/core/dfn.js
@@ -10,11 +10,44 @@ import {
   showError,
   toMDCode,
 } from "./utils.js";
+import {
+  validateCommonName,
+  validateDOMName,
+  validateMimeType,
+} from "./dfn-validators.js";
 import { registerDefinition } from "./dfn-map.js";
 import { slotRegex } from "./inline-idl-parser.js";
 import { sub } from "./pubsubhub.js";
 
 export const name = "core/dfn";
+
+const knownTypesMap = new Map([
+  ["abstract-op", { requiresFor: false }],
+  ["attribute", { requiresFor: false, validator: validateDOMName }],
+  [
+    "attr-value",
+    {
+      requiresFor: true,
+      associateWith: "an HTML attribute",
+      validator: validateDOMName,
+    },
+  ],
+  ["element", { requiresFor: false, validator: validateDOMName }],
+  [
+    "element-state",
+    {
+      requiresFor: true,
+      associateWith: "an HTML attribute",
+      validator: validateCommonName,
+    },
+  ],
+  ["event", { requiresFor: false, validator: validateCommonName }],
+  ["http-header", { requiresFor: false }],
+  ["scheme", { requiresFor: false, validator: validateCommonName }],
+  ["media-type", { requiresFor: false, validator: validateMimeType }],
+]);
+
+const knownTypes = [...knownTypesMap.keys()];
 
 export function run() {
   for (const dfn of document.querySelectorAll("dfn")) {
@@ -22,11 +55,44 @@ export function run() {
     registerDefinition(dfn, titles);
 
     const [linkingText] = titles;
-    // Matches attributes and methods, like [[some words]](with, optional, arguments)
-    // but ignores legacy data-cite="foo#bar" (e.g., a link to a slot in the ES6 spec)
-    if (!dfn.dataset.cite && slotRegex.test(linkingText)) {
-      processAsInternalSlot(linkingText, dfn);
+
+    let shouldExport = false;
+    let type = "";
+
+    // It's a legacy cite or redefining a something it doesn't own, so it gets no benefit.
+    if (dfn.dataset.cite && /.+#./.test(dfn.dataset.cite)) {
+      continue;
     }
+
+    switch (true) {
+      // class defined type (e.g., "<dfn class="element">)
+      case knownTypes.some(name => dfn.classList.contains(name)):
+        type = knownTypes.find(name => dfn.classList.contains(name));
+        validateDefinition(dfn, linkingText, type);
+        shouldExport = true;
+        break;
+
+      // Internal slots: attributes+ methods (e.g., [[some words]](with, optional, arguments))
+      case slotRegex.test(linkingText):
+        shouldExport = false;
+        type = processAsInternalSlot(linkingText, dfn);
+        break;
+    }
+
+    // If the Editor explicitly asked for it to be exported, so let's export it.
+    if (dfn.classList.contains("export")) shouldExport = true;
+
+    if (shouldExport && !dfn.dataset.hasOwnProperty("noexport")) {
+      dfn.dataset.export = "";
+    }
+
+    // Get closest type from context
+    if (!type) {
+      /** @type {HTMLElement} */
+      const closestType = dfn.closest("[data-dfn-type]");
+      type = closestType?.dataset.dfnType;
+    }
+    if (!dfn.dataset.dfnType && type) dfn.dataset.dfnType = type;
 
     // Only add `lt`s that are different from the text content
     if (titles.length === 1 && linkingText === norm(dfn.textContent)) {
@@ -36,6 +102,21 @@ export function run() {
   }
   sub("plugins-done", addContractDefaults);
 }
+
+function validateDefinition(dfn, linkingText, type) {
+  const entry = knownTypesMap.get(type);
+  if (entry.requiresFor && !dfn.dataset.dfnFor) {
+    const msg = docLink`Definition of type "\`${type}\`" requires a ${"[data-dfn-for]"} attribute.`;
+    const { associateWith } = entry;
+    const hint = docLink`Use a ${"[data-dfn-for]"} attribute to associate this with ${associateWith}.`;
+    showError(msg, name, { hint, elements: [dfn] });
+  }
+
+  if (entry.validator) {
+    entry.validator(linkingText, type, dfn, name);
+  }
+}
+
 /**
  *
  * @param {string} title
@@ -66,8 +147,7 @@ function processAsInternalSlot(title, dfn) {
   // If it ends with a ), then it's method. Attribute otherwise.
   const derivedType = title.endsWith(")") ? "method" : "attribute";
   if (!dfn.dataset.dfnType) {
-    dfn.dataset.dfnType = derivedType;
-    return;
+    return derivedType;
   }
 
   // Perform validation on the dfn's type.
@@ -82,7 +162,9 @@ function processAsInternalSlot(title, dfn) {
       derivedType
     )}"?`;
     showError(msg, name, { hint, elements: [dfn] });
+    return "dfn";
   }
+  return dfnType;
 }
 
 function addContractDefaults() {
--- a/src/core/dfn.js
+++ b/src/core/dfn.js
@@ -96,10 +96,10 @@ function computeTypeAndExport(dfn, linkingText) {
       break;
   }
 
-  // Get closest type from context
   // If the Editor explicitly asked for it to be exported, so let's export it.
   if (dfn.classList.contains("export")) shouldExport = true;
   
+  // Get closest type from context
   if (!type) {
     /** @type {HTMLElement} */
     const closestType = dfn.closest("[data-dfn-type]");
--- a/src/core/dfn.js
+++ b/src/core/dfn.js
@@ -98,7 +98,7 @@ function computeTypeAndExport(dfn, linkingText) {
 
   // If the Editor explicitly asked for it to be exported, so let's export it.
   if (dfn.classList.contains("export")) shouldExport = true;
-  
+
   // Get closest type from context
   if (!type) {
     /** @type {HTMLElement} */
--- a/src/core/dfn.js
+++ b/src/core/dfn.js
@@ -60,23 +60,7 @@ export function run() {
     }
 
     const [linkingText] = titles;
-
-    let { type, shouldExport } = computeTypeAndExport(dfn, linkingText);
-
-    // If the Editor explicitly asked for it to be exported, so let's export it.
-    if (dfn.classList.contains("export")) shouldExport = true;
-
-    if (shouldExport && !dfn.dataset.hasOwnProperty("noexport")) {
-      dfn.dataset.export = "";
-    }
-
-    // Get closest type from context
-    if (!type) {
-      /** @type {HTMLElement} */
-      const closestType = dfn.closest("[data-dfn-type]");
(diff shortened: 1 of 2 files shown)


src/core/dfn.js
```diff
@@ -10,24 +10,90 @@ import {
...
+    if (shouldExport && !dfn.dataset.hasOwnProperty("noexport")) {
+      dfn.dataset.export = "";
+    }
+
+    // Get closest type from context
+    if (!type) {
+      /** @type {HTMLElement} */
+      const closestType = dfn.closest("[data-dfn-type]");
+      type = closestType?.dataset.dfnType;
+    }
```
sidvishnoi, line 94: Move this into a function?

src/core/dfn.js
```diff
@@ -10,24 +10,90 @@ import {
...
+  validateDOMName,
+  validateMimeType,
+} from "./dfn-validators.js";
 import { registerDefinition } from "./dfn-map.js";
 import { slotRegex } from "./inline-idl-parser.js";
 import { sub } from "./pubsubhub.js";
 
 export const name = "core/dfn";
 
+const knownTypesMap = new Map([
```
sidvishnoi, line 24:
```suggestion
/** @type {Map<string, { requiresFor: boolean, validator?: (...args: any[])=> boolean,  associateWith?: string}>}  */
const knownTypesMap = new Map([
```
(Might not be entirely correct, please check locally)

src/core/dfn.js
```diff
@@ -36,6 +102,21 @@ export function run() {
...
+  const entry = knownTypesMap.get(type);
+  if (entry.requiresFor && !dfn.dataset.dfnFor) {
+    const msg = docLink`Definition of type "\`${type}\`" requires a ${"[data-dfn-for]"} attribute.`;
+    const { associateWith } = entry;
+    const hint = docLink`Use a ${"[data-dfn-for]"} attribute to associate this with ${associateWith}.`;
+    showError(msg, name, { hint, elements: [dfn] });
+  }
+
+  if (entry.validator) {
+    entry.validator(linkingText, type, dfn, name);
```
sidvishnoi, line 116: I wonder if it would make sense to throw from validators and showError here?

src/core/dfn.js
```diff
@@ -10,24 +10,90 @@ import {
...
+    if (dfn.dataset.cite && /.+#./.test(dfn.dataset.cite)) {
+      continue;
     }
 
+    switch (true) {
+      // class defined type (e.g., "<dfn class="element">)
+      case knownTypes.some(name => dfn.classList.contains(name)):
+        type = knownTypes.find(name => dfn.classList.contains(name));
+        validateDefinition(dfn, linkingText, type);
+        shouldExport = true;
```
sidvishnoi, line 72: Should we remove the classes after processing to avoid potential CSS styling clashes?

sidvishnoi (commented):
Will do another pass in a while, but looking good!

src/core/dfn.js
marcoscaceres: The errors might be quire custom per validator, so I'd like to just leave them inside the validators themselves.

src/core/dfn.js
marcoscaceres: No, as I expect people will want to actually use them for styling. That's a feature :)

src/core/dfn.js
marcoscaceres: Not sure if it's worth it, tbh. It will be a like a two line function so I'd be inclined just to leave it here.

sidvishnoi: I meant line +59 to +94

marcoscaceres: Ah, yes, could do the whole block.

src/core/dfn.js
```diff
@@ -36,6 +70,61 @@ export function run() {
   }
   sub("plugins-done", addContractDefaults);
 }
+
+/**
+ * @param {HTMLElement} dfn
+ * @param {string} linkingText
+ * @return {Object}
+ * */
+function computeTypeAndExport(dfn, linkingText) {
```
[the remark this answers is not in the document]
sidvishnoi: Or maybe move the assignments to the end of function.

src/core/dfn.js
```diff
@@ -36,6 +70,61 @@ export function run() {
...
+  // Get closest type from context
+  if (!type) {
+    /** @type {HTMLElement} */
+    const closestType = dfn.closest("[data-dfn-type]");
+    type = closestType?.dataset.dfnType;
+  }
+  if (!dfn.dataset.dfnType && type) dfn.dataset.dfnType = type;
+}
+
+function validateDefinition(dfn, linkingText, type) {
```
sidvishnoi, line 114: Add JsDoc for this?

tests/spec/core/dfn-spec.js
```diff
@@ -329,4 +329,218 @@ describe("Core — Definitions", () => {
...
+      `;
+      const ops = makeStandardOps(null, body);
+      const doc = await makeRSDoc(ops);
+      const dfn = doc.querySelector("#attributes dfn");
+      expect(dfn.dataset.dfnType).toBe("attribute");
+      expect(dfn.dataset.export).toBe("");
+
+      // Check validation error
+      const errors = doc.respec.errors.filter(err => err.plugin === "core/dfn");
+      expect(errors).toHaveSize(1);
```
sidvishnoi, line 391: Let's also test for error content. Ditto for rest.

tests/spec/core/dfn-spec.js
```diff
@@ -329,4 +329,218 @@ describe("Core — Definitions", () => {
...
+      `;
+      const ops = makeStandardOps(null, body);
+      const doc = await makeRSDoc(ops);
+      const dfn = doc.querySelector("#attr-value dfn");
+      expect(dfn.dataset.dfnFor).toBe("input/type");
+      expect(dfn.dataset.dfnType).toBe("attr-value");
+      expect(dfn.dataset.export).toBe("");
+
+      // Check validation error
+      const errors = doc.respec.errors.filter(err => err.plugin === "core/dfn");
```
sidvishnoi, line 415: Can move this into a small helper function in nearest `describe()` block, as it's repeated.

tests/spec/core/dfn-spec.js
```diff
@@ -329,4 +329,218 @@ describe("Core — Definitions", () => {
...
+      expect(errors).toHaveSize(1);
+      expect(errors[0].message).toContain("-element");
+    });
+
+    it("handles element states", async () => {
+      const body = `
+        <section>
+          <h2>Element states</h2>
+          <p id="element-state">
+            <dfn data-dfn-for="input" data-dfn-type="element-state">Password</dfn>
```
sidvishnoi, line 448: I think we're supposed to be using `class="element-state"` here?

tests/spec/core/dfn-spec.js
```diff
@@ -329,4 +329,218 @@ describe("Core — Definitions", () => {
...
+      const doc = await makeRSDoc(ops);
+      const dfn = doc.querySelector("#http-header dfn");
+      expect(dfn.dataset.dfnType).toBe("media-type");
+      expect(dfn.dataset.export).toBe("");
+
+      // Check validation error
+      const errors = doc.respec.errors.filter(err => err.plugin === "core/dfn");
+      expect(errors).toHaveSize(1);
+    });
+  });
```
sidvishnoi, line 545: Let's also add some tests on what happens if we provided multiple classes (constructive as well as clashing).

sidvishnoi (changes requested):
Nearly there!

src/core/dfn.js
```diff
@@ -10,24 +10,58 @@ import {
...
     const titles = getDfnTitles(dfn);
     registerDefinition(dfn, titles);
 
-    const [linkingText] = titles;
-    // Matches attributes and methods, like [[some words]](with, optional, arguments)
-    // but ignores legacy data-cite="foo#bar" (e.g., a link to a slot in the ES6 spec)
-    if (!dfn.dataset.cite && slotRegex.test(linkingText)) {
-      processAsInternalSlot(linkingText, dfn);
+    // It's a legacy cite or redefining a something it doesn't own, so it gets no benefit.
+    if (dfn.dataset.cite && /.+#./.test(dfn.dataset.cite)) {
```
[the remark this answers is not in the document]
marcoscaceres: Let's go with 1, as otherwise .test() will needlessly test the string "undefined".

src/core/dfn.js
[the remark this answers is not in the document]
marcoscaceres: Hmmm, kinda like it all here, tbh. This handles type and export and leaves the lt stuff to the the outer function.

tests/spec/core/dfn-spec.js
marcoscaceres: I wanter to avoid that tbh, as it makes it harder to change text strings. The strings themselves don't matter, just that an error is generated.

tests/spec/core/dfn-spec.js
marcoscaceres: oops, yes.

marcoscaceres, line 448:
```suggestion
            <dfn class="element-state" data-dfn-for="input">Password</dfn>
```

src/core/dfn.js
```diff
@@ -10,24 +10,58 @@ import {
   showError,
   toMDCode,
 } from "./utils.js";
+import {
+  validateCommonName,
+  validateDOMName,
+  validateMimeType,
+} from "./dfn-validators.js";
```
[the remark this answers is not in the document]
marcoscaceres: Per the explicit  imports for this one...

tests/spec/core/dfn-spec.js
sidvishnoi: Don't have to test full string. Can test for particular nouns to be existing in message.

marcoscaceres: Right, I had just checks for "-attribute", for example. I can add it them back.

src/core/dfn.js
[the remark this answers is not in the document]
marcoscaceres: Rationale is that they already contain the word "validate*" otherwise, `validator.validate*` is somewhat redundant.

tests/spec/core/dfn-spec.js
marcoscaceres: Yeah, I deliberately left this case undefined. I wanted to get some experience with them before handling that particular case (i.e. first one wins right now). I can at least test that.

marcoscaceres: Ok, cool. This we backwards... it was giving the names in the Map priority.

src/core/dfn.js
```diff
@@ -36,6 +71,68 @@ export function run() {
...
+function computeTypeAndExport(dfn, linkingText) {
+  let shouldExport = false;
+  let type = "";
+
+  switch (true) {
+    // class defined type (e.g., "<dfn class="element">)
+    case knownTypes.some(name => dfn.classList.contains(name)):
+      // First one wins
+      type = [...dfn.classList].find(className =>
+        knownTypes.includes(className)
```
sidvishnoi, line 88:
```suggestion
        knownTypesMap.has(className)
```

src/core/dfn.js
```diff
@@ -36,6 +71,68 @@ export function run() {
...
+      type = processAsInternalSlot(linkingText, dfn);
+      break;
+  }
+
+  // If the Editor explicitly asked for it to be exported, so let's export it.
+  if (dfn.classList.contains("export")) shouldExport = true;
+
+  if (shouldExport && !dfn.hasAttribute("data-noexport")) {
+    dfn.dataset.export = "";
+  }
```
sidvishnoi, line 106: Nit: Let's move this assignment to the bottom of function, so we've all getters together, followed by setters.

tests/spec/core/dfn-spec.js
```diff
@@ -10,6 +10,9 @@ import {
 
 describe("Core — Definitions", () => {
   afterAll(flushIframes);
+
+  const findDfnErrors = err => err.plugin === "core/dfn";
```
sidvishnoi, line 14: I was thinking for something like following, but this is ok too.
```suggestion
  const findDfnErrors = doc => doc.respec.errors.filter(err => err.plugin === "core/dfn");
```

sidvishnoi (approved):

src/core/dfn.js
marcoscaceres: better!

marcoscaceres (commented):
Woot woot. Looking pretty swish!

src/core/dfn.js
```diff
@@ -36,6 +71,66 @@ export function run() {
...
+    type = closestType?.dataset.dfnType;
+  }
+  if (!dfn.dataset.dfnType && type) dfn.dataset.dfnType = type;
+
+  // If the Editor explicitly asked for it to be exported, so let's export it.
+  if (dfn.classList.contains("export")) shouldExport = true;
+
+  if (shouldExport && !dfn.hasAttribute("data-noexport")) {
+    dfn.dataset.export = "";
+  }
```
sidvishnoi, line 112: Nit: My OCD. Please run through linter.

```suggestion
  // If the Editor explicitly asked for it to be exported, so let's export it.
  if (dfn.classList.contains("export")) shouldExport = true;
  
  if (!type) {
    /** @type {HTMLElement} */
    const closestType = dfn.closest("[data-dfn-type]");
    type = closestType?.dataset.dfnType;
  }

  if (!dfn.dataset.dfnType && type) dfn.dataset.dfnType = type;
  if (shouldExport && !dfn.hasAttribute("data-noexport")) {
    dfn.dataset.export = "";
  }
```

src/core/dfn.js
```diff
@@ -36,6 +71,66 @@ export function run() {
...
+      shouldExport = false;
+      type = processAsInternalSlot(linkingText, dfn);
+      break;
+  }
+
+  // Get closest type from context
+  // If the Editor explicitly asked for it to be exported, so let's export it.
+  if (dfn.classList.contains("export")) shouldExport = true;
+  
+  if (!type) {
```
sidvishnoi, line 103:
```suggestion
  // If the Editor explicitly asked for it to be exported, so let's export it.
  if (dfn.classList.contains("export")) shouldExport = true;
  
  // Get closest type from context
  if (!type) {
```

PR #3706, opened 2021-07-23
perf(core/xref): get query id in parallel

sidvishnoi:
Haven't measured, but gotta be faster?


sidvishnoi:
It's fast as in takes less time, but maybe not better. Tested on ~11k items. Left is for loop (older), right is Promise.all (newer).

No CPU throttling:
image

at 4x slowdown:
image

sidvishnoi:
The slowest part is `{id, ...entry}` btw 😄

marcoscaceres:
@sidvishnoi, when you are around, let me know what you want to do with this?

sidvishnoi:
Closing as it has no major benefit. From the Chrome Devtools Performance Panel, `Promise.all` on many `objectHash` calls blocks the main thread (though only with really large number of calls). Might be beneficial to limit the "max concurrency", but meh. Will send a PR to reduce the `{id, ...entry}` overhead though.

PR #3712, opened 2021-07-27
refactor(core/dfn): split computing type / export

marcoscaceres:


marcoscaceres:
Oops. Bit premature it seems. Will send again.

PR #3717, opened 2021-07-28
feat(core/dfn): add support for task source types

marcoscaceres:
This automatically detects and exports "task source" definitions, which will make them easier for find and reference.


marcoscaceres:
@saschanaz, is it ok if ping you for the occasional review? @sidvishnoi is taking a break. 

I won't bother you with refactors or small code changes.

saschanaz (approved):
LGTM, is this something supported by specref/xref?

src/core/dfn.js
```diff
@@ -108,6 +114,13 @@ function computeType(dfn, linkingText) {
...
+function processAsTaskSource(dfn) {
+  // Task sources are exported by default.
+  if (!dfn.matches(".no-export, [data-noexport]")) {
+    dfn.dataset.export = "";
+  }
+}
+
 // Deal with export/no export
 function computeExport(dfn) {
   switch (true) {
```
saschanaz, line 126: I'm not sure how I feel about this pattern 🤔😂

marcoscaceres: At first, I was like... that's weird/not right... but now I really love how it split up the code into different cases. 

Eventually, once ECMAScript gets pattern matching, we can switch them all over to that:
https://github.com/tc39/proposal-pattern-matching

PR #3727, opened 2021-08-02
perf(core/xref): reuse obj when adding query id

sidvishnoi:
As noted in https://github.com/w3c/respec/pull/3706#issuecomment-886071313


src/core/xref.js
```diff
@@ -172,6 +172,7 @@ function getRequestEntry(elem) {
   const forContext = getForContext(elem, isIDL);
 
   return {
+    id: "", // TBD
```
sidvishnoi, line 175: To ensure the shape of object returned is same (minor optimization, and also to make TypeScript happier).

src/core/xref.js
```diff
@@ -71,8 +71,8 @@ export async function run(conf) {
   const queryKeys = [];
   for (const elem of elems) {
     const entry = getRequestEntry(elem);
-    const id = await objectHash(entry);
-    queryKeys.push({ ...entry, id });
+    entry.id = await objectHash(entry);
```
sidvishnoi, line 74: This will mean now the objectHash also hashes `id: ""`, but it's ok as we just need a checksum ID for caching.

src/core/xref.js
sidvishnoi: Should add this as a comment instead 😄 
```suggestion
    // Add an empty `id` to ensure the shape of object returned is same (minor optimization, and also makes TypeScript happy).
    id: "",
```

marcoscaceres (approved):

PR #3731, opened 2021-08-03
fix(core/inlines): match task source multiline

marcoscaceres:


src/core/inlines.js
```diff
@@ -216,7 +216,7 @@ function inlineAnchorMatches(matched) {
   const forContext = isFor ? norm(isFor) : null;
   let type = "dfn";
   switch (true) {
-    case /task source$/i.test(matched):
+    case /task\s+source$/i.test(matched):
```
sidvishnoi, line 219: Test on norm() maybe?

marcoscaceres: yeah, I was thinking of doing that as a followup... I'm not sure why `norm()` wasn't run on "matched" before being sent into the function, tbh.

PR #3732, opened 2021-08-03
Logius profile

PHaasnoot:
Logius is the digital government service of the Netherlands Ministry of the Interior and Kingdom Relations.
Logius created (and intents to develop further) a Logius ReSpec profile for publishing documents, but one that is generic enough so that in the future more Dutch government organizations, including Geonovum,can use it.

In this version certain fixed urls's in the style and header modules are made configurable, so that other organizations don't have to build their own profile. Also there is added some extra markdown functionality.

(configuration options are documented here : https://github.com/Logius-standaarden/respec/wiki )


marcoscaceres:
@PHaasnoot thanks for preparing this. 

Some immediate things:

  * please see the new profile structure, for example [w3c.js](https://github.com/w3c/respec/blob/develop/profiles/w3c.js) 
  * Please exclude the built files, we will generate and check those in automatically.
  * current_issues.md probably shouldn't be in the root... maybe move to the Logius wiki?
  * Ultimately, it's up to you to use the "nl_" property prefixes, at lot don't look like they would be needed. Are you sure you want to do that? (your call)

src/core/structure.js
```diff
@@ -2,6 +2,7 @@
 // Module core/structure
 //  Handles producing the ToC and numbering sections across the document.
 
+
```
marcoscaceres, line 5:
```suggestion
```

src/logius/conformance.js
```diff
@@ -0,0 +1,106 @@
...
+    keywordInterpretation(keywords, plural) {
+      return html`<p>
+        ${plural ? "De trefwoorden" : "Het trefwoord"} ${keywords} in dit document
+        ${plural ? "moeten" : "moet"} worden geïnterpreteerd als in
+        <a href="https://tools.ietf.org/html/bcp14">BCP 14</a>
+        ${renderInlineCitation("RFC2119")} ${renderInlineCitation("RFC8174")}
+        als, en alleen als deze in hoofdletters zijn weergegeven, zoals hier getoond.
+      </p>`;
+    },
+  },
```
marcoscaceres, line 64: this seems to be the only bit that is different from W3C/Conformance... I wonder if we should move this into the W3C/conformance module and then just rename that "core/conformance.js"?

src/logius/defaults.js
```diff
@@ -0,0 +1,98 @@
...
+linter.register(
+  noHttpPropsRule,
+  privsecSectionRule,
+  noHeadinglessSectionsRule,
+  checkPunctuation,
+  localRefsExist
+);
+
+// todo: check if licenses are already defined
+// const cgbg = new Set(["BG-DRAFT", "BG-FINAL", "CG-DRAFT", "CG-FINAL"]);
```
marcoscaceres, line 22:
```suggestion
```

src/logius/fix-md-elements.js
```diff
@@ -0,0 +1,84 @@
...
+        table.classList.add(conf.nl_markdownTableClass);
+      })
+  );
+}
+
+// todo check non happy flows
+function addFigureImg(conf) {
+  if (
+    !conf.nl_markdownEmbedImageInFigure ||
+    conf.nl_markdownEmbedImageInFigure == false
```
marcoscaceres, line 31: These are the same, no? 
```suggestion
    !conf.nl_markdownEmbedImageInFigure
```

src/logius/fix-md-elements.js
```diff
@@ -0,0 +1,84 @@
...
+          if (!img.getAttribute("title")) {
+            cloneImg.setAttribute("title", caption);
+          }
+          figure.appendChild(cloneImg);
+          figure.appendChild(figcaption);
+          img.parentNode.insertBefore(figure, img);
+          img.remove();
+        }
+      })
+  );
```
marcoscaceres, line 68: I think you want something like the following. You will need to import `addId` and `showError` from utils, and hyperHTML as well (this is untested... it's just a suggestion): 

```suggestion
  [...document.querySelectorAll("section[data-format=markdown] img")]
  .filter(img => !img.closest("figure"))
  .forEach(img => {
    if (!img.alt) {
      const msg = " ... warning message about missing alt ... ";
      const hint = " ... helpful text here.... ";
      showError(msg, name, { elements: [img], hint });
    }
    const alt = img.alt;
    const figure = html`<figure>
      ${img.clone(true)}
      <figcaption>${alt}</figcaption>
    </figure>`;
    addId(figure);
    img.replaceWith(figure);
  });
```

src/logius/github.js
```diff
@@ -0,0 +1,53 @@
+// @ts-check
+
+import { showWarning } from "../core/utils.js";
+/**
+ * this module fixes some peculiarities in core/github for the Dutch context
```
marcoscaceres, line 5: I wonder if we can discuss this as a fix to the Core module? I wasn't able to figure out what you were trying to do here, but we can probably fix whatever the issue is in core to make it work without requiring this module?

src/logius/inlines.js
```diff
@@ -0,0 +1,349 @@
+// @ts-check
+// Module core/inlines
```
marcoscaceres, line 2: Likewise, we might be able to add the bits you need to the core one? This is using things you probably don't need (e.g., the WebIDL inline parser)

src/logius/seo.js
```diff
@@ -0,0 +1,201 @@
+// @ts-check
+// Module w3c/seo
```
marcoscaceres, line 2: This module is still using a lot of W3C stuff...

src/logius/splitmarkdownheaders.js
```diff
@@ -0,0 +1,125 @@
+// @ts-check
+/*
+* Logius addition
```
marcoscaceres, line 3: Possibly something we can fix in the Core module?

src/logius/templates/headers.js
```diff
@@ -0,0 +1,354 @@
+// @ts-check
```
marcoscaceres, line 1: this module contains quite a bit of w3c stuff still... might need a cleanup.

src/logius/templates/sotd.js
```diff
@@ -0,0 +1,160 @@
...
+      : conf.isWV
+      ? renderWV(conf, opts)
+      : conf.isBASIS
+      ? renderBASIS(conf, opts)
+      : ""}
+    ${renderGovernance(conf)} ${opts.additionalSections}
+  `;
+};
+
+export function renderPreview(conf) {
```
marcoscaceres, line 50: Might want to localize or remove this one.

PHaasnoot:
> 
> Some immediate things:
> 
> * please see the new profile structure, for example [w3c.js](https://github.com/w3c/respec/blob/develop/profiles/w3c.js)
[changed]
> * Please exclude the built files, we will generate and check those in automatically.
[changed]
> * current_issues.md probably shouldn't be in the root... maybe move to the Logius wiki?
[changed]
> * Ultimately, it's up to you to use the "nl_" property prefixes, at lot don't look like they would be needed. Are you sure you want to do that? (your call)
[we will discuss this]
> * revert changes to package-lock.json
[changed]

src/logius/inlines.js
edwinwisse: moved nl stuff to core version.

src/logius/seo.js
edwinwisse: Switched logius profile to the w3c version.

src/logius/fix-md-elements.js
```diff
@@ -0,0 +1,56 @@
...
+  [...document.querySelectorAll("[data-format=markdown]:not(body)")].forEach(
+    section =>
+      section.querySelectorAll("table").forEach(table => {
+        // table.classList.add("complex");
+        table.classList.add(conf.nl_markdownTableClass);
+      })
+  );
+}
+
+function checkImgAlt(conf) {
```
marcoscaceres, line 28: For this, you might want to consider creating a simple linter rule. For example:

https://github.com/w3c/respec/blob/develop/src/core/linter-rules/no-headingless-sections.js

src/logius/splitmarkdownheaders.js
```diff
@@ -0,0 +1,123 @@
+// @ts-check
+/*
+ * Logius addition
+ * this script splits markdown level 1 headers ("#") in seperate sections
```
marcoscaceres, line 4: Note that if you hit bugs with the markdown parser itself, you can also file bugs at:
https://github.com/markedjs/marked

profiles/logius.js
```diff
@@ -0,0 +1,85 @@
...
+  import("../src/w3c/conformance.js"), // added dutch additions into w3c version
+  import("../src/core/dfn.js"),
+  import("../src/core/pluralize.js"),
+  import("../src/core/examples.js"),
+  // import("../src/core/issues-notes.js"),
+  import("../src/logius/issues-notes.js"),
+  // todo insert requirements ?
+  import("../src/core/best-practices.js"),
+  import("../src/core/figures.js"),
+  import("../src/core/webidl.js"), // todo skipped by geonovum
```
marcoscaceres, line 38: I think it's unlikely you will need WebIDL... it's very W3C specific. 
```suggestion
```

profiles/logius.js
```diff
@@ -0,0 +1,85 @@
...
+  import("../src/core/biblio.js"),
+  import("../src/core/link-to-dfn.js"),
+  import("../src/core/xref.js"),
+  import("../src/core/data-cite.js"),
+  import("../src/core/webidl-index.js"),
```
marcoscaceres, line 43: As above... 

```suggestion
```

profiles/logius.js
```diff
@@ -0,0 +1,85 @@
...
+  import("../src/core/render-biblio.js"),
+  import("../src/core/dfn-index.js"),
+  import("../src/core/contrib.js"),
+  import("../src/core/fix-headers.js"),
+  import("../src/core/structure.js"),
+  import("../src/core/informative.js"), // solved handhaaf core version ipv  geonovum version
+  import("../src/core/id-headers.js"),
+  import("../src/core/caniuse.js"),
```
marcoscaceres, line 51: Probably don't need this either: 
```suggestion
```

it shows data from https://caniuse.com

profiles/logius.js
```diff
@@ -0,0 +1,85 @@
...
+  import("../src/core/mdn-annotation.js"),
```
marcoscaceres, line 52: Similarly... you probably don't need: 

```suggestion
```

that's for data from https://developer.mozilla.org/

profiles/logius.js
```diff
@@ -0,0 +1,85 @@
...
+  // todo geonovum insert leafletfigures.js
+  import("../src/ui/save-html.js"),
+  import("../src/ui/search-specref.js"),
+  import("../src/ui/search-xref.js"),
+  import("../src/ui/about-respec.js"),
+  import("../src/core/seo.js"),
+  import("../src/w3c/seo.js"), // changed from logius/seo.js to w3c/seo.js since there was no notable difference
+  import("../src/core/highlight.js"),
+  import("../src/core/webidl-clipboard.js"), // skipped by geonovum
```
marcoscaceres, line 61: As above... this is for WebIDL, so seems quite unlikely you will need it.  

```suggestion
```

profiles/logius.js
```diff
@@ -0,0 +1,85 @@
...
+  import("../src/core/data-tests.js"),
```
marcoscaceres, line 62: Probably same with this one... I should have moved this "w3c", as it's for "[web platform tests](https://web-platform-tests.org)" 😅
```suggestion
```

profiles/logius.js
```diff
@@ -0,0 +1,85 @@
...
+  import("../src/core/custom-elements/index.js"),
+  /* Linters must be the last thing to run */
+  import("../src/core/linter-rules/check-charset.js"),
+  import("../src/core/linter-rules/check-punctuation.js"),
+  import("../src/core/linter-rules/check-internal-slots.js"),
+  import("../src/core/linter-rules/local-refs-exist.js"),
+  import("../src/core/linter-rules/no-headingless-sections.js"),
+  import("../src/core/linter-rules/no-unused-vars.js"),
+  import("../src/core/linter-rules/privsec-section.js"),
+  import("../src/core/linter-rules/wpt-tests-exist.js"),
```
marcoscaceres, line 78: This is also for "web platform tests", so probably can go... 

```suggestion
```

marcoscaceres (commented):
Starting to look good... I'm guessing the /deprecated/ files will be removed also?

src/logius/fix-md-elements.js
sanderke: Thanks for the tip. A linter rule has been added to core and the "Occurred at" message has been localised.

sanderke:
Tests were blocked by the presence of Logius builds. They have been removed.

PHaasnoot:
A new PR will be made based on cleaned up code

PR #3747, opened 2021-08-19
fix(core/biblio): expire entries based on server response

### LINKED ISSUE #3746: Biblio cache might need clearing more often
## Description of problem

Getting reports that some biblio entries are not getting updated in some browsers. 

They should be getting updated. It might be safer to not cache them for now.

marcoscaceres:
Closes #3746


src/core/biblio-db.js
```diff
@@ -172,9 +192,18 @@ export const biblioDB = {
...
     const db = await this.ready;
-    const isInDB = await this.has(type, details.id);
-    const store = db.transaction(type, "readwrite").store;
+    let isInDB = await this.has(type, details.id);
     // update or add, depending of already having it in db
+    // or if it's expired
+    if (isInDB) {
+      const entry = await this.get(type, details.id);
+      if (entry?.expires < Date.now()) {
+        const { store } = db.transaction(type, "readwrite");
```
sidvishnoi, line 201: Aren't we suppose to `await tx.done`? Maybe `db.objectStore` would be nicer?

src/core/biblio-db.js
```diff
@@ -46,6 +47,23 @@ async function openIdb() {
       db.createObjectStore("reference", { keyPath: "id" });
     },
   });
+  // Clean the database of expired biblio entries.
+  const now = Date.now();
+  for (const storeName of [...ALLOWED_TYPES]) {
+    const store = db.transaction(storeName, "readwrite").store;
+    const range = IDBKeyRange.lowerBound(now);
+    let result = await store.openCursor(range);
```
sidvishnoi, line 55: Nice!

src/core/biblio-db.js
marcoscaceres: > Aren't we suppose to await tx.done? 

I don't think it has a .done()? is that something Jake added to the library? 

> Maybe db.objectStore would be nicer?

sorry, not sure what you are suggesting here... happy to change it if there is something better.

PR #3768, opened 2021-09-06
Do not set data-dfn-for on top level IDL objects

### LINKED ISSUE #3764: bogus data-dfn-for on IDL dictionaries and enums
## Important info

- URL to affected spec: 
- ReSpec version: 26.13.2

## Description of problem

the [`PermissionDescription` dfn](https://w3c.github.io/permissions/#dom-permissiondescriptor) generated from the IDL fragment has a `data-dfn-for` set to "PermissionDescriptor" - this ends up making it idenfied in [xref](https://respec.org/xref/?term=PermissionDescriptor) (via webref) as `{{PermissionDescriptor/PermissionDescriptor}}` instead of just `{{PermissionDescriptor}}`

(presumably the permission spec should also not rely on automatic IDL-dfn, but respec should not generate bogus dfn in that situation)

The same seems to be true for enums (`PermissionName` in that same spec)

no `data-dfn-for` should be generated.

dontcallmedom:


marcoscaceres:
@dontcallmedom, not sure how much you want to spend on this... I can try to fix it later this week. 

Few notes/todo:

- [ ]  We have defined `topLevelEntities` in some other module, which should relocate to + export from the WebIDL module.
- [ ]  We should add a some tests.

dontcallmedom:
I'm supportive of DRYing and testing the code, but at the moment, this bug means quite a few IDL names from the permission API creates linking errors, so I would prefer if we could hot fix this and worry about the rest separately.

marcoscaceres:
Ok, no problem. @dontcallmedom, can you:
```
npm run lint -- --fix
````

dontcallmedom:
linted

marcoscaceres:
I'll update merge it once it's green, but will leave the bug open so I remember to add the tests, etc.

marcoscaceres (approved):

marcoscaceres:
Released respec@26.13.4

PR #3782, opened 2021-09-21
feat(core/caniuse): snapshot caniuse data

marcoscaceres:
Now exports caniuse data to TR docs.

### DIFF
--- a/src/core/caniuse.js
+++ b/src/core/caniuse.js
@@ -41,7 +41,10 @@ export function prepare(conf) {
     return; // no feature to show
   }
 
-  document.head.appendChild(html`<style class="removeOnSave">
+  document.head.appendChild(html`<style
+    id="caniuse-stylesheet"
+    class="${options.removeOnSave ? "removeOnSave" : ""}"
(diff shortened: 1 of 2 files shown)


mdarosa:
Why do I keep getting these emails?

src/core/caniuse.js
```diff
@@ -80,21 +85,25 @@ export async function run(conf) {
...
+  if (options.removeOnSave) {
+    // Will remove the browser support cells.
+    headDlElem
+      .querySelectorAll(".caniuse-browser")
+      .forEach(elem => elem.classList.add("removeOnSave"));
+    sub("beforesave", outputDoc => {
+      html.bind(outputDoc.querySelector(".caniuse-stats"))`
+        <a href="${featureURL}">caniuse.com</a>`;
+    });
+  }
```
sidvishnoi, line 98: Should we remove it entirely instead?

marcoscaceres: For that case, the actual configuration option can be set to null.

PR #3789, opened 2021-09-28
fix(w3c/templates/headers): en translations for Process 2021

marcoscaceres:


sidvishnoi:
This might make more sense to be with the actual usage.

marcoscaceres:
Unfortunately, that's part of the stuff I can't split :( I'd rather take this in, then add the bits that use it.

sidvishnoi (approved):

PR #3791, opened 2021-09-28
feat(core/github): add private excludeGithubLinks option

marcoscaceres:
Also adds some additional URLs to `normalizedGHObj`, which are needed by Process 2021.

### DIFF
--- a/src/core/github.js
+++ b/src/core/github.js
@@ -153,10 +153,6 @@ export async function run(conf) {
     repoURL: ghURL.href,
     apiBase: githubAPI,
     fullName: `${org}/${repo}`,
-    issuesURL: issueBase,
-    pullsURL: newProps.pullBase,
-    newIssuesURL: new URL("./new", issueBase).href,
-    commitHistoryURL: commitHistoryURL.href,
   };
   resolveGithubPromise(normalizedGHObj);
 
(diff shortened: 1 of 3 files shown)


sidvishnoi (approved):

src/core/github.js
```diff
@@ -138,11 +121,42 @@ export async function run(conf) {
       showWarning(msg, name);
     }
   }
+  if (!conf.excludeGithubLinks) {
```
sidvishnoi, line 124: Should we also exclude editor draft URL?

src/core/github.js
```diff
@@ -138,11 +121,42 @@ export async function run(conf) {
...
+  }
   const normalizedGHObj = {
     branch,
     repoURL: ghURL.href,
     apiBase: githubAPI,
     fullName: `${org}/${repo}`,
+    issuesURL: issueBase,
+    pullsURL: newProps.pullBase,
+    newIssuesURL: new URL("./new", issueBase).href,
+    commitHistoryURL: commitHistoryURL.href,
```
marcoscaceres, line 159: Can do these clean as a followup when they get used... 

```suggestion
```

src/core/github.js
marcoscaceres: Still under discussion... eventually, but it's up to the working group and requires a culture change at the W3C.

marcoscaceres: Or do you mean when this option is added?

sidvishnoi: Yes, when this option is added

marcoscaceres: After the 2021 stuff, I'm going to remove conf.excludeGithubLinks, as this module shouldn't be affecting what goes in the headers.

PR #3793, opened 2021-09-29
fix(w3c/headers-spec): wrapped header in <details>

marcoscaceres:
As part of 2021, we are wrapping the header in `<details>`. 

I'm setting it to be open by default for now, but eventually there will be a site-wide setting to remember if it should be open or closed.


marcoscaceres:
Merging this one, as it's trivial and will allow me to work on the others.

Issue #3533, opened 2021-05-13
Use shortname when downloading


marcoscaceres:
**Is your feature request related to a problem? Please describe.**
When exporting, it's frustrating that every file is called "index.html"... 

**Describe the solution you'd like**
For the short name to be used, if available.

PR #3808, opened 2021-10-05
feat(ui/save-html): add shortName, publishDate to file name

sidvishnoi:
Closes https://github.com/w3c/respec/issues/3533
`backlog--`

### DIFF
--- a/src/ui/save-html.js
+++ b/src/ui/save-html.js
@@ -1,7 +1,7 @@
 // @ts-check
 // Module ui/save-html
 // Saves content to HTML when asked to
-import { getIntlData, showWarning } from "../core/utils.js";
+import { ISODate, getIntlData, showWarning } from "../core/utils.js";
 import { html } from "../core/import-maps.js";
 import { rsDocToDataURL } from "../core/exporter.js";
 import { ui } from "../core/ui.js";
@@ -30,7 +30,7 @@ const l10n = getIntlData(localizationStrings);
 const downloadLinks = [
   {
     id: "respec-save-as-html",
-    fileName: "index.html",
+    fileName: (name = "index") => `${name}.html`,
     title: "HTML",
     type: "text/html",
     get href() {
@@ -39,7 +39,7 @@ const downloadLinks = [
   },
   {
     id: "respec-save-as-xml",
-    fileName: "index.xhtml",
+    fileName: (name = "index") => `${name}.xhtml`,
     title: "XML",
     type: "application/xml",
     get href() {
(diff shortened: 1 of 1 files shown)


sidvishnoi:
Tests failing as https://status.w3.org/incidents/lfwjx05yvldr

src/ui/save-html.js
```diff
@@ -63,38 +63,46 @@ const downloadLinks = [
...
 ];
 
-function toDownloadLink(details) {
+/**
+ * @param {typeof downloadLinks[0]} details
+ */
+function toDownloadLink(details, conf) {
   const { id, href, fileName, title, type } = details;
+  const filename = conf.shortName
+    ? `${conf.shortName}-${ISODate.format(conf.publishDate)}`
```
marcoscaceres, line 72: Is publishDate assured? Or should we not added it if missing?

marcoscaceres (approved):
Worth adding a test?

src/ui/save-html.js
sidvishnoi: image
Should exist on each, but maybe good idea to add a failsafe.

sidvishnoi:
> Worth adding a test?

Would need to add tests for rest of the UI stuff too (we've zero UI/pill tests as of now).

src/ui/save-html.js
```diff
@@ -63,38 +63,46 @@ const downloadLinks = [
...
 ];
 
-function toDownloadLink(details) {
+/**
+ * @param {typeof downloadLinks[0]} details
+ */
+function toDownloadLink(details, conf) {
   const { id, href, fileName, title, type } = details;
+  const filename = conf.shortName
+    ? `${conf.shortName}-${ISODate.format(conf.publishDate || new Date())}`
```
marcoscaceres, line 72: Maybe format the date as a W3C date? Or does it come out nice? (Sorry not at computer, but can check tomorrow)

sidvishnoi: I looks like: `{shortName}-2021-10-05`, which I think is a good file name.

marcoscaceres:
I might have a little play with this tomorrow. Now I’m kinda wanting to add the specStatus too 🥰

src/ui/save-html.js
marcoscaceres: nah, that's good enough 👍

marcoscaceres, line 72: I think we want this:

```suggestion
    ? `${conf.specStatus}-${conf.shortName}-${concatDate(conf.publishDate || new Date())}`
```

So then it comes out like this:
Screen Shot 2021-10-06 at 3 37 36 pm

Which was the original motivation.

PR #3810, opened 2021-10-06
tests(core/xref): use new WebIDL spec URL

sidvishnoi:


tests/spec/core/xref-spec.js
```diff
@@ -21,6 +21,7 @@ describe("Core — xref", () => {
       id: "service-workers-1",
       href: "https://www.w3.org/TR/service-workers-1/",
     },
+    webidl: { id: "WebIDL", href: "https://webidl.spec.whatwg.org/" },
```
sidvishnoi, line 24: Looks like Specref hasn't updated yet.

marcoscaceres: Yep, I’m working on that patch 😅

PR #3813, opened 2021-10-07
feat(w3c/headers): support dual + document license

marcoscaceres:
Simplifies the copyright notice template, and moves some of the logic out template.


marcoscaceres:
Yep, will add those now...

PR #3815, opened 2021-10-08
fix(tocIntroductory): remove tocIntroductory warning

marcoscaceres:


sidvishnoi:
Also need to update docs for this iirc.

PR #3820, opened 2021-10-12
New specStatus: editor-draft-finding

darobin:
When the `draft-finding` status was created, all TAG documents were in TAG datedspace, and those assumptions carried over to which headers ought to be used. Since the TAG now does a lot of work on GitHub, but there is still a notion of a "draft finding" that needs to be maintained, this introduces a new `specStatus: editor-draft-finding`. This matches the usage that is emulated in https://w3ctag.github.io/privacy-principles/ for instance.


marcoscaceres:
In respec@26.16.0 🎉🚀

PR #3846, opened 2021-10-29
BREAKING CHANGE: switch to Node 16 (LTS) + update deps

marcoscaceres:
For the dev server, macOS Monterey seems to claim port 5000 for itself for the "airplay" server.


sidvishnoi:
I need to use Node 16 in spec-prod else `yarn add respec` fails.

sidvishnoi:
spec-prod now uses Node 16. https://github.com/w3c/spec-prod/pull/113.

PR #3858, opened 2021-11-03
fix(w3c/templates/headers): open details based on fixup.js

marcoscaceres:
As fixup.js is loaded late, this prevents the `<details>` element from randomly closing after the document loads by initializing in the right state.


tests/spec/w3c/headers-spec.js
```diff
@@ -40,6 +40,18 @@ describe("W3C — Headers", () => {
     expect(dl).toBeTruthy();
   });
 
+  it("removes the open attribute from details on save", async () => {
+    const opts = makeStandardOps({ specStatus: "WD" });
+    const doc = await makeRSDoc(opts);
+    doc.querySelector(".head details").open = true;
+    const exportedDoc = new DOMParser().parseFromString(
```
sidvishnoi, line 47: We've a helper to get Exported doc

marcoscaceres: oh, forgot... having a look.

marcoscaceres: You sure? at least one other test does the same?

marcoscaceres: found it... `getExportedDoc()` ... that seems like overkill.. it creates an iframe, etc.

sidvishnoi: We can make that one better then.

marcoscaceres: follow up, me thinks?

marcoscaceres: (there is nothing super wrong with other one... it's fit for purpose if running doc is needed... for this, just need to check something in the dom from the generated output).

sidvishnoi: I would prefer to use getExportedDoc here and follow-up with the improvement to getExportedDoc later.

marcoscaceres: That doesn't work for the reason I mentioned... if the document runs in the iframe, then fixup.js adds "open" to it, so the test would fail.

sidvishnoi: Ah! Understood.

sidvishnoi (approved):

dontcallmedom:
Pubrules requires documents to be published with `<details open=''>` which I believe this change prevents from happening (see [related spec-prod failure](https://lists.w3.org/Archives/Public/public-tr-notifications/2021Nov/0082.html))

PR #3870, opened 2021-11-09
WIP: WPT integration

marcoscaceres:
Work in progress to show WPT results and addition to caniuse. 

cc @sideshowbarker


src/w3c/wpt.js
```diff
@@ -0,0 +1,37 @@
+// Show WPT results
+export const name = "w3c/wpt";
+
+/**
+ *
+ * @param {RespecConfig} conf
+ */
+export async function run(conf) {
+  // if (!conf.wpt) return;
+  // Gett the run ids first
```
sideshowbarker, line 10:
```suggestion
  // Get the run ids first
```

mera15 (changes requested):
profiles/w3c.js
wpt-results

jhonmurga (commented):
``

marcoscaceres:
Problem we still have is some features only being exposed on particular platforms/os combinations… see Web Share API, which is on Firefox Windows and Android. Similarly, for Chrome: only supported on Android. 

Not much we can do about that, as wpt.fyi doesn’t test every platform combo… but it makes things a little messy/confusing for somebody that doesn’t know.

PR #3874, opened 2021-11-11
fix(core/id-headers): improve and localize aria-label

### LINKED ISSUE #3853: Accessible names for permalinks
## Important info

- URL to affected spec: https://w3c.github.io/epub-specs/epub33/core/
- ReSpec version: 26.16.1

## Description of problem

Another respec-related issue that came up in the epub accessibility horizontal review was the lack of meaningful labels for the permalinks. Currently they all get an `aria-label` attribute with the value '§'.

I've patched this for epub in a post-processing step that identifies what is getting permalinked and using that label with the number. For example, "Permalink for Section #", "Permalink for Example #", etc.

It looks like this problem was manually fixed in the last WCAG release to use the entire heading value, so you get "Permalink for #.#.# Blah blah title".

That looks simpler to implement, but I'm not sure if it would be annoying to users given its repetition of the entire title. We'll need to figure out what approach is best before a solution can be proposed.

For more info about the original issue, see https://github.com/w3c/epub-specs/issues/1827

marcoscaceres:
closes #3853

Also adds permalinks to all introductory sections... I don't see any reason why those shouldn't have permalink also. 

@mattgarrish, this now adds:

 * Permalink for Section N.N. 
 * Permalink for Appendix N.N.


tests/spec/core/id-headers-spec.js
```diff
@@ -31,13 +37,25 @@ describe("Core - ID headers", () => {
...
-    it("doesn't add sections links to introductory sections", () => {
-      expect(doc.querySelector(".introductory h2 a.self-link")).toBeFalsy();
+    it("adds sections links to introductory sections", () => {
+      expect(doc.querySelector(".introductory h2 a.self-link")).toBeTruthy();
     });
 
-    it("(aria) labels section links", () => {
-      const queryForSymbol = "h2 a[aria-label=§].self-link";
-      expect(doc.querySelector(queryForSymbol)).toBeTruthy();
+    it("adds aria-labels section, including section type", () => {
```
sidvishnoi, line 44:
```suggestion
    it("adds aria-label with section type and number", () => {
```

tests/spec/core/id-headers-spec.js
```diff
@@ -11,14 +11,20 @@ describe("Core - ID headers", () => {
...
     <section id="sotd" class="notoc">
-    <p>...</p>
-    <section>
-      <h3>Level 3 heading</h3>
+      <p>...</p>
       <section>
-        <h4>Level 4 heading</h4>
+        <h3>Level 3 heading</h3>
+        <section>
+          <h4>Level 4 heading</h4>
```
sidvishnoi, line 18: Let's add a test for this too, given there's a change here.

src/core/id-headers.js
```diff
@@ -20,8 +28,15 @@ export function run(conf) {
       id = h.parentElement.id || h.id;
     }
     if (!conf.addSectionLinks) continue;
-    h.appendChild(html`
-      <a href="${`#${id}`}" class="self-link" aria-label="§"></a>
+    let label = h.closest(".appendix")
+      ? l10n.perma_for_appendix
+      : l10n.perma_for_section;
+    const sectionNumber = h.querySelector(":scope > bdi.secno");
+    if (sectionNumber) {
```
sidvishnoi, line 35: What are the cases where the `bdi.secno` might not exist?

src/core/id-headers.js
```diff
@@ -20,8 +28,15 @@ export function run(conf) {
       id = h.parentElement.id || h.id;
     }
     if (!conf.addSectionLinks) continue;
-    h.appendChild(html`
-      <a href="${`#${id}`}" class="self-link" aria-label="§"></a>
+    let label = h.closest(".appendix")
+      ? l10n.perma_for_appendix
+      : l10n.perma_for_section;
```
sidvishnoi, line 33: I feel it might be better if we have something like:
`label = l10n.permalinkLabel(type = "appendix" | "section", sectionNumber)`

src/core/id-headers.js
```diff
@@ -20,8 +28,15 @@ export function run(conf) {
...
+    const sectionNumber = h.querySelector(":scope > bdi.secno");
+    if (sectionNumber) {
+      label += ` ${norm(sectionNumber.textContent)}`;
```
sidvishnoi, line 36: Nit: maybe we should we add the sectionNumber as a data attribute on `<bdi>` in `core/structure`? Could be nice as we've also similar thing in `core/dfn-index`.

sidvishnoi (changes requested):

src/core/id-headers.js
marcoscaceres: Anything that is "introductory" (e.g., "Abstract", "Status of This Document").

src/core/id-headers.js
sidvishnoi: Yeah, it's same same.
https://github.com/w3c/respec/blob/34242b5f079ba209433966aebb9c6255d35a9d0e/src/core/dfn-index.js#L225-L227

sidvishnoi: Though, feel free to ignore this suggestion. Non-blocking.

marcoscaceres:
Merging, as feedback is all addressed.

mattgarrish:
Thanks!

PR #3908, opened 2021-12-09
fix(core/pubsubhub): account for all message types

marcoscaceres:
Events were being broadcast to no one... this now forces every registered `pub()` to have a listener of a particular name.

Hopefully we can get rid of the eventing model in the future. It creates too many interdependent parts.


tests/spec/core/validators-spec.js
```diff
@@ -30,6 +31,13 @@ describe("Core - Validators", () => {
     it("generates an error if the element name is not valid", () => {
       const elements = ["my element", "crypto$", "🪳", "-something", ""];
       for (const element of elements) {
+        sub(
```
sidvishnoi, line 34: We should probably use a spy here, and/or use jasmine's `toHaveBeenCalledWith`

src/core/pubsubhub.js
```diff
@@ -12,9 +12,14 @@ export const name = "core/pubsubhub";
 
 const subscriptions = new Map();
 
+/**
+ *
+ * @param {EventTopic} topic
+ * @param  {...any} data
```
sidvishnoi, line 18: Optional TypeScript improvement: We can make EventTopic an interface and make these functions generic over that. Then we would get proper type checking in all pubsub calls.

src/core/examples.js
```diff
@@ -113,7 +111,6 @@ export function run() {
       const selfLink = div.querySelector("a.self-link");
       selfLink.href = `#${div.id}`;
       example.replaceWith(div);
-      if (!inAside) pub("example", report);
```
sidvishnoi, line 116: Nice! _Begone!_

src/core/pubsubhub.js
marcoscaceres: Yeah, was thinking the same... but I need to refactor all the call sites... they only use a single argument.

I found a huge bug also with pubsubhub.js... it wasn't designed to handle async functions, and now we have a mess for supporting "conf.beforeSave".

tests/spec/core/validators-spec.js
marcoscaceres: Nearly went there, but this seemed cleaner to me... the spy adds a lot of indirection.

marcoscaceres: Got rid of the actual test part, as it's not important for these.

marcoscaceres: Plus a lot of these tests still check "respec-offending-element" :( will need to rewrite them later.

tests/unit/SpecHelper.js
```diff
@@ -14,6 +14,11 @@ export function makePluginDoc(
   plugins,
   { config = {}, head = `<meta charset="UTF-8" />`, body = "" } = {}
 ) {
+  plugins = [
+    `/src/core/ui.js`, // Needed for "start-all" event
+    `/src/core/dfn.js`, // Needed for "plugins-done" event,
```
sidvishnoi, line 19: Nit: Let's either move this to `allPlugins` line below or move the `/src/core/base-runner.js` here,

sidvishnoi (approved):

PR #3909, opened 2021-12-09
feat(core/beforeSave): support modifications before save

marcoscaceres:
Adds a `beforeSave` configuration option, which takes an array of sync JS functions. Unfortunately, due to a limitation of pubsubhub, we can't support async functions yet. But that should be ok for now.


src/core/beforeSave.js
```diff
@@ -0,0 +1,40 @@
...
+      let pos = 0;
+      for (const fn of conf.beforeSave) {
+        try {
+          fn(ownerDocument);
+        } catch (err) {
+          const nameOrPosition = `\`${fn.name}\`` || `at position ${pos}`;
+          const msg = docLink`Function ${nameOrPosition}\` threw an error during processing of ${"[beforeSave]"}.`;
+          const hint = "See developer console.";
+          showError(msg, name, { hint });
+          console.error(err);
```
sidvishnoi, line 32: I think we support passing `{error}` in `showError`? I might be wrong. We should support that if not.

src/core/beforeSave.js
```diff
@@ -0,0 +1,40 @@
...
+    return;
+  }
+
+  sub(
+    "beforesave",
+    ({ ownerDocument }) => {
+      let pos = 0;
+      for (const fn of conf.beforeSave) {
+        try {
+          fn(ownerDocument);
```
sidvishnoi, line 26: Should it also have access to `conf`?

src/core/beforeSave.js
```diff
@@ -0,0 +1,40 @@
...
+    )
+  ) {
+    const msg = docLink`${"[beforeSave]"} configuration option must be an array of synchronous JS functions.`;
+    showError(msg, name);
+    return;
+  }
+
+  sub(
+    "beforesave",
+    ({ ownerDocument }) => {
```
sidvishnoi, line 22: Nit: let's move this into a named function for clarity.

tests/spec/core/beforeSave-spec.html
```diff
@@ -0,0 +1,32 @@
+<!DOCTYPE html>
+<script class="remove">
+  const modifier1 = exportDoc => {
```
sidvishnoi, line 3: Nit:
```suggestion
  const modifier1 = exportedDoc => {
```

tests/spec/core/beforeSave-spec.html
```diff
@@ -0,0 +1,32 @@
...
+    exportDoc.body.appendChild(p);
+  };
+  const respecConfig = {
+    specStatus: "unofficial",
+    shortName: "i",
+    editors: [
+      {
+        name: "Foo",
+        url: "https://foo.com/",
+      },
```
sidvishnoi, line 22:
```suggestion
      { name: "Foo", url: "https://foo.com/" },
```
(_OCD_)

tests/spec/core/beforeSave-spec.js
```diff
@@ -0,0 +1,55 @@
...
+    const p2 = exportedDoc.querySelector("#p2");
+    expect(p1.nextElementSibling).toBe(p2);
+  });
+
+  it("complains if it's not passed an array", async () => {
+    const ops = makeStandardOps({ beforeSave: "not a array" });
+    const doc = await makeRSDoc(ops);
+    const errors = doc.respec.errors.filter(
+      err => err.plugin === "core/beforeSave"
+    );
```
sidvishnoi, line 31: Nit: can move this repeated code block into a function `getErrors(doc)` to the parent `describe` block.

sidvishnoi (commented):
Looks good. What prompted for this feature though?
(In future, I would love if we can have https://github.com/w3c/respec/pull/3320.)

src/core/beforeSave.js
marcoscaceres: I don't think so, tbh. The conf is already global and we want to discourage people poking at the private stuff.

src/core/beforeSave.js
marcoscaceres: ah, let me check...

sidvishnoi: We don't. We can however use `Error.cause`. Let's do that as follow up.

profiles/w3c.js
```diff
@@ -59,6 +59,7 @@ const modules = [
   import("../src/core/anchor-expander.js"),
   import("../src/core/custom-elements/index.js"),
   import("../src/core/web-monetization.js"),
+  import("../src/core/beforeSave.js"),
```
sidvishnoi, line 62: Let's follow the file naming conventions we already have (i.e. `before-save.js`)

sidvishnoi, line 62: Might want to add it to other profiles too as it's a core feature.

marcoscaceres: I'll let there other folks do that, as it's up to them to maintain their own profiles.

tests/spec/core/beforeSave-spec.js
marcoscaceres: Added UIMessageFilters, and errorFilters, and warningFilters exports... wdyt?

tests/spec/SpecHelper.js
```diff
@@ -56,6 +56,34 @@ export function makeRSDoc(opts, src, style = "") {
...
+    this.type = type;
+  }
+  /**
+   * @param {string} pluginName
+   * @returns (Document) => Array<RespecError>
+   */
+  filter(pluginName) {
+    if (this.has(pluginName)) {
+      return this.get(pluginName);
+    }
```
sidvishnoi, line 77: Not following why we might want a Map here.

marcoscaceres: You always need a map :) ... but I'm thinking I'll just make it internal, not extension.

src/core/beforeSave.js
marcoscaceres: Yeah, good idea... been meaning to play with `.cause`

marcoscaceres:
Going ahead with this one, as I need `UIMessageFilters` for another patch I'm working on. 

But we can revise UIMessageFilters there, but hopefully it's all fine now.

Issue #3510, opened 2021-05-06
Update icons to use SVG


marcoscaceres:
In ReSpec's menu, we use emojis for the menu icons, which doesn't look great and is inconsistent across platforms... we should use images there instead, like:

https://iconic.app

PR #3919, opened 2021-12-20
fix(core/ui): improve warning/errors + margin fixes

marcoscaceres:
Small improvement... just splits out the hints a bit more.

Screen Shot 2021-12-20 at 6 50 26 pm

Could use more work but better than before.


marcoscaceres:
need to check what's broken in the tests... but you get the idea.

sidvishnoi:
We should add that left padding on everything, not only on hint I think. Below (current) looks weird:
image

src/styles/ui.css.js
```diff
@@ -184,6 +188,16 @@ export default css`
...
 
+.respec-warning-list > li:last-child,
+.respec-error-list > li:last-child {
+  margin-bottom: 0;
+}
+
+.respec-warning-list > li:first-child,
+.respec-error-list > li:first-child {
+  margin-top: 0;
+}
```
sidvishnoi, line 199: How about using [lobotomized owl selector](https://alistapart.com/article/axiomatic-css-and-lobotomized-owls/):
```suggestion
.respec-warning-list > li + li,
.respec-error-list > li + li {
  margin-top: 0.5rem;
}
```

(Will also need following change at line 184/188)
```diff
 .respec-warning-list > li,
 .respec-error-list > li {
   padding: 0.4em 0.7em;
+  margin: 0;
 }
```

src/styles/ui.css.js
```diff
@@ -359,4 +376,11 @@ export default css`
   margin: 0 0.5em 0.5em;
   border-bottom-width: 1px;
 }
+
+.respec-hint {
+  display: inline-block;
+  padding-top: .2cm;
+  padding-left: .7cm;
```
sidvishnoi, line 383: Oddly specific.

sidvishnoi (changes requested):

src/styles/ui.css.js
```diff
@@ -14,6 +14,10 @@ export default css`
   background: transparent;
 }
 
+.respec-modal .close-button:focus {
+  outline:none;
```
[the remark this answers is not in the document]
marcoscaceres: yeah, maybe something less ugly here... I really don't like how it puts the blue outline on the X tho.

src/styles/ui.css.js
marcoscaceres: This looks neat, but I'm not following this? 

The current problem is the padding seen here: 

Screen Shot 2021-12-21 at 5 59 53 pm
 
If the above fixes that, then sure... but li:first-child and li:last-child is more explicit IMO? (i.e., the rest are all fine)

src/styles/ui.css.js
marcoscaceres: heh, yeah... this whole thing needs a rewrite (maybe I'll do that over the break)... I figure, we are already doing `cm` here, so why not.

marcoscaceres:
yeah, true about putting padding on everything....

src/styles/ui.css.js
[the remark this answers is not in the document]
sidvishnoi: Blue outline: I think that depends on browser.
Something less ugly: Maybe https://github.com/w3c/respec/issues/3510 can help. We can change SVG params or use two different icons to show focus.

But let's keep the outline for now.

src/styles/ui.css.js
sidvishnoi: Above (my suggestion) fixes the white margin issue.

> li:first-child and li:last-child is more explicit IMO

But longer. I think the lobotomized owl selector is fairly common (it's basically saying: add a `margin-top` on every `li` that is preceded by a `li` - so that excludes the first; above link has good diagrams if you want to see how it works).
_[Even tailwind people use it](https://tailwindcss.com/docs/space)._

src/styles/ui.css.js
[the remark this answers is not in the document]
marcoscaceres: I’ll see if I can at least make a softer outline.

src/styles/ui.css.js
marcoscaceres: “lobotomized owl” hehe, never heard of this. I’ll take a look - same with tailwind (never heard of it).

src/styles/ui.css.js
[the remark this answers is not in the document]
marcoscaceres: Ok, will keep the ugly outline for now :(

marcoscaceres:
Pretty happy with this now... much nicer to work with.

src/styles/ui.css.js
```diff
@@ -359,4 +351,15 @@ export default css`
   margin: 0 0.5em 0.5em;
   border-bottom-width: 1px;
 }
+
+p:is(.respec-hint,.respec-occurrences) {
+  display: block;
+  padding-top: 1em;
```
sidvishnoi, line 357: This feels a bit too much to me.
image

Perhaps:
```suggestion
  margin-top: 0.5em;
```
image

sidvishnoi (approved):
Nice! Much nicer than before. One nit on spacing.
I think using SVG icons would help further, but not needed.

Issue #3754, opened 2021-08-24
Don't require sections - let ReSpec handle them


marcoscaceres:
**Is your feature request related to a problem? Please describe.**
Apparently, wrapping thing in `<section>` elements is a feature some folks don't like about ReSpec.

ReSpec has been opinionated about `<section>` from the very beginning, but the Markdown module automatically wrap headings into `<sections>`. 

**Describe the solution you'd like**

I wonder if we could take the `Builder` and the `structure()` method from Markdown.js and apply it to documents. 

This might actually already work by simply enabling: `format: "markdown"`, but it might be nice as a more general opt in, without running everything through the markdown parser.

marcoscaceres:
Yeah, `Builder` and `structure()` do the right thing when `format: "markdown"` is set on a regular document.

A thing to fix is that the headings ids include the section numbers (e.g., "x2-1-i-m-a-sub-section"). However, the ToC uses the section ID, which is always correct e.g. `"i-m-a-sub-section"`.

PR #3927, opened 2021-12-22
fix(w3c/abstract): find abstract from markdown

### LINKED ISSUE #3897: Reuse abstract section if present when in markdown mode
## Important info

- ReSpec version: 28.0.5

## Description of problem

Because editing HTML is troublesome, and not all editors are good at HTML, I am trying to convert some ReSpec documents into Markdown by adding `format: "markdown"` to the config, but encountered some problems.

The first issue is that the Abstract and sotd sections do not seem to support Markdown. See the [code](https://github.com/xfq/testing/blob/9ab55220464ebfa86d98a564a95f32dcdb46225b/simple-ruby-markdown-rs-v1/index.html#L14-L24) and [rendered version](https://xfq.github.io/testing/simple-ruby-markdown-rs-v1/) of a document for an example. The style of the rendered document seems broken.

So I tried to turn the Abstract and sotd parts into HTML, but the Markdown parts in the body text were not rendered at all. See the [code](https://github.com/xfq/testing/blob/main/simple-ruby-markdown-rs-v2/index.html) and [rendered version](https://xfq.github.io/testing/simple-ruby-markdown-rs-v2/) of a document for details. Because I can't use Markdown for some of the main text (for example, I'm not able to add an `id` to an `li` in Markdown), I used HTML for those parts.

Maybe I'm using it wrong, but I'm not sure how to debug this. Can I get a pointer to information on writing specs in Markdown?

marcoscaceres:
Closes #3897

### DIFF
--- a/src/w3c/abstract.js
+++ b/src/w3c/abstract.js
@@ -33,7 +33,7 @@ const l10n = getIntlData(localizationStrings);
 export async function run() {
   const abstract = findAbstract();
   if (!abstract) {
-    const msg = 'Document must have one `<section id="abstract">` in the body.';
+    const msg = 'Document must have one `<section id="abstract">`.';
     showError(msg, name);
     return;
   }
--- a/src/w3c/abstract.js
+++ b/src/w3c/abstract.js
@@ -1,7 +1,8 @@
 // @ts-check
 // Module w3c/abstract
 // Handle the abstract section properly.
-import { getIntlData, showError } from "../core/utils.js";
+import { getIntlData, norm, showError } from "../core/utils.js";
+import { html } from "../core/import-maps.js";
 export const name = "w3c/abstract";
 
 const localizationStrings = {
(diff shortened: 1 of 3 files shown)


sidvishnoi:
Shouldn't this be solved via https://github.com/w3c/respec/issues/3754 instead?

marcoscaceres:
> Shouldn't this be solved via #3754 instead?

Yeah, though it's fixing a different thing. This one is also kinda special: it needs the section to have the ID,  it has to be a H2 and has traditionally (vomit) not required the H2 because it gets magically added.

marcoscaceres:
> #3754 

put differently, I'm sure this one will break #3754 🤣

I might hack on #3754 tomorrow

sidvishnoi:
Ah! Understood. I think then the error message should say use `<section id="abstract">` or a heading with text "Abstract" (with l10n)?

src/w3c/abstract.js
```diff
@@ -30,18 +31,33 @@ const localizationStrings = {
 const l10n = getIntlData(localizationStrings);
 
 export async function run() {
-  const abs = document.getElementById("abstract");
-  if (!abs) {
-    const msg = 'Document must have one element with `id="abstract"`.';
+  const abstract = findAbstract();
+  if (!abstract) {
+    const msg = 'Document must have one `<section id="abstract">` in the body.';
```
sidvishnoi, line 36: Nit: "in the body" seems a bit redundant, no?
```suggestion
    const msg = 'Document must have one `<section id="abstract">`.';
```

marcoscaceres:
> or a heading with text "Abstract" (with l10n)?

Yeah, something like that could work. Although, the key thing is just the `<section id="abstract">`, the `<h2>` can be whatever...

sidvishnoi (approved):

PR #3928, opened 2021-12-23
fix(w3c/linter-rules/required-sections): support other locales

marcoscaceres:
Forgot that this would get upset if the document was not in English.

### DIFF
--- a/src/w3c/linter-rules/required-sections.js
+++ b/src/w3c/linter-rules/required-sections.js
@@ -13,6 +13,7 @@ import {
   showError,
   showWarning,
 } from "../../core/utils.js";
+import { lang } from "../../core/l10n.js";
 import { recTrackStatus } from "../headers.js";
 
 const ruleName = "required-sections";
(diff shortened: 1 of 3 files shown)


src/w3c/linter-rules/required-sections.js
```diff
@@ -40,15 +55,20 @@ export function run(conf) {
     return;
   }
 
+  // We can't check for headers unless we also have a translation
+  if (!localizationStrings[lang]) {
+    return;
```
sidvishnoi, line 60: Should add a warning here. A `console.warn` if not `showWarning`..

src/w3c/linter-rules/required-sections.js
```diff
@@ -40,15 +55,20 @@ export function run(conf) {
     return;
   }
 
+  // We can't check for headers unless we also have a translation
+  if (!localizationStrings[lang]) {
```
sidvishnoi, line 59: There's a little caveat here. I think we should normalize the `lang` here like [`getIntlData` does](https://github.com/w3c/respec/blob/0067ad45facb56565463a5fe18b8eb97ffc222a2/src/core/utils.js#L181).

src/w3c/linter-rules/required-sections.js
marcoscaceres: Not sure about that ... there is nothing the spec Editor can do about it? 

Also, spec translations are not canonical, so the warnings is just nice to have in this context.

src/w3c/linter-rules/required-sections.js
marcoscaceres: yeah, can run it over the same code.

src/w3c/linter-rules/required-sections.js
marcoscaceres: Went with the  console.warn ... I think that works for now.

sidvishnoi: Yeah, console.warn is good. As the author (or us) will know what's happening if that happens. Less unexpected.

sidvishnoi (approved):

PR #3929, opened 2021-12-23
BREAKING CHANGE: Support <section>less editing

marcoscaceres:
Ok, draft of what it looks like to edit without sections.

Todo: 
- [ ] add tests for sections.js 

This allows:

```
<h2>Abstract</h2>
<p>Wrapping things in sections is hard.</p> 

<h3>My section</h3>
<p>bla bla</p> 

<h2>New h2</h2>
<p>And so on...</p>
```

This also now id's headings to match their initial text, instead of the section.

Sections get id also, but "section-whatever". 

Open questions: 

 * maybe the section should retain the canonical id (probably, yes...). I'll sleep on it.


marcoscaceres:
Actually, yeah... I shouldn't have changed the id generation... will send again.

PR #3930, opened 2021-12-23
BREAKING CHANGE: support <section>less editing

### LINKED ISSUE #3754: Don't require sections - let ReSpec handle them

marcoscaceres:
Closes #3754

Allows editing documents without requiring `<section>`.

This PR is mostly just a refactor. It moves "Builder" to its own module "sections"...


src/core/sections.js
```diff
@@ -0,0 +1,118 @@
...
+    section.remove();
+    elem.append(...section.childNodes);
+  } else {
+    elem.textContent = "";
+  }
+  elem.appendChild(structuredInternals);
+}
+
+export function run() {
+  restructure(document.body);
```
marcoscaceres, line 117: Should we make it configurable? Like, run if "markdown" and "????".... or, just make it be the standard operating mode?

src/core/markdown.js
```diff
@@ -182,118 +176,6 @@ function workaroundBlockLevelMarkdown(element, selector) {
...
-
-function structure(fragment, doc) {
-  function process(root) {
-    const stack = new Builder(doc);
-    while (root.firstChild) {
-      const node = root.firstChild;
-      if (node.nodeType !== Node.ELEMENT_NODE) {
-        root.removeChild(node);
-        continue;
-      }
```
marcoscaceres, line 256: This is the only thing I had to remove... it was discarding too much stuff (well, stuff that was causing tests to fail that were expecting comment nodes... but I think it also trashed TEXT nodes too).

tests/spec/w3c/linter-rules/required-sections-spec.js
```diff
@@ -11,7 +11,11 @@ import { noTrackStatus } from "../../../../src/w3c/headers.js";
 import { requiresSomeSectionStatus } from "../../../../src/w3c/linter-rules/required-sections.js";
 
 describe("w3c — required-sections", () => {
+  beforeAll(() => {
+    jasmine.DEFAULT_TIMEOUT_INTERVAL *= 2;
```
marcoscaceres, line 15: I think the w3c server was being slow tonight... I don't think we actually need this.

tests/spec/w3c/headers-spec.js
```diff
@@ -1962,9 +1962,6 @@ describe("W3C — Headers", () => {
       <section id="first-sub-section">
         <h3>Testing</h3>
       </section>
-      <p id="p3">
```
marcoscaceres, line 1965: The structuring code doesn't like this kind of thing... it makes a new section or something... this kind of thing shouldn't happen anyway, so 🤷‍♂️

tests/spec/core/markdown-spec.js
```diff
@@ -22,7 +22,7 @@ describe("Core - Markdown", () => {
     });
     const foo = doc.getElementById("foo");
     expect(foo).toBeTruthy();
-    expect(foo.textContent).toBe("1. Foo");
+    expect(foo.textContent.trim()).toBe("1. Foo");
```
marcoscaceres, line 25: Now that we don't throw out text nodes, I think the markdown parser generates some extraneous whitespace after the headings. No big deal though.

src/core/sections.js
```diff
@@ -0,0 +1,118 @@
+/**
+ * Module core/sections
+ *
+ * Adds sections to the document, based on the heading structure.
```
sidvishnoi, line 4:
```suggestion
 * Adds <section>s to the document, based on the heading structure.
```

src/core/sections.js
```diff
@@ -0,0 +1,118 @@
...
+          stack.addElement(node);
+      }
+    }
+    return stack.root;
+  }
+  return process(fragment);
+}
+
+/**
+ * Re-structure DOM around elem whose markdown has been processed.
```
sidvishnoi, line 98: Comment needs update. Unrelated to markdown now.

profiles/w3c.js
```diff
@@ -32,11 +32,12 @@ const modules = [
   import("../src/core/link-to-dfn.js"),
   import("../src/core/xref.js"),
   import("../src/core/data-cite.js"),
-  import("../src/core/webidl-index.js"),
   import("../src/core/render-biblio.js"),
   import("../src/core/dfn-index.js"),
   import("../src/core/contrib.js"),
+  import("../src/core/sections.js"),
```
sidvishnoi, line 38: Should add this one to other profiles I think. It's a core breaking change.

src/core/markdown.js
```diff
@@ -29,17 +29,11 @@
...
- *       <section>
- *         <h3>Subtitle</h3>
- *         <p>Here's some text.</p>
- *       </section>
- *       <section>
- *         <h3>Another subtitle</h3>
- *         <p>More text.</p>
- *       </section>
- *     </section>
+ *       <h3>Subtitle</h3>
```
sidvishnoi, line 33: Comment body above needs an update too.

src/core/sections.js
```diff
@@ -0,0 +1,118 @@
...
+        case "section":
+          stack.addSection(node, process);
+          break;
+        default:
+          stack.addElement(node);
+      }
+    }
+    return stack.root;
+  }
+  return process(fragment);
```
sidvishnoi, line 94: Nit: we can un-nest this nested function I guess.

src/core/sections.js
sidvishnoi: I think making it default should be fine, as long as resultant output is same.

tests/spec/core/id-headers-spec.js
```diff
@@ -33,7 +33,7 @@ describe("Core - ID headers", () => {
 
   it("sets an id on header", () => {
     const h2 = doc.querySelector("#t0 h2");
-    expect(h2.id).toBe("foo");
+    expect(h2.id).toBe("x1-foo");
```
sidvishnoi, line 36: Oh. This seems to be a breaking change. IIUC, it's supposed to be adding id `"foo"` to `<section>` now (but that section already has an id)?

tests/spec/core/id-headers-spec.js
```diff
@@ -46,7 +46,7 @@ describe("Core - ID headers", () => {
       expect(intro.getAttribute("aria-label")).toBe("Permalink for Section");
       const custom = doc.querySelector("#custom-id a.self-link");
       expect(custom.getAttribute("aria-label")).toBe(
-        "Permalink for Section 2."
+        "Permalink for Section 3."
```
sidvishnoi, line 49: What happened here?

tests/spec/core/markdown-spec.js
sidvishnoi: Should investigate I think.

tests/spec/core/sections-spec.js
```diff
@@ -0,0 +1,87 @@
...
+      <h5 id="h5">Section 5</h5>
+      <h6 id="h6">Section 6</h6>
+    `;
+    const ops = makeStandardOps(null, body);
+    const doc = await makeRSDoc(ops);
+    for (let i = 2; i <= 6; i++) {
+      const h = doc.getElementById(`h${i}`);
+      const section = h.parentElement;
+      expect(section.localName).toBe("section");
+      expect(section.id).toBe(`section-${i}`);
```
sidvishnoi, line 22: Nit: add `withContext`.

tests/spec/core/sections-spec.js
```diff
@@ -0,0 +1,87 @@
...
+    }
+  });
+
+  it("it handles when some headings are wrapped", async () => {
```
sidvishnoi, line 26:
```suggestion
  it("handles when some headings are wrapped", async () => {
```

tests/spec/core/sections-spec.js
```diff
@@ -0,0 +1,87 @@
...
+      const section = h.parentElement;
+      expect(section.localName).toBe("section");
+      expect(section.id).toBe(`section-${i}`);
+    }
+  });
+
+  it("retains different nodes when wrapping", async () => {
+    const body = `
+      <h2 id="h2">Section 2</h2>
+        <p>paragraph 2</p><!--comment 2-->
```
sidvishnoi, line 53: Is this indent on `<p>` intentional/meaningful?

tests/spec/w3c/headers-spec.js
sidvishnoi: I think this should be wrapped into a section of its own?

sidvishnoi: Section without heading that the linter shouts at.

tests/spec/w3c/linter-rules/required-sections-spec.js
sidvishnoi: Yeah, let's not forget to remove this. Or, we should probably set it at config level as it's a "timeout".

tests/spec/core/id-headers-spec.js
sidvishnoi: Concerned with any sort of xref/webref breakage mainly.

tests/spec/core/sections-spec.js
marcoscaceres: Nope. Will fix.

tests/spec/core/markdown-spec.js
marcoscaceres: Yeah, we have some inconsistency on how the ids are being assigned to sections and headers somewhere... I'll investigate further.

That also explains `.trim()` in a few places above... the `<section>`s are now getting the ids that were on the heading elements.

tests/spec/w3c/headers-spec.js
marcoscaceres: I think that gets messy/confusing, tbh... plus it's too much of an edge case to worry about.

sidvishnoi: Let's skip it then

tests/spec/core/id-headers-spec.js
marcoscaceres: This is consistent with "normal" mode, so it should be fine. Also, both ids (the one on the section and the one on the header are retained, so nothing will break.)

tests/spec/core/id-headers-spec.js
marcoscaceres: This is correct... removing the `<p>BLAH</p>` above (which we won't support), causes the sections to be ordered correctly... so this is now "3. Pass":

Screen Shot 2022-01-31 at 12 06 58 pm

marcoscaceres:
Ok, I think we are good to go here.

sidvishnoi (approved):
LGTM. Not sure if feat or breaking change anymore though.

marcoscaceres:
> LGTM. Not sure if feat or breaking change anymore though.

Same... I figured it's a breaking change because it breaks the profiles... so if someone out there has their own profile, this would break.

PR #3941, opened 2022-01-04
Added additional German translations.

init-dcat-ap-de:
The only questionable translation in my eyes is for "Edited in place". 

I translated it as "zuletzt geändert am" which means "last change at". I guess this is the meaning, since the text is used between the publishing date and the modified date.

The Japanese translation seems to be "Updated without revision", at least according to https://www.deepl.com/translator#ja/en/%E6%94%B9%E7%89%88%E3%81%AA%E3%81%8F%E6%9B%B4%E6%96%B0 

For "This document is also available in this non-normative format:" there seems to be no l10n variable.


marcoscaceres (approved):

marcoscaceres:
Thanks @init-dcat-ap-de! 

> For "This document is also available in this non-normative format:" there seems to be no l10n variable.

Good point. Sent https://github.com/w3c/respec/pull/3953/files

PR #3953, opened 2022-01-11
fix(w3c/headers): allow localizing multiple_alternates

marcoscaceres:
Small refactor to allow localization of "multiple_alternates".


src/w3c/templates/headers.js
```diff
@@ -25,6 +25,11 @@ const localizationStrings = {
     latest_recommendation: "Latest Recommendation:",
     message_topic: "… message topic …",
     more_details_about_this_doc: "More details about this document",
+    multiple_alternates(plural) {
+      return `This document is also available in ${
+        plural ? "these non-normative formats" : "this non-normative format"
```
sidvishnoi, line 30: Maybe we can use the `pluralize` library we've?

sidvishnoi (approved):
LGTM. One optional suggestion.
(Don't forget to update commit message before merge!)

src/w3c/templates/headers.js
marcoscaceres: It can only do single words 😢

sidvishnoi: I was thinking something like:
```js
      const count = plural ? 1 : 0;
      return `This document is also available in ${pluralize('this', count)} non-normative ${pluralize('format', count)}:`;
```
But never mind, not important.

PR #3957, opened 2022-01-17
fix(w3c/abstract): regression, keep supporting legacy div#abstract

### LINKED ISSUE #3955: New requirement to use section rather than div for SOTD & abstract
For some reason (i don't remember seeing any announcement, and can't find one), Respec now complains if you have used a div rather than a section tag for the abstract and SOTD text, and it seems that this needs to be changed to allow publication of the document. Why was this changed?

The i18n WG/IG is actively maintaining over 40 editor's drafts, so this is a pain, especially given that around 20 of those ED are typically republished without having to open the HTML and edit it (because the changed text is automatically sucked in from elsewhere).  Presumably this is also going to trip up a lot of other folks publishing documents at W3C, and make them re-edit their docs when they are trying to publish.

Why can't respec simply rewrite the div to a section automatically?

marcoscaceres:
fixes #3955


marcoscaceres:
test failures seem unrelated... they are also occurring in:
https://github.com/w3c/respec/runs/4823946674?check_suite_focus=true

sidvishnoi (approved):

PR #3967, opened 2022-01-21
fix(core/linter-rules/a11y): update axe to 4.* + fix missing .

marcoscaceres:


sidvishnoi:
Any new rules that we might want to enable/disable?

sidvishnoi (approved):
Any new rules that we might want to enable/disable? Ok otherwise.

marcoscaceres:
I don’t have much experience with Axe, but maybe ok by my reading (plus looking at the complete list)? https://www.deque.com/blog/announcing-axe-core-4-0/

We can always disable things after if anyone complains.

PR #3972, opened 2022-01-24
fix(core/issues-notes): show non-GH issues in summary

marcoscaceres:
@jyasskin, are they ok to just show up like this?: 

Screen Shot 2022-01-25 at 9 37 43 am

Happy to add anything else you need. 

If you'd like issues numbers, I could also either untangle the `data-number` from Github or we could add a new `"data-issue-number"` for internal issues.


jyasskin:
Ah, cool, and the `title` attribute would be able to give the issue something like the github issue's title. Yep, this is perfect. I don't need issue numbers if I can give the elements `id`s.

jyasskin (approved):

marcoscaceres:
Ok, released respec@28.2.4 ... should be available in a few mins.

PR #3975, opened 2022-01-27
fix(styles/respec.css): margin space

marcoscaceres:
Desktop: 
Screen Shot 2022-01-27 at 12 26 55 pm

Mobile:
Screen Shot 2022-01-27 at 12 27 37 pm

### DIFF
--- a/src/styles/respec.css.js
+++ b/src/styles/respec.css.js
@@ -222,15 +222,19 @@ aside.example .marker > a.self-link {
   border: none;
   color: inherit;
   position: relative;
-  top: -2.9em;
+  top: -2.75em;
   left: -1.4em;
   margin-bottom: -2em;
+  display: block;
+  font-size: 100%;
 }
 
 :is(h2, h3, h4, h5, h6) + a.self-link::before {
   content: "§";
   opacity: 0.5;
   text-decoration: none;
+  line-height: 1.2em;
+  vertical-align: top;
 }
 
 @media (max-width: 767px) {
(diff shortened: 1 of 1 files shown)


marcoscaceres:
I'm merging this a quick fix, as it's affecting all specs. 

There is still a small bug when the headings are too long, but I think to fix that we need to add wrapper div around the `hN + a` and lay these out with flex or grid. 

Screen Shot 2022-01-27 at 12 29 56 pm

src/styles/respec.css.js
```diff
@@ -222,15 +222,19 @@ aside.example .marker > a.self-link {
   border: none;
   color: inherit;
   position: relative;
-  top: -2.9em;
+  top: -2.75em;
   left: -1.4em;
   margin-bottom: -2em;
+  display: block;
```
sidvishnoi, line 228: This should be an inline-block I think. It makes entire heading (even the whitespace after the heading) a link clickable area (and makes text selection difficult).

src/styles/respec.css.js
```diff
@@ -240,7 +244,7 @@ aside.example .marker > a.self-link {
 
   /* Don't position self-link in headings off-screen */
   :is(h2, h3, h4, h5, h6) + a.self-link {
-    left: 95%;
+    left: 101%;
```
sidvishnoi, line 247: We can maybe follow Bikeshed here and always keep the self-link left of heading? Might simplify things.

PR #3978, opened 2022-01-28
fix(core/webidl): Misleading links on Exposed extended attribute values

### LINKED ISSUE #3976: Misleading links on Exposed extended attribute values
## Important info

- URL to affected spec: any that uses `[Exposed]`
- ReSpec version: 28.2.2

## Description of problem
Respec [adds links to the values of the `Exposed` extended attribute](https://github.com/w3c/respec/blob/1314564705e987e05e4d3d31a376c4bbb370b9e7/src/core/webidl.js#L51) that link to interfaces; but in practice, the names are distinct from interfaces, and only indirectly linked to them: e.g. a `Exposed=Worker` doesn't imply your global object is an instance of `WorkerGlobalScope` - it instead indicates that you're exposed in any object surfaced with the `Global=Worker` extended attribute.

I think it would be cleaner to have no link at all.

marcoscaceres:
closes #3976


dontcallmedom (approved):
LGTM

PR #3979, opened 2022-01-28
BREAKING CHANGE: don't show logo or allow usage of W3C status if not in a W3C WG

### LINKED ISSUE #3965: Prevent W3C logo/template showing up when no group is specified
Sometimes people take spec templates and use W3C templates for things that are not in a working group. 

Prevent use of the W3c templates when it's not a valid working group. So it should use base.css instead.

marcoscaceres:
closes #3965

This prevents usage of W3C Rec Track status and W3C logo on documents that are not attached to a working group. 

The only exception is for "ED", which some groups sometimes use as "unofficial" before publishing. 

It means that everyone now needs to specify a `group` if they want to edit a Rec Track document.

### DIFF
--- a/src/w3c/group.js
+++ b/src/w3c/group.js
@@ -6,37 +6,14 @@
  * `wgURI`, and `wgPatentURI` options.
  */
 
-import {
-  codedJoinAnd,
-  docLink,
-  fetchAndCache,
-  showError,
-  showWarning,
-} from "../core/utils.js";
+import { docLink, fetchAndCache, showError } from "../core/utils.js";
 
 export const name = "w3c/group";
 
 const W3C_GROUPS_API = "https://respec.org/w3c/groups/";
-const LEGACY_OPTIONS = ["wg", "wgURI", "wgId", "wgPatentURI", "wgPatentPolicy"];
-
 export async function run(conf) {
-  const usedLegacyOptions = LEGACY_OPTIONS.filter(opt => conf[opt]);
-
   if (!conf.group) {
-    if (usedLegacyOptions.length) {
-      const outdatedOptionsStr = codedJoinAnd(LEGACY_OPTIONS);
-      const msg = `Configuration options ${outdatedOptionsStr} are deprecated.`;
-      const hint = docLink`Please use the ${"[group]"} configuration option instead.`;
-      showWarning(msg, name, { hint });
-    }
     return;
-  }
-
-  if (usedLegacyOptions.length) {
-    const outdatedOptionsStr = codedJoinAnd(usedLegacyOptions);
-    const msg = docLink`Configuration options ${outdatedOptionsStr} are superseded by ${"[group]"} and will be overridden by ReSpec.`;
-    const hint = docLink`Remove them from the document's ${"[respecConfig]"} to silence this warning.`;
-    showWarning(msg, name, { hint });
   }
 
   const { group } = conf;
--- a/src/w3c/style.js
+++ b/src/w3c/style.js
@@ -107,7 +107,7 @@ export function run(conf) {
       break;
     case "UD":
     case "UNOFFICIAL":
-      styleFile += "UD";
+      styleFile = "W3C-UD";
       break;
     case "FINDING":
     case "DRAFT-FINDING":
--- a/src/w3c/style.js
+++ b/src/w3c/style.js
@@ -5,12 +5,7 @@
 //  - specStatus: the short code for the specification's maturity level or type (required)
 
 import { W3CNotes, recTrackStatus, registryTrackStatus } from "./headers.js";
-import {
-  createResourceHint,
-  docLink,
-  linkCSS,
-  showWarning,
-} from "../core/utils.js";
+import { createResourceHint, linkCSS } from "../core/utils.js";
 import { html } from "../core/import-maps.js";
 import { sub } from "../core/pubsubhub.js";
 export const name = "w3c/style";
--- a/src/w3c/style.js
+++ b/src/w3c/style.js
@@ -88,7 +88,7 @@ function styleMover(linkURL) {
 
 export function run(conf) {
   const canonicalStatus = conf.specStatus?.toUpperCase() ?? "";
-  let styleFile = "W3C-";
+  let styleFile = "";
   const canUseW3CStyle =
     [
       ...recTrackStatus,
--- a/src/w3c/style.js
+++ b/src/w3c/style.js
@@ -112,7 +112,7 @@ export function run(conf) {
       break;
     case "UD":
     case "UNOFFICIAL":
-      styleFile = "W3C-UD";
+      styleFile += "UD";
       break;
     case "FINDING":
     case "DRAFT-FINDING":
--- a/src/w3c/style.js
+++ b/src/w3c/style.js
@@ -4,7 +4,13 @@
 // CONFIGURATION
 //  - specStatus: the short code for the specification's maturity level or type (required)
 
-import { createResourceHint, linkCSS, showWarning } from "../core/utils.js";
+import { W3CNotes, recTrackStatus, registryTrackStatus } from "./headers.js";
+import {
+  createResourceHint,
+  docLink,
+  linkCSS,
+  showWarning,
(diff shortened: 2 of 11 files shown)


src/w3c/style.js
```diff
@@ -81,28 +87,32 @@ function styleMover(linkURL) {
...
     case "CG-FINAL":
     case "BG-DRAFT":
     case "BG-FINAL":
-      styleFile = conf.specStatus.toLowerCase();
+      styleFile = canonicalStatus.toLowerCase();
       break;
+    case "UD":
     case "UNOFFICIAL":
-      styleFile += "UD";
+      styleFile = "W3C-UD";
```
marcoscaceres, line 115:
```suggestion
      styleFile += "UD";
```

src/w3c/style.js
```diff
@@ -81,26 +87,30 @@ function styleMover(linkURL) {
...
 
 export function run(conf) {
-  if (!conf.specStatus) {
-    const msg = "`respecConfig.specStatus` missing. Defaulting to 'base'.";
-    conf.specStatus = "base";
-    showWarning(msg, name);
-  }
-
+  const canonicalStatus = conf.specStatus?.toUpperCase() ?? "";
   let styleFile = "W3C-";
```
sidvishnoi, line 91: Nit:
```suggestion
  let styleFile = "";
```
And use `W3C-` prefix explicitly without `+=`.

sidvishnoi, line 91: Should extract this into a function some time.

tests/spec/w3c/defaults-spec.js
```diff
@@ -74,4 +80,61 @@ describe("W3C — Defaults", () => {
     expect(rsConf.license).toBe("c0");
     expect(rsConf.specStatus).toBe("ED");
   });
+
+  it("doesn't show the W3C logo if no group or an invalid group is specified", async () => {
+    const ops = makeStandardOps({
+      specStatus: "WD",
+    });
```
sidvishnoi, line 87: OCD: 
```suggestion
    const ops = makeStandardOps({ specStatus: "WD" });
```
(and ditto on rest)

tests/spec/w3c/defaults-spec.js
```diff
@@ -74,4 +80,61 @@ describe("W3C — Defaults", () => {
...
+    const docNoGroup = await makeRSDoc(ops);
+    expect(docNoGroup.querySelector("img[alt='W3C']")).toBeNull();
+
+    const opsBadGroup = makeStandardOps({
+      specStatus: "WD",
+      group: "not a real group",
+    });
+    const docBadGroup = await makeRSDoc(opsBadGroup);
+    expect(docBadGroup.querySelector("img[alt='W3C']")).toBeNull();
```
sidvishnoi, line 96: Should move into separate `it()`.

tests/spec/w3c/headers-spec.js
```diff
@@ -105,6 +107,7 @@ describe("W3C — Headers", () => {
       const ops = makeStandardOps();
       const newProps = {
         specStatus: "REC",
+        group: "webapps",
```
sidvishnoi, line 110: How about we default to `group: "webapps"` in makeStandardOps and override when needed?

tests/spec/w3c/style-spec.js
```diff
@@ -172,15 +181,25 @@ describe("W3C - Style", () => {
...
+        const doc = await makeRSDoc(ops);
+        const query = `link[href^='${expectedURL}']`;
+        const elem = doc.querySelector(query);
+        const context = JSON.stringify(conf);
+        expect(elem).withContext(context).toBeTruthy();
+        expect(elem.href).withContext(context).toBe(expectedURL);
+      }
+    },
+    jasmine.DEFAULT_TIMEOUT_INTERVAL * 3
+  );
```
sidvishnoi, line 202: Not sure, but maybe we can switch for-loop order like:
```js
for (const { specStatus, expectedURL, group } of statuses) {
  it(`styles with specStatus: ${specStatus}; group: ${group}`, async() => {
    // makeRsDoc();
    // expect() ...
    // ...
  });
}
```

Should have cleaner assertion failures and not require custom timeouts then. If above works, might make sense to collect them together in a `describe()` block.

sidvishnoi (changes requested):

tests/spec/w3c/headers-spec.js
marcoscaceres: I thought of doing that... but I feel we should be further limiting `makeStandardOps()` because I can never remember what defaults it actually returns, which are probably out of date and can cause us to miss things: In the tests, I prefer we are explicit even if it means more repetition.

tests/spec/w3c/style-spec.js
marcoscaceres: Ironically, that's what it was originally 😬... I'm 50/50 on this one... without wasting too much time one it, maybe I could get it to produce all the documents are once, then `Promise.all()` the documents, and do the checks that way. 

That, in theory should be faster... and `.context()` can tell us which one failed.

src/w3c/style.js
marcoscaceres: Yeah, the += was bugging me too... I went back and forth on that.

src/w3c/group.js
```diff
@@ -6,39 +6,16 @@
...
 export async function run(conf) {
-  const usedLegacyOptions = LEGACY_OPTIONS.filter(opt => conf[opt]);
-
   if (!conf.group) {
-    if (usedLegacyOptions.length) {
-      const outdatedOptionsStr = codedJoinAnd(LEGACY_OPTIONS);
-      const msg = `Configuration options ${outdatedOptionsStr} are deprecated.`;
-      const hint = docLink`Please use the ${"[group]"} configuration option instead.`;
-      showWarning(msg, name, { hint });
-    }
```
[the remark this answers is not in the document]
marcoscaceres: I think we already disallowed that right? "wg" is in the LEGACY_OPTIONS.... or is it used elsewhere? I'd like to kill it with this PR.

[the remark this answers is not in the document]
sidvishnoi: We allow setting it directly (this warning tells users not to), but pretty sure some are still doing it. No harm in keeping warning I think.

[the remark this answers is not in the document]
sidvishnoi: This would actually go when we stop using conf.foo everywhere and instead use something local.

tests/spec/w3c/style-spec.js
sidvishnoi: 😬 didn't see what was before. I thought you just added timeout

src/w3c/group.js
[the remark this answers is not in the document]
marcoscaceres: Nah, I'd like to ban it at the W3C level. There is zero reason for anyone to set it directly and would just lead to confusion.

tests/spec/w3c/style-spec.js
marcoscaceres: Creating the docs all at once ended up making the test twice as long, so reverted to your solution 👍.

marcoscaceres:
About `makeStandardConfig()`, this is what I mean:

```JS
{
        editors: [
          {
            name: "Person Name",
          },
        ],
        specStatus: "ED",
        edDraftURI: "https://foo.com",
        shortName: "Foo",
        previousMaturity: "CR",
        previousPublishDate: "1999-01-01",
        errata: "https://github.com/tabatkins/bikeshed",
        implementationReportURI: "https://example.com/implementationReportURI",
        perEnd: "1999-01-01",
        lint: false,
        definitionMap: {},
}
```

there is way too much stuff there... `definitionMap` shouldn't be there, `implementationReportURI` and `errata` shouldn't either... and so on... a lot of stuff is handled by w3c/default.js.

src/w3c/group.js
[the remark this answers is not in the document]
sidvishnoi: ☠️

marcoscaceres:
going ahead an merging as all the feedback has been addressed.

PR #3982, opened 2022-01-30
fix(w3c/headers): header date was not localized

### LINKED ISSUE #3981: Human Readable Date Inconsistant And Not Localized
## Important info

- ReSpec version: 28.2.8

## Description of problem

grafik

The dates in the top part of the document are 
1. inconsistant
The first date uses the English name of the month (March) while the second date uses the German name (Januar = January).
Both should use the German name, since the language is set to German.

2. not localized
The correct localization in German would be
`1. März 2022`  and `27. Januar 2022`

The leading `0` is not used in texts and the `.` after the day is mandatory.

marcoscaceres:
Closes #3981

cc @init-dcat-ap-de

Now renders as: 
Screen Shot 2022-01-31 at 10 42 09 am


sidvishnoi:
The linked issue states:

> The leading 0 is not used in texts 

I think this would fail W3C requirements? Can we use current `year: "numeric", month: "long", day: "2-digit"` for `en-*` and `dateStyle: "long"` for other languages to better support localization?

src/w3c/headers.js
```diff
@@ -98,12 +98,16 @@ import cgbgHeadersTmpl from "./templates/cgbg-headers.js";
...
 import { pub } from "../core/pubsubhub.js";
 import sotdTmpl from "./templates/sotd.js";
 
 export const name = "w3c/headers";
 
-const W3CDate = new Intl.DateTimeFormat(["en-AU"], {
+// We use an "Australian Date" because it omits the ","
+// after the day of the month, which is required by the W3C.
+const dateLang = lang === "en" || lang.startsWith("en-") ? "en-AU" : lang;
+const W3CDate = new Intl.DateTimeFormat([dateLang], {
```
sidvishnoi, line 110:
```suggestion
const W3CDate = new Intl.DateTimeFormat(dateLang, {
```
Or,
```suggestion
const W3CDate = new Intl.DateTimeFormat([dateLang, "en-AU"], {
```

marcoscaceres:
> I think this would fail W3C requirements? Can we use current year: "numeric", month: "long", day: "2-digit" for en-* and dateStyle: "long" for other languages to better support localization?

Yeah, we could...

marcoscaceres:
I just made it us "numeric" instead, just to keep it as consistent as possible.

sidvishnoi (approved):

init-dcat-ap-de:
Wow, thank you for the fast fix. The first date is now perfect, but the second one is still inconsistant:

grafik

PR #3985, opened 2022-01-31
chore(workflows/codeql-analysis): ignore builds

marcoscaceres:
I think this is right?


sidvishnoi:
> I think this is right?

Umm. No. This is only preventing the action run when builds folder is changed. We want to ignore builds from analysis. So, some param to CodeQL action.

marcoscaceres:
No joy... reading the docs, we might need an actual config file:
https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-code-scanning#specifying-directories-to-scan

.github/codeql/codeql-config.yml
```diff
@@ -0,0 +1,4 @@
+paths:
+  - src
+  - profiles
+  - js
```
sidvishnoi, line 4: Maybe this instead? This will include tools and worker too.
```suggestion
paths-ignore:
  - builds
  - tests
```

marcoscaceres: Good point.

sidvishnoi (approved):

PR #3997, opened 2022-02-01
fix(core/inlines): render missing inline expansion + better warn msg

marcoscaceres:
When [[[expansion]]] was not found, nothing was being rendered in its place... which is not great. 

Better to render something, even if it's not found.


sidvishnoi:
I think we should add a `data-original-text` for all the inlines that we parse. We can then use original text in error messages as well as in this case.

marcoscaceres:
Hmm... I went down that route for a bit, but then felt like a lot of duplication. But I guess we always end up back at the original input. 

I also just wanted to fix this one, I think the others may be ok... this is the most "offensive" one.

src/core/render-biblio.js
```diff
@@ -285,15 +285,17 @@ function decorateInlineReference(refs, aliases) {
...
-    const badrefs = [
-      ...document.querySelectorAll(
-        `a.bibref[href="#bib-${ref.toLowerCase()}"]`
-      ),
-    ].filter(({ textContent: t }) => t.toLowerCase() === ref.toLowerCase());
-    const msg = `Bad reference: [\`${ref}\`] (appears ${badrefs.length} times)`;
-    showError(msg, name);
-    console.warn("Bad references: ", badrefs);
+function warnBadRefs(refs) {
+  refs.forEach(({ ref }) => {
```
sidvishnoi, line 289: Nit: While we're here, let's use `for .. of`

sidvishnoi (approved):

sidvishnoi:
Should use `data-original-text` sometime though. I think it'll be helpful when showing xref failures. Currently, in error messages' `elements`, we render element name. We could instead/also show original text there so it's easier to reason about.

marcoscaceres:
> Should use data-original-text sometime though. I think it'll be helpful when showing xref failures. 

Yeah, I agree... we can then also show aliased "|" things.

I'll rewrite it.

src/core/data-cite.js
```diff
@@ -181,6 +181,9 @@ export async function run() {
       linkElem(elem, linkProps, citeDetails);
     } else {
       const msg = `Couldn't find a match for "${originalKey}"`;
+      if (elem.dataset?.matchedText) {
+        elem.textContent = elem.dataset?.matchedText;
```
marcoscaceres, line 185:
```suggestion
        elem.textContent = elem.dataset.matchedText;
```

src/core/data-cite.js
```diff
@@ -181,6 +181,9 @@ export async function run() {
...
+      }
```
marcoscaceres, line 186:
```suggestion
      if (elem.dataset?.matchedText) {
        elem.textContent = elem.dataset.matchedText;
      }
```

PR #4002, opened 2022-02-02
chore: support development with pnpm

sidvishnoi:
pnpm uses Node's resolve algorithm, so some packages failed to resolve (npm flattens dependency tree, so all resolved fine there).

This PR is allowing development with pnpm for those who want to use it. npm will continue to work fine. If someone wants to use pnpm locally, they can:
```bash
rm -rf node_modules
pnpm import # imports package-lock.json and creates a pnpm-lock.yaml
pnpm install
# and everything works fine as is
```

When updating a dependency in `package-lock.json`, we can:
```bash
npm i --package-lock-only # this will not touch node_modules
pnpm import && pnpm install
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,6 +57,7 @@
     "rollup-plugin-minify-html-literals": "^1.2.6",
     "rollup-plugin-terser": "^7.0.2",
     "serve": "^13.0.2",
+    "serve-handler": "^6.1.3",
     "sniffy-mimetype": "^1.1.1",
     "typescript": "^4.5.5",
     "vnu-jar": "^21.10.12",
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,6 +5,7 @@
   "requires": true,
   "packages": {
     "": {
+      "name": "respec",
       "version": "29.0.4",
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -57,6 +57,7 @@
     "rollup-plugin-minify-html-literals": "^1.2.6",
     "rollup-plugin-terser": "^7.0.2",
     "serve": "^13.0.2",
+    "serve-handler": "^6.1.3",
```
sidvishnoi, line 60: We would've found this failing when we removed the last package that has `serve-handler` as a dependency.

marcoscaceres:
neat! but didn't pnpm ship with Node 17? We can't use 17 yet (at least, I don't have pnpm installed?).

sidvishnoi:
With Node 16, if you run `corepack enable`, it'll enable pnpm (which is installed already but hidden).

package.json
marcoscaceres: I don't even remember what it does...

sidvishnoi: It's used in dev-server

PR #4003, opened 2022-02-04
Adding German Translations for rfc2119Keywords

init-dcat-ap-de:
In German, the words KANN, SOLL and MUSS are different in plural, so I had to add them to the list.
I hope the umlaute are not a problem.


marcoscaceres (approved):

marcoscaceres:
> I hope the umlaute are not a problem.

They shouldn't be... but let us know :)

PR #4007, opened 2022-02-07
fix(linter-rules/no-unused-dfns): improve hint

marcoscaceres:


src/core/linter-rules/no-unused-dfns.js
```diff
@@ -21,7 +21,15 @@ const localizationStrings = {
       return `Found definition for "${text}", but nothing links to it. This is usually a spec bug!`;
     },
     get hint() {
-      return docLink`Either remove the definition, or add a \`"lint-ignore"\` CSS class. If you meant to export it, add the ${"[export|#data-export]"} CSS.`;
+      return docLink`
+      You can do one of the following:
+
+        * Add a \`class="lint-ignore"\` attribute the definition.
+        * Either remove the definition or change \`<dfn>\` to another type of HTML element.
+        * If you meant to ${"[export|#data-export]"} the definition, add \`class="export"\` to definition.
```
marcoscaceres, line 29:
```suggestion
        * If you meant to ${"[export|#data-export]"} the definition, add \`class="export"\` to the definition.
```

sidvishnoi:
When I click the ReSpec warning pill, I get this in console (and error list doesn't open):
image

sidvishnoi:
The error message is 👍🏽 though. We should see where else we can have such more helpful messages.

sidvishnoi:
Also, in terminal, it becomes a `<pre><code>`:
image

sidvishnoi:
`docLink` should probably include a dedent logic.

marcoscaceres:
Fixed the `<pre><code>` issue by re-indenting the hint.

sidvishnoi:
Does it fix cli usage too?

marcoscaceres:
Actually, no... still there. I can live with that tho.

sidvishnoi:
If we add dedent (npm package) into docLink, I think we should be able to fix both. I can take over if you want.

marcoscaceres:
We have `reindent()`, which should do the job... the problem is that the hint end ups as HTML, instead of staying as markdown.

So, I think the fix is still just to reindent() the hint, before it becomes HTML (and make sure the error's hint remains markdown instead of HTML).

But happy for you to take it over. I won't be able to return to it until next week.

marcoscaceres:
I'd still like to get the error message out today tho, as a lot of specs are seeing this message and people are not understanding it.

sidvishnoi:
I'll fix the ci msg within 2-3 hours and release myself.

sidvishnoi:
Web UI:
image

And CLI:
image

([CLI before for comparison](https://user-images.githubusercontent.com/8426945/152963324-faf385e5-6577-49f1-89fe-b223d147edb9.png))

sidvishnoi (approved):

PR #4009, opened 2022-02-07
BREAKING CHANGE(utils): remove humanDate + isoDate, use W3CDate 

### LINKED ISSUE #3981: Human Readable Date Inconsistant And Not Localized

marcoscaceres:
Closes #3981

Fixes localization of modificationDate + moved W3CDate to util. 

(yeah, I should have done this in 2 PRs... but no time :( )


marcoscaceres:
Moved part of this into https://github.com/w3c/respec/pull/4010

marcoscaceres:
Ok, marking as breaking change... removed `utils.isoDate()` (not used) and `utils.humanDate()` (not localized properly). 

Now use `W3CDate` everywhere.

src/core/templates/show-people.js
```diff
@@ -83,7 +81,7 @@ export default function showPeople(conf, propName) {
  * @param {Person} person
  */
 function personToHTML(person) {
-  const l10n = localizationStrings[lang];
+  const l10n = getIntlData(localizationStrings);
```
sidvishnoi, line 84: Should move this out of this function.

src/core/utils.js
```diff
@@ -29,6 +29,18 @@ export const ISODate = new Intl.DateTimeFormat(["en-ca-iso8601"], {
   day: "2-digit",
 });
 
+// We use an "Australian Date" because it omits the ","
+// after the day of the month, which is required by the W3C.
+const dateLang =
+  docLang === "en" || docLang.startsWith("en-") ? "en-AU" : docLang;
+
+export const W3CDate = new Intl.DateTimeFormat(dateLang, {
```
sidvishnoi, line 37: As we're calling .format everywhere, maybe we should export a function formatDateW3C that accepts a string|Date instead?

tests/spec/core/utils-spec.js
```diff
@@ -239,32 +239,6 @@ describe("Core - Utils", () => {
     });
   });
 
-  describe("humanDate", () => {
```
sidvishnoi, line 242: Should move W3CDate tests here now that it's part of core/utils?

src/core/utils.js
marcoscaceres: Yeah, `.format()` everywhere, but seems like having a wrapper is not necessary. I'd rather just have a single object we work from (checks and conversion are where needed). 

The places where it's doing the wrong thing probably need validateAndRecover()

tests/spec/core/utils-spec.js
marcoscaceres: Probably, yeah.

sidvishnoi: We don't have specific tests for W3CDate, so let's ignore this.

PR #4018, opened 2022-02-09
fix(styles/respec.css): don't offset some headers

marcoscaceres:


src/styles/respec.css.js
```diff
@@ -223,7 +223,7 @@ aside.example .marker > a.self-link {
   align-items: center;
 }
 
-:is(h2, h3, h4, h5, h6):not(#toc h2, #abstract h2, #sotd h2) {
+:is(h2, h3, h4, h5, h6):not(#toc > h2, #abstract > h2, #sotd > h2, h2#subtitle) {
```
sidvishnoi, line 226: Nit: I think we should remove this `not` and override using cascade the old way.

sidvishnoi (approved):

marcoscaceres:
Will eventually rewrite all this... current solution is borked...

PR #4027, opened 2022-02-14
fix(core/ui): don't reparse markdown

marcoscaceres:
The double-parsing can cause the markdown parser to crash.


src/core/ui.js
```diff
@@ -276,15 +276,14 @@ function rsErrorToHTML(err) {
...
       }</strong> times at:</p>
-    ${err.elements.map(generateMarkdownLink)}
-    `
+    ${markdownToHtml(err.elements.map(generateMarkdownLink).join(""))}`
     : "";
   const details = err.details
     ? `\n\n<details>\n${err.details}\n</details>\n`
     : "";
   const msg = markdownToHtml(`**${err.message}**`, { inline: true });
   const text = `${msg}${hint}${elements}${details}${plugin}`;
```
sidvishnoi, line 285: Let's call this `html` then?

src/core/ui.js
```diff
@@ -276,15 +276,14 @@ function rsErrorToHTML(err) {
     ? `<p class="respec-occurrences">Occurred <strong>${
         err.elements.length
       }</strong> times at:</p>
-    ${err.elements.map(generateMarkdownLink)}
-    `
+    ${markdownToHtml(err.elements.map(generateMarkdownLink).join(""))}`
```
sidvishnoi, line 279: Nit: let's do `join('\n')` instead?

sidvishnoi:
Any example when it crashed?

sidvishnoi (approved):

src/core/ui.js
marcoscaceres: called it result instead, as we use "html" for actual parsed html.

marcoscaceres:
> Any example when it crashed?

Oh, argh... it was a weird one... I should have kept it.

PR #4036, opened 2022-02-16
fix(w3c/cgbg-headers): show Feedback: links

marcoscaceres:
Feedback links are missing from WICG docs.


sidvishnoi (approved):

sidvishnoi:
We should see if we can have a larger common template shared between cgbg and regular W3C header. They get out of sync as we often forget this file exists.

marcoscaceres:
Yeah. Thinking exactly the same... just no time to do it.

PR #4046, opened 2022-02-20
fix(w3c/headers): always add History or FPWD

### LINKED ISSUE #4043: generation of History URI fails for FPWD
## Important info

- URL to affected spec:
https://w3c.github.io/vsso/spec/vsso-core-re.html?specStatus=FPWD&publishDate="2022-02-22"&shortName="vsso-core"
- ReSpec version: Click the ReSpec pill in top-right corner (e.g., 25.12.0).
31.0.3

## Description of problem

The generated History link does not contain the expected URI, for this spec it should be
https://www.w3.org/standards/history/vsso-core
Console shows that there's a HTTP request
HEAD https://www.w3.org/standards/history/vsso-core
that gives a 404 (which is obviously expected for a FPWD, it exists only after publication)

Something special should happen after https://github.com/w3c/respec/blob/main/src/w3c/headers.js#L747
for the FPWD case?

marcoscaceres:
closes #4043

### DIFF
--- a/src/w3c/headers.js
+++ b/src/w3c/headers.js
@@ -735,7 +735,8 @@ async function deriveHistoryURI(conf) {
   );
 
   // If it's on the Rec Track or it's TR worthy, then allow history override.
-  if (conf.historyURI && canShowHistory) {
+  // Also make a an exception for FPWD.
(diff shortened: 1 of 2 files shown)


marcoscaceres:
Taking as hot fix.

PR #4052, opened 2022-02-23
fix(w3c/cgbg-headers): handle subtitle correctly

marcoscaceres:
Oops, the subtitle was not being handled correctly... cgbg template strikes again :D


sidvishnoi (approved):
LGTM

PR #4080, opened 2022-03-14
feat(core/caniuse): group by device type

marcoscaceres:
Now looks like: 

Screen Shot 2022-03-14 at 4 41 58 pm


sideshowbarker (approved):

src/core/caniuse.js
```diff
@@ -133,18 +141,43 @@ function validateBrowsers({ caniuse }) {
...
   const results = json.result;
-  const out = results.map(({ browser: browserId, version, caniuse }) => {
-    const { name } = BROWSERS.get(browserId);
+  const groups = new Map([
+    ["desktop", []],
+    ["mobile", []],
+  ]);
+  const toBrowserCell = browserCellRenderer(feature);
+  results.reduce(toBrowserCell, groups);
+  const out = [...groups.keys()]
```
sidvishnoi, line 152: Can use `.entries()` instead (or shorthand `...groups.filter().map()`)

src/core/caniuse.js
```diff
@@ -133,18 +141,43 @@ function validateBrowsers({ caniuse }) {
...
+    .filter(key => groups.get(key).length)
+    .map(
+      key =>
+        html`<div class="${`caniuse-group`}">
```
sidvishnoi, line 156:
```suggestion
        html`<div class="caniuse-group">
```

src/core/caniuse.js
```diff
@@ -133,18 +141,43 @@ function validateBrowsers({ caniuse }) {
...
+          <div class="caniuse-browsers">${groups.get(key)}</div>
+          <div class="caniuse-type"><span>${key}</div>
```
sidvishnoi, line 158: Not sure how this works for screen reader users. Should this be like a section heading? (We can move position with CSS)

sidvishnoi:
Is this the intended design on mobile?


I think the more info link should come at bottom at least.

src/core/caniuse.js
```diff
@@ -14,21 +14,29 @@ export const name = "core/caniuse";
...
 
 export const BROWSERS = new Map([
-  ["and_chr", { name: "Android Chrome", path: "chrome" }],
-  ["and_ff", { name: "Android Firefox", path: "firefox" }],
-  ["and_uc", { name: "Android UC", path: "uc" }],
-  ["chrome", { name: "Chrome" }],
-  ["edge", { name: "Edge" }],
-  ["firefox", { name: "Firefox" }],
-  ["ios_saf", { name: "iOS Safari", path: "safari-ios" }],
+  ["and_chr", { name: "Android Chrome", path: "chrome", type: "mobile" }],
```
sidvishnoi, line 17: Maybe create two maps? 
(Can merge when testing input for valid browsers)

src/styles/caniuse.css.js
```diff
@@ -4,11 +4,44 @@ const css = String.raw;
 // Prettier ignore only to keep code indented from level 0.
 // prettier-ignore
 export default css`
+
 .caniuse-stats {
   display: flex;
-  flex-wrap: wrap;
+  column-gap: 1cm;
```
sidvishnoi, line 10: Maybe use em/rem?

marcoscaceres: Sure, I'll find an equivalent. 1cm looks good to base off.

src/core/caniuse.js
marcoscaceres: > Can use .entries() instead 

Also returns an iterator :( 

> or shorthand ...groups.filter().map())

No sure what you mean? move it to another function?

marcoscaceres:
> Is this the intended design on mobile?

Yeah, at least for. We plan to iterate on the design as we get feedback. 

It might make sense to reverse back to column and rows in contained space.

src/core/caniuse.js
marcoscaceres: I think it's maybe "labelled-by"?

src/core/caniuse.js
sidvishnoi: `out = [...groups].filter(([key, values]) => {})`

src/core/caniuse.js
marcoscaceres: oh, you mean DESKTOP_BROWSERS and MOBILE_BROWSERS. I'll take a look.

sidvishnoi (approved):

PR #4107, opened 2022-03-28
fix(w3c/headers): history page doesn't exist yet for First DNOTE/NOTE/DRY

deniak:
New specifications that are being published on Note/Registry track don't have an history page until the first version is being published. That PR will display the history link for these docs.


deniak:
I believe that PR should do the right thing but I can't figure out why the test on the history link is failing. @sidvishnoi, @marcoscaceres, do you have an idea?

src/w3c/headers.js
```diff
@@ -731,7 +731,7 @@ async function deriveHistoryURI(conf) {
 
   const canShowHistory = conf.isEd || trStatus.includes(conf.specStatus);
 
-  if (conf.historyURI && canShowHistory) {
+  if (conf.historyURI && !canShowHistory) {
```
sidvishnoi, line 734: Good catch! That's what is breaking the broken tests.
The test needs to set `respecConfig.group`, otherwise ReSpec defaults to `specStatus: "base"`, which isn't part of trStatus.

src/w3c/headers.js
```diff
@@ -745,8 +745,11 @@ async function deriveHistoryURI(conf) {
   );
 
   // If it's on the Rec Track or it's TR worthy, then allow history override.
-  // Also make a an exception for FPWD.
-  if ((conf.historyURI && canShowHistory) || conf.specStatus === "FPWD") {
+  // Also make a an exception for FPWD, DNOTE, NOTE and DRY.
+  if (
+    (conf.historyURI && canShowHistory) ||
+    ["FPWD", "DNOTE", "NOTE", "DRY"].includes(conf.specStatus)
+  ) {
```
sidvishnoi, line 752: These exceptions shouldn't be needed as `trStatus` includes them.

src/w3c/headers.js
sidvishnoi: Ignore rest of the tests. They're broken due to xref data change.

deniak: Thanks! Adding the `webapps` group to the options solved the issue.

src/w3c/headers.js
deniak: these are actually special use cases. A new specification can start as a FPWD, DNOTE, NOTE or DRY. In this case, the history link doesn't exist yet so adding the exception will display it without checking if it exists or not.

sidvishnoi: I'm not following. (Maybe I'm just sleepy)
1. As per line 742, `historyURI` can be a `respecConfig.historyURI` or one inferred via `shortName`.
2. `canShowHistory` is true for `DNOTE, NOTE, DRY` (they're part of `trStatus`).
   - `["FPWD", "DNOTE", "NOTE", "DRY"].includes(conf.specStatus)` is a subset of `canShowHistory`
4. When exceptions path is followed (i.e. if `conf.historyURI` is falsy), we end up setting `conf.historyURI = "https://www.w3.org/standards/history/{shortName}"`. Is that intended?

deniak: Let's say you want to publish a FPWD and you expect respec to infer the history link so you don't add a `respecConfig.historyURI`.
In this case, without the exception, respec will do a `HEAD` to find the history link, which will return a 404 since it doesn't exist yet, and `conf.historyURI` will be empty.

sidvishnoi: In that case, we'll end up having `"https://www.w3.org/standards/history/{shortName}"` as `History:` header link, right? Is that intended?

Unless author manually disables History link with `respecConfig.historyURI = null`, we'll have a non-existing history link for FPWD et. al?

sidvishnoi (approved):
Either way, this works as intended so LGTM.
We've to simplify this headers file someday.

sidvishnoi:
Shipped in v32.1.1

PR #4112, opened 2022-04-01
hotfix(core/inline-idl-parser): use WebIDL in lowercase

dontcallmedom:
Otherwise, respec complains about unknown 'WebIDL' reference

See https://w3c.github.io/mediacapture-region/ where this manifests itself (cf https://github.com/w3c/mediacapture-region/pull/39#pullrequestreview-928481678)


sidvishnoi (approved):
This would be a hotfix. Would need to spend some time on making aliasing work consistently. This is probably due to some conflict in order of operations (biblio IDB vs network).

PR #4122, opened 2022-04-13
feat: add Webrecorder profile

edsu:
I don't know if it is common practice for respec users to push profiles back upstream. But if it is here is a profile that the [Webrecorder](https://webrecorder.net) project is [using](https://webrecorder.github.io/wacz-spec/1.1.1/).


edsu:
Feedback on this PR would be appreciated, even if the answer is to close it because sharing profiles is not something that is desirable.

src/webrecorder/abstract.js
```diff
@@ -0,0 +1,76 @@
+// @ts-check
+// Module w3c/abstract
```
marcoscaceres, line 2:
```suggestion
// Module webrecorder/abstract
```

src/webrecorder/conformance.js
```diff
@@ -0,0 +1,87 @@
+// @ts-check
+// Module w3c/conformance
```
marcoscaceres, line 2:
```suggestion
// Module webrecorder/conformance
```

marcoscaceres:
Hi @edsu, including it should be fine. Would you mind removing the changes to the `build/` directory? Those should only get generated during deployment.

profiles/webrecorder.js
```diff
@@ -0,0 +1,75 @@
...
+  import("../src/core/dfn.js"),
+  import("../src/core/pluralize.js"),
+  import("../src/core/examples.js"),
+  import("../src/core/issues-notes.js"),
+  import("../src/core/best-practices.js"),
+  import("../src/core/figures.js"),
+  import("../src/core/webidl.js"),
+  import("../src/core/biblio.js"),
+  import("../src/core/link-to-dfn.js"),
+  import("../src/core/xref.js"),
```
sidvishnoi, line 30: You might not need `xref`, `ui/search-xref` modules, unless you reference definitions and terms from W3C specs. This should reduce bundle size and avoid unintended dependency on W3C related stuff.

Also potentially:
- `mdn-annotation`
- `webidl`
- `webidl-index`
- `linter-rules/wpt-tests-exist`.

sidvishnoi:
Also, please add `webrecorder` to the release script so it gets added to `./builds/respec-webrecorder.js` during release:
https://github.com/w3c/respec/blob/740123075705afc1eae4093eac06c9b33255a2d1/tools/release.js#L363

And also to: https://github.com/w3c/respec/blob/740123075705afc1eae4093eac06c9b33255a2d1/tools/netlify.js#L13

marcoscaceres:
Closing due to inactivity. Please feel free to send again though. We sent some suggestions above.

PR #4132, opened 2022-04-20
fix(w3c/sotd): use `crEnd` instead of incorrect `CREnd`

deniak:
CR requires an end date which can be set with the `crEnd` option. However, that option is not taken into account.
That PR fixes the SOTD template to use the right option.


sidvishnoi (approved):

sidvishnoi:
Released in v32.1.3

PR #4134, opened 2022-04-21
fix(w3c): allow trStatus specStatus for TAG

darobin:
The TAG can now publish in TR, and intends to do so relatively regularly. This PR removes the restriction. This is blocking publication of an FPWD.

### DIFF
--- a/src/w3c/defaults.js
+++ b/src/w3c/defaults.js
@@ -103,7 +103,7 @@ function processLogos(conf) {
 }
 
 function validateStatusForGroup(conf) {
-  const { specStatus, groupType } = conf;
+  const { specStatus, groupType, group } = conf;
 
   if (!specStatus) {
     const msg = docLink`The ${"[specStatus]"} configuration option is required.`;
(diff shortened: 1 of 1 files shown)


src/w3c/defaults.js
```diff
@@ -152,13 +152,6 @@ function validateStatusForGroup(conf) {
       break;
     }
     case "other":
-      if (group === "tag" && !tagStatus.includes(specStatus)) {
```
sidvishnoi, line 155: I wonder if we should update `tagStatus` to include `FPWD` (and others) instead.
This here is a safety net mainly to give a better error message in case things go wrong. I haven't touched W3C specific details since some time, and Marcos is away, so I'm not really sure it's the best way.

sidvishnoi: But YOLO!

sidvishnoi (approved):

src/w3c/defaults.js
sidvishnoi: Or maybe use `[...trStatus, ...tagStatus].includes(specStatus)`

darobin: Hey @sidvishnoi I see you just approved, but I actually made it `[...trStatus, ...tagStatus].includes(specStatus)` instead, as that's a good idea IMHO.

sidvishnoi (approved):

sidvishnoi:
Released in v32.1.4

darobin:
Thanks, you rock.

Issue #1620, opened 2018-04-10
Include support to "list of tables"


andrea-perego:
I wonder whether there's any plan to support "list of tables" - similarly to what done for ["table of figures"](https://github.com/w3c/respec/wiki/figure).

PR #4200, opened 2022-07-02
feat(core/tables): auto-number tables, list tables and allow shortcode refs

### LINKED ISSUE #4004: Auto-numbering tables like examples
I am always frustrated when I want to link to or mention a table, like in a GitHub issue, saying that "Table 3 has an issue at ...". Tables are not automatically numbered and are not linkable like examples are. Tables can have an `id` attribute and it can be used for linking but this needs one to look at the HTML source and also does not solve the fact that the tables are not numbered.

I would like to use a CSS class for the table that auto generates the table number and by using the `title` attribute, hyperlinks the table URL to the "[Table 3](http://exampe.com/myspec#my-amazing-table): My Amazing Table" . A concrete example would be https://respec.org/docs/#example-a-basic-respec-document but for tables. A hacky solution is shown in Additional Context, item 2.

**Describe alternatives you've considered**
One can manually add a caption for the tables, manually insert a number (and maintain that number) and add a link. This can be scripted rather easily. In the case of the [W3C WoT Thing Description](https://w3c.github.io/wot-thing-description/) which has already some custom rendering scripts in place, it would not be a lot of work and would be invisible for the contributors of the spec (since they have to use those rendering scripts anyway).

My organization is a member of W3C and W3C seems to be funding the development. For a separate contribution, I would need to trigger some people with no guarantees of a favorable answer. A one-time contribution would be most likely due to how my organization does payments _most of the time_. I do not plan to make a personal financial contribution but I can support the development with my skills. My organization can also support someone in my team to make a PR with this feature but no guarantees on the maintenance of this feature.

**Additional context**

1. Currently a table looks like this:
image
for the following HTML
```html
    <table class="def" id="tab-namespaces">
    <thead>
        <tr>
            <th>Vocabulary</th>
            <th>Namespace IRI</th>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>Core</td>
            <td><code>https://www.w3.org/2019/wot/td#</code></td>
        </tr>
        <tr>
            <td>Data Schema</td>
            <td><code> 	https://www.w3.org/2019/wot/json-schema#</code></td>
        </tr>
        <tr>
            <td>Security</td>
            <td><code>https://www.w3.org/2019/wot/security#</code></td>
        </tr>
        <tr>
            <td>Hypermedia Controls</td>
            <td><code>https://www.w3.org/2019/wot/hypermedia#</code></td>
        </tr>
    </tbody>
    </table>
```

2. Ideally something like this would be nice

```html
<table class="def" id="tab-namespaces" title="Namespaces used in a Thing Description">
<!-- line below is ideally auto-generated -->
      <caption style="text-align:left;margin-bottom: 0;" class="def"><a href="#tab-namespaces">Table 1</a>: Namespaces used in a Thing Description</caption>
    <thead>
        <tr>
            <th>Vocabulary</th>
            <th>Namespace IRI</th>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>Core</td>
            <td><code>https://www.w3.org/2019/wot/td#</code></td>
        </tr>
        <tr>
            <td>Data Schema</td>
            <td><code> 	https://www.w3.org/2019/wot/json-schema#</code></td>
        </tr>
        <tr>
            <td>Security</td>
            <td><code>https://www.w3.org/2019/wot/security#</code></td>
        </tr>
        <tr>
            <td>Hypermedia Controls</td>
            <td><code>https://www.w3.org/2019/wot/hypermedia#</code></td>
        </tr>
    </tbody>
    </table>
```

image

3. A related issue: https://github.com/w3c/wot-thing-description/issues/1256

ProgramMax:
This commit adds auto-numbering to tables to mirror the behavior of
figures.

There is a notable difference. A table needs to include class="numbered"
to have the autonumbering applied. Without this class, the number will
be skipped. This leaves room for tables that shouldn't be referenced.

Closes #4004
Closes https://github.com/w3c/respec/issues/1620


ProgramMax:
Pinging @marcoscaceres since you commented on #4004.

If a table isn't marked class="numbered", that number is skipped.
For example "Table 1, Example Table, Table 3".
I think maybe the behavior should be different though? Like "Table 1, Example Table, Table 2". And the non-numbered tables shouldn't count?

ProgramMax:
Oh also, I updated example_aom.html to be a visible proof-of-concept.
But I think maybe this file is supposed to be a guideline for AOM so I shouldn't update it?

ProgramMax:
Also, a lot of this code is very similar to the code that handles figures. Maybe (in a follow-up) a lot of common code can be refactored & shared.

src/core/tables.js
```diff
@@ -0,0 +1,163 @@
...
+    list_of_tables: "Lista de Figuras",
+  },
+  zh: {
+    table: "图 ",
+    list_of_tables: "规范中包含的图",
+  },
+  de: {
+    table: "Abbildung",
+    list_of_tables: "Abbildungsverzeichnis",
+  },
```
ProgramMax, line 45: Is there a way to get translations for these?
I copied from the "figure" code.

ProgramMax:
Is there a way to bump the bot? The Karma Chrome Headless test seems to have failed a flaky test which is unrelated to this change.

src/core/tables.js
sidvishnoi: For now, we can just keep the English one.

examples/example_aom.html
```diff
@@ -83,9 +83,72 @@ <h2>
         Link to [[[#figure]]] ... the three square-brackets "expand" a local reference.   
       </p>
     </section>
+    <section id="example-tables">
```
sidvishnoi, line 86: Let's not modify this file.

src/core/tables.js
```diff
@@ -0,0 +1,163 @@
...
+  const tot = [];
+  document.querySelectorAll("table").forEach((table, i) => {
+    const caption = table.querySelector("caption");
+
+    if (caption) {
+      decorateTable(table, caption, i);
+      tot.push(getTableOfTablesListItem(table.id, caption));
+    } else {
+      const msg = "Found a `<table>` without a `<caption>`.";
+      showWarning(msg, name, { elements: [table] });
```
sidvishnoi, line 80: This can be too much of a breaking change. It might be better to add it as a linter rule instead.

src/core/tables.js
```diff
@@ -0,0 +1,163 @@
...
+    }
+  });
+  return tot;
+}
+
+/**
+ * @param {HTMLElement} table
+ * @param {HTMLElement} caption
```
sidvishnoi, line 88: Nit:
```suggestion
 * @param {HTMLTableElement} table
 * @param {HTMLTableCaptionElement} caption
```

src/core/tables.js
```diff
@@ -0,0 +1,163 @@
...
+ * @param {number} i
+ */
+function decorateTable(table, caption, i) {
+  const title = caption.textContent;
+  addId(table, "table", title);
+  // set proper caption title
+  wrapInner(caption, html`<span class="table-title"></span>`);
+  if (table.classList.contains("numbered")) {
```
sidvishnoi, line 96: I'm not sure if we should make it opt-in as above or support opt-out like we do in rest of modules (like `.notoc, .no-export, .nohighlight`). Maybe I'm just not comfortable with `numbered`.

src/core/tables.js
```diff
@@ -0,0 +1,163 @@
...
+
+/**
+ * @param {string} tableId
+ * @param {HTMLElement} caption
+ * @return {HTMLElement}
+ */
+function getTableOfTablesListItem(tableId, caption) {
+  const totCaption = caption.cloneNode(true);
+  totCaption.querySelectorAll("a").forEach(anchor => {
+    renameElement(anchor, "span").removeAttribute("href");
```
sidvishnoi, line 109: Should remove `id` too to avoid duplicate IDs.

src/core/tables.js
```diff
@@ -0,0 +1,163 @@
...
+  }
+}
+
+/**
+ * process all tables
+ */
+function collectTables() {
+  /** @type {HTMLElement[]} */
+  const tot = [];
+  document.querySelectorAll("table").forEach((table, i) => {
```
sidvishnoi, line 72: Let's run over `"table.numbered"` only and avoid internal checks later on.

src/core/tables.js
```diff
@@ -0,0 +1,163 @@
...
+ * @param {string} tableId
+ * @param {HTMLElement} caption
+ * @return {HTMLElement}
+ */
+function getTableOfTablesListItem(tableId, caption) {
+  const totCaption = caption.cloneNode(true);
+  totCaption.querySelectorAll("a").forEach(anchor => {
+    renameElement(anchor, "span").removeAttribute("href");
+  });
+  return html`<li class="totline">
```
sidvishnoi, line 111: Let's remove this `totline` class as it's not used for styling or any other thing.

```suggestion
  return html`<li>
```

src/core/tables.js
sidvishnoi, line 45: We can add the translations later as we learn more languages 😄

src/core/tables.js
```diff
@@ -0,0 +1,163 @@
+// @ts-check
+// Module core/tables
+// Handles tables in the document.
+// Generates a Table of Tables wherever there is a #tot element.
```
sidvishnoi, line 4: Might be better to be more explicit `#table-of-tables`.
(`#tof` was probably a mistake; `#toc` is okay)

src/core/anchor-expander.js
```diff
@@ -93,6 +97,34 @@ function processFigure(matchingElement, id, a) {
   }
 }
 
+function processTable(matchingElement, id, a) {
+  const caption = matchingElement.querySelector("caption");
```
sidvishnoi, line 101: Nit:
```suggestion
function processTable(matchingTable, id, a) {
  const caption = matchingTable.querySelector("caption");
```

src/core/anchor-expander.js
```diff
@@ -93,6 +97,34 @@ function processFigure(matchingElement, id, a) {
...
+  if (matchingElement.classList.contains("numbered")) {
+    children = children.filter(
+      // @ts-ignore
+      node => !node.classList || !node.classList.contains("table-title")
+    );
+    // drop an empty space at the end.
+    children.pop();
+  }
+  a.append(...children);
+  a.classList.add("tab-ref");
```
sidvishnoi, line 121:
```suggestion
  a.classList.add("table-ref");
```

src/core/tables.js
sidvishnoi: Ah, this PR also allows referencing unnumbered tables. Never mind above check then.

sidvishnoi:
Thanks for the PR @ProgramMax! I think we should be a able to merge this with a few changes :)

src/core/tables.js
sidvishnoi: Actually, it's a list 😄 
So, `#list-of-tables` or `#tables-list`.

src/core/tables.js
ProgramMax: The conversation in #4004 seemed to prefer an opt-in since there might be tables that are used for things other than tabular data.

src/core/tables.js
ProgramMax: I misunderstood why we would want tables that aren't numbered.

Tabular data (like what will show up in the list of tables) will all be numbered.
Only tables used for non-tabular data are unnumbered. But they shouldn't be linkable or included in the list of tables.
So I agree that enumerating "table.numbered" is the better option.

src/core/tables.js
ProgramMax: Ahh got ya.
I misunderstood that these were perhaps languages where we had a translator available.
But it sounds more like when a translation is requested we find a translator.

ProgramMax: I've removed the anchor-to-table translation test. Let me know if there is a good way to preserve it. I suppose I could keep it in English?

src/core/tables.js
ProgramMax: I could use help here. The TypeScript compiler noted that the parameters were not HTMLTableElement.
I could use the TypeScript 'as' keyword to cast from HTMLElement to HTMLTableElement. But these are JavaScript files, not TypeScript.

This code originally comes from the figure code. Apparently figure and figcaption tags are both HTMLElement. There is no HTMLFigureElement for example. So that code didn't warrant an upcast.

Thoughts?

sidvishnoi: I took the liberty to update above in https://github.com/w3c/respec/pull/4200/commits/9175d7528c2e8c83c15e9bc35579451f5147f616

sidvishnoi (approved):
This looks good. Thanks @ProgramMax!

Though, I would prefer if we can split the linter rule into a separate PR (just to keep commits concise in case we need to revert etc.). I'll leave it up to you.

sidvishnoi:
I've added the docs for [`no-captionless-tables`](https://github.com/w3c/respec/wiki/no-captionless-tables).

@ProgramMax Can you add the docs for `.numbered`? I've added the [template](https://github.com/w3c/respec/wiki/numbered).

src/core/tables.js
```diff
@@ -0,0 +1,129 @@
...
+  /** @type {NodeListOf<HTMLTableElement>} */
+  const tables = document.querySelectorAll("table.numbered");
+  tables.forEach((table, i) => {
+    const caption = table.querySelector("caption");
+    // there is a separate linter rule to catch numbered tables without captions
+    if (caption) {
+      decorateTable(table, caption, i);
+      listOfTables.push(getListOfTablesListItem(table.id, caption));
+    }
+  });
```
sidvishnoi, line 49: Question: If some table doesn't have a caption, we'll jump from Table 1 to Table 3? Should it be table 2 instead?

sidvishnoi:
```suggestion
  tables
  // there is a separate linter rule to catch numbered tables without captions
    .filter((table) => !!table.querySelector("caption"))
    .forEach((table, i) => {
      const caption = table.querySelector("caption");
      decorateTable(table, caption, i);
      listOfTables.push(getListOfTablesListItem(table.id, caption));
  });
```

sidvishnoi (changes requested):
Question: If some table doesn't have a caption, we'll jump from Table 1 to Table 3? Should it be table 2 instead?

src/core/tables.js
```diff
@@ -0,0 +1,129 @@
...
+    listOfTablesElement.classList.add("introductory");
+  } else if (previousSections.some(sec => sec.classList.contains("appendix"))) {
+    listOfTablesElement.classList.add("appendix");
+  }
+}
+
+/**
+ * @param {Element} element
+ */
+function getPreviousSections(element) {
```
marcoscaceres, line 109: I wonder if this should move to utils, as figures.js has this code too.

src/core/tables.js
```diff
@@ -0,0 +1,129 @@
...
+      sections.push(previous);
+    }
+  }
+  return sections;
+}
+
+/**
+ * @param {Element} element
+ */
+function* iteratePreviousElements(element) {
```
marcoscaceres, line 123: same here...

marcoscaceres:
Looks great @ProgramMax. Just a couple of little suggestions.

src/core/tables.js
ProgramMax: I think this solves the same problem that class="numbered" already solved. But perhaps this solves it in a more elegant way.

If someone is using a table for layout (for example), that table should not be included in the list of tables. Having a per-table opt-in of class="numbered" is a way to indicate this is a genuine table with tabular data.

Ideally, all of these good tables have a caption. They can later be referenced by table number or caption. If a table does not have a caption, it can only be referenced by table number. This matters when generating the list of tables.

We decided not to generate an error and only make a linter rule for captionless tables. I interpreted this as being lenient on the caption requirement. But if we decide not to consider captionless tables at all, we don't need to worry about how they are referenced in the list of tables.

Further, someone using tables for non-tabular data is unlikely to provide a caption. So maybe we can skip the class="numbered" requirement and instead replace it with a caption requirement.

What do you think?
(I approve of this change either way. But if we accept it, we should probably make other changes to reflect its impact.)

ProgramMax: Correction: It was changed from a warning to a linter rule. Not from an error to a linter rule.

ProgramMax:
- Because we now iterate table.numbered instead of table (then checking for class="numbered"), the problem of "table 1, example, table 3" is now solved. It is now "table 1, example, table 2".
- I have removed the no-captionless-tables linter rule. I will put it in a separate pull request to keep the commit log clear.
- I will update the wiki and this PR once we resolve a decision: Should the presence of a caption be the required opt-in or should class="numbered"?

If the presence of a caption is the opt-in, we no longer need the awkward class. However, existing ReSpec users might be surprised when their tables change without them knowing. Maybe this is fine.

If we only process tables with captions (by accepting the proposed changes above, which I approve) then we're *this close* to using the caption as an opt-in. There is a case where an existing ReSpec user has a table with a caption but is not prepared for that table to be parsed. For example, they might be manually numbering their tables and ReSpec will apply an additional number.

In either case, if we accept the proposed changes above we should also update the anchor tag to no longer worry about captionless tables.

ProgramMax:
The Pull Request to add the linter rule is: https://github.com/w3c/respec/pull/4211
This PR should land first.

src/core/tables.js
```diff
@@ -0,0 +1,111 @@
...
+/**
+ * process all tables
+ */
+function collectTables() {
+  /** @type {HTMLLIElement[]} */
+  const listOfTables = [];
+  /** @type {NodeListOf<HTMLTableElement>} */
+  const tables = document.querySelectorAll("table.numbered");
+  const tablesArray = Array.from(tables);
+  tablesArray
```
sidvishnoi, line 49:
```suggestion
  [...tables]
```

sidvishnoi (approved):

ProgramMax:
Thank you for the help there. I'll update the wiki to include docs.

ProgramMax:
I'm trying to edit the wiki to add documentation.
But I don't see buttons for edit / new page. Is there a permissions issue? Are those only available to approved accounts?

sidvishnoi:
@ProgramMax The permission was to "anyone with write access to repository". I've eased it now.
Try editing now? https://github.com/w3c/respec/wiki/numbered/_edit

ProgramMax:
Thank you. Will do.
Actually, I learned there is a secret git repo for the wiki. I was able to `git clone git@github.com:w3c/respec.wiki.git`. So I think I might be able to send a pull request and get approval without a blanket "anyone can edit anything" permission. Is it okay if I try the git repo method? I haven't used it before and want to learn. :)

sidvishnoi:
The git method won't allow for sending a pull request, but if you do a git push, it'll be pushed as is. It's same as GitHub UI, but more git-ty.

Also, https://ReSpec.org/docs/ will be automatically updated as you push/edit wiki.

Issue #4206, opened 2022-07-05
Add captions to images in Markdown mode


xfq:
**Is your feature request related to a problem? Please describe.**

There is no easy way to add captions to images in Markdown mode.

**Describe the solution you'd like**

It would be nice to have some easy way to add a caption when using Markdown. There are packages like [markdown-it-image-figures](https://github.com/Antonio-Laguna/markdown-it-image-figures#options) for [markdown-it](https://www.npmjs.com/package/markdown-it), but I don't know if [marked](https://marked.js.org/) has something similar.

**Additional context**

Many images in W3C specs have captions, like:

* [clreq](https://w3c.github.io/clreq/#increased_inter_character_spacing)
* [jlreq](https://w3c.github.io/jlreq/#characters_used_for_japanese_composition)
* [MiniApp for IoT](https://w3c.github.io/miniapp-iot/#switch-panel)
* [uievents-code](https://w3c.github.io/uievents-code/#keyboard-sections)
* [Motion Path](https://www.w3.org/TR/motion-1/)

**Describe any alternatives you've considered**

Use HTML directly.

PR #4207, opened 2022-07-05
feat(core/markdown): render images with title as `<figure>`

sidvishnoi:
Closes https://github.com/w3c/respec/issues/4206

```md
![This is an alt](fig.png "This is a title")
```

becomes:

```html
<figure>
    <img src="fig.png" alt="This is an alt">
    <figcaption>This is a title</figcaption>
</figure>
```

### DIFF
--- a/src/core/markdown.js
+++ b/src/core/markdown.js
@@ -37,6 +37,19 @@ class Renderer extends marked.Renderer {
     return html.replace("<pre>", `<pre title="${title}" class="${className}">`);
   }
 
+  image(href, title, text) {
+    if (!title) {
+      return super.image(href, title, text);
+    }
+    const html = String.raw;
+    return html`
+      <figure>
+        <img src="${href}" alt="${text}" />
+        <figcaption>${title}</figcaption>
+      </figure>
+    `;
+  }
+
   /**
    * @param {string} infoString
    */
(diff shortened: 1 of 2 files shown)


xfq (approved):
LGTM. Thanks for implementing this so quickly!

marcoscaceres:
hmm... are we effectively creating custom markdown here? that might result in issues in the future.

sidvishnoi:
This is standard markdown (from parsing perspective). Just the renderer is custom.

Standard markdown behavior is:
```md
![This is an alt](fig.png "This is a title")
```
```html
<img alt="This is an alt" title="This is a title" src="fig.png">
```

Instead of image with title, which is kinda useless for non-mouse users, we're turning it into a figure. 
i.e. `[title]` => `<figcaption>`

Might be better to make it configurable via `respecConfig`?

marcoscaceres:
>This is based on https://github.com/Antonio-Laguna/markdown-it-image-figures#options BTW (as described in linked issue).

Yep, saw :) 

> Maybe instead of treating it as a behavior change in Markdown, consider it a ReSpec syntax? We've support for [custom heading IDs already](https://github.com/w3c/respec/blob/b8709ef34c9ec47d2139b865100ee907c06c8cae/src/core/markdown.js#L76-L83).

Ok, but this one is a little different because we share it with Bikeshed.

> Also, just to be clear...

Ok, let's roll with that. The risk that it will come back to bite us is fairly minimal, but wanted to caution.

marcoscaceres (approved):

PR #4232, opened 2022-08-03
fix(styles/ui): respect prefers-reduced-motion, cooler loading animation

### LINKED ISSUE #4193: turn off (or make quicker) the animations to respect reduced-motion preferences
## Description of problem

Certain actions in specs cause a keyframes "pop" animation

https://w3c.github.io/aria/

Animation occurs even when reduce-motion OS preference is on

A spec shouldn't need animation. Suggest to use media query to prevent this animation when reduce motion is enabled

* go to spec
* click on informative or normative references link

fycdev:
### Disable animation of respec-pill when prefers-reduced-motion is true

Fixes #4193 

Not sure if this requires a test and if so would it be under `tests/spec/core/ui-spec.js`?

### DIFF
--- a/src/styles/ui.css.js
+++ b/src/styles/ui.css.js
@@ -49,28 +49,29 @@ export default css`
 }
 
 #respec-pill:disabled {
-  font-size: 2.8px;
-  text-indent: -9999em;
-  border-top: 1.1em solid rgba(40, 40, 40, 0.2);
-  border-right: 1.1em solid rgba(40, 40, 40, 0.2);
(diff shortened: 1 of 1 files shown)


fycdev:
Will also be updating the `respect-pill` button styles alongside this fix as described by @marcoscaceres 👍

sidvishnoi (approved):
Love that morphing effect, @fycdev!

sidvishnoi:
And no, no need for tests for this.

marcoscaceres (approved):
lgtm! Nice one @fycdev!

PR #4239, opened 2022-08-12
fix(core/issues-notes): ignore svg descendents from processing

### LINKED ISSUE #4238: Ignore SVG styles and classes

I am trying to embed SVG renderings of music notation into a ReSpec Document. The SVG contains `class` attributes on the elements that intersect with known ReSpec classes (e.g., `.note`), which then processes the SVG and inserts HTML. 

e.g., it will take: 

`<g data-id="n1ck10te" data-class="note" class="note">`

and change this to:

```
<div class="note" role="note" id="issue-container-generatedID-3">
    <div role="heading" class="note-title marker" id="h-note-3" aria-level="5">
        <span>Note</span><!---0.688706%-->
    </div>
    <g data-id="nve1u6i" data-class="note" class=""></g>
</div>
```

This obviously breaks the SVG rendering. I cannot control the output of the SVG and, indeed, `note` seems like a perfectly valid class for music notation renderings.

The solution might be one, or both of:

 - Ignore all `class` attributes on elements with `<svg>` as an ancestor;
 - Add a CSS class on elements that tells ReSpec to completely skip processing all descendants 

**Additional context**

I can personally see no good reason for ReSpec to be processing anything within an `<svg>` block, so at a minimum it should probably ignore SVG entirely.

I have no work-around. I have tried various combinations of `exclude` or `ignore` classes, but nothing seems to work. I also can't see any obvious way to fix this with a different profile, or plugins, since the processing of these classes runs pretty deep into the core as far as I can tell. 

Yes, I probably have the skills and can try to implement this, but with two caveats:

1. I am not familiar with the ReSpec codebase, so would need pointers on where to look; and 
2. I would need a design decision from the community on which solution is preferable (or indeed, if another solution is desired)

NB: I apologize if this is a duplicate of a message I sent to the mailing list. That seems to not work very well.

ahankinson:
This changes the behaviour of the issues and notes selector to ignore any elements that have an SVG element as an ancestor. It does this by filtering the selected elements and looking to see if the `.closest` method returns null; if it is null, it means that the element is not within an SVG block; if it is not null, it is filtered out of the list of elements for further processing.

A simple unit test is included that checks to see if there are any `<div>` elements that are within an `<svg>` element (`<div>` is not a valid child of `<svg>`). If the selector is null it means the test passed.

Fixes #4238.

### DIFF
--- a/src/core/issues-notes.js
+++ b/src/core/issues-notes.js
@@ -94,7 +94,7 @@ const l10n = getIntlData(localizationStrings);
  * @property {string} bodyHTML
  * @property {GitHubLabel[]} labels
 
- * @param {NodeListOf<HTMLElement>} ins
+ * @param {HTMLElement[]} ins
  * @param {Map<string, GitHubIssue>} ghIssues
  * @param {*} conf
  */
@@ -371,9 +371,8 @@ export async function run(conf) {
   const allEls = document.querySelectorAll(query);
 
   const issuesAndNotes = Array.from(allEls).filter(itm => {
-    // Ignores any elements that were selected that is an ancestor of the
-    // <svg> element.
-    return !itm.closest("svg");
+    // Removes any elements that are not HTML Elements (e.g., SVG nodes)
+    return itm instanceof HTMLElement;
   });
 
   if (!issuesAndNotes.length) {
--- a/src/core/issues-notes.js
+++ b/src/core/issues-notes.js
@@ -368,7 +368,14 @@ async function fetchAndStoreGithubIssues(github) {
 export async function run(conf) {
   const query = ".issue, .note, .warning, .ednote";
   /** @type {NodeListOf<HTMLElement>} */
-  const issuesAndNotes = document.querySelectorAll(query);
+  const allEls = document.querySelectorAll(query);
+
+  const issuesAndNotes = Array.from(allEls).filter(itm => {
+    // Ignores any elements that were selected that is an ancestor of the
+    // <svg> element.
+    return !itm.closest("svg");
+  });
+
   if (!issuesAndNotes.length) {
     return; // nothing to do.
   }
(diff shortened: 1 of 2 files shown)


src/core/issues-notes.js
```diff
@@ -368,7 +368,14 @@ async function fetchAndStoreGithubIssues(github) {
 export async function run(conf) {
   const query = ".issue, .note, .warning, .ednote";
   /** @type {NodeListOf<HTMLElement>} */
-  const issuesAndNotes = document.querySelectorAll(query);
+  const allEls = document.querySelectorAll(query);
+
+  const issuesAndNotes = Array.from(allEls).filter(itm => {
+    // Ignores any elements that were selected that is an ancestor of the
+    // <svg> element.
+    return !itm.closest("svg");
```
sidvishnoi, line 376: This check seems a bit expensive to me, as `closest()` will end up traversing until `<html>`. I think we can do the following safely:

```suggestion
    return !(itm instanceof HTMLElement);
```

image

tests/spec/core/issues-notes-spec.js
```diff
@@ -420,4 +420,22 @@ describe("Core — Issues and Notes", () => {
...
+           xmlns="http://www.w3.org/2000/svg">
+        <!-- this element should not be selected even with the "note" class -->
+        <rect width="100%" height="100%" fill="red" class="note" />
+      </svg>`,
+    };
+
+    const doc = await makeRSDoc(ops);
+    // <div> is not a valid child of <svg>, so if it appears there it means ReSpec has added it.
+    const svgNote = doc.querySelector("svg div");
+    expect(svgNote).toBeNull();
```
sidvishnoi, line 439: 👍🏽

src/core/issues-notes.js
ahankinson: One minor change -- it was `return itm instanceof HTMLElement;` instead of `return !(itm instanceof HTMLElement);`. Otherwise, *only* non-HTMLElement nodes were selected.

sidvishnoi (approved):

sidvishnoi:
I'll merge and release tomorrow. Thanks for quickly sending the PR @ahankinson!

Issue #4156, opened 2022-05-11
Checks for privacy and security sections don't work if lang is set to en-US


jnurthen:
## Description of problem

Checks for privacy and security sections don't work if lang is set to en-US. Changing the Document language to en correctly shows the warnings but having the lang set to en-US does not.

Document languages are recommended to be as specific as possible to en-US is correct.

**URL to affected spec or repo:**

https://w3c.github.io/aria/

**What happened (e.g., it crashed)?:**

Gives console warning Missing localization strings for en-US.

**Expected behavior (e.g., it shouldn't crash):**

Should complain about no separate security / privacy sections

xfq:
The reason may be because [ReSpecs's language matching method](https://github.com/w3c/respec/blob/471d6f5b20aab17e0b9db969418ef53f4b568d4c/src/core/utils.js#L176-L208) is relatively simple. Ideally, the algorithm in [BCP 47](https://www.rfc-editor.org/info/bcp47) should be used.

See https://github.com/wooorm/bcp-47-match for an example.

PR #4277, opened 2022-10-01
fix(core/utils:`getIntlData`): match language subtag too

sidvishnoi:
Fixes https://github.com/w3c/respec/issues/4156

A full BCP 47 checker might be too much here.

Also add `getIntlDataForKey` util to test for translation availability.

### DIFF
--- a/src/w3c/linter-rules/required-sections.js
+++ b/src/w3c/linter-rules/required-sections.js
@@ -9,13 +9,12 @@ import {
   InsensitiveStringSet,
   docLink,
   getIntlData,
+  getIntlDataForKey,
   norm,
-  resolveLanguageAlias,
   showError,
   showWarning,
 } from "../../core/utils.js";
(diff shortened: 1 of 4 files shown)


src/w3c/linter-rules/required-sections.js
```diff
@@ -59,8 +58,10 @@ export function run(conf) {
   }
 
   // We can't check for headers unless we also have a translation
-  if (!localizationStrings[resolveLanguageAlias(lang)]) {
-    console.warn(`Missing localization strings for ${lang}.`);
+  if (!getIntlDataForKey(localizationStrings, "privacy_considerations")) {
+    const msg = `Cannot check for required sections as translations are not available.`;
```
sidvishnoi, line 62:
```suggestion
  if (!getIntlDataForKey(localizationStrings, "privacy_considerations")) {
    // We can't check for headers unless we also have a translation
    const msg = `Cannot check for required sections as translations are not available.`;
```

Issue #4295, opened 2022-10-17
Question of normative vs. informative references


johanneswilm:
We need some user help in this PR: https://github.com/w3c/input-events/pull/139

The I18N group told us to link for a specific definition into their spec. Respec says that that definition is not normative, but they say they have never had an issue with their spec not being normative enough. I've looked through the documentation, and 

1. I cannot find how to to make an informative reference to a single term in a normative part of a spec (I did figure out how to make an informative reference to an entire document).
2. It seems to me that the reference should actually be normative and not just informative and so we'd need your advice on how we should proceed with that if that.

tidoust:
On the notion of "normative enough", the official definition of [normative](https://www.w3.org/Consortium/Patent-Policy-20200915/#dfn-norm) for W3C specs is in the Patent Policy and it only applies to documents that are published on the Recommendation track. By this definition, documents published as Notes do not contain normative terms and statements. The Internationalization Glossary is authored with ReSpec. ReSpec adds an `informative` class to the `<body>` as a result. This is what makes all definitions it contains identified as "informative".

The W3C Process provides guarantees of stability (and reviews) for normative portions of documents published on the Recommendations track. There are no such guarantees for Notes and informative content. Nothing in the process prevents the Internationalization working group from issuing a new version of the glossary tomorrow without the definition you're trying to reference.

Now, in practice, the evaluation of whether a normative reference to an external term or specification is acceptable (i.e. "normative enough") is [up to the Director when a spec transitions on the Recommendation track](https://www.w3.org/2013/09/normative-references). The Director may well evaluate that it is appropriate to have, in a normative section, a normative reference to a term defined in a document that is not on the Recommendation track.

As far as I can tell, there is only one other browser spec right now that references the Internationalization Glossary normatively: [Secure Payment Confirmation](https://w3c.github.io/secure-payment-confirmation/). That spec has not yet transitioned to Candidate Recommendation so I think that the Director has not yet reviewed that reference.

Back to the issue at hand, it seems fine to me that ReSpec issues a warning when an author references an informative term from within a normative section, but I agree that there should be a way to bypass this warning and have ReSpec generate:
- a real link to "grapheme", instead of a link flagged with a warning message.
- a normative reference to I18N-GLOSSARY in the list of normative references (currently, ReSpec does not seem to add any entry at all, even looking at the list of informative references)

marcoscaceres:
[...]

PR #4329, opened 2022-12-02
feat(linter-rules/informative-dfn): warn on refs to informative dfns

dontcallmedom:
There are situations where making a normative reference to a definition in an informative section is either OK or unavoidable (e.g. because updating the referenced spec is unlikely to be possible). Moving the relevant check in a linter make it easier to silence either individual or all instances of the warning.

see also #4295

### DIFF
--- a/src/core/xref.js
+++ b/src/core/xref.js
@@ -25,7 +25,6 @@ import {
   nonNormativeSelector,
   norm as normalize,
   showError,
-  showWarning,
 } from "./utils.js";
 import { possibleExternalLinks } from "./link-to-dfn.js";
 import { sub } from "./pubsubhub.js";
@@ -37,6 +36,8 @@ const profiles = {
 };
 
 export const API_URL = "https://respec.org/xref/";
+
+export const informativeRefsInNormative = [];
 
 if (
   !document.querySelector("link[rel='preconnect'][href='https://respec.org']")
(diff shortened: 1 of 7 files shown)


sidvishnoi (approved):

src/core/xref.js
```diff
@@ -439,9 +440,8 @@ function addToReferences(elem, cite, normative, term, conf) {
     return;
   }
 
-  const msg = `Normative reference to "${term}" found but term is defined "informatively" in "${cite}".`;
-  const title = "Normative reference to non-normative term.";
-  showWarning(msg, name, { title, elements: [elem] });
+  // This is used by the informative-dfn linter
+  informativeRefsInNormative.push({ term, cite, element: elem });
```
sidvishnoi, line 444: Nit: Maybe call it "spec"?
```suggestion
  informativeRefsInNormative.push({ term, spec: cite, element: elem });
```

src/core/xref.js
```diff
@@ -38,6 +37,8 @@ const profiles = {
 
 export const API_URL = "https://respec.org/xref/";
 
+export const informativeRefsInNormative = [];
```
sidvishnoi, line 40: Not sure how TypeScript types behave right now, but should add this to avoid `any[]`. (something like this, I have not checked below in IDE).
```suggestion
/** @type {{ term: string; spec: string; element: HTMLAnchorElement }[]} */
export const informativeRefsInNormative = [];
```

src/core/xref.js
```diff
@@ -38,6 +37,9 @@ const profiles = {
 
 export const API_URL = "https://respec.org/xref/";
 
+/** @type {{ term: string; spec: string; element: HTMLAnchorElement }[]} */
```
sidvishnoi, line 40:
```suggestion
/** @type {{ term: string; spec: string; element: HTMLElement }[]} */
```

PR #4335, opened 2022-12-08
fix(core/inlines): don't expand inlines in SVG

### LINKED ISSUE #4334: Respec insertion of `<abbr>` in SVG hides the text in the abbr
## Description of problem

When Respec wraps abbreviation text in an SVG inside an `<abbr>` element that text disappears.

https://github.com/w3c/dapt/pull/86 - unfortunately the PR Preview is broken, so you need to check out the branch to see this happening locally, and remove the worksaround `class="exclude"` on the DAPT abbreviation in `index.html`.

When the abbreviation "DAPT" is defined in the Respec document, and a `<text>` element containing "DAPT Script" is included in an SVG figure, the SVG only renders the text "Script", because the text "DAPT" was wrapped in an `<abbr>`.

The text "DAPT Script" should appear, and there should be no `<abbr>` elements inserted into the SVG.

_Possibly_ some alternative to `<abbr>` in SVG is possible, like a hover-over tooltip.

1.  Create a Respec document that defines an abbreviation using an `<abbr>` element, e.g. for the text "DAPT".
2.  Create an SVG with a `<text>` element that includes the abbreviated text (e.g. "DAPT Script").
3.  Include the SVG in the Respec document as a `<figure>` containing a `<div data-include="path/to/figure.svg"></div>` to bring the SVG document in.
4.  View the document: the abbreviated text inside the SVG is wrapped inside an `<abbr>` element, which is not rendered.

**Things I tried:**

I tried to add `class="exclude"` on the `<div>` that includes the SVG, and I tried wrapping that `<div>` in another `<div class="exclude">` but neither thing stopped the Respec script from finding and wrapping the abbreviated text.

I successfully stopped Respec from replacing the abbreviated text throughout the whole document, including the SVG, by adding `class="exclude"` to the originating abbreviation, but that is a non-ideal solution.

nigelmegitt:
Tests that `<abbr>` element is not generated within an svg element.

Closes #4334.

Inspired by @sidvishnoi 's https://github.com/w3c/respec/issues/4334#issuecomment-1342616362

What I haven't thought much about, or checked:
* whether _other_ inline expansions than `<abbr>` are needed but also blocked by this change.

I _assume_ this will block all inline expansions, but I'm not sure if there is some need for others apart from `<abbr>`. Since all the tests pass, I think nobody ever created a test case for inlines being expanded in SVG. I guess if someone is relying on that behaviour some other change will be needed, to exclude specific inlines from specific subtrees.

### DIFF
--- a/src/core/inlines.js
+++ b/src/core/inlines.js
@@ -274,7 +274,7 @@ export function run(conf) {
 
   // PROCESSING
   // Don't gather text nodes for these:
-  const exclusions = ["#respec-ui", ".head", "pre"];
+  const exclusions = ["#respec-ui", ".head", "pre", "svg"];
   const txts = getTextNodes(document.body, exclusions, {
     wsNodes: false, // we don't want nodes with just whitespace
   });
--- a/tests/spec/core/inlines-spec.js
+++ b/tests/spec/core/inlines-spec.js
@@ -124,6 +124,32 @@ describe("Core - Inlines", () => {
     expect(abbr.title).toBe("included abbr");
   });
 
+  it("excludes generating abbr elements in svg content", async () => {
+    const body = `
+      <section>
+        <h2>
+          <abbr title="expanded abbreviation">
+            EA
+          </abbr>
+        </h2>
+        <div id="test">
+          <svg version="1.1"
+            xmlns="http://www.w3.org/2000/svg">
+            <text>SVG EA</text>
+          </svg>
+          <p>HTML EA</p>
+        </div>
+      </section>
+    `;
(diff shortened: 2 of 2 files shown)


sidvishnoi (approved):
Thanks @nigelmegitt!

nigelmegitt:
I'm not authorized to merge this - hopefully someone else can either merge it for me, or authorize me? This is my first pull request to this repo so I'm not sure what's normal procedure in this project.

nigelmegitt:
Thanks @sidvishnoi just saw your comment after posting mine. Guess that answers the question!

PR #4343, opened 2022-12-14
feat(w3c): support specStatus "ED" for the TAG

darobin:
The TAG can now produce REC-track docs, and in general tends to use a different method of drafting compared to what things were like ten years ago.

I added `ED` in this way rather than by adding it to eg. `tagStatus` or `trStatus` as that would break other things that those are used for.

### DIFF
--- a/src/w3c/defaults.js
+++ b/src/w3c/defaults.js
@@ -155,10 +155,10 @@ function validateStatusForGroup(conf) {
     case "other":
       if (
         group === "tag" &&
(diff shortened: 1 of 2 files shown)


marcoscaceres:
Seems ok. I wonder if it's worth adding a test though?

darobin:
Fair point, gone and done it!

sidvishnoi (approved):

PR #4344, opened 2022-12-16
fix(core/dfn): run `addContractDefaults` before linters

### LINKED ISSUE #4323: Bogus "unused definition" linter report
## Description of problem
When a definition doesn't explicitly set `data-export` in the source markup, but its exportability is inferred by ReSpec from its definition type, the "no-unused-dfns" linter complains that the definition is unused.

The linter claims the definition isn't used (which is true but should not be raised since it's an exported definition).

It should not warn.

```html
<!doctype html>
<html>
  <title>Element definition</title>
  <script>var respecConfig = {};</script>
  <script src="https://www.w3.org/Tools/respec/respec-w3c.js"></script>
  <body>
  <section>
  <h2>Test Section</h2>
  <p>There is  1 element: <dfn class="element">foo</dfn></p>
```

The problem seems to be that `addContractDefaults()` method in `core/dfn.js` is invoked after the linter runs (presumably because it's only called after all plugins are done). If so, it sounds like fixing this would require running the linter under a different execution group than the current "plugins".

sidvishnoi:
Fixes #4323 

Not a huge fan of this approach, but 🤷🏽‍♂️.


sidvishnoi:
I'll instead add a new plugin that runs before linters..

src/core/base-runner.js
```diff
@@ -21,9 +21,14 @@ export async function runAll(plugs) {
   await preProcess(respecConfig);
 
   const runnables = plugs.filter(p => isRunnableModule(p));
-  runnables.forEach(
-    plug => !plug.name && console.warn("Plugin lacks name:", plug)
-  );
+  runnables.forEach((plug, i) => {
+    if (!plug.name) {
+      console.warn("Plugin lacks name:", plug);
```
marcoscaceres, line 26:
```suggestion
      console.warn(`Plugin ${i} lacks name:`, plug);
```

Issue #3426, opened 2021-03-30
Apparent requirement for nested includes and/or simpler markdown inclusion


SteveALee:
# Background

I'm working on the W3C note [Content Usable](https://w3c.github.io/coga/content-usable/) in order to also render the [Design Guide](https://w3c.github.io/coga/content-usable/#design_for_everyone) section as a set of web pages in the [W3C WAI website](https://www.w3.org/WAI/).

The WAI website is built using the Jekyll SSG and the design guide section is currently a single HTML + respec file. The basic goal is to refactor the Design Guide into a number of files that can be consumed by the SSG with minimal processing. The structure will also be processed by respec and saved as usual for the W3C note.

**Is your feature request related to a problem? Please describe.**

# Matching includes to section structure

So the structure of the Design guide is as follows:
 
- heading Design Guide
- section intro
  -  heading Objective 1
  - section intro
       - heading Pattern 1
       - section 1 intro
       - section 2 
       - 
       -  heading Pattern 2
       -   
    - heading Objective 2
    - section

With each heading and section(s) being contents of one file.  

file DG:
- heading Design Guide
- section intro
- include file O_1

file O_1
-  heading Objective 1
- section intro
- include file P_1

File P_1:
 - heading Pattern 1
 - section 1 intro
 - section 2 

etc

This would be naturally suggest nested includes but they are not supported. 

Flattening the structure requires markup or content that should be in the included file to be in the including file. Plus I found it impossible to get the section tags correctly nested. This all complicates file processing and a 1 to 1, section to file mapping appears to be the natural solution. 

# Markdown

Ideally the files will all contain markdown but as the content is currently HTML that will be OK. 

**Describe the solution you'd like**

- Nested includes so files map 1 to 1 to sections
- top level just a list of includes - don't mind what tag - div would do, especially if auto replace is used.
- Each file can start with a Heading and I don't mind if the section tag gets added automatically as part of the include (as it seems to right now) or we have to add them to the content (but not if markdown).

And also when using markdown:

- markup declared at the point of inclusion - ie data-attrib on the include

**Describe alternatives you've considered**

- Flattening the includes and playing around with missing/extra section end tags to get the correct structure. 
- tried markdown but i could get the include markdown to only work in limited circumstances when working throughthe permutations. That might well me my lack of understanding.

Any suggestions are most welcome on how to implement this now, or with some tweaks to respec for nested include of html and markdown.

[...]

PR #4372, opened 2023-01-29
feat(core/data-include): support nested includes

sidvishnoi:
For performance reasons, limited to maximum depth of 3.

> **Warning** Does not take cycles into account (so, be careful).

See https://github.com/w3c/wcag2ict/issues/84#issuecomment-1407208896 and https://github.com/w3c/respec/issues/3426


sidvishnoi:
@shawnthompson Can you verify it works when you replace:
```html
  <script src="https://www.w3.org/Tools/respec/respec-w3c" class="remove" defer>
```
with
```html
  <script src="https://deploy-preview-4372--respec-pr.netlify.app/respec-w3c.js" class="remove" defer>
```

shawnthompson:
@sidvishnoi I just verified and it works as expected.

Thank you very much.

PR #4402, opened 2023-03-17
tests(core/xref): update broken xref-based tests

sidvishnoi:


sidvishnoi:
This will likely not be needed when reffy/webref runs again (it runs once a day for TR specs). https://github.com/w3c/webref/commit/4a5b4cf79c8042b6283f07780dc181c8029da85d added resource-hints, but now the spec redirects back to HTML.

PR #4409, opened 2023-03-30
fix(w3c/style): add `crossorigin` attribute to W3C logo preload

deniak:
The W3C logo on respec source documents doesn't appear correctly on Chrome which reports the following warning/error:
* warning: A preload for 'https://www.w3.org/StyleSheets/TR/2021/logos/W3C' is found, but is not used because the request credentials mode does not match. Consider taking a look at crossorigin attribute.
* error: Access to image at 'https://www.w3.org/StyleSheets/TR/2021/logos/W3C' from origin 'https://w3c.github.io' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.

MDN describes that the [preload `link` should match the credential mode](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/preload#cors-enabled_fetches) used to fetch the resource so I think it's missing the `crossorigin` attribute.

### DIFF
--- a/src/core/utils.js
+++ b/src/core/utils.js
@@ -69,8 +69,8 @@ export function createResourceHint(opts) {
       if ("as" in opts) {
         linkElem.setAttribute("as", opts.as);
       }
-      if ("crossorigin" in opts) {
-        linkElem.crossOrigin = opts.crossorigin;
+      if (opts.corsMode) {
+        linkElem.crossOrigin = opts.corsMode;
       }
       break;
   }
--- a/src/core/utils.js
+++ b/src/core/utils.js
@@ -69,6 +69,9 @@ export function createResourceHint(opts) {
       if ("as" in opts) {
         linkElem.setAttribute("as", opts.as);
       }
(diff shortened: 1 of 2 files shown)


sidvishnoi (approved):

src/core/utils.js
```diff
@@ -69,6 +69,9 @@ export function createResourceHint(opts) {
       if ("as" in opts) {
         linkElem.setAttribute("as", opts.as);
       }
+      if (opts.corsMode) {
+        linkElem.crossOrigin = opts.corsMode;
+      }
```
sidvishnoi, line 74: @deniak Any reason to not use the same as line 64 above?
```suggestion
      if (opts.corsMode || url.origin !== document.location.origin) {
        linkElem.crossOrigin = opts.corsMode || "anonymous";
      }
```

sidvishnoi: I guess this the approach in this PR is more explicit. So, that's ok.

Issue #4324, opened 2022-11-24
Missing dfn error needs HTML escaping


dontcallmedom:
## Description of problem
When opening the error list in the respec pill on the following document:
```html
<!doctype html>
<html>
  <title>XSS in error reporting</title>
  <script>var respecConfig = {};</script>
  <script src="https://www.w3.org/Tools/respec/respec-w3c.js"></script>
  <body>
  <section>
  <h2>Test Section</h2>
  <p><a>&lt;script&gt;alert("xss");&lt/script>&gt;</a>
```
an alert dialog showing xss pops up

**What happened (e.g., it crashed)?:**
The markup, HTML-escaped in the document, is interpereted literally when included in the ReSpec error list.

**Expected behavior (e.g., it shouldn't crash):**
It should keep the escaped markup escaped in the error list.

PR #4431, opened 2023-04-20
fix(core/dfn-index): escape HTML/XML tags

### LINKED ISSUE #4430: External reference of XML elements
## Description of problem

https://w3c.github.io/dapt/#index-defined-elsewhere

DAPT specification references XML element defined in TTML2, such as `<ttp:feature>`, and writing as `<dfn data-cite="ttml2#profile-vocabulary-feature"><code>&lt;ttp:feature&gt;</code></dfn>` using character reference for brackets.

In Terms defined by reference section, these are listed as `<ttp:feature></ttp:feature>` but not using character reference.
dfn-index.js in respec uses [`${{ html: text }}`](https://github.com/w3c/respec/blob/a950478bca8bcfc3f540ca8d40a4e5e01ae5cd7d/src/core/dfn-index.js#L296) for this part.

show term correctly using character reference.

1.  ...

nigelmegitt:
A fix for #4430 : when the normalized text is rendered as html e.g. in `dfn-index.js` line 298 within `renderExternalTermEntry` using `html: text` this prevents creation of what the UA thinks are HTML elements from terms defined with angle brackets at the start and end.

An alternative fix could be to replace `html: text` with `text: text` but I think that would break cases where code blocks are deliberately put around the term in the index.

This fix seems to pass all the tests.

---

- Fixes #4430
- Fixes https://github.com/w3c/respec/issues/4324

### DIFF
--- a/src/core/utils.js
+++ b/src/core/utils.js
@@ -174,9 +174,7 @@ export function xmlEscape(str) {
  * @param {string} str
  */
 export function norm(str) {
-  return str
-    .trim()
-    .replace(/\s+/g, " ");
+  return str.trim().replace(/\s+/g, " ");
 }
 
 /**
--- a/src/core/utils.js
+++ b/src/core/utils.js
@@ -174,7 +174,11 @@ export function xmlEscape(str) {
  * @param {string} str
  */
 export function norm(str) {
-  return str.trim().replace(/\s+/g, " ").replace(/</g, '&lt;').replace(/>/g, '&gt;');
+  return str
+      .trim()
+      .replace(/\s+/g, " ")
+      .replace(/</g, '&lt;')
+      .replace(/>/g, '&gt;');
 }
 
 /**
--- a/src/core/utils.js
+++ b/src/core/utils.js
@@ -174,7 +174,7 @@ export function xmlEscape(str) {
  * @param {string} str
  */
 export function norm(str) {
-  return str.trim().replace(/\s+/g, " ");
+  return str.trim().replace(/\s+/g, " ").replace(/</g, '&lt;').replace(/>/g, '&gt;');
 }
 
 /**
--- a/src/core/utils.js
+++ b/src/core/utils.js
@@ -175,10 +175,10 @@ export function xmlEscape(str) {
  */
 export function norm(str) {
   return str
-      .trim()
-      .replace(/\s+/g, " ")
-      .replace(/</g, '&lt;')
-      .replace(/>/g, '&gt;');
+    .trim()
+    .replace(/\s+/g, " ")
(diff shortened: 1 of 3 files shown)


nigelmegitt:
Ideally I would have added a test for this but I wasn't sure the best way to do that. Hints accepted!

sidvishnoi:
In terms of new test, it'd be be great if you can add the `<ttp:feature>` reference mentioned in #4430: https://github.com/w3c/respec/blob/a950478bca8bcfc3f540ca8d40a4e5e01ae5cd7d/tests/spec/core/dfn-index-spec.js#L159

---

Updating the `norm` function here seems a bit risky to me, even though the tests pass. `norm` is used in a lot more places, that might not be tested well enough.

Can we just update the part where this is rendered (consider adding the test first):
https://github.com/w3c/respec/blob/a950478bca8bcfc3f540ca8d40a4e5e01ae5cd7d/src/core/dfn-index.js#L338

sidvishnoi:
Actually, if you use following, we get fix for #4324 too:
```diff
diff --git a/src/core/xref.js b/src/core/xref.js
index 4e0e9e8b..e53048e0 100644
--- a/src/core/xref.js
+++ b/src/core/xref.js
@@ -25,6 +25,7 @@ import {
   nonNormativeSelector,
   norm as normalize,
   showError,
+  xmlEscape,
 } from "./utils.js";
 import { possibleExternalLinks } from "./link-to-dfn.js";
 import { sub } from "./pubsubhub.js";
@@ -189,7 +190,7 @@ function getRequestEntry(elem) {
 export function getTermFromElement(elem) {
   const { lt: linkingText } = elem.dataset;
   let term = linkingText ? linkingText.split("|", 1)[0] : elem.textContent;
-  term = normalize(term);
+  term = xmlEscape(normalize(term));
   return term === "the-empty-string" ? "" : term;
 }
```

as `getTermFromElement` is used in both places (and only in these two places).

nigelmegitt:
Thanks @sidvishnoi, trying that now...

nigelmegitt:
@sidvishnoi thanks for the tips, that appears to work. I reverted my change, created the tests first, verified that they failed, implemented your proposal, and verified that the tests passed and that with an example document the behaviour was indeed as expected. 

I'm not sure how to write the test for #4324 because it seems to rely on a behaviour within the Respec "pill" specifically, so I cannot confirm that this does fix it, however it seems likely that it would.

sidvishnoi:
Let's skip tests for #4324, I verified locally.

sidvishnoi:
Released in [`v34.0.1`](https://github.com/w3c/respec/releases/tag/v34.0.1).

PR #4495, opened 2023-07-26
fix(respecDocWriter): use default `userDataDir`

deniak:
It doesn't seem necessary to generate a different user data directory for each document.
spec-generator relies on `RespecDocWriter` to generate snapshots from respec documents and the recent increase of requests completely filled up the tmp partition.

### DIFF
--- a/tools/respecDocWriter.js
+++ b/tools/respecDocWriter.js
@@ -1,11 +1,10 @@
 /**
  * Exports toHTML() method, allowing programmatic control of the spec generator.
  */
-import { mkdtemp, readFile } from "fs/promises";
 import { fileURLToPath } from "url";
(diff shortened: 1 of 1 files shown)


sidvishnoi:
I was initially concerned that this will add history entries to user's profile when running locally, but it did not affect existing profiles for me 👍🏽 

The idea for using new dir was meant fresh state in each request - so no indexedDb or cache could interfere across builds. Maybe we should add this as an optional parameter to `toHTML()`? - Defaulting to current one, and spec-generator can override?

This works too, no strong opinions.

sidvishnoi (approved):
Ok to merge. We can add parameter if needed in future.

Issue #4538, opened 2023-09-13
"TypeError: handle.executionContext is not a function" at respecDocWriter.js:308


nondebug:
## Description of problem

After https://github.com/w3c/respec/issues/4535 was resolved I tried re-running checks on https://github.com/w3c/gamepad/pull/168 and got a different error. Generate Static HTML step fails with "TypeError: handle.executionContext is not a function" 

Failed run: https://github.com/w3c/gamepad/actions/runs/6169448759/job/16743712129

**What happened (e.g., it crashed)?:**

      file:///home/runner/work/_actions/w3c/spec-prod/v2/node_modules/.pnpm/respec@34.1.7_typescript@5.2.2/node_modules/respec/tools/respecDocWriter.js:308
          return await handle.executionContext().evaluate(obj => {
                              ^
      
      TypeError: handle.executionContext is not a function
          at stringifyJSHandle (file:///home/runner/work/_actions/w3c/spec-prod/v2/node_modules/.pnpm/respec@34.1.7_typescript@5.2.2/node_modules/respec/tools/respecDocWriter.js:308:25)
          at Array.map (<anonymous>)
          at file:///home/runner/work/_actions/w3c/spec-prod/v2/node_modules/.pnpm/respec@34.1.7_typescript@5.2.2/node_modules/respec/tools/respecDocWriter.js:320:51
          at file:///home/runner/work/_actions/w3c/spec-prod/v2/node_modules/.pnpm/puppeteer-core@21.2.0/node_modules/puppeteer-core/lib/esm/third_party/mitt/index.js:1:243
          at Array.map (<anonymous>)
          at Object.emit (file:///home/runner/work/_actions/w3c/spec-prod/v2/node_modules/.pnpm/puppeteer-core@21.2.0/node_modules/puppeteer-core/lib/esm/third_party/mitt/index.js:1:226)
          at CDPPage.emit (file:///home/runner/work/_actions/w3c/spec-prod/v2/node_modules/.pnpm/puppeteer-core@21.2.0/node_modules/puppeteer-core/lib/esm/puppeteer/common/EventEmitter.js:82:22)
          at #addConsoleMessage (file:///home/runner/work/_actions/w3c/spec-prod/v2/node_modules/.pnpm/puppeteer-core@21.2.0/node_modules/puppeteer-core/lib/esm/puppeteer/common/Page.js:625:14)
[...]

PR #4539, opened 2023-09-13
fix(tools/respecDocWriter): `handle.executionContext` is not a function

sidvishnoi:
Fixes https://github.com/w3c/respec/issues/4538

Not sure when this got removed.


sidvishnoi:
Headless tests should include tests for warnings/errors too 😄

PR #4561, opened 2023-10-13
feat(core/seo): support adding git revision in meta

dontcallmedom:
This allows to set a git revision of the source used to the generated content. Bikeshed does this, but with 'document-revision' as a meta name - that said, 'revision' is the name recommended in https://wiki.whatwg.org/wiki/MetaExtensions

Typically, this would be set when generating the static HTML (a possible separate improvement to spec-prod)


sidvishnoi (approved):

src/w3c/seo.js
```diff
@@ -33,6 +33,18 @@ export const requiresCanonicalLink = new Set([
 ]);
 
 export async function run(conf) {
+  if (conf.gitRevision) {
```
sidvishnoi, line 36: I wonder if it should be a part of `conf.github` though - but that can mean we lose setting `conf.github` as string. Maybe better as `gitRevision`.

sidvishnoi:
@dontcallmedom I'll move this to `core/seo` instead of limiting to `w3c/seo`.

PR #4585, opened 2023-11-17
fix(core/highlight-vars): scope highlighted var to `.algorithm` container

dontcallmedom:
This matches what bikeshed does https://speced.github.io/bikeshed/#var-and-algorithms


sidvishnoi (approved):

tests/unit/core/highlight-vars-spec.js
```diff
@@ -14,18 +14,29 @@ describe("Core - highlightVars", () => {
       <ol>
         <li><var id="section1-bar">bar</var>
         <li><var>a
-         foo</foo>
+         foo</var>
```
sidvishnoi, line 17: Good catch!

PR #4617, opened 2023-12-21
l10n(w3c/level): add some more strings; improve docs

marcoscaceres:


src/w3c/level.js
```diff
@@ -3,19 +3,45 @@
...
+ * @module w3c/level
  */
 
 import { getIntlData, showError } from "../core/utils.js";
 export const name = "w3c/level";
-
+/**
+ * @type {LocalizationStrings}
+ */
 const localizationStrings = {
```
sidvishnoi, line 15:
```suggestion
const localizationStrings = {
```

src/w3c/level.js
```diff
@@ -3,19 +3,45 @@
...
+  zh: {
+    level: "级别",
+  },
 };
 
+/**
+ * Localization data based on the user's language preference.
+ *
+ * @type {{ level: string }}
+ */
```
sidvishnoi, line 37:
```suggestion
/**
 * Localization data based on the user's language preference.
 *
 * @type {{ level: string }}
 */
```

Issue #4522, opened 2023-08-25
Contextualizing definitions for export 


ghost:
## Description of problem

The terminology include file is being processed, but restrictReferences() is failing, which results in unused terms appearing in the final document, each of which generates a warning.

**URL to affected spec or repo:**

https://w3c.github.io/vc-use-cases/

**What happened (e.g., it crashed)?:**

```
ReferenceError: $ is not defined
    at restrictReferences (common.js:123:5)
    at ln (utils.js:294:28)
    at Ls (data-include.js:54:16)
    at data-include.js:98:7
    at async Promise.all (/vc-use-cases/index 34)
    at async Ps (data-include.js:107:6)
    at async Object.run (data-include.js:77:8)
    at async base-runner.js:72:27
```

**Expected behavior (e.g., it shouldn't crash):**

Unused terms should be removed.

**Optional, steps to reproduce:**

See https://github.com/w3c/vc-use-cases/blob/2023-08-25-cleanup/index.html and https://github.com/w3c/vc-use-cases/blob/2023-08-25-cleanup/terms.html.

Fails under Chrome Version 116.0.5845.96 (Official Build) (x86_64) on MacBook Pro, macOS Monterey 12.6.7.

jandrieu:
Also see https://github.com/tobie/pr-preview/issues/144

marcoscaceres:
This appears to be an issue with `restrictReferences()` function in that spec's included common.js stuff (not part of ReSpec). It's trying to use "$", which I guess is jQuery? However, the main document is not including jQuery (ReSpec doesn't use or include jQuery anymore, and hasn't for many years).

marcoscaceres:
@jandrieu (and maybe @msporny), it would be amazing if there is stuff that is needed in common.js from the VC specs could move into ReSpec proper... I don't know what common.js is doing, but it's better if we just get ReSpec to handle whatever that is.

marcoscaceres:
@jandrieu, and just as an aside.... I think to fix the `$` issue, just add `<script src="https://ajax.googleapis.com/ajax/libs/jquery/3.7.1/jquery.min.js"></script>` to the doc above the ReSpec script.

msporny:
> @jandrieu, and just as an aside.... I think to fix the `$` issue, just add `<script src="https://ajax.googleapis.com/ajax/libs/jquery/3.7.1/jquery.min.js"></script>` to the doc above the ReSpec script.

[...]

PR #4620, opened 2023-12-21
feat(core/dfn): add `data-defines` on some patterns of term/definition association

dontcallmedom:
As discussed in #4522
See also https://github.com/w3c/reffy/pull/1444


src/core/dfn.js
```diff
@@ -237,4 +237,37 @@ function addContractDefaults() {
   for (const dfn of exportableDfns) {
     dfn.dataset.export = "";
   }
+
+  // - Sets data-defines on well-known definition content patterns
+
+  // A dt with class dfn-desc (on in a dl with such a class)
```
tidoust, line 243:
```suggestion
  // A dt with class dfn-desc (or in a dl with such a class)
```

src/core/dfn.js
```diff
@@ -237,4 +237,37 @@ function addContractDefaults() {
...
+  // containing a definition
+  // indicates that the following dd or div element contains its prose content
+  /** @type NodeListOf<HTMLElement> */
+  const describedDTs = document.querySelectorAll(
+    "dl.dfn-desc dt:has(dfn[data-dfn-type]), dt.dfn-desc:has(dfn[data-dfn-type])"
```
tidoust, line 248: What's the use of selecting only dfns that have a `data-dfn-type` attribute? In other words, what does it filter out?

src/core/dfn.js
```diff
@@ -237,4 +237,37 @@ function addContractDefaults() {
...
+  );
+  for (const dt of describedDTs) {
+    const dfnId = dt.querySelector("dfn[id]").id;
```
tidoust, line 251: This code expects that there will always be one definition with an `id` attribute and yet you didn't use that in your filter to select the dt elements. If the `id` is always set, the `[id]` could be dropped. If not, the filter above should be updated to restrict to dfns that have an `id` attribute, I think.

src/core/dfn.js
```diff
@@ -237,4 +237,37 @@ function addContractDefaults() {
...
+    if (dfnContent && dfnId && ["DIV", "DD"].includes(dfnContent.tagName)) {
+      dfnContent.dataset.defines = `#${dfnId}`;
+    }
+  }
+
+  // a non-dt element with class dfn-desc containing a definition
+  // indicates that the said element contains its prose content
+  /** @type NodeListOf<HTMLElement> */
+  const otherDescriptionContainers = document.querySelectorAll(
+    ":not(dt):not(dl).dfn-desc:has(dfn[data-dfn-type])"
```
tidoust, line 265: Same comments on `data-dfn-type` and lack of `id` as above.

src/core/dfn.js
dontcallmedom: It makes apply only to definitions that follow the definition contract (which at this stage of the processing should all have a `data-dfn-type` set).

src/core/dfn.js
dontcallmedom:
```suggestion
    const dfnId = dt.querySelector("dfn[data-dfn-type]").id;
```
good point (I had started by hooking on `[id]` but didn't think of updating this)

src/core/dfn.js
```diff
@@ -237,4 +237,37 @@ function addContractDefaults() {
...
+  );
+  for (const el of otherDescriptionContainers) {
+    const dfnId = el.querySelector("dfn[id]").id;
```
dontcallmedom, line 268:
```suggestion
    const dfnId = el.querySelector("dfn[data-dfn-type]").id;
```

src/core/dfn.js
tidoust: But is there any way to produce a definition that would *not* follow the definition contract at this stage of the processing?

(Totally fine to keep the selector in any case)

tidoust (approved):

src/core/dfn.js
dontcallmedom: I don't know, but feels like the semantic -right thing to do 😅

marcoscaceres:
I'm not loving `dfn-desc` to be honest... it's not as author friendly as it could be and not in line with other CSS named we use in bikeshed and respec (e.g., `<dl class="algorithm">`). Could we just use "`definition`" instead:

```html
<p class="definition">
   This is a <dfn>term</dfn> that is being defined.
</p>

<p>
   This is some term <dfn>term</dfn> that is being defined.
   And <span class="definition">this is some <dfn>other</dfn> that is being defined</spa>
</p>
```

We can still use `data-dfn-desc` as the thing, but ReSpec should add that... that's what we do for `class="export"`: it becomes `data-export=""` after processing.

src/core/dfn.js
```diff
@@ -237,4 +237,37 @@ function addContractDefaults() {
   for (const dfn of exportableDfns) {
     dfn.dataset.export = "";
   }
+
+  // - Sets data-defines on well-known definition content patterns
```
marcoscaceres, line 241: let's move this to it's own function... this function is already getting a bit too long.

marcoscaceres (changes requested):
Suggested some changes...

src/core/dfn.js
```diff
@@ -238,3 +243,37 @@ function addContractDefaults() {
...
+    if (dfnContent && dfnId && ["DIV", "DD"].includes(dfnContent.tagName)) {
+      dfnContent.dataset.defines = `#${dfnId}`;
+    }
+  }
+
+  // a non-dt element with class definition containing a definition
+  // indicates that the said element contains its prose content
+  /** @type NodeListOf<HTMLElement> */
+  const otherDescriptionContainers = document.querySelectorAll(
+    ":not(dt):not(dl).definition:has(dfn[data-dfn-type])"
```
sidvishnoi, line 271: Nit: I wonder if we should just iterate over `.definition:has(dfn[data-dfn-type])` and do the check/processing in a single loop. Keeps selectors simpler, and avoids some duplication (the `dataset.defines = #{dfnId}` part).

tests/spec/core/dfn-spec.js
```diff
@@ -655,5 +655,32 @@ describe("Core — Definitions", () => {
...
+        <section>
+          <h2>Definition and its description</h2>
+          <p id="desc1" class="definition">
+            A <dfn>definition</dfn> can also have an associated description
+          </p>
+          <dl class="definition">
+            <dt><dfn>different convention</dfn></dt>
+            <dd id="desc2">Different conventions can be applied to associate a term with its description</dd>
+          </dl>
+          <dl>
```
sidvishnoi, line 670: Thinking out loud:
Do we want/need to handle the cases with multiple dt or multiple dd?
Multiple dt: as an alias.. sounds like job for `data-lt`
Multiple dd: why?

tests/spec/core/dfn-spec.js
```diff
@@ -655,5 +655,32 @@ describe("Core — Definitions", () => {
...
+            <dt class="definition"><dfn>local convention</dfn></dt>
```
sidvishnoi, line 671: Can we expect `dt.definition` to contain anything other than `dfn`? Seems a bit redundant markup if not. I know we need to process both dt and dfn for different reasons, so this convention feels a bit unnatural..

tidoust: Thinking about it, I wonder whether it wouldn't be more intuitive for authors to write:

```html
<dt><dfn>local convention</dfn></dt>
<dd class="definition">The prose definition of the previous dfn</dd>
```

This would help capture additional cases automatically as well, such as a `<dfn>` in a heading followed by a paragraph that contains the actual definition. In other words, the code could look for the `<dfn>` within the element that has a `class="definition"` attribute, and if not found look for the closest `<dfn>` before that element.

If it's attached to the `<dt>`, a `hasdefinition` class might be clearer.

dontcallmedom: I've updated the code to match that suggestion, which I agree seems more author-friendly and flexible.

For `<dl>/<dt>`, I have moved to having a new `hasdefinitions` class, only for `<dl>` - individual entries can be managed using the regular "definition" class.

src/core/dfn.js
dontcallmedom: the new approach solves this as well

src/core/dfn.js
```diff
@@ -238,3 +243,32 @@ function addContractDefaults() {
     dfn.dataset.export = "";
   }
 }
+
+// - Sets data-defines on well-known definition content patterns
+function addDefinitionPointers() {
+  // A dl with class hasdefinitions marks all next siblings of dt with the class
+  // definition
+  /** @type NodeListOf<HTMLElement> */
+  const describedDTs = document.querySelectorAll("dl.hasdefinitions dt");
```
sidvishnoi, line 252: Maybe `dl.definitions`?

src/core/dfn.js
```diff
@@ -238,3 +243,32 @@ function addContractDefaults() {
...
+  for (const dt of describedDTs) {
+    dt.nextElementSibling.classList.add("definition");
+  }
+
+  // an element with class "definition" is marked as defining the term
+  // found either it the element itself, or in the closest previous sibling with
+  // a defined term
+  /** @type NodeListOf<HTMLElement> */
+  const definitionContainers = document.querySelectorAll(".definition");
+  for (const el of definitionContainers) {
```
sidvishnoi, line 262: I think we should exclude (or lint) those that already have `data-defines` set to a given local dfn.

src/core/dfn.js
```diff
@@ -238,3 +243,32 @@ function addContractDefaults() {
...
+    let curEl = el;
+    let dfn;
+    while (curEl) {
+      dfn = curEl.querySelector("dfn[data-dfn-type]");
```
sidvishnoi, line 266: A comment explaining why `[data-dfn-type]` would be be nice.

src/core/dfn.js
sidvishnoi: I don't see any reason why `data-dfn-type` won't be set, given `addContractDefaults` is run before:
https://github.com/w3c/respec/blob/1d198ab4abd4dab72b77b59998790c39795c34e7/src/core/dfn.js#L219-L227


But it still feels like semantic right thing to do to me too 😄😆

src/core/dfn.js
sidvishnoi: Could be follow-up.

sidvishnoi (approved):
LGTM 🎉 

Just one nit: `dl.hasdefinitions` should probably become `dl.definitions`.

tests/spec/core/dfn-spec.js
marcoscaceres: Yeah, this is probably the right thing to do here (as an explicit opt-in) for this new feature. 

We should eventually also cater for:

```HTML
<dl> 
<div class="definition">
  <dt><dfn>dd</dfn><dt>
  <dd>d1</dd>
  <dd>d2</dd>
</div>
</dl>
```

src/core/dfn.js
```diff
@@ -238,3 +243,32 @@ function addContractDefaults() {
     dfn.dataset.export = "";
   }
 }
+
+// - Sets data-defines on well-known definition content patterns
+function addDefinitionPointers() {
+  // A dl with class hasdefinitions marks all next siblings of dt with the class
```
marcoscaceres, line 249: I don't dislike this (`dl. definitions`), but I feel this is overstepping based on what we have previously discussed. I still think we should start with an extremely simple / basic system and build it up very incrementally (and eventually get to this, but I don't think we are there yet).

src/core/dfn.js
marcoscaceres: This feels like it should be a CSS query, not a for loop.

tests/spec/core/dfn-spec.js
```diff
@@ -655,5 +655,29 @@ describe("Core — Definitions", () => {
       expect(errors[0].message).toContain("Declares both");
       expect(errors[1].message).toContain("but also has a");
     });
+
+    it("assigns data-defines on well-known patterns", async () => {
+      const body = `
+        <section>
+          <h2>Definition and its <dfn>description</dfn></h2>
+          <p id="desc3" class="definition">A description explains what a term is.</p>
```
marcoscaceres, line 663: I think this is too magical. We should only rely on sibling order where HTML itself relies on order (i.e., only for `dt+dd`). Other should require explicit association through nesting (and/or `data-defines`).

src/core/dfn.js
dontcallmedom: a `<dl>` containing definitions is exactly what the [VC spec (our initial customer) is using](https://w3c.github.io/vc-use-cases/#terminology) AFAICT; if we're not going to support that convention, I'm not sure we need a respec patch at all at the moment (since reffy recognizes `data-defines` without any additional help).

tests/spec/core/dfn-spec.js
dontcallmedom: a heading with a `<dfn>` followed by a nearby definition of the term is pretty frequent pattern, but I certainly wouldn't object to leave that to a future improvement. @tidoust?

src/core/dfn.js
dontcallmedom: the reason I was liking "hasdefinitions" more than just "definitions" is that it is more easily distinguishable from the other pattern (it's easy to see that "hasdefinitions" is different from "definition", but it would really easy to confuse "definition" and "definitions" which generate quite different behavior).

sidvishnoi: Makes sense, but still feels a little odd to me. Maybe I can get used to it :)

tests/spec/core/dfn-spec.js
tidoust: Totally fine to start without it!

src/core/dfn.js
dontcallmedom: I've included that in my latest commit

marcoscaceres (approved):
I think this is good as a start. 

We might need to deal with edge cases like:
```HTML
<p class="definition">
  <dfn>a</dfn> ... <dfn>b</dfn>
</p>
```

src/core/dfn.js
```diff
@@ -238,3 +243,35 @@ function addContractDefaults() {
...
   }
 }
+
+// - Sets data-defines on well-known definition content patterns
+function addDefinitionPointers() {
+  // A dl with class hasdefinitions associated the dfn in each dt
+  // the definition in the following sibling element
+  /** @type NodeListOf<HTMLElement> */
+  const describedDTs = document.querySelectorAll(
+    "dl.hasdefinitions dt:has(dfn[data-dfn-type])"
```
marcoscaceres, line 253:
```suggestion
    "dl.definitions dt:has(dfn[data-dfn-type])"
```

tests/spec/core/dfn-spec.js
```diff
@@ -655,5 +655,26 @@ describe("Core — Definitions", () => {
...
     });
+
+    it("assigns data-defines on well-known patterns", async () => {
+      const body = `
+        <section>
+          <h2>Definitions</h2>
+          <p id="desc1" class="definition">
+            A <dfn>definition</dfn> can also have an associated description
+          </p>
+          <dl class="hasdefinitions">
```
marcoscaceres, line 666:
```suggestion
          <dl class="definitions">
```

src/core/dfn.js
marcoscaceres: Wasn't loving hasdefinitions either tbh. I updated the PR to use "definitions".

src/core/dfn.js
marcoscaceres: Right, and that might be ok. But let's settle on the convention for:

* container (p, span) has .definition
* and multiple: dl.definitions

And explicitly marked as such.

dontcallmedom:
is there anything blocking this or should I go ahead and merge it at this point?

sidvishnoi:
OK by me. @marcoscaceres?
Let me know I'll merge and release.

sidvishnoi:
Going ahead with the merge given Marcos approved already.

PR #4623, opened 2023-12-28
refactor(w3c/level): improve method docs

marcoscaceres:


sidvishnoi:
I'm not finding the benefit of this particular class of refactor. How does typescript help here?

marcoscaceres:
The `LocalizationStrings` helps remind us to add all the appropriate languages. So, although it's not super useful here, it's going to be useful in other places where we have localization language gaps.

src/w3c/level.js
```diff
@@ -28,6 +30,11 @@ const localizationStrings = {
   },
 };
 
+/**
+ * Localization data based on the user's language preference.
+ *
+ * @type {{ level: string }}
```
sidvishnoi, line 36: This type cast removes benefits of TS here I think.

sidvishnoi (changes requested):
I'm in favor of grouping all these PRs into bigger chunks.

Nits to be followed:
1. Make this comment appear in one line.
   ```diff
   - /**
   - * @type {LocalizationStrings}
   - */
   + /** @type {LocalizationStrings} */
   ```
2. Remove this "Localization data based on the user's language preference." comment as it's noisy.

src/w3c/level.js
```diff
@@ -9,7 +9,9 @@
 
 import { getIntlData, showError } from "../core/utils.js";
 export const name = "w3c/level";
-
+/**
+ * @type {LocalizationStrings}
+ */
```
marcoscaceres, line 14:
```suggestion
/** @type {LocalizationStrings} */
```

sidvishnoi (approved):

PR #4625, opened 2023-12-28
refactor(w3c/abstract): docs and l10n improvements

marcoscaceres:


src/w3c/abstract.js
```diff
@@ -1,58 +1,69 @@
...
+import { lang as docLang } from "../core/l10n.js";
 import { html } from "../core/import-maps.js";
+
 export const name = "w3c/abstract";
 
+/**
+ * Localization strings for different languages.
+ *
+ * @type {LocalizationStrings}
+ */
```
sidvishnoi, line 20:
```suggestion
/** @type {LocalizationStrings} */
```

src/w3c/abstract.js
```diff
@@ -1,58 +1,69 @@
...
+  nl: { abstract: "Samenvatting" },
+  es: { abstract: "Resumen" },
+  de: { abstract: "Zusammenfassung" },
 };
+
+/**
+ * Localization data based on the user's language preference.
+ *
+ * @type {{ abstract: string }}
+ */
```
sidvishnoi, line 35: This shouldn't be needed.
```suggestion
```

sidvishnoi (changes requested):

PR #4626, opened 2023-12-28
refactor(core/10n): small docs fixes and js improvements

marcoscaceres:


src/core/l10n.js
```diff
@@ -10,7 +10,9 @@
 export const name = "core/l10n";
 
 const html = document.documentElement;
-if (html && !html.hasAttribute("lang")) {
+// Explicitly default lang and dir on <html> if not set.
+// We assume English and ltr as default for international standards.
```
sidvishnoi, line 14: When could `html` be undefined with ReSpec?

marcoscaceres: I know it's come up in super weird edge cases where the document is an iframe or something. But yeah, this is more just defensive than anything else (I'm sure everything will break if this ever happens).

sidvishnoi (approved):

PR #4664, opened 2024-03-06
fix(core/dfn): add contractDefaults, dfnPointers before running linters

sidvishnoi:
Fixes https://github.com/w3c/respec/issues/4323
Closes https://github.com/w3c/respec/pull/4344 (supersedes)

### DIFF
--- a/src/core/dfn-contract.js
+++ b/src/core/dfn-contract.js
@@ -1,0 +1,61 @@
+export const name = "core/dfn-contract";
+
+export function run() {
+  addContractDefaults();
+  addDefinitionPointers();
+}
+
+function addContractDefaults() {
+  // Find all dfns that don't have a type and default them to "dfn".
+  /** @type NodeListOf<HTMLElement> */
+  const dfnsWithNoType = document.querySelectorAll(
+    "dfn:is([data-dfn-type=''],:not([data-dfn-type]))"
+  );
+  for (const dfn of dfnsWithNoType) {
+    dfn.dataset.dfnType = "dfn";
+  }
+
+  // Per "the contract", export all definitions, except where:
+  //  - Explicitly marked with data-noexport.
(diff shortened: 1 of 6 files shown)


src/core/dfn-contract.js
```diff
@@ -0,0 +1,61 @@
...
+  // Find all dfns that don't have a type and default them to "dfn".
+  /** @type NodeListOf<HTMLElement> */
+  const dfnsWithNoType = document.querySelectorAll(
+    "dfn:is([data-dfn-type=''],:not([data-dfn-type]))"
+  );
+  for (const dfn of dfnsWithNoType) {
+    dfn.dataset.dfnType = "dfn";
+  }
+
+  // Per "the contract", export all definitions, except where:
```
marcoscaceres, line 18: This doesn't seem right... we should only export definitions explicitly marked as export, no?

src/core/dfn-contract.js
```diff
@@ -0,0 +1,61 @@
...
+  //  - Explicitly marked with data-noexport.
+  //  - The type is "dfn" and not explicitly marked for export (i.e., just a regular definition).
+  //  - definitions was included via (legacy) data-cite="foo#bar".
+  /** @type NodeListOf<HTMLElement> */
+  const exportableDfns = document.querySelectorAll(
+    "dfn:not([data-noexport], [data-export], [data-dfn-type='dfn'], [data-cite])"
+  );
+  for (const dfn of exportableDfns) {
+    dfn.dataset.export = "";
```
marcoscaceres, line 27: I'm so confused... we were already doing this? I guess if it didn't break or do the wrong things it's ok? 🤷

sidvishnoi: Not sure what you mean, but I just moved this content out from dfn.js so it can run before linters instead of after "plugins-done" event.

marcoscaceres (approved):

PR #4676, opened 2024-04-08
fix(styles/respec.css): defer default styles to W3C

### LINKED ISSUE #4667: table.simple style is not dark-mode friendly
The respec style sheet includes
```css
.simple tr:nth-child(even) {
  background: #f0f6ff;
}
```

when using dark mode, the cell background remains light colored (which isn't great); more problematically, the text in the cell is also very light, making it unreadable.

The rule above should either have a set text color, or better, use variable-defined colors to work with dark mode.

marcoscaceres:
We should just let the W3C's base.css handle styling things on the page. That way we get consistent styling across all docs, and don't need to worry if base.css changes... and we can upstream styling issues to the W3C

Closes #4667

### DIFF
--- a/src/styles/respec.css.js
+++ b/src/styles/respec.css.js
@@ -18,14 +18,6 @@ export default css`
 }
 
 /* --- INLINES --- */
-:is(h1, h2, h3, h4, h5, h6, a) abbr {
-  border: none;
-}
-
-dfn {
-  font-weight: bold;
-}
-
 a.internalDFN {
   color: inherit;
   border-bottom: 1px solid #99c;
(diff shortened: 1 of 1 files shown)


sidvishnoi (approved):

sidvishnoi:
I think it should still be a breaking change, as the `table.simple` design changed? In case people are relying on that to be styled like this..

tfrancart:
@marcoscaceres which additionnal CSS should we include to have our tables properly styled ? we use respec in a non-W3C spec (http://linked-statistics.github.io/xkos/xkos-best-practices.html) and that change broke our document. Thanks

TimvdLippe:
Since we were using these styles in our fork as well, I opened https://github.com/w3c/tr-design/pull/372 to upstream these changes

PR #4680, opened 2024-04-09
chore!: require Node.js v20 instead of v18

sidvishnoi:
Node.js 20 is active LTS since 2023-10-24.

Also replaced some hardcoded node-version values with .nvmrc


sidvishnoi:
As we're going for a next major version, I'll also include https://github.com/w3c/respec-web-services/pull/177 and https://github.com/w3c/respec-web-services/pull/178 into ReSpec

marcoscaceres (approved):

PR #4689, opened 2024-04-12
fix(core/linter/a11y): import axe from jsdelivr instead of unpkg

sidvishnoi:
unpkg is down, and I'm not sure it's maintained. Tests are failing for a11y linter, and we get occasional timeouts. So, moving to "bigger player".

### DIFF
--- a/src/core/linter-rules/a11y.js
+++ b/src/core/linter-rules/a11y.js
@@ -89,7 +89,7 @@ async function getViolations(opts) {
 function importAxe() {
   const script = document.createElement("script");
   script.classList.add("remove");
-  script.src = "https://unpkg.com/axe-core@4/axe.min.js";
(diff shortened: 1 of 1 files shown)


marcoscaceres (approved):

src/core/linter-rules/a11y.js
```diff
@@ -89,7 +89,7 @@ async function getViolations(opts) {
 function importAxe() {
   const script = document.createElement("script");
   script.classList.add("remove");
-  script.src = "https://unpkg.com/axe-core@4/axe.min.js";
+  script.src = "https://cdn.jsdelivr.net/npm/axe-core@4/axe.min.js";
```
marcoscaceres, line 92: Maybe we should host this too

sidvishnoi: Maybe sometime in future

PR #4700, opened 2024-04-24
fix!(w3c/style): override color-scheme with meta tag

### LINKED ISSUE #4687: How to deal with custom styles in dark mode?
## Description of problem

* Code: https://github.com/w3c/bp-i18n-specdev/pull/130
* Preview: https://deploy-preview-130--bp-i18n-specdev.netlify.app/

1. Open https://deploy-preview-130--bp-i18n-specdev.netlify.app/#lang_misc when the system is in light mode.
2. Select 'dark' in the lower left corner of the spec.
3. The colour is incorrect.

If the OS is in dark mode, the colour is correct.

Colours should follow the `@media (prefers-color-scheme: dark)` media query.

marcoscaceres:
Removes the opt in `respecConfig.darkMode` option, allowing HTML's  meta tag to be used instead.

### DIFF
--- a/src/w3c/style.js
+++ b/src/w3c/style.js
@@ -114,30 +114,34 @@ export function run(conf) {
   // Make sure the W3C stylesheet is the last stylesheet, as required by W3C Pub Rules.
   sub("beforesave", styleMover(finalStyleURL));
 
+  // Add color scheme meta tag and style
   /** @type HTMLMetaElement */
   let colorScheme = document.querySelector("head meta[name=color-scheme]");
   if (!colorScheme) {
-    colorScheme = html`<meta name="color-scheme" content="light dark" />`;
+    // Default to light mode during transitional period.
+    colorScheme = html`<meta name="color-scheme" content="light" />`;
     document.head.appendChild(colorScheme);
   }
-  const darkModeStyleUrl = getStyleUrl("dark.css");
   const css = `:root {
     color-scheme: ${colorScheme.content};
   }`;
   document.head.appendChild(
-    html`<style>
+    html`<style id="respec-color-scheme-declaration">
       ${css}
     </style>`
   );
-  document.head.appendChild(
-    html`<link
-      rel="stylesheet"
-      href="${darkModeStyleUrl.href}"
-      media="(prefers-color-scheme: dark)"
-    />`
-  );
-  // As required by W3C Pub Rules.
-  sub("beforesave", styleMover(darkModeStyleUrl));
+  if (colorScheme.content.includes("dark")) {
+    const darkModeStyleUrl = getStyleUrl("dark.css");
+    document.head.appendChild(
+      html`<link
+        rel="stylesheet"
+        href="${darkModeStyleUrl.href}"
+        media="(prefers-color-scheme: dark)"
+      />`
+    );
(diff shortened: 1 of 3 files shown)


src/w3c/style.js
```diff
@@ -114,28 +114,30 @@ export function run(conf) {
...
-    // As required by W3C Pub Rules.
-    sub("beforesave", styleMover(darkModeStyleUrl));
+  /** @type HTMLMetaElement */
+  let colorScheme = document.querySelector("head meta[name=color-scheme]");
+  if (!colorScheme) {
+    colorScheme = html`<meta name="color-scheme" content="light dark" />`;
+    document.head.appendChild(colorScheme);
   }
+  const darkModeStyleUrl = getStyleUrl("dark.css");
+  const css = `:root {
```
marcoscaceres, line 124: If the meta tag sets this, any idea why this is needed, @sidvishnoi ? Just for completeness?

sidvishnoi:
Let's still keep this behind darkMode flag for a while? It was added to prevent specs that use custom styles to opt-in after getting adapted.

marcoscaceres:
I still don’t feel that a configuration flag is the right approach. It should then be we inject the meta to default to “light”, allowing spec authors to then opt into light/dark. 

Then the transitional path is that we announce this, and give everyone a year to check their specs before switching.

marcoscaceres:
Should be good to go now, @sidvishnoi.

src/w3c/style.js
sidvishnoi: I think we can remove this now. It existed before I added that meta tag. Though, check if any browser support differences.

src/w3c/style.js
```diff
@@ -114,18 +114,24 @@ export function run(conf) {
...
+    document.head.appendChild(colorScheme);
+  }
+  const css = `:root {
+    color-scheme: ${colorScheme.content};
+  }`;
+  document.head.appendChild(
+    html`<style id="respec-color-scheme-declaration">
+      ${css}
+    </style>`
+  );
```
marcoscaceres, line 132:
```suggestion
```

tests/spec/w3c/style-spec.js
```diff
@@ -238,6 +250,22 @@ describe("W3C - Style", () => {
...
+    /** @type HTMLStyleElement? */
+    const styleElem = doc.querySelector(
+      "style#respec-color-scheme-declaration"
+    );
+    expect(styleElem).toBeTruthy();
+
+    // Check if the style content includes the correct color-scheme value
+    const actualCSSContent = styleElem?.textContent;
+    expect(actualCSSContent?.includes("color-scheme: dark light")).toBeTruthy();
+  });
```
marcoscaceres, line 267:
```suggestion
```

tests/spec/w3c/style-spec.js
```diff
@@ -238,6 +250,7 @@ describe("W3C - Style", () => {
     expect(linkBase.nextElementSibling).toBe(linkDarkMode);
   });
 
+
```
marcoscaceres, line 253:
```suggestion
```

sidvishnoi (approved):

PR #4709, opened 2024-05-10
fix(core/inlines): support generic nullable type in inline var

### LINKED ISSUE #4708: Generic nullable type not rendered in inline var syntax
## Description of problem

https://github.com/w3c/gamepad/pull/205#pullrequestreview-2050627610

It wasn't recognized as a variable name:
image

If I remove the ? then it renders correctly.

|gamepads| should be styled like a variable and should show the type sequence<Gamepad?> on hover.

nondebug:
Fixes #4708


sidvishnoi (approved):
Unrelated tests failing, should wait for them to pass before merge.

sidvishnoi:
Will fix unrelated tests failing in Firefox separately.

Issue #4718, opened 2024-05-28
Sandboxing issue in Ubuntu 24.04


pchampin:
Since I upgraded to Ubuntu 24.04, the command-line version of respec is not working. I found a workaround, but I need help to propose a more sustainable solution.

## Description of problem

Running `respec` in the command line, I get the following error message:

    [FATAL] Error: Failed to launch the browser process!
    [257756:257756:0523/090825.033795:FATAL:zygote_host_impl_linux.cc(127)] No
    usable sandbox! Update your kernel or see https://chromium.googlesource.com/
    chromium/src/+/main/docs/linux/suid_sandbox_development.md for more information
    on developing with the SUID sandbox. If you want to live dangerously and need an
    immediate workaround, you can try using --no-sandbox.

    TROUBLESHOOTING: https://pptr.dev/troubleshooting

        at ChildProcess.onClose (file:///home/pa/.nvm/versions/node/v18.17.1/lib/node_modules/respec/node_modules/@puppeteer/browsers/lib/esm/launch.js:258:24)
        at ChildProcess.emit (node:events:526:35)
        at ChildProcess._handle.onexit (node:internal/child_process:291:12)

## Workaround

Here is the part of the Ubuntu 24.04 release note that seems to be relevant

https://discourse.ubuntu.com/t/ubuntu-24-04-lts-noble-numbat-release-notes/39890#unprivileged-user-namespace-restrictions-15

The following workaround did the trick

    echo 0 | sudo tee /proc/sys/kernel/apparmor_restrict_unprivileged_userns

but it affects all applications on the system, so that's no ideal.

It seems that if Respec was running chromium with the `--no-sanbox` option, this would solve the problem. (Note that I had a similar issue with an electron base application, and using the `--no-sandbox` option did solve the problem).

sidvishnoi:
@pchampin There's a `--disable-sandbox` option in `respec` CLI that maps to puppeteer `--no-sandbox`. If that works for you, we can perhaps make that the default way forward.

pchampin:
> here's a `--disable-sandbox` option in `respec` CLI

Aaaaahhhh! Any reason for not calling it `--no-sandbox`, then? :wink: 

It would make sense as the error message explicitly suggests this: `you can try using --no-sandbox`.

marcoscaceres:
Can we switch spec-prod to run on macOS temporarily? Running without sandbox sounds scary.

marcoscaceres:
[...]

PR #4722, opened 2024-06-04
feat(CLI): add `--no-sandbox` as alias to `--disable-sandbox`

sidvishnoi:
Closes https://github.com/w3c/respec/issues/4718

`--disable-sandbox` is used enough already (searched GitHub), so adds alias without removing it.

Here's the mapping (making it work with our CLI args processing tool):
| Flag | `disableSandBox` | 
| --- | --- |
| no flags | false |
| `--disable-sandbox` | true |
| `--no-sandbox` | true |

### DIFF
--- a/tools/respec2html.js
+++ b/tools/respec2html.js
@@ -184,7 +184,12 @@ cli
     "Abort if ReSpec generates warnings (or errors).",
     false
   )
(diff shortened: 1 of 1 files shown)


marcoscaceres (approved):
Oh dear...

PR #4726, opened 2024-06-05
fix(w3c/headers): don't crash on save if `.head details` doesn't exist

sidvishnoi:
See https://github.com/w3c/geolocation-api/pull/158#issuecomment-2149219993

### DIFF
--- a/src/w3c/templates/headers.js
+++ b/src/w3c/templates/headers.js
@@ -190,7 +190,7 @@ export default (conf, options) => {
    */
   sub("beforesave", doc => {
     const details = doc.querySelector(".head details");
-    details.open = true;
+    if (details) details.open = true;
   });
   return html`<div class="head">
     ${conf.logos.length


src/w3c/templates/headers.js
```diff
@@ -190,7 +190,7 @@ export default (conf, options) => {
    */
   sub("beforesave", doc => {
     const details = doc.querySelector(".head details");
-    details.open = true;
+    if (details) details.open = true;
```
marcoscaceres, line 193: Would this work too? 
```suggestion
    details?.open = true;
```

marcoscaceres (approved):

src/w3c/templates/headers.js
sidvishnoi: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Invalid_assignment_left-hand_side#using_optional_chaining_as_assignment_target

https://github.com/tc39/proposal-optional-chaining-assignment

marcoscaceres: Right, I had my suspicions

PR #4753, opened 2024-07-10
No special treatment of plain anchors <a id="...">

### LINKED ISSUE #4747: Empty anchor (`<a id="foo></a>`) tag creates superfluous reference entry
## Description of problem

Source showing the problem:

https://github.com/ralfhandl/respec-experiments/blob/c94de45be239521a56f3c35f438c6d9d1689aab7/src/example.html#L94

Superfluous/wrong reference entry "css" is created:

image

Just create the anchor, don't create a buggy reference entry.

1. clone repo
2. check out branch
3. view file in browser

ralfhandl:
Ignore `<a id="...">` in addition to `<a href="...">` and some other anchor variants.

This style of anchors is the typical (only?) way to create a local link target in Markdown documents.

Fixes #4747

### DIFF
--- a/src/core/link-to-dfn.js
+++ b/src/core/link-to-dfn.js
@@ -67,7 +67,7 @@ export async function run(conf) {
 
   /** @type {NodeListOf<HTMLAnchorElement>} */
   const localAnchors = document.querySelectorAll(
-    "a[data-cite=''], a:not([href]):not([data-cite]):not(.logo):not(.externalDFN)"
+    "a[data-cite=''], a:not([href]):not([id]):not([data-cite]):not(.logo):not(.externalDFN)"
   );
   for (const anchor of localAnchors) {
     if (!anchor.dataset?.linkType && anchor.dataset?.xrefType) {


sidvishnoi:
This check is too strong. I think this needs some fixing only in `core/xref`.

ralfhandl:
> I think this needs some fixing only in `core/xref`.

Question is what to check for: anchor elements with an `id` and no content, as anchor elements with content are used for magic cross-referencing?

sidvishnoi:
Is this intended to be used for defining IDs for sharing URL with fragments only? Also, why not using a `<dfn>` there as it looks like a use of dfn there?

If it's needed regardless, we'd need to update `core/link-to-dfn` to add that check to ensure it doesn't go into either of `possibleExternalLinks` or `badLinks`. Then might also need to update [`core/anchor-expander`](https://github.com/speced/respec/blob/0409aa09b262cd828191324cdaace143fb39e420/src/core/anchor-expander.js#L12) so it doesn't try to use other special expansion behavior.

Also, might be ok to use `<a name="foo"></a>` instead of `<a id="foo"></a>`? That might be easier to bypass ReSpec's processing.

ralfhandl:
> might be ok to use `<a name="foo"></a>` instead of `<a id="foo"></a>`

Using `name` runs into the same issues as using `id`, and `name` is deprecated, so not where I'd like to go.

ralfhandl:
What seems to work is rewriting
~~~md
<a name="oasSecurity"></a>security
~~~
into
~~~md
<a id="oasSecurity" class="logo"></a>security
~~~
as anchors with this class are already ignored in https://github.com/speced/respec/blob/0409aa09b262cd828191324cdaace143fb39e420/src/core/link-to-dfn.js#L70

Does this hack have foreseeable negative side-effects?

PR #4786, opened 2024-09-02
Prevent creating links from inline links inside links

dontcallmedom:


src/core/inlines.js
```diff
@@ -145,9 +145,10 @@ function inlineXrefMatches(matched, text) {
   }
 
   const node = idlStringToHtml(ref);
-  // If it's inside a dfn, it should just be coded, not linked.
+  // If it's inside a dfn or a `a`, it should just be coded, not linked.
   // This is because dfn elements are treated as links by ReSpec via role=link.
-  const renderAsCode = !!text.parentElement.closest("dfn");
+  const renderAsCode =
+    !!text.parentElement.closest("dfn") || !!text.parentElement.closest("a");
```
tidoust, line 151: I suggest to merge the selectors:

```suggestion
    !!text.parentElement.closest("dfn,a");
```

tidoust (approved):
One editorial suggestion but that looks good to me!

PR #4788, opened 2024-09-07
fix(core/xref) : preserve the filename in URL

### LINKED ISSUE #4787: Incorrect generated xref link
## Description of problem

The IsolatedContext IDL attribute resolves correctly in bikeshed and [respec.org/xref](https://respec.org/xref/?term=IsolatedContext&types=extended-attribute) after https://github.com/w3c/browser-specs/pull/1467, but it doesn't resolve to the correct URL when generating a spec with respec.

respec.org/xref generates its link by [resolving](https://github.com/speced/respec-web-services/blob/main/static/xref/script.js#L150) the search uri against the spec's url, which does the right thing.

When respec looks up the xref info from respec.org/xref/search it gets:
```
{
    "shortname": "isolated-contexts",
    "spec": "isolated-contexts",
    "type": "extended-attribute",
    "normative": true,
    "uri": "#IsolatedContext"
}
```
It then [resolves](https://github.com/speced/respec/blob/main/src/core/xref.js#L402) `uri` against `https://partial`, and pulls the path and hash out of that, which in this case ends up setting the path to '/'. Later, when generating the anchor's href, the specification URL's path gets [overridden](https://github.com/speced/respec/blob/main/src/core/data-cite.js#L59) because of the '/'.

https://github.com/WICG/direct-sockets/pull/71

`{{IsolatedContext}}` resolved to the wrong link.

1. Add `xref: ["isolated-contexts"],` to respecConfig
2. Include `{{IsolatedContext}}` in a spec
3. Check the href that IsolatedContext resolves to

tidoust:
Fixes #4787.
The URL of the spec may end with a filename but the code assumed that it would always be a path, and thus that resolving `./` against the URL of the spec would be a no-op. The code now makes sure resolve the URL of the spec against an empty string when the xref search does not return a more specific path.

### DIFF
--- a/src/core/xref.js
+++ b/src/core/xref.js
@@ -400,7 +400,10 @@ function addDataCite(elem, query, result, conf) {
   // but sometimes we get lucky and we get an absolute URL from xref
   // which we can then use in other places (e.g., data-cite.js)
   const url = new URL(uri, "https://partial");
-  const { pathname: citePath } = url;
+  let { pathname: citePath } = url;
+  // final resolution will be against the URL of the spec, which may end with
+  // a filename. That filename must be preserved if there's no specific path.
+  if (citePath === "/") citePath = "";
   const citeFrag = url.hash.slice(1);
   const dataset = { cite, citePath, citeFrag, type };
   if (forContext) dataset.linkFor = forContext[0];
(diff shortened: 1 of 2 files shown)


sidvishnoi (approved):
Thanks @tidoust!

PR #4921, opened 2025-03-27
fix(*/conformance): correct grammar when single RFC2119 term used

daniel-montalvo:


src/dini/conformance.js
```diff
@@ -20,7 +20,7 @@ const localizationStrings = {
         ${plural ? "are" : "is"} to be interpreted as described in
         <a href="https://www.rfc-editor.org/info/bcp14">BCP 14</a>
         ${renderInlineCitation("RFC2119")} ${renderInlineCitation("RFC8174")}
-        when, and only when, they appear in all capitals, as shown here.
+        when, and only when, it appears in all capitals, as shown here.
```
sidvishnoi, line 23: Still need to handle plurals case (in all profiles).
```suggestion
        when, and only when, ${plural ? "they appear" : "it appears"} in all capitals, as shown here.
```

builds/respec-aom.js
sidvishnoi, line 1: Please remove all the files from builds directly from PR. They're generated automatically on release.

sidvishnoi (changes requested):
Good find, @daniel-montalvo! Need two small changes though.

src/dini/conformance.js
sidvishnoi: Don't forget to run `pnpm lint --fix` after these changes!

sidvishnoi (approved):

Issue #4871, opened 2025-01-09
color contrast issues


pkra:
While working on https://github.com/w3c/aria/pull/2410 I found some contrast issues that can also be observed on https://respec.org/docs/.

- TOC
  - light & dark mode: "Tables of Content" has insufficient contrast (due to `var(--tocsidebar-heading-text)`)
  - dark mode: `.toc code {#005a9c}`  leads to insufficient contrast
- highlight.js: in dark mode, the color theme causes contrast issues in most code snippets, in particular html.
- WebIDL:  in light mode, "WebIDL" has insufficient contrast (`.idlHeader { color: var(--defrow-border,#fff)}`)

I'll update this issue if I find more.

pkra:
Thanks @sidvishnoi. 

Is there a way to configure the highlight.js theme in the respec configuration? It offers some themes with better contrast.

SamB:
Hmm. The dark mode PR (<https://github.com/speced/respec/pull/3236>) said this:

> This gives a "good enough" support for dark mode. Not all styles are perfected yet.
> And some styles can't be overridden by the manual theme toggle (they only respect `@media (prefers-color-scheme: dark) {...}`).
> I've tried to use W3C stylesheets's CSS variables wherever possible to avoid this, but at least styles for code syntax highlighting (including WebIDL), and definition panel aren't supported in that theme, so they won't fully respect the manual theme toggle.

Apparently relating to the following blocks of code:

https://github.com/speced/respec/blob/19b2c0e260e9bbaf2a8a8b8149bb648c3f02c6b6/src/styles/highlight.css.js#L27-L43

https://github.com/speced/respec/blob/19b2c0e260e9bbaf2a8a8b8149bb648c3f02c6b6/src/styles/algorithms.css.js#L24-L31

It seems that the issue is that https://github.com/w3c/tr-design/blob/gh-pages/src/fixup.js can disable W3C's dark-mode stylesheet but nobody else's.

<https://github.com/speced/respec/issues/4687> is basically about how to handle it here, <https://github.com/w3c/tr-design/issues/349> was about having fixup.js set a global class, but that (afaik) won't ultimately work for viewing a pre-processed spec with JavaScript turned off. I filed https://github.com/w3c/tr-design/issues/363 requesting an explanation of how to use the feature.

pkra:
https://github.com/w3c/tr-design/pull/364 will fix some of this.

TimvdLippe:
[...]

PR #4925, opened 2025-04-01
Support dark mode toggle for SVGs and highlight.js

niklasl:
Syntax highlighting already works with automatic preferences, but has not played well with neither default (no dark mode) settings, nor worked with the manual scheme toggle. (See also #4871.)

This fixes that using CSS (so that default behavior works without JS).

SVG behavior is by default based on user-agent preferences, and has therefore also not played well with the manual toggle. With this change, that is now handled by setting `color-scheme: light` or `dark` on the body element, based on the sidebar-provided manual scheme toggle.

If no meta element indicating dark color-scheme support is present, light mode is set on the body. This instructs SVGs to render in light mode when embedded in documents with no dark mode support, even when the user agent indicates dark mode. (The `color-scheme` CSS property can of course still be set to other values for various sections in a document, if so desired.)

A darkmode example document is also added, containing syntax-highlighted code and embedding an SVG document supporting light/dark color-scheme preferences.


sidvishnoi:
Sorry for the delay in review. I've some concerns:
- `:has` is a relatively new selector (although supported since 2023-ish), and for the generated docs, we try to use well supported features
- Relying on that selector seems bit brittle as there's no guarantee if that'll be changed.
- The which scheme is being used (and toggle) work is something that should be done in tr-design repo (I saw your comment in relevant issues there)

I don't feel too comfortable with the approach here of `:has` selector on a `input[type=checkbox]`.

Regardless, I'd suggest removing the `examples/darkmode.html`, and add `examples/figure1-lightdark.svg` to some existing example.

marcoscaceres (approved):
Thanks for your patience on this, @niklasl. Over a year waiting for a review is not OK, and I'm sorry about that.

Reviewing this fresh, the `:has()` concern from the earlier review is no longer relevant: `:has()` has been baseline since December 2023 and is supported in all browsers we target.

The concern about relying on fixup.js's `input[name='color-scheme']` structure is valid in principle, but the alternative is doing nothing, which means dark mode remains broken for syntax highlighting and linked SVGs. That's not acceptable. Additionally, the W3C team coordinates closely with us on tr-design changes, so we shouldn't be overly concerned about them changing fixup.js's structure without checking with us first.

The code is clean, well-commented, and the CSS variable deduplication is a nice improvement on its own. The example file is helpful for testing.

Approving this. Thank you for contributing.

niklasl:
@marcoscaceres Thank you! I'm glad this was possible to move forward with.

PR #4926, opened 2025-04-02
Report all `console.error` arguments

TimvdLippe:
While attempting to debug an issue with the `respec-mermaid` plugin, we discovered that the plugin logs multiple messages in the same `console.error` call:

```
console.error('respec-mermaid error: Failed to generate figure.',
  e, mermaidSource);
```

Unfortunately only the first message would appear in the console logs when running Respec. That's because it would get the `[0]` index of the args array and would report that as text.

Instead of only reporting the first argument, instead loop over all argument text parts and report them individually on the console. We can then use this to figure out why our mermaid figures are not correctly rendering on CI (they are fine on local).

### DIFF
--- a/tools/respecDocWriter.js
+++ b/tools/respecDocWriter.js
@@ -338,7 +338,7 @@ function handleConsoleMessages(page, onError, onWarning) {
   page.on("console", async message => {
     const args = await Promise.all(message.args().map(stringifyJSHandle));
     const msgText = message.text();
-    const text = args.filter(msg => msg !== "undefined")[0] || "";
+    const text = args.filter(msg => msg !== "undefined") || [{}];
     const type = message.type();
     if (
       (type === "error" || type === "warning" || type === "warn") &&
@@ -351,12 +351,16 @@ function handleConsoleMessages(page, onError, onWarning) {
       // https://github.com/GoogleChrome/puppeteer/issues/1939
       return;
     }
-    switch (type) {
-      case "error":
-        return onError(JSON.parse(text));
-      case "warn":
-      case "warning":
-        return onWarning(JSON.parse(text));
+    for (const textParts of text) {
+      switch (type) {
+        case "error":
+          onError(JSON.parse(textParts));
+          break;
+        case "warn":
+        case "warning":
+          onWarning(JSON.parse(textParts));
+          break;
+      }
     }
   });
 }


TimvdLippe:
This is the plugin implementation in question: https://github.com/w3c/respec-mermaid/blob/e400850c90eddaaa6f11528dc476f85af92cb09a/index.js#L48-L49

sidvishnoi:
@TimvdLippe In this particular case, I think it might be better to update the mermaid plugin to make use of [ReSpec's error/warning logging mechanism](https://github.com/speced/respec/wiki/extension-utils-object).

If that is not enough, we can look into this PR. I'm concerned we may end of logging details we don't want to (like logs from other scripts or from browser).

TimvdLippe:
TIL about that utils object. I opened https://github.com/w3c/respec-mermaid/pull/10 and hopefully this gets published into a new version of the `respec-mermaid` plugin that we can use.

TimvdLippe:
Unfortunately with the update of the respec-mermaid plugin (which any case is a good one), the log output is still not verbose enough to understand what's going on:

```
[INFO] [Timeout: 7823ms] Processing ReSpec document...
Error:  Failed to generate figure
   Count: 1
  Plugin: core/post-process/createFigures
[INFO] [Timeout: 7560ms] Processed document.
```

It doesn't seem to want to print the stacktrace, even though it has a `cause`. I am considering updating the logic here to allow that when verbose logging. And also to print the elements to give some more information.

TimvdLippe:
We decided to drop the respec-mermaid plugin, since it doesn't work with processed Markdown. That's because Mermaid uses `>` in graph notation to denote relations. But in Markdown, if these characters are in a `<code>` block, they are escaped to `&gt;`. Therefore, this will never work if the graphs themselves are part of Markdown files that Respec needs to process. Thus, we are generating the `.svg` files beforehand and include those, avoiding the dynamic generation.

sidvishnoi:
Suggestion: you can consider using data-include with type text to import mermaid files containing `<>`, and then use the mermaid plugin as `postProcess`. This does mean you'll have to add move mermaid diagrams to separate text files.

TimvdLippe:
I did try that before and unfortunately it still escapes.

Tested it with the following:

```html
<section data-include="../diagrammen/parkeervergunning-inzien.mermaid" data-include-format="text"></section>
```

This still escapes the text:

```
    rect lavender
    B-&gt;&gt;+MO: tonenVergunningenVraag
    MO-&gt;&gt;V: opvragenVergunningenVraag
    V--&gt;&gt;MO: opvragenVergunningenAntwoord
    V-&gt;&gt;L2: Log gegevensverwerking (opvragenVergunningen)
    MO--&gt;&gt;B: tonenVergunningenAntwoord
    MO-&gt;&gt;L1: Log gegevensverwerking (tonenVergunningen)
    end
```

It seems like the escaping happens regardless of the `data-include-format`.

PR #4958, opened 2025-05-13
fix(core/inlines): don't process text in inline script, style

sidvishnoi:
Fixes https://github.com/speced/respec/issues/4957


sidvishnoi:
@xfq can you confirm by using this script (temporarily):
`https://deploy-preview-4958--respec-pr.netlify.app/respec-w3c.js`

PR #4962, opened 2025-05-16
fix(core/pre-process, core/post-process): Run async functions in order

### LINKED ISSUE #4961: preProcess and postProcess do not execute "in order" as documented for async functions
## Description of problem

Given the following `preProcess` and `postProcess` examples within a respec config:

```js
var respecConfig = {
  // ...
  preProcess: [
    async () => {
      await new Promise((resolve) => setTimeout(resolve, 1000));
      console.log("pre async");
    },
    () =>
      new Promise((resolve) => setTimeout(resolve, 1000)).then(() =>
        console.log("pre promise")
      ),
    () => {
      console.log("pre sync");
    },
  ],
  postProcess: [
    async () => {
      await new Promise((resolve) => setTimeout(resolve, 1000));
      console.log("post async");
    },
    () =>
      new Promise((resolve) => setTimeout(resolve, 1000)).then(() =>
        console.log("post promise")
      ),
    () => {
      console.log("post sync");
    },
  ]
```

The above config defines 3 functions for each of `preProcess` and `postProcess`:

- One defined as an async function, which resolves after a 1-second delay
- One defined as a function that returns a promise, which resolves after a 1-second delay (the same behavior as the first function, just defined differently)
- One defined as a function that returns immediately

## Expected Behavior

Based on the documentation (which states functions run in order), I expect this:

```
pre async
pre promise
pre sync
post async
post promise
post sync
```

## Actual Behavior

What actually happens is this:

```
14:18:39.971 pre sync
14:18:40.979 pre async
14:18:40.980 pre promise
14:18:41.119 post sync
14:18:42.119 post async
14:18:42.120 post promise
```

(Timestamps are included above to demonstrate that the async functions finish around a second after the sync function starts, indicating they start around the same time, given the async functions' 1-second wait time.)

## Root Cause

This occurs because the implementations for both properties use `Promise.all` on a mapped array which has already immediately invoked each function up-front. To preserve order, these functions' invocations would need to be sequentially chained instead.

**Aside:** It is worth noting that the documentation itself does not seem to explicitly indicate that `preProcess` and `postProcess` support promises; it seems it was never updated when that functionality was added.

---

I would be interested in working on a PR to fix this if one would be accepted. (Edit: I initially had problems getting tests to run, but forcing karma to run with `--browsers Firefox` helped in my case, since I don't have Chrome installed on this machine.)

kfranqueiro:
Fixes #4961.

- Updates pre/post-process tests to include verification of invocation order
  - Adds a counter shared between invocations that is incremented and output for each one
  - Changes existing `toBe` expectations to `toContain` for pass/fail without order
  - Adds new test with precise `toBe` expectations including order
- Revises code flow of pre- and post-process to use `for...of` with an `await` on each iteration, instead of `map` that invokes every function up-front followed by `Promise.all`
  - Note: the changes to the plugins are easier to read with whitespace ignored (`?w=1`).

The new test fails when run without the fix commit.

**Question:** if this gets merged, would it also be acceptable for me to make an attempt at updating the preProcess and postProcess docs to mention that they support promises?


marcoscaceres:
@sidvishnoi, I don't think we should block on the timing issue... we should figure out some other way of dealing with it as issues come up.

PR #5003, opened 2025-07-11
Patch to address Mermaid issue 1766

KDean-Dolphin:
Mermaid diagrams that include line breaks in their labels break the ReSpec workflow when validator.nu is called. This is because the validator expects pure XML in the SVG but line breaks are rendered with unclosed \<br\> tags. See https://github.com/mermaid-js/mermaid/issues/1766.

Replacing HTML \<br\> with XHTML \<br/\> is essentially impossible until the very end; there are multiple points in Mermaid and ReSpec Mermaid that roll back the replacement. For example, [line 43](https://github.com/w3c/respec-mermaid/blob/8b2eecf7ebe415bbf96fa7c4cb1775f5bd2e43d1/index.js#L43) of index.js in ReSpec Mermaid is as follows:

```javascript
template.innerHTML = cleanedSvg;
```

The assignment hands control over to the DOM, which promptly reverts to \<br\>, even though the content is XHTML.

This patch takes the final HTML generated by ReSpec, searches for all \<svg\> tags, and replaces the \<br\> tags within them with \<br /\>.

The end result is an XML-conformant SVG.

### DIFF
--- a/tools/respec2html.js
+++ b/tools/respec2html.js
@@ -287,7 +287,37 @@ async function run(source, destination, options, log) {
     );
   }
 
-  await write(destination, html);
+  // Patch to address Mermaid issue https://github.com/mermaid-js/mermaid/issues/1766.
+  let reduceHTML = html;
+  let svgSafeHTML = "";
+
+  // Keep going until no more SVG tags.
+  while (reduceHTML !== "") {
+    const svgStart = reduceHTML.search(/<svg( [\S\s]*?|)>/);
+    if (svgStart !== -1) {
+      svgSafeHTML += reduceHTML.slice(0, svgStart);
+      reduceHTML = reduceHTML.slice(svgStart);
+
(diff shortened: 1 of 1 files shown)


sidvishnoi:
I wonder if we can use `DOMParser` along with `XMLSerializer` instead of custom string manipulation.. I think we can get these out through puppeteer.

KDean-Dolphin:
DOMParser is one of many culprits in the processing chain that, when given `<br />`, rewrites it as `<br>`.

KDean-Dolphin:
@sidvishnoi It turns out not to be a bug after all. See my comment on the Mermaid issue [here](https://github.com/mermaid-js/mermaid/issues/1766#issuecomment-2706655915).

I've been through multiple layers of the process, and every layer rewrites the `<br/>` as `<br>`, courtesy of the JavaScript DOM engine. Everywhere that I've inserted `<br/>` gets undone. Accordingly, I'm closing this.

PR #5014, opened 2025-07-22
fix(core/dfn): prevent adding `data-local-lt` to `data-lt`

### LINKED ISSUE #5007: Local text defined in `data-local-lt` incorrectly copied to `data-lt`
## Description of problem

https://w3c.github.io/rdf-concepts/spec/#dfn-rdf-string

Definition defines an additional linking text aimed at local usage in a `data-local-lt` attribute. Source is:

```html
<dfn id="dfn-rdf-string" class="export" data-local-lt="string">RDF string</dfn>
```

The additional linking text ends up appearing in a `data-lt` attribute as well, defeating the purpose of using `data-local-lt` in the first place. Generated HTML:

```html
<dfn id="dfn-rdf-string" class="export" data-local-lt="string"
  data-export="" data-lt="RDF string|string" data-plurals="rdf strings|strings"
  data-dfn-type="dfn">
    RDF string
</dfn>
```

The local linking text should not be copied to `data-lt`.

See code example above!

dontcallmedom:
Closes #5007

### DIFF
--- a/src/core/dfn.js
+++ b/src/core/dfn.js
@@ -72,7 +72,7 @@ export function run() {
     computeExport(dfn);
 
     // Only add `lt`s that are different from the text content and local-lts
-    const localLt = (dfn.dataset.localLt || "").split("|");
+    const localLt = (dfn.dataset.localLt || "").split("|").map(norm);
     const lt = titles.filter(t => !localLt.includes(t));
     if (lt.length > 1 || linkingText !== norm(dfn.textContent)) {
       dfn.dataset.lt = lt.join("|");


src/core/dfn.js
```diff
@@ -71,11 +71,11 @@ export function run() {
     computeType(dfn, linkingText);
     computeExport(dfn);
 
-    // Only add `lt`s that are different from the text content
-    if (titles.length === 1 && linkingText === norm(dfn.textContent)) {
-      continue;
+    // Only add `lt`s that are different from the text content and local-lts
+    const lt = titles.filter(t => !((dfn.dataset.localLt || "").split("|").includes(t)));
```
sidvishnoi, line 75: Might be a good idea to add a `const localLt` to avoid splitting on each `filter` callback.

dontcallmedom: done in https://github.com/speced/respec/pull/5014/commits/ae73f00da97d1c5dd11a7637ed7937078274c320

src/core/dfn.js
```diff
@@ -71,11 +71,12 @@ export function run() {
     computeType(dfn, linkingText);
     computeExport(dfn);
 
-    // Only add `lt`s that are different from the text content
-    if (titles.length === 1 && linkingText === norm(dfn.textContent)) {
-      continue;
+    // Only add `lt`s that are different from the text content and local-lts
+    const localLt = (dfn.dataset.localLt || "").split("|");
```
tidoust, line 75: `getDfnTitles` calls `norm` on individual entries to trim spaces. Since you're going to compare strings, that seems needed as well?

 ```suggestion
    const localLt = (dfn.dataset.localLt || "").split("|").map(norm);
```

sidvishnoi:
(Ignore failing tests about History, some rate limiting issue likely)

PR #5028, opened 2025-08-06
Add flattenRefs feature

### LINKED ISSUE #5027: Documents with no normative/informative split on references
There are a great many documents that don't have any normative force.  We've been using ReSpec for TAG findings for example.  In these documents, references always fall into the "informative" category, with no normative references at all.

The current arrangement adds a useless extra heading in the "References" appendix.

I'd like an option to disable the normative/informative grouping.  

This might warn if both are in use.

I've found that it is possible to disable the heading with some CSS:

```css
#toc a.tocxref[href="#references"] + ol.toc,
#informative-references > .header-wrapper {
  display: none;
}
```

Obviously, this is somewhat brittle.  It would be better if those elements simply did not exist.

I'm willing to contribute the code.

martinthomson:
This adds a boolean configuration option to disable the separation of references into Normative and Informative sections.

This option, when enabled, warns if references are sorted into these groupings.

Closes #5027.

(This includes a rebuild of files in the `builds/` directory.  I'm not sure what the expectations are there.  The developer guide wasn't clear.  I wasn't able to find where the requisite documentation was either, so I'd appreciate a pointer to that.)


marcoscaceres (changes requested):
Blocking based on needing more info. But this should already work without needing this change.

PR #5085, opened 2025-11-05
fix(w3c/headers): allow forcing `/TR` namespace for TAG/AB

ylafon:
Fixes https://github.com/speced/respec/issues/5084
The solution rely on using canonicalURI = "TR" in the ReSpec config.

### DIFF
--- a/src/w3c/headers.js
+++ b/src/w3c/headers.js
@@ -667,9 +667,10 @@ function derivePubSpace(conf) {
 
   if (
     (conf.group === "tag" || conf.group === "ab") &&
-    conf.canonicalURI === "TR") {
-      return `/TR`
-    }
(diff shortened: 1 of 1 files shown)


src/w3c/headers.js
```diff
@@ -665,6 +665,12 @@ function derivePubSpace(conf) {
       return `/Submission`;
   }
 
+  if (conf.groupType === "other" &&
+    (conf.group === "tag" || conf.group === "ab") &&
+    conf.canonicalURI === "TR") {
```
sidvishnoi: Can we require `conf.canonicalURI === "TR"` (at line 653) for all those who want to opt-in, not just TAG/AB (skipping group/groupType checks here)?

PR #5105, opened 2026-02-26
feat(core/rs-changelog): let `repo` attr override `github` config

daniel-montalvo:
@sidvishnoi 

This is one more that the ARIA monorepo setup would greatly benefit from.

It enables a new `repo` parameter  to be used with the `rs-changelog` custom element. This is for monorepos which still keep child repos in the respecConfig (mainly for issue base, editor draft, and other generated properties), but are using a single repo to generate all specs. For child specs we need to specify that they want to pull the commit history from the monorepo instead of from the child repo.

### DIFF
--- a/tests/spec/core/custom-elements/rs-changelog-spec.js
+++ b/tests/spec/core/custom-elements/rs-changelog-spec.js
@@ -55,11 +55,13 @@ describe("Core - Custom Elements - <rs-changelog>", () => {
     const doc = await makeRSDoc(ops);
 
     const commits = doc.querySelectorAll("rs-changelog li");
-    expect(commits).toHaveSize(8);
+    expect(commits).toHaveSize(2);
     const firstCommit = commits[0];
     // Verify links point to the custom repo, not the config repo
     const links = firstCommit.querySelectorAll("a");
-    expect(links).toHaveSize(1);
-    expect(links[0].href).toBe("https://github.com/org/corp/commit/80aad0a");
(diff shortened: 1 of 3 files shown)


sidvishnoi (commented):
Looks good, left a small suggestion.

tests/spec/core/custom-elements/rs-changelog-spec.js
```diff
@@ -48,4 +48,18 @@ describe("Core - Custom Elements - <rs-changelog>", () => {
...
+    const ops = makeStandardOps(conf, body);
+    const doc = await makeRSDoc(ops);
+
+    const commits = doc.querySelectorAll("rs-changelog li");
+    expect(commits).toHaveSize(8);
+    const firstCommit = commits[0];
+    // Verify links point to the custom repo, not the config repo
+    const links = firstCommit.querySelectorAll("a");
+    expect(links).toHaveSize(1);
+    expect(links[0].href).toBe("https://github.com/org/repo/commit/80aad0a");
```
sidvishnoi:
```suggestion
    expect(links[0].href).toBe("https://github.com/org/corp/commit/80aad0a");
```

I think we will need to update `toHTML` as well, as the listed commits should point to repo provided by `rs-changelog`'s `repo` attribute.

sidvishnoi (approved):
Thanks @daniel-montalvo!
Edit: Released in [v35.8.0](https://github.com/speced/respec/releases/tag/v35.8.0).

PR #5124, opened 2026-03-26
feat: add possibility to amend configuration in a plugin

TimvdLippe:
We have a `preProcess` plugin that needs to store some information for later use when a document is rendered. Currently, it is not possible to update the `initialUserConfig` config object, since any update to the `config` object is ignored. These updates in plugins are visible to other plugins, but not to the rendered document.

There is a way to update the object with `"amend-user-config"`, but that requires publishing to the internal system. That system isn't accessible to a plugin.

Therefore, add a new method that can be called with an `Object` that forwards it to the internal publication system. That way, a plugin can update configuration (like some of the core plugins do) and update the `initialUserConfig`.


marcoscaceres (commented):
The use case is legitimate and the implementation is correct — `pub("amend-user-config", ...)` is exactly what core plugins use (e.g. `caniuse.js`), and routing it through `makePluginUtils` is the right pattern.

A few things worth addressing:

**Type is too wide**

```js
/** @type {(configUpdates: Object) => void} */
```

`Object` (uppercase) is the constructor type and accepts `null` in some contexts. The other entries in `makePluginUtils` use more specific types. Suggest:

```js
/** @type {(configUpdates: Record<string, unknown>) => void} */
```

**Missing test**

There's no test verifying that calling `utils.amendConfiguration({ key: "value" })` from a `preProcess` function results in that key appearing in `#initialUserConfig`. The existing `include-config-spec.js` tests would be the right place to add one.

**Missing docs**

`preProcess` and `postProcess` docs mention the `utils` argument but don't document what's on it. Worth adding `amendConfiguration` to those pages (and the wiki) so authors know it exists.

The type nit and docs are minor — happy to see this land once there's a test.

marcoscaceres:
Lol, sorry! the first comment was right. Deleted the second.

marcoscaceres (approved):

marcoscaceres (approved):

marcoscaceres:
Gah, sorry, one more time (I needed to land a hotfix before). I think there’s also a checkbox somewhere in GitHub that allows me also to do the updates.

TimvdLippe:
https://github.com/orgs/community/discussions/5634

> Currently, if a PR is created from a fork living under an organisation (rather than an individual) it's not possible to allow edits by maintainers.

That's the case with ours, since I created it from the Logius-Standaarden fork :/

marcoscaceres:
All good... merged to avoid the rebase ping pong.

PR #5138, opened 2026-03-28
fix(core/github): file-a-bug → new issue chooser form

### LINKED ISSUE #3457: Link target of 'File a Bug' is not new issue but a list of issues
Background: Timed-Text WG is updating our spec building system (still using old XML/XSLT based one) to match with Process 2020 -ish panel, and we started building XSLT using texts copied from respec like parts of 'Participate:' etc. This point was pointed during that work. 

Current respec output points `<repository>/issues/` for 'File a bug' link, which will navigate you to a list of issues. 
Bikeshed outputs two links with a text `File an issue (open issues)` which will navigate you to both new issue creation page and a list of issues.
I'm personally not sure which is better, but I'd agree that current output as a set of `File a bug` and link to a list of issues is mismatching.

marcoscaceres:
### `src/core/github.js` — Fix #3457 (High priority since 2021)

The "File an issue" link in spec headers was pointing to `/issues` (the list) instead of `/issues/new/choose` (the new issue form). `newIssuesURL` was already computed at line 215 but not used for this link. Also extracted the URL construction to a shared variable to avoid duplication.

### `README.md`

- Primary user documentation links (User's guide, WebIDL guide) now point to `respec.org/docs` instead of raw wiki URLs
- `specref.org` link upgraded to HTTPS


marcoscaceres:
@copilot please rebase

marcoscaceres:
@copilot can you rebase and remove all the changes from the build directory. Those shouldn’t be in this PR. Can you lint and prettier?

sidvishnoi (approved):

PR #5149, opened 2026-03-29
fix(w3c/sotd): match CSS selectors used in base.css

daniel-montalvo:
These updates the selector for better matching of the CSS classes used in base.css as per [Sample spec guidance on handling amendments](https://w3c.github.io/tr-design/src/README#amendment)


marcoscaceres (changes requested):
Thanks for catching this — you're right that the hyphenated classes don't exist in `base.css`, so those selectors would never match. Good find.

A few suggestions inline to match the class order convention in `base.css` (which uses `.addition.proposed` and `.correction.proposed`, base class first).

Would you also be willing to add a test for this behaviour? The relevant file is `tests/spec/w3c/headers-spec.js` — there is currently no test coverage for the updateable-rec selectors at all. A test could verify that when a document has elements with `.addition.proposed` and `.correction.proposed` classes, the correct SOTD paragraph is rendered. That would prevent a silent regression in a future refactor.

One broader note: the updatable REC process these classes support has been used (we used it in Geolocation) but proved very painful in practice — the tooling and editor burden are significant. Discussion on simplifying or rethinking it has stalled somewhat, but the issues are worth being aware of: [w3c/process#866](https://github.com/w3c/process/issues/866) and [w3c/process#589](https://github.com/w3c/process/issues/589).

Thanks again for the contribution.

marcoscaceres (approved):

PR #5168, opened 2026-04-08
Add configurable MCP support for ReSpec report workflows

danielcamposramos:
## Summary

This PR adds a configurable stdio MCP companion for ReSpec so AI tools and editors can scaffold, preflight, validate, and build reports from repo-local profiles.

## What Changed

- add `respec-mcp` stdio entrypoints and MCP core/server modules
- add repo-local profile/config discovery via `respec-mcp.config.json` and `respec-mcp/profiles/*.json`
- add Docker runtime for Node 24 + Chromium environments
- add focused MCP tests
- add LLM authoring guidance for W3C and Community Group report workflows, including standards-transition patterns and examples
- expose the authoring guidance path in MCP responses so clients can discover and apply it

## Why

The goal is to let Community Groups and similar spec repositories keep local policy and templates while reusing one upstream MCP runtime. The added authoring guidance is meant to make the MCP useful for real standards-oriented writing and review, not only for rendering HTML.

## Validation

- `docker build -t respec-mcp:local .`
- `docker run --rm --entrypoint node respec-mcp:local /app/node_modules/jasmine/bin/jasmine.js --random=false /app/tests/mcp.cjs`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,39 @@
+FROM node:24-bookworm-slim
+
+ENV PUPPETEER_CACHE_DIR=/root/.cache/puppeteer
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+  ca-certificates \
+  fonts-liberation \
+  libasound2 \
+  libatk-bridge2.0-0 \
+  libatk1.0-0 \
+  libcups2 \
+  libdbus-1-3 \
+  libdrm2 \
+  libgbm1 \
+  libglib2.0-0 \
+  libgtk-3-0 \
+  libnspr4 \
+  libnss3 \
+  libu2f-udev \
+  libx11-6 \
+  libx11-xcb1 \
+  libxcb1 \
+  libxcomposite1 \
+  libxdamage1 \
+  libxext6 \
+  libxfixes3 \
+  libxkbcommon0 \
+  libxrandr2 \
+  xdg-utils \
+  && rm -rf /var/lib/apt/lists/*
+
+WORKDIR /app
+
+COPY . .
+
+RUN corepack enable \
+  && pnpm install --frozen-lockfile
+
+ENTRYPOINT ["node", "tools/respec-mcp.js", "--repo-root", "/workspace"]
--- a/README.md
+++ b/README.md
@@ -38,6 +38,30 @@ if you think something is broken, [file a bug](https://github.com/speced/respec/
 problems encountered with ReSpec, or on why it may be failing to produce pubrules-compliant
 content.
 
+## MCP server
+
+ReSpec now ships a stdio MCP companion entrypoint, `respec-mcp`, for AI-assisted
+spec authoring and validation. The MCP server discovers repo-local configuration from
+`respec-mcp.config.json` and `respec-mcp/profiles/*.json`, so Community Groups and
+other spec repos can add local policy without forking ReSpec.
+
+Use:
+
+```bash
+respec-mcp --repo-root /path/to/spec-repo
+```
+
+Or via Docker:
+
+```bash
+docker build -t respec-mcp:local .
+docker run --rm -i -v "$PWD":/workspace -w /workspace respec-mcp:local
+```
+
+See [docs/MCP.md](docs/MCP.md) for the config contract, tool list, and AI client examples.
+For LLM-assisted authoring guidance, including W3C/Community Group report patterns,
+see [docs/MCP_LLM_AUTHORING_GUIDE.md](docs/MCP_LLM_AUTHORING_GUIDE.md).
+
 ## Bibliographical references
 
 Bibliographical references have been moved out of ReSpec. You want to use
--- a/docs/MCP.md
+++ b/docs/MCP.md
@@ -1,0 +1,107 @@
+# ReSpec MCP
+
+## Summary
+
+`respec-mcp` is a stdio MCP server that wraps ReSpec's existing rendering API and
+adds repo-local profile discovery. The goal is to let specification repos keep their
+own policy, templates, and build paths while reusing one upstream MCP runtime.
+
+LLM operators should also read [MCP_LLM_AUTHORING_GUIDE.md](MCP_LLM_AUTHORING_GUIDE.md)
+before drafting or revising W3C/Community Group reports.
+
+## Repo-local configuration
+
+Each consuming repo can provide:
+
+- `respec-mcp.config.json`
+- `respec-mcp/profiles/*.json`
+
+Minimal config example:
+
+```json
+{
+  "default_profile": "example-cg",
+  "profile_directory": "respec-mcp/profiles",
+  "source_root": "reports/source",
+  "build_root": "reports/build"
+}
+```
+
+Minimal profile example:
+
+```json
+{
+  "profile_id": "example-cg",
+  "allowed_statuses": ["CG-DRAFT", "CG-FINAL"],
+  "default_status": "CG-DRAFT",
+  "repo_metadata_source": "w3c.json",
+  "status_templates": {
+    "CG-DRAFT": "respec-mcp/templates/cg-draft.html",
+    "CG-FINAL": "respec-mcp/templates/cg-final.html"
+  },
+  "required_sections": ["Abstract", "Introduction"],
+  "required_links": ["https://www.w3.org/community/example/"],
+  "forbidden_phrases": ["W3C Recommendation"]
+}
+```
+
+## Tools
+
+- `respec_list_profiles`
+- `respec_scaffold`
+- `respec_build`
+- `respec_validate`
+- `respec_preflight`
+
+All tools accept `repo_root` and may accept `profile`, `status`, `source`, `output`,
+and `overrides` depending on the operation.
+
+## Local execution
+
+```bash
+node tools/respec-mcp.js --repo-root /path/to/spec-repo
+```
+
+## Docker execution
+
+```bash
+docker build -t respec-mcp:local .
+docker run --rm -i -v /path/to/spec-repo:/workspace -w /workspace respec-mcp:local
+```
+
+## AI client usage
+
+Point the client to the stdio command and set the repo root at process start or per
+tool call.
+
+Authoring discipline matters as much as rendering. For W3C/CG reports, clients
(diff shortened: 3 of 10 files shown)


sidvishnoi:
Can you share a video demo to show how it works?

danielcamposramos:
> Can you share a video demo to show how it works?

Hi @sidvishnoi,

Thanks for looking at this. We built this MCP as a side project within the [PM-KR Community Group](https://www.w3.org/community/pm-kr/) to support our own report authoring workflow. The group's focus is on the main specification work, so we don't have the bandwidth to record a video demo right now — but the MCP is functional and has already been used in practice.

How to use it:
The server runs over stdio, so any MCP-aware client can connect.

VSCode agents (Claude Code, Cline, GitHub Copilot, etc.) — add to your MCP settings:


{
  "mcpServers": {
    "respec-mcp": {
      "command": "node",
      "args": ["tools/respec-mcp.js", "--repo-root", "/path/to/your/spec-repo"],
      "transport": "stdio"
    }
  }
}

Docker (includes Node 24 + headless Chromium for respec_build):

docker build -t respec-mcp:local .
docker run --rm -i respec-mcp:local

CLI (directly, for scripting or CI):

node tools/respec-mcp.js --repo-root /path/to/your/spec-repo

Once connected, the agent gets five tools: 
respec_list_profiles, respec_scaffold, respec_preflight, respec_validate, and respec_build. 
Each consuming repo provides its own respec-mcp.config.json and respec-mcp/profiles/*.json to define allowed statuses, required sections, forbidden phrases, templates, and build paths — so group-specific policy stays in the repo, not in the MCP server.

Live artifact
The [PM-KR Standards Track Transition Report](https://github.com/w3c-cg/pm-kr/raw/refs/heads/main/reports/source/pm-kr-standards-track-transition-report.html) was drafted using this setup with LLM assistance (Codex). The PM-KR repo at [w3c-cg/pm-kr](https://github.com/w3c-cg/pm-kr) has a working respec-mcp.config.json and profile that can serve as a reference for other groups.

Happy to answer any questions about the implementation or integration patterns.

marcoscaceres:
Will take a look soon…

marcoscaceres:
Thanks for putting this together, Daniel. The profile/policy pattern is a genuinely interesting idea: letting each repo declare its own allowed statuses, required sections, forbidden phrases, and templates is a clean way to handle CG governance without forking ReSpec. The references in the authoring guide are solid too. I checked them all and they hold up.

That said, I have concerns in two areas: security and packaging.

## Security

**Path traversal / arbitrary file write:** There are no containment checks on any resolved paths. Every tool that accepts `repo_root`, `source`, or `output` passes them straight through `path.resolve()` without verifying the result stays within the repo root. For example, in `resolveScaffoldPath()`:

```js
if (output) {
  return path.resolve(state.repoRoot, output);
}
```

An MCP client (i.e., an LLM) sending `output: "../../.bashrc"` or `output: "/etc/crontab"` would write to those locations. The `mkdir({ recursive: true })` call before each `writeFile` makes this worse, since it creates any intermediate directories needed. The same applies to `readSourceText()`, which will read any file on the filesystem if given an absolute path.

Since MCP tool inputs come from LLMs (which can be manipulated via prompt injection from document content), this is a meaningful attack surface. Every `path.resolve()` result that feeds into `readFile`, `writeFile`, or URL construction needs to be validated to stay within the repo root.

**Unrestricted URL navigation:** `toSourceUrl()` allows `source` to be any URL scheme. If an LLM provides `source: "https://evil.example.com/exploit.html"`, Puppeteer will navigate Chromium to that URL. Combined with `--disable-sandbox`, this is a code execution risk. The server should restrict `source` to `file://` URLs rooted within `repoRoot`, or relative paths within it.

**`repo_root` is LLM-overridable:** Every tool's Zod schema accepts `repo_root: z.string().optional()`. An LLM can override the CLI-provided `--repo-root` to any directory (e.g., `/`). The CLI-provided root should be the enforced boundary, not something the client can override per-call.

**Prototype pollution via `overrides`:** The schema uses `z.record(z.any())` and the values are spread directly into objects:

```js
const merged = {
  ...repoDefaults,
  ...profileDefaults,
  ...overrides,  // LLM-controlled
};
```

Keys like `__proto__`, `constructor`, `prototype` should be filtered before spreading. V8's spread operator currently doesn't copy `__proto__` from plain objects, but relying on engine-specific behavior for security is fragile.

## Design / Packaging

**Separate package, not bundled in core:** `@modelcontextprotocol/sdk` pulls in ~30 transitive dependencies including Express 5, Hono, jose (JWT), pkce-challenge (OAuth), and more. These are added as production `dependencies`, meaning every consumer of ReSpec as an npm package inherits the entire tree even if they never use the MCP server. This would work better as a separate package that depends on `respec` for `respecDocWriter.js` but ships independently.

**`validate` vs `preflight` are identical:** Both call `renderAndAssess(input, options, { writeOutput: false })`. From an MCP client's perspective, there's no behavioral difference. Either differentiate them (e.g., preflight does fast source-only checks without rendering, validate does the full Puppeteer render) or merge them into one.

**Compliance checks are shallow:** Required sections are detected via `htmlLower.includes(section.toLowerCase())`, which means the word "Introduction" appearing anywhere in the document body would satisfy a required-section check for `<h2>Introduction</h2>`. Same issue for required links and forbidden phrases. For forbidden phrases in particular, a false positive could block a build unnecessarily.

Happy to discuss any of this further.

marcoscaceres:
After thinking about this more, I think the right next step would be to incubate this in a separate repository rather than merging it into the ReSpec core package. There's precedent for this in the `speced` org: both ReSpec and Bikeshed split out specialized concerns into their own repos (e.g., [`respec-web-services`](https://github.com/speced/respec-web-services), [`bikeshed-data`](https://github.com/speced/bikeshed-data), [`bikeshed-boilerplate`](https://github.com/speced/bikeshed-boilerplate)).

A separate repo like `speced/respec-mcp` would let you iterate on the profile/policy design and address the security concerns without adding ~30 transitive dependencies to every ReSpec install. It can import `respecDocWriter.js` from `respec` as a dependency and ship on its own release cycle.

Happy to help set that up if you'd like.

marcoscaceres:
Closing this out... but happy to keep discussing here.

danielcamposramos:
Hi @marcoscaceres — thanks for the careful review. You were right that this belonged in its own repo. I took the feedback seriously and rebuilt it as a standalone package:

**→ https://github.com/danielcamposramos/respec-mcp**

Every concern you and Copilot raised is addressed. Summary for the record:

### Security (your blockers)

- **Path containment.** Every path input (`source`, `output`, `repo_metadata_source`, templates, profile directory) resolves through a `resolveWithinRoot` helper that rejects anything escaping `--repo-root` (including absolute paths outside the root, `..` traversal, and null bytes). Unit tests cover each case.
- **URL restriction for Puppeteer.** `source` accepts only relative paths inside the repo root or `file://` URLs pointing inside it. `http(s)`, `data:`, `javascript:`, and other schemes are rejected before Puppeteer is ever handed a URL.
- **`repo_root` is no longer part of the tool input schema.** The CLI flag is the sole boundary — the MCP client cannot override it per call.
- **Prototype-pollution hardening.** `overrides`, `template_defaults`, and `respec_defaults` merge into an `Object.create(null)` target with a key filter that drops `__proto__`, `constructor`, and `prototype`. Template placeholder resolution also refuses to traverse into `__proto__`. Tests include a `JSON.parse('{"__proto__": {...}}')` adversarial input.

### Design

- **`preflight` vs `validate` are now distinct.** `preflight` is source-only, no Puppeteer — fast feedback before editing. `validate` is the full ReSpec render via Puppeteer with diagnostics but no write. `build` is render + write. The README table makes the trade-off explicit.
- **Compliance checks use DOM parsing (linkedom), not `String.includes()`.** Required sections are matched against `<h1>..<h6>` text (ReSpec's `.secno` / `.self-link` nodes are stripped first so auto-numbering like "1. Introduction" still matches `Introduction`). Required links are matched against `<a href>` values. Forbidden phrases use word-boundary matching on visible text with scripts/styles removed — so "commit" doesn't match "uncommittable" and a `W3C Recommendation` string inside a `<script>` no longer trips a false positive.

### Packaging

- **Standalone npm package.** `respec` is a runtime dependency (via `respec/tools/respecDocWriter.js`), not a parent. ReSpec core ships none of the MCP SDK's transitive tree.
- **Dockerfile runs as a non-root user** (uid 10001) and bakes `--disable-sandbox` into the ENTRYPOINT so Chromium starts under an unprivileged UID.
- **`docs/` is in the published `files` list,** so MCP responses referencing the authoring guide resolve for npm consumers. The guide is also exposed as an MCP resource (`respec-mcp://authoring-guide`) so clients can fetch it over the wire.

### Copilot's items

- `file://` sources are read via `fileURLToPath` + `readFile`, not `fetch()`.
- Template placeholders with dotted keys (`{{editors.primary.name}}`) now resolve against nested objects.
- Dockerfile: see above — non-root + `--disable-sandbox`.
- `docs/` shipped in npm package: see above.

### Also

- Added W3C explainer support per https://www.w3.org/TR/explainer-explainer/: authoring-guide section, and a second worked profile (`examples/example-explainer/`) that enforces the TAG explainer skeleton via `required_sections` and blocks spec-like status claims via `forbidden_phrases`.
- 34 unit specs (security, template, compliance, core) + 3 integration specs (full Puppeteer render) — all passing. CI matrix runs Node 20 / 22 / 24.

If you're up for taking this under the `speced/` umbrella (transfer or fork — whichever is simpler for you), I'd be glad to hand it over. Happy to iterate on naming, API, or anything else. No pressure either way; I'll maintain it under my own namespace in the interim so PM-KR and similar groups can use it.

marcoscaceres:
Do you want some time for the community to experiment? I think it would be great to see what you all come up with. Having built several MCPs, I know how rapidly they can evolve. 

How about you check-in in 6 weeks or so once you’ve had a chance for the design to stabilize and have gained some usage experience across the CG? No objections to setting up a repo for you. But would like to be sure you’ve gained significant experience with it first. Then we can bring it over piecemeal and go hard on privacy, security, performance, etc. audits to make it generally available if it proves useful. 

I’ll try to find time to play with it and provide feedback.

PR #5177, opened 2026-04-13
fix(core/render-biblio): sanitize spaces in reference IDs

### LINKED ISSUE #5102: The names of the reference contain spaces
## Description of problem

https://w3c.github.io/ruby-tts-req/

When the name of a [reference](https://respec.org/docs/#references-0) contains spaces (like [1](https://github.com/w3c/ruby-tts-req/blob/310c7ffc192406d2f1553513a68bf494948c2d1a/index.html#L80) and [2](https://github.com/w3c/ruby-tts-req/blob/310c7ffc192406d2f1553513a68bf494948c2d1a/index.html#L74)), the generated HTML `id` attribute value will also contain spaces, which violates HTML syntax.

Either explicitly prohibit spaces in the names of references, or replace spaces with hyphens/underscores in the generated IDs.

marcoscaceres:
Reference names with spaces (e.g. 'Ruby TTS Req') produced invalid HTML IDs containing spaces. Consolidates all bib- ID generation through a bibRefId() helper that replaces whitespace with hyphens.

Closes #5102


marcoscaceres:
Hmmm... definitely need a test for this one. 

The examples are wild:

```
	  'Tokyo Ghoul: re': {
              title: "Tokyo Ghoul: re, Vol. 8, Chapter 76 “Daso (惰疎)",
              authors: ["Sui Ishida"], 
              id: "Tokyo Groul: re 8-76",
              date: "2016",
              publisher: "Shueisha",
          },
          'Transliteration Training Course': {
              title: "Textbook for the Volunteer Transliteration Training Course: Basics (In Japanese, 音訳ボランティア養成講習会テキスト　基礎課程編)",
	      href: "https://naiiv-books.net/shopdetail/000000000023/",
            date: "March 2010",
            publisher: "National Council of Japan for the Visually Impaired (In Japanese, 全国視覚障害者情報提供施設協会)",
          },
```

As I never expected ids from bib-ref to be used this way (they follow the short name rules from w3c)... but sure, why not!

marcoscaceres:
@copilot can you add tests. See my comment above?

src/core/render-biblio.js
```diff
@@ -8,6 +8,11 @@ import { html } from "./import-maps.js";
 
 export const name = "core/render-biblio";
 
+/** @param {string} ref */
+function bibRefId(ref) {
+  return `bib-${ref.toLowerCase().replace(/\s+/g, "-")}`;
```
marcoscaceres: @copilot, is it enough to just replace spaces? Are there other weird things that could be used for IDs by people that could end up breaking HTML? Should we be more thorough here? 

If yes, can you implement it?

src/core/utils.js
```diff
@@ -393,6 +393,25 @@ export function addHashId(elem, prefix = "") {
...
+/**
+ * Converts a string to a slug suitable for use in an HTML id attribute:
+ * lowercases (unless noLC is true), decomposes Unicode, strips diacritics,
+ * replaces runs of non-word characters with "-", and trims leading/trailing
+ * hyphens.
+ * @param {string} txt
+ * @param {boolean} [noLC] - when true, skip lowercasing
+ * @returns {string}
+ */
+export function toId(txt, noLC = false) {
```
marcoscaceres, line 405: @copilot, as we are adding this new util, you might want to test it pretty thoroughly along with the other exported util functions

marcoscaceres:
@copilot wonderful! But I’m worried toId is to generic a name (and it looks like told)… maybe call it toSafeHTMLId() or something better?

sidvishnoi (approved):
:shipit:

sidvishnoi (approved):
GitHub asking for review even when I just did, so, :shipit: again.

PR #5185, opened 2026-04-13
refactor!: remove legacy AMD require shim from expose-modules

marcoscaceres:
## Summary

- Removes the legacy AMD `window.require` shim from `expose-modules.js`
- `expose()` becomes a no-op (call sites unchanged, just stop registering modules)
- Worker access migrated from `require(["core/worker"])` to `document.respec.worker`
- Removes dead `require()` fallback from `respecDocWriter.js`

## Breaking change

External code that called `window.require(["core/exporter"], ...)` or `window.require(["core/worker"], ...)` will no longer work. Use `document.respec.toHTML()` and `document.respec.worker` instead.

## Test plan

- [x] ~~Modified Web platform tests~~ — no observable spec rendering change


src/core/expose-modules.js
```diff
@@ -1,30 +1,10 @@
...
  */
-export function expose(name, object) {
-  if (!inAmd) {
-    window.require.modules[name] = object;
-  }
+export function expose(_name, _object) {
+  // No-op. Previously exposed modules via a legacy AMD require shim
+  // for the deprecated w3c-common profile. Kept as a no-op so existing
+  // expose() call sites don't need to be removed yet.
 }
```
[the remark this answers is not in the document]
marcoscaceres: @copilot can you find a fix?

marcoscaceres:
@copilot can you check/fix what's going on in CI?

src/core/worker.js
```diff
@@ -78,7 +77,4 @@ async function createWorker() {
 
 export const workerPromise = createWorker();
 
-expose(
-  name,
-  workerPromise.then(worker => ({ worker }))
-);
+window.respecWorker = workerPromise;
```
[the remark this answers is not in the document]
marcoscaceres: The builds/ are CI-generated on release, not committed to PRs. They will be rebuilt automatically.

sidvishnoi: Not following this one. Why add to window?
Btw, if we really need to add it, we've `document.respec`.

sidvishnoi (commented):
Looks good, just one question on adding `respecWorker` on window.

Also, let's mark this as a breaking change?

marcoscaceres:
@copilot Can you rebase, lint, and prettier this and do what Sid suggested?

src/core/worker.js
marcoscaceres: Good point. The TS6 merge resolved this — the test fixture now uses `document.respec.worker` instead. The `window.respecWorker` assignment was removed.

marcoscaceres:
@copilot please rebase

marcoscaceres:
@copilot can you take a look at CI and fix the issue?

src/core/respec-global.js
```diff
@@ -41,6 +43,10 @@ class ReSpec {
     return this._respecDonePromise;
   }
 
+  get worker() {
+    return workerPromise;
+  }
```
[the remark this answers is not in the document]
marcoscaceres: This thread was resolved without a reply comment. Reviewing whether the concern was addressed in subsequent commits. Leaving this reply for traceability — Copilot should re-verify whether this was properly fixed.

marcoscaceres:

tools/respecDocWriter.js
```diff
@@ -237,42 +237,15 @@ async function generateHTML(page, timer, version, url) {
...
-      require(["core/exporter"], resolve, err => {
-        reject(new Error(err.message));
-      });
-    });
-    const dataURL = rsDocToDataURL("text/html");
-    const encodedString = dataURL.replace(/^data:\w+\/\w+;charset=utf-8,/, "");
-    return decodeURIComponent(encodedString);
-  } else {
-    return await document.respec.toHTML();
-  }
```
sidvishnoi, line 275: Maybe ok to keep this for backward compatibility? Or maybe if `document.respec.toHTML` doesn't exist, exit with an error message: "please use older version of respec CLI/API, or use newer ReSpec script"?

src/core/respec-global.js
```diff
@@ -41,6 +43,10 @@ class ReSpec {
     return this._respecDonePromise;
   }
 
+  get worker() {
```
sidvishnoi, line 46: Who's the user of this? Do we really want to add this?
I mean, internally, we can import from `./worker.js` when needed. So, who the consumers of this API are?

tools/respecDocWriter.js
marcoscaceres: Good call. Keeping the `require` shim just for backward compat feels like the wrong direction (the whole point of the PR is to remove it), but silently failing is not great either.

I think the error message approach is better. Something like:

```js
if (!document.respec?.toHTML) {
  throw new Error(
    "document.respec.toHTML is not available. " +
    "Please use a newer version of ReSpec (v20.10+), " +
    "or use an older version of the ReSpec CLI that supports legacy AMD modules."
  );
}
return await document.respec.toHTML();
```

That way users of very old ReSpec get a clear, actionable error instead of a cryptic `TypeError`. I will add this.

src/core/respec-global.js
marcoscaceres: You are right to question this. I added it as a 1:1 replacement for the old `expose("core/worker", ...)` call, but looking at it again, there are no external consumers. Internally, `highlight.js` imports `workerPromise` directly from `./worker.js`, which is the correct path.

The old AMD `expose()` was an implementation detail, not a deliberate public API. Promoting it to `document.respec.worker` creates new API surface without a use case.

I will remove the `worker` property (and its test) from this PR.

tools/respecDocWriter.js
marcoscaceres: Applied: added a guard that throws a clear error when `document.respec.toHTML` is unavailable, directing users to either upgrade ReSpec or use an older CLI version.

src/core/respec-global.js
marcoscaceres: Agreed and removed: the `worker` property has been removed from the public API surface since it has no external consumers.

sidvishnoi (approved):

marcoscaceres:
@copilot Can you check and fix the ci issue?

marcoscaceres:
@copilot Can you check ci again? Still busted

PR #5204, opened 2026-04-18
fix: make target background of references accessible in dark mode

marcoscaceres:
## Summary

- Adds dark mode styling for `#references :target` highlight background
- Covers both `@media (prefers-color-scheme: dark)` and the manual dark mode toggle
- Adds fallback value for `--borderedblock-bg` CSS variable (for non-W3C profiles)

Based on @TimvdLippe's work in #5165, with additional fixes for the manual toggle selector and CSS variable fallback.

Closes #5165

## Test plan

- [x] ~~Modified Web platform tests~~ — CSS-only change, no observable behavior change beyond styling


src/styles/respec.css.js
```diff
@@ -58,6 +58,26 @@ a[href].respec-offending-element {
...
 }
 
+@media (prefers-color-scheme: dark) {
+  #references :target {
+    background: color-mix(
+      in srgb,
+      var(--borderedblock-bg, #303030) 70%,
+      transparent
+    );
+    animation: pop 0.4s ease-in-out 0s 1;
```
sidvishnoi: Don't need to add `animation` separately, it's already on `#references :target`

sidvishnoi (changes requested):

src/styles/respec.css.js
marcoscaceres: Good catch — the base `#references :target` rule already sets the animation, and the dark mode overrides only need to change `background`. The cascade handles the rest. I'll remove the `animation` line from both dark mode blocks.

I'll also extract the shared `color-mix` value into a JS const to avoid the duplication between the `@media` and `body:has()` selectors.

src/styles/respec.css.js
```diff
@@ -1,5 +1,7 @@
 /* ReSpec specific CSS */
 const css = String.raw;
+const defined = String.raw;
```
sidvishnoi: Why this particular constant?

sidvishnoi (approved):

src/styles/respec.css.js
```diff
@@ -1,5 +1,7 @@
...
+const darkRefTargetBg = defined`color-mix(in srgb, #eaf3ff 15%, transparent)`;
```
marcoscaceres: yeah, don't need that.

```suggestion
const darkRefTargetBg = String.raw`color-mix(in srgb, #eaf3ff 15%, transparent)`;
```

src/styles/respec.css.js
marcoscaceres: Acknowledged — if the `animation` property is already on `#references :target`, adding it separately is redundant. Will check the CSS and remove the duplicate if confirmed.

marcoscaceres:

PR #5210, opened 2026-04-18
fix(core/highlight): use textContent instead of innerText

### LINKED ISSUE #4695: Highlight.js and <details>
## Description of problem

When using progressive disclosure using the details element in a ReSpec document, Highlight.js ceases to function and erases the content from the rendering tree.

https://w3c.github.io/sustyweb/drafts/ (Note as a temporary fix we're using nohighlight as a workaround.)

The code inside the pre element is erased and an empty tag is rendered instead of syntax being highlighted.

The syntax should be highlighted, as it is for any code that would exist outside of the details element.

1.  Create a ReSpec draft document.
2. Add a **details** element with some syntax inside.
<details>
<summary>Show / Hide content.</summary>
<pre class="js"><code>
!function(e,t){"use strict";"object"==typeof module&&"object"==typeof module.exports?module.exports=e.document?t(e,!0):function(e){if(!e.document)throw new Error("jQuery requires a window with a document");return t(e)}:t(e)}("undefined"!=typeof window?window:this,function(g,e){"use strict";var t=[],r=Object.getPrototypeOf,s=t.slice,v=t.flat?function(e){return t.flat.call(e)}:function(e){return t.concat.apply([],e)},u=t.push,i=t.indexOf
</code></pre>
</details>

- This test showcases a basic ReSpec with it non-functional [test.txt](https://github.com/w3c/respec/files/15023847/test.txt)
- This test showcases a basic HTML file with it functional [test2.txt](https://github.com/w3c/respec/files/15023848/test2.txt)

**Note:** I have verified that this issue does not occur in a normal HTML document using highlight.js therefore it possibly is due to ReSpec or its implementation of highlight?

marcoscaceres:
## Summary

`innerText` returns empty for content inside closed `<details>` elements because it reads only rendered (visible) text. `textContent` reads the DOM regardless of visibility, so syntax highlighting now works inside `<details>`.

One-line change in `highlightElement()`.

Fixes #4695

## Test plan

- [x] 8 highlight tests pass
- [x] ESLint clean
- [x] ~~Modified Web platform tests~~ — browser behaviour fix in ReSpec, no WPT scope

### DIFF
--- a/src/core/highlight.js
+++ b/src/core/highlight.js
@@ -23,7 +23,7 @@ async function highlightElement(elem) {
   const languages = getLanguageHint(elem.classList);
   let response;
   try {
-    response = await sendHighlightRequest(elem.innerText, languages);
+    response = await sendHighlightRequest(elem.textContent, languages);
   } catch (err) {
     console.error(err);
     return;


marcoscaceres:
Duplicate — fix already included in #5208

marcoscaceres:
@copilot please fix the missing test: add a test case to `tests/spec/core/highlight-spec.js` that places a `<pre class="js">` block inside a closed `<details>` element and verifies that the code inside is still syntax-highlighted (i.e., the `<pre>` gets `data-highlighted` set and contains `<span>` elements from highlight.js). This is the regression case that the `textContent` fix in this PR addresses.

PR #5214, opened 2026-04-18
fix(core/link-to-dfn): skip THIS_SPEC substitution when shortName is absent

### LINKED ISSUE #3429: Weird xref error message when shortName is missing from config
## Important info

- URL to affected spec: None. Fake spec used for testing purpose in Reffy.
- ReSpec version: 26.5.0

## Description of problem

**What happened?:**
We run a test in Reffy with a fake and mostly empty ReSpec spec to make sure that switching to the latest version of ReSpec won't create any issue. Something changed between 26.3.0 and 26.5.0 that made the test fail.

One of the reasons was the switch to `api.specref.org` and that one is totally normal.

The other reason left me puzzled for a few minutes. The fake spec has an empty ReSpec config and a simple IDL interface definition exposed on `Window`, and this makes ReSpec create a weird `[__SPEC__HTML] Reference not found.` entry in the list of references, along with the following two errors:

```
(core/xref): Couldn't match "Window" to anything in the document or in any other document cited in this specification: __spec__html. Learn more about this error or see how to cite to resolve the error. Occurred at: 1.
(core/render-biblio): Bad reference: [__SPEC__HTML] (appears 0 times)
```

This seems triggered by the lack of shortname for the spec, meaning that adding a `shortName` or a `github` entry to the config fixes the problem.

**Expected behavior:**
The fake config is clearly invalid, so I obviously don't mind if ReSpec reports errors.

I would just expect it to report something like "No shortName for the spec", and to still be able to generate references to the HTML spec (simply because the notion of spec shortname seems totally orthogonal to xref...)

Here is an example of spec that creates the issue:

```html
<script>respecConfig = {};</script>
<script src='https://www.w3.org/Tools/respec/respec-w3c'></script>
<div id="abstract"></div>
<pre class='idl'>[Exposed=Window] interface Foo { attribute DOMString bar; };</pre>
```

And here is one way to fix the config (ReSpec still complains about the lack of editors, which is good, but correctly generates the HTML entry in the list of references):

```html
<script>respecConfig = { shortName: 'test' };</script>
<script src='https://www.w3.org/Tools/respec/respec-w3c'></script>
<div id="abstract"></div>
<pre class='idl'>[Exposed=Window] interface Foo { attribute DOMString bar; };</pre>
```

marcoscaceres:
## Problem

When `shortName` is absent from `respecConfig`, `updateReferences` built a degenerate regex:

```js
new RegExp(String.raw`^([?!])?${regExpEscape("")}\b([^-])`, "i")
// → /^([?!])?\b([^-])/i
```

This matched the first character of every `data-cite` value, corrupting e.g. `"HTML"` into `"__SPEC__HTML"`, causing:

```
(core/xref): Couldn't match "Window" … __spec__html
(core/render-biblio): Bad reference: [__SPEC__HTML]
```

## Fix

Build the regex conditionally — only when `shortName` is non-empty. Reference tracking (normative/informative) still runs in all cases.

## Testing

New test: "does not corrupt data-cite values when shortName is missing".

Closes #3429


src/core/link-to-dfn.js
```diff
@@ -342,22 +342,25 @@ function showLinkingError(elems) {
...
   /** @type {NodeListOf<HTMLElement>} */
   const elems = document.querySelectorAll(
     "dfn[data-cite]:not([data-cite='']), a[data-cite]:not([data-cite=''])"
   );
   for (const elem of elems) {
-    elem.dataset.cite = (elem.dataset.cite ?? "").replace(
-      regex,
-      `$1${THIS_SPEC}$2`
-    );
+    if (regex) {
```
sidvishnoi, line 358: If we also check for `elem.dataset.cite` here,  does it cause issues?

sidvishnoi (approved):

src/core/link-to-dfn.js
marcoscaceres: It wouldn't cause issues, but it would be redundant: the `querySelectorAll` selector on line 354 already excludes elements with empty `data-cite` via `:not([data-cite=''])`, so `elem.dataset.cite` is always truthy inside the loop.

The regression test at the bottom of the diff covers the real risk (empty `shortName` producing a degenerate regex). That test would catch any future path where `data-cite` somehow gets corrupted, regardless of the guard structure.

PR #5217, opened 2026-04-20
fix(core/inlines): render self-citation as <cite> with spec title

### LINKED ISSUE #2363: Self-citing references show in bibliography
## Important info

* URL to affected spec: https://w3c.github.io/json-ld-syntax/
* ReSpec version: 24.21.7

## Description of problem

What happened (e.g., it crashed)?:  Reference to self shows up in References ([JSON-LD11], in this case).

But what I was expecting?: According to https://github.com/w3c/respec/commit/9604ca8ca1486996b1c21d2a7cbe24ac641b43a1, as I understand it, definitions and references to a spec with the same short-name should be filtered from the References section. However, reference does show up.

marcoscaceres:
Closes #2363

- When `[[shortName]]` matches the spec's own `shortName`, render as `<cite>Spec Title</cite>` instead of adding to bibliography
- Prevents a spec from citing itself in its own references section
- The `<cite>` element gets a `title` attribute with the shortName for traceability


marcoscaceres:

marcoscaceres (commented):
Approving.

PR #5225, opened 2026-04-23
Security: Prototype pollution vector via query-parameter configuration override

tomaioo:
## Summary

Security: Prototype pollution vector via query-parameter configuration override

## Problem

**Severity**: `High` | **File**: `src/core/override-configuration.js:L14`

Configuration keys from `document.location.search` are converted directly into object properties and merged with `Object.assign(config, overrideProps)` without blocking special keys like `__proto__`, `constructor`, or `prototype`. An attacker-controlled URL can inject these keys and alter object prototypes or behavior of downstream config consumers.

## Solution

Reject dangerous property names before `Object.fromEntries`/`Object.assign` (e.g., deny `__proto__`, `prototype`, `constructor`), and consider using a schema allowlist for overridable config keys.

## Changes

- `src/core/override-configuration.js` (modified)


marcoscaceres (commented):
Thanks for the fix — the vulnerability is real. I verified that `Object.fromEntries` + `Object.assign` with `__proto__` as a key does replace the target object's prototype chain (though it doesn't pollute `Object.prototype` globally). The `constructor` case also breaks `config.constructor`. The filter is the right approach.

Could you add a test? Something like:

```js
it("rejects dangerous property names from query parameters", async () => {
  const url = new URL(simpleURL.href);
  url.searchParams.set("__proto__", '{"polluted":true}');
  url.searchParams.set("constructor", '{"polluted":true}');
  url.searchParams.set("prototype", '{"polluted":true}');
  url.searchParams.set("specStatus", "RSCND");
  url.searchParams.set("group", "webapps");
  const doc = await makeRSDoc(makeStandardOps(), url);
  const { respecConfig: conf } = doc.defaultView;
  // Dangerous keys must not land on config
  expect(conf).not.toHaveProperty("polluted");
  expect(Object.getPrototypeOf(conf)).toBe(Object.prototype);
  // Normal overrides still work
  expect(conf.specStatus).toBe("RSCND");
});
```

One minor note on severity: I'd call this **moderate** rather than high, since exploitation requires tricking a user into clicking a crafted URL to a spec document (no remote exploitation without user interaction). But the fix is correct regardless.

marcoscaceres (changes requested):
See https://github.com/speced/respec/pull/5225#pullrequestreview-4175238821

Just needs a test.

PR #5226, opened 2026-04-23
fix(core/xref): normalize versioned spec names in data-cite context

### LINKED ISSUE #5224: Same spec version resolution mess
## Description of problem

Weird bug;

```HTML
        <p>
          <a data-cite="SERVICE-WORKERS-1#scope-match-algorithm">Service
          Workers</a>. 
        </p>
         <aside class="note" data-cite="service-workers">
           [=service worker=]
          </aside>
```
Gives: 

> Couldn't find "service worker" in this document or other cited documents: [csp3], [dom], [encoding], [fetch], [html], [image-resource], [infra], [mediaqueries-5], [mimesniff], [screen-orientation], [url], and [webidl].
How to fix: [See search matches for "service worker"](https://respec.org/xref/?term=service+worker&types=_CONCEPT_) or [Learn about this error](https://respec.org/docs/#error-term-not-found).
. Occurred 1 times at:
[<a>](/manifest/index.html#respec-offender-no-matching-definition-found) element

There's some weirdness with resolving service-workers but also, SERVICE-WORKERS-1 ... makes a mess.

marcoscaceres:
Closes #5224

`getSpecContext()` in `xref.js` only lowercased spec names from `data-cite` attributes — it did no version-suffix normalization. So `data-cite="SERVICE-WORKERS-1"` became `service-workers-1`, which the xref API couldn't match via its `shortname` field (`service-workers`).

The fix expands each versioned spec name (matching `-\d+$`) to include both the original and the unversioned form in the same priority level. This preserves the versioned form for correct biblio resolution while also enabling shortname matching on the server.

Additionally, the "term not found" error hint now shows the raw `data-cite` value when the failing element has a local `data-cite` ancestor, so authors can spot spec shortname typos immediately.

### DIFF
--- a/src/core/xref.js
+++ b/src/core/xref.js
@@ -210,6 +210,29 @@ export function getTermFromElement(elem) {
 }
 
 /**
+ * @param {string} spec
+ */
+function stripVersionSuffix(spec) {
+  return spec.replace(/-\d+$/, "");
+}
+
+/**
+ * Expands versioned spec names to include the unversioned form.
+ * E.g., ["service-workers-1"] becomes ["service-workers-1", "service-workers"].
+ * @param {string[]} specs
+ */
+function expandVersionedSpecs(specs) {
+  return [
+    ...new Set(
+      specs.flatMap(s => {
+        const stripped = stripVersionSuffix(s);
+        return stripped !== s ? [s, stripped] : [s];
+      })
+    ),
+  ];
+}
+
+/**
  * Get spec context as a fallback chain, where each level (sub-array) represents
  * decreasing priority.
  * @param {HTMLElement} elem
@@ -226,7 +249,7 @@ function getSpecContext(elem) {
     const cite = (dataciteElem.dataset.cite ?? "")
       .toLowerCase()
       .replace(/[!?]/g, "");
-    const cites = cite.split(/\s+/).filter(s => s);
+    const cites = expandVersionedSpecs(cite.split(/\s+/).filter(s => s));
     if (cites.length) {
       specs.push(cites);
     }
@@ -242,7 +265,9 @@ function getSpecContext(elem) {
     const bibrefs = closestSection
       ? closestSection.querySelectorAll("a.bibref")
       : [];
-    const inlineRefs = [...bibrefs].map(el => el.textContent.toLowerCase());
+    const inlineRefs = expandVersionedSpecs(
+      [...bibrefs].map(el => el.textContent.toLowerCase())
+    );
     if (inlineRefs.length) {
       specs.push(inlineRefs);
     }
@@ -500,7 +525,13 @@ function showErrors({ ambiguous, notFound }) {
     const originalTerm = getTermFromElement(elems[0]);
     const formUrl = getPrefilledFormURL(originalTerm, query);
     const specsString = joinAnd(specs, s => `**[${s}]**`);
-    const hint = howToFix(formUrl, originalTerm);
+    let hint = howToFix(formUrl, originalTerm);
+    /** @type {HTMLElement | null} */
+    const closestCite = elems[0].closest("[data-cite]");
+    if (closestCite && closestCite !== document.body) {
+      const citeAttr = (closestCite.dataset.cite ?? "").replace(/`/g, "");
+      hint += ` A parent element has \`data-cite="${citeAttr}"\` — check that the spec shortname is correct.`;
+    }
     const forParent = query.for ? `, for **"${query.for}"**, ` : "";
     const msg = `Couldn't find "**${originalTerm}**"${forParent} in this document or other cited documents: ${specsString}.`;
     const title = "No matching definition found.";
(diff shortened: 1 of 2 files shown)


sidvishnoi (commented):
Concern: what if xref data has a term defined in both versoned spec and non-versioned? Can lead to "ambiguous dfn" errors.
It might not be the case, but worth checking the data and adding a test for.

src/core/xref.js
```diff
@@ -500,7 +525,13 @@ function showErrors({ ambiguous, notFound }) {
     const originalTerm = getTermFromElement(elems[0]);
     const formUrl = getPrefilledFormURL(originalTerm, query);
     const specsString = joinAnd(specs, s => `**[${s}]**`);
-    const hint = howToFix(formUrl, originalTerm);
+    let hint = howToFix(formUrl, originalTerm);
+    /** @type {HTMLElement | null} */
+    const closestCite = elems[0].closest("[data-cite]");
+    if (closestCite && closestCite !== document.body) {
+      const citeAttr = (closestCite.dataset.cite ?? "").replace(/`/g, "");
+      hint += ` A parent element has \`data-cite="${citeAttr}"\` — check that the spec shortname is correct.`;
```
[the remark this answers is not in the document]
marcoscaceres: This thread was resolved without a reply comment. Reviewing whether the concern was addressed in subsequent commits. Leaving this reply for traceability — Copilot should re-verify whether this was properly fixed.

marcoscaceres:

marcoscaceres:
@sidvishnoi Your concern is addressed in commit be15aa2. The unversioned form (e.g. `service-workers` for `service-workers-1`) is now added as a **separate lower-priority fallback level** in the spec context chain, not at the same level as the versioned form. The xref server only tries the unversioned shortname when the versioned one yields no result — so both can coexist in the data without triggering ambiguous-dfn errors.

`dedupeSpecContext` was also updated to track all previously-seen specs across levels (not just the immediate previous), preventing duplicates when the unversioned form appears elsewhere in the context chain.

A test was added to verify no "ambiguous" or "not found" errors are raised when using versioned spec names in `data-cite`.

src/core/xref.js
```diff
@@ -229,6 +244,20 @@ function getSpecContext(elem) {
...
+      // at a lower priority so the server only falls back to them when the
+      // versioned name yields no result.  This prevents ambiguous-dfn errors if
+      // a term is indexed under both shortname variants.
+      const unversioned = [
+        ...new Set(
+          cites
+            .map(stripVersionSuffix)
+            .filter((stripped, i) => stripped !== cites[i])
+        ),
+      ];
```
sidvishnoi: Looks like a use-case for [`Set.difference`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/difference).

sidvishnoi (commented):
Will review again in detail...

src/core/xref.js
marcoscaceres: Applied — both occurrences in `getSpecContext` now use `Set.difference` to compute the unversioned fallback specs:

```js
const unversioned = [
  ...new Set(cites.map(stripVersionSuffix)).difference(new Set(cites)),
];
```

Much cleaner than the `.map().filter()` + positional index approach. Thanks for the suggestion!

src/core/xref.js
```diff
@@ -242,9 +267,20 @@ function getSpecContext(elem) {
...
+      ...new Set([...bibrefs].map(el => el.textContent.toLowerCase())),
+    ];
+    if (inlineRefList.length) {
+      specs.push(inlineRefList);
+      // Same versioned → unversioned fallback for inline bibrefs.
+      const unversioned = [
+        ...new Set(inlineRefList.map(stripVersionSuffix)).difference(
+          new Set(inlineRefList)
+        ),
+      ];
```
sidvishnoi: Nit: consider moving this into a small util function (within this file, as used twice).

src/core/xref.js
```diff
@@ -260,11 +296,11 @@ function getSpecContext(elem) {
...
   const unique = [];
+  /** @type {Set<string>} tracks all specs seen in higher-priority levels */
+  const seen = new Set();
   for (const level of specs) {
-    const higherPriority = unique[unique.length - 1] || [];
-    const uniqueSpecs = [...new Set(level)].filter(
-      spec => !higherPriority.includes(spec)
-    );
+    const uniqueSpecs = [...new Set(level)].filter(spec => !seen.has(spec));
+    uniqueSpecs.forEach(s => seen.add(s));
```
sidvishnoi: If we're ok with some more recent JS support, we can call `.filter` iterator directly on `Set.values()` (without converting it to array). I think we did it already in some other PR.

sidvishnoi (approved):

src/core/xref.js
marcoscaceres: Applied in db12c058: extracted `getUnversionedFallbacks(specs)` util used by both call sites.

src/core/xref.js
marcoscaceres: Applied in db12c058: `dedupeSpecContext` now uses `new Set(level).values().filter(s => !seen.has(s))` directly instead of spreading to array first.

PR #5263, opened 2026-05-04
fix(w3c/headers): derive /TR pubSpace for multi-group specs

### LINKED ISSUE #4688: "Latest published version" in editor's draft points to wrong place
In https://w3c.github.io/geolocation-api/, the "Latest published version:" points to https://www.w3.org/geolocation/, which is not a version of this spec. I think it should point to https://www.w3.org/TR/geolocation/.

marcoscaceres:
Closes #4688

When `group` is an array (e.g., `["das", "webapps"]`), `conf.groupType` becomes an array like `["wg", "wg"]`. The strict equality check `conf.groupType === "wg"` always fails for arrays, so `derivePubSpace` returned `undefined`, producing an incorrect or empty `latestVersion`.

Now checks both string and array forms of `groupType`.

Visually verified in Safari with `group: ["das", "webapps"], specStatus: "ED", shortName: "geolocation"`: "Latest published version" correctly shows `https://www.w3.org/TR/geolocation/`.

### DIFF
--- a/src/w3c/headers.js
+++ b/src/w3c/headers.js
@@ -684,7 +684,11 @@ function derivePubSpace(conf) {
   if (conf.isNoTrack && !conf.isCGBG && !conf.isTagFinding) {
     return "";
   }
-  if (trStatus.includes(specStatus ?? "") || conf.groupType === "wg") {
+  const groupType = conf.groupType;
+  const isWG =
+    groupType === "wg" ||
+    (Array.isArray(groupType) && groupType.includes("wg"));
+  if (trStatus.includes(specStatus ?? "") || isWG) {
     return `/TR`;
   }
 
--- a/tests/spec/w3c/headers-spec.js
+++ b/tests/spec/w3c/headers-spec.js
@@ -1469,6 +1469,27 @@ describe("W3C — Headers", () => {
       expect(latestVersionEl.textContent.trim()).toBe("none");
     });
 
+    it("derives /TR latestVersion for multi-group ED specs", async () => {
+      const ops = makeStandardOps({
+        shortName: "multi-group-test",
+        specStatus: "ED",
+        group: ["das", "webapps"],
+        edDraftURI: "https://example.com/ed/",
+      });
+      const doc = await makeRSDoc(ops);
+
+      const terms = [...doc.querySelectorAll("dt")];
+      const latestVersion = terms.find(
+        el => el.textContent.trim() === "Latest published version:"
+      );
+      expect(latestVersion).toBeTruthy();
+      const latestVersionLink =
+        latestVersion.nextElementSibling.querySelector("a");
+      expect(latestVersionLink.href).toBe(
+        "https://www.w3.org/TR/multi-group-test/"
+      );
+    });
+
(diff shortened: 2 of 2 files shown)


src/w3c/headers.js
```diff
@@ -684,7 +684,11 @@ function derivePubSpace(conf) {
   if (conf.isNoTrack && !conf.isCGBG && !conf.isTagFinding) {
     return "";
   }
-  if (trStatus.includes(specStatus ?? "") || conf.groupType === "wg") {
+  const groupType = conf.groupType;
+  const isWG =
+    groupType === "wg" ||
+    (Array.isArray(groupType) && groupType.includes("wg"));
+  if (trStatus.includes(specStatus ?? "") || isWG) {
```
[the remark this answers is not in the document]
marcoscaceres: Added. The test uses `group: ["das", "webapps"]` with `specStatus: "ED"` and asserts the latestVersion link resolves to `https://www.w3.org/TR/multi-group-test/`.

marcoscaceres:

marcoscaceres:

src/w3c/headers.js
```diff
@@ -684,7 +684,11 @@ function derivePubSpace(conf) {
   if (conf.isNoTrack && !conf.isCGBG && !conf.isTagFinding) {
     return "";
   }
-  if (trStatus.includes(specStatus ?? "") || conf.groupType === "wg") {
+  const groupType = conf.groupType;
+  const isWG =
+    groupType === "wg" ||
+    (Array.isArray(groupType) && groupType.includes("wg"));
```
jyasskin, line 690: Are there any cases where there are heterogenous group types, and the correct space isn't `/TR`?

marcoscaceres: @jyasskin Good question. In practice I haven't seen a multi-group spec with heterogeneous types (e.g., WG + CG). But looking at the code path:

The `specStatus` provides the primary gate. If someone sets a CG/BG status (`CG-DRAFT`, `CG-FINAL`, etc.), the `isCGBG` flag is set earlier in the pipeline and `derivePubSpace` handles it via the switch statement (returning `/community/reports/...`), regardless of `groupType`. So a CG-status spec won't accidentally get `/TR` even if one group in the array happens to be a WG.

I don't think the heterogeneous case is reachable in valid configurations. Per [W3C Process §6.3.1](https://www.w3.org/policies/process/20250818/#editors-draft):

> "Working Groups and Interest Groups may make available Editor's drafts."

`ED` is a WG/IG status. CGs use `CG-DRAFT`/`CG-FINAL`. A CG wouldn't appear alongside a WG for the same deliverable because they operate under different IPR frameworks (CLA vs W3C Patent Policy). If a WG is in the group array, the spec is on the Recommendation track and `/TR` is correct.

That said, if you have a concrete scenario in mind I'd be happy to add a guard. But I think the `specStatus` + `groupType` combination makes this self-correcting already.

jyasskin: I don't have a concrete case. IG/WG notes would correctly be placed in /TR. I was just flagging a common bug pattern with this sort of code.

sidvishnoi (approved):

PR #5275, opened 2026-05-09
fix(styles): add width:100% to .header-wrapper for flexbox layout

marcoscaceres:
Closes #5150

Without an explicit width, flex containers with small content can wrap alongside adjacent elements instead of taking a full row.


marcoscaceres:

PR #5276, opened 2026-05-09
fix(core/issues-notes): warn about githubusercontent images in issues

### LINKED ISSUE #3825: using class="issue" will include githubusercontents URL
## Important info

- spec: editing in PR / https://github.com/w3c/imsc-hrm/pull/11
- Respec version: ReSpec 26.16.0

## Description of problem

we have a github issue with image attached and included in the initial comment, and we wanted to include it by using class="issue" feature of respec.
output from github API contains a img with source URL using githubusercontent, and that line causes an error on W3C publication.

it should be hard to fix that url and place contents locally during processing, so I'd request this to be documented...

marcoscaceres:
Closes #3825

GitHub issue bodies often contain images hosted on user-images.githubusercontent.com. These URLs are rejected by W3C publication tools (pubrules). Warns spec authors so they can download and self-host the images.

### DIFF
--- a/src/core/issues-notes.js
+++ b/src/core/issues-notes.js
@@ -190,6 +190,15 @@ function handleIssues(ins, ghIssues, conf) {
         body = document
           .createRange()
           .createContextualFragment(ghIssue.bodyHTML);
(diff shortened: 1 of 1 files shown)


marcoscaceres:

PR #5277, opened 2026-05-09
fix(core/structure): propagate custom heading ID to section element

### LINKED ISSUE #4947: Sections getting different ids from their headings
## Description of problem

https://w3ctag.github.io/privacy-principles/#hl-sensitive-information

The [source](https://github.com/w3ctag/privacy-principles/blob/a274b7db8f1b6bda65c01874015f59a9a0da8193/index.html#L1548) for this heading is

```markdown
## Sensitive Information {#hl-sensitive-information}
```

This generates:

```html
<section id="sensitive-information"><div class="header-wrapper">
<h3 id="hl-sensitive-information"><bdi class="secno">2.4 </bdi>Sensitive Information</h3>
<a class="self-link" href="#hl-sensitive-information" aria-label="Permalink for Section 2.4"></a> ...
```

This causes https://w3c.github.io/webref/ed/headings/privacy-principles.json to omit the `id` that the self-link refers to. (Filed as https://github.com/w3c/reffy/issues/1824)

Only 1 `id` should be generated for this section/heading, and it should match the `id` assigned by the [custom heading id](https://respec.org/docs/#custom-heading-ids).

marcoscaceres:
Closes #4947

When a heading has a custom ID (from {#custom-id} markdown syntax), moves it to the wrapping section element instead of generating a separate auto-ID from the heading text. Ensures section and heading share one canonical ID for fragment navigation and tools like webref that expect section-level IDs.

### DIFF
--- a/src/core/structure.js
+++ b/src/core/structure.js
@@ -164,7 +164,12 @@ function getSectionTree(parent) {
       continue;
     }
     const title = header.textContent;
-    addId(section, undefined, title);
+    if (header.id && !section.id) {
+      section.id = header.id;
+      header.removeAttribute("id");
+    } else {
+      addId(section, undefined, title);
+    }
     sections.push({
       element: section,
       header,


marcoscaceres:

PR #5278, opened 2026-05-09
fix(core/caniuse): use dl markup and CSS Grid layout

### LINKED ISSUE #4090: Caniuse feature improvements
The caniuse feature could use a couple of improvements, which may help with both layout and accessibility: 
 
 * instead of being divs, we should probably use another `<dl>`
 * we should redo the layout using CSS grid instead.
 * The mobile view still needs improvement. In particular, the "more info" link should be on it's on row.

marcoscaceres:
Closes #4090

Replaces div-based layout with semantic dl (dt for browser type, dd for browser cells). Switches from flexbox to CSS Grid with auto-fit columns. More info link gets its own row on mobile.

### DIFF
--- a/src/core/caniuse.js
+++ b/src/core/caniuse.js
@@ -170,21 +170,15 @@ async function processJson(json, { feature }) {
   ]);
   const toBrowserCell = browserCellRenderer(feature);
   results.reduce(toBrowserCell, groups);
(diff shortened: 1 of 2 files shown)


marcoscaceres:

PR #5281, opened 2026-05-09
fix(core/structure): propagate custom heading ID to section element

### LINKED ISSUE #4947: Sections getting different ids from their headings

marcoscaceres:
Closes #4947

When a heading has an explicit id and its section has none, move
the id from the heading to the section.

### DIFF
--- a/src/core/structure.js
+++ b/src/core/structure.js
@@ -164,7 +164,12 @@ function getSectionTree(parent) {
       continue;
     }
     const title = header.textContent;
-    addId(section, undefined, title);
+    if (header.id && !section.id) {
+      section.id = header.id;
+      header.removeAttribute("id");
+    } else {
+      addId(section, undefined, title);
+    }
     sections.push({
       element: section,
       header,
--- a/tests/spec/core/id-headers-spec.js
+++ b/tests/spec/core/id-headers-spec.js
@@ -46,15 +46,21 @@ describe("Core - ID headers", () => {
       let ariaLabel = anchor.getAttribute("aria-label");
       expect(ariaLabel).toBe("Permalink for this Section");
 
-      const custom = doc.querySelector("#custom-id + a.self-link");
+      const custom = doc.querySelector(
+        "#custom-id > div.header-wrapper > h2 + a.self-link"
+      );
       ariaLabel = custom.getAttribute("aria-label");
       expect(ariaLabel).toBe("Permalink for Section 3.");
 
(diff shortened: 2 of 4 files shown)


marcoscaceres:
@copilot can you fix whatever is causing CI to fail?

marcoscaceres:
The change this PR makes (moving the author's heading id onto the section, and generating a fresh section-number-based id for the heading) does not reduce the number of ids emitted; it swaps which element holds the author's one, and changes the ids in every ReSpec document. The reason it was worth doing, that webref omitted the custom id, no longer applies: w3c/reffy#1824 landed and webref now exposes it as `alternateIds`. See the comment on #4947.

marcoscaceres:
Reopening: I closed this too early.

The close reasoning was that the reported breakage is gone, since w3c/reffy#1824 landed and webref now exposes the author's id as `alternateIds`. That part still holds, but it was never @martinthomson's point. On #4947 he argues that the section and the heading are genuinely different targets, demonstrable with a `:target` rule, and that the useful anchor is the section, which is the direction this PR takes: the author's id moves onto the `<section>`.

Where this PR differs from his suggestion is the heading's generated id. This PR leaves it to be generated from the numbered heading text, which is the `x2-2-another-title` shape ReSpec already produces, so it moves whenever sections are renumbered. He suggests a reserved space instead, "a single-character plus hyphen, like `h-sensitive-information`", which would be predictable.

So the open question is whether to adopt that reserved prefix here. Worth noting either way: this changes the ids emitted by every ReSpec document, and @tabatkins' original question was why a second id is generated at all.

PR #5282, opened 2026-05-09
fix(core/caniuse): mark up browser support groups as a dl

### LINKED ISSUE #4090: Caniuse feature improvements

marcoscaceres:
Closes #4090

The browser support groups were nested `<div>`s; they are now a `<dl>` with a `<dt>` label and `<dd>` of browser cells per group, which is what the issue asked for and reads better to assistive technology. The existing flex layout and its legend-on-a-rule look are kept as-is, and on narrow screens the groups stack full width with the "More info" link on its own row.

Not switched to CSS grid, despite the issue suggesting it: the flex layout already produced the intended appearance, and several grid rewrites each regressed it (misaligned group labels, the rule striking through the label text, pills scattered across columns). Two rendering bugs found along the way are fixed here too. The "More info" link carried `caniuse-cell`, which sets white text with a background gradient built from a variable only defined on the support-level classes, so it rendered white on white; and the group labels used `var(--bg, white)`, but `--bg` is not defined anywhere in ReSpec, so they painted a white patch on dark-mode pages.

### DIFF
--- a/src/core/caniuse.js
+++ b/src/core/caniuse.js
@@ -170,21 +170,15 @@ async function processJson(json, { feature }) {
   ]);
   const toBrowserCell = browserCellRenderer(feature);
   results.reduce(toBrowserCell, groups);
-  const out = [...groups]
+  const entries = [...groups]
     .filter(([, arr]) => arr.length)
     .map(
       ([key, arr]) =>
-        html`<div class="caniuse-group">
-          <div class="caniuse-browsers">${arr}</div>
-          <div class="caniuse-type"><span>${key}</div>
-        </div>`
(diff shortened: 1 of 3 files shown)


marcoscaceres:
Before: 
Screenshot 2026-08-18 at 5 52 49 PM

After: 
Screenshot 2026-08-18 at 5 50 49 PM

PR #5285, opened 2026-05-09
fix(core/title): detect id="title" collision and use fallback

### LINKED ISSUE #4712: More flexible title id generation
## Description of problem

https://w3c.github.io/wot-thing-description/ontology/td

**What happened (e.g., it crashed)?:** ReSpec always inserts the id title for the document title. However, if there is already a section id with value title in the document, it causes issues with linking to that section.

**Expected behavior (e.g., it shouldn't crash):** In such cases, respec should display a warning. It can also try to iterate and add `title2` etc. Even better would be a configuration option to set a custom title id.

1. Go to https://w3c.github.io/wot-thing-description/ontology/td#title
2. Expect to land at section 4.3.12 but the window moves to the top of the page, i.e. to the document title

Note: I am not sure why the title needs an id. Maybe removing that addition of the id can make sense?

marcoscaceres:
Closes #4712

When another element already uses id="title", emit an error and
fall back to id="document-title" for the h1.


src/core/title.js
```diff
@@ -51,7 +51,18 @@ export function run(conf) {
   }
 
   // Decorate the spec title
-  if (!h1Elem.id) h1Elem.id = "title";
+  if (!h1Elem.id) {
```
marcoscaceres, line 54: This needs a test. @copilot, can you add a test for this?

PR #5290, opened 2026-05-10
fix(w3c/headers): always show history link without checking if it exists

### LINKED ISSUE #4486: History link is not present if there's a shortname change
## Description of problem

From time to time, a document changes its shortname to a new one. In such cases, the new history link (using the new shortname) will only exist after the publication of the document. Until then, respec will not display the history link as [it's checking if the link does return a 200](https://github.com/w3c/respec/blob/main/src/w3c/headers.js#L763).

While it's good to check if links don't return 404s, it looks like this is not necessary here. Pubrules will ignore the return code of the history link if there's a shortname change.

marcoscaceres:
Closes #4486


marcoscaceres (commented):
Approving

PR #5291, opened 2026-05-10
fix(core/utils): register unquoted form for quoted enum-value dfns

### LINKED ISSUE #4615: Defining quoted enum values should be supported
## Description of problem

Given an enum with, say, the enum-value of "standard", it would be nice to support:

```
<dfn>"standard"</dfn>
```

That's not supported (or doesn't work). 

It should be possible to support that, if the context is an enum.

marcoscaceres:
Closes #4615

### DIFF
--- a/tests/spec/core/webidl-spec.js
+++ b/tests/spec/core/webidl-spec.js
@@ -1048,6 +1048,63 @@ enum EnumBasic {
       );
       expect(doc.getElementById("idl-def-enumbasic")).toBeTruthy();
     });
+
+    it("links quoted enum value definitions to IDL", async () => {
+      // Authors sometimes write <dfn>"value"</dfn> with surrounding quotes to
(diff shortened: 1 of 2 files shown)


tests/spec/core/webidl-spec.js
```diff
@@ -1048,6 +1048,63 @@ enum EnumBasic {
...
+
+    it("links quoted enum value definitions to IDL", async () => {
+      // Authors sometimes write <dfn>"value"</dfn> with surrounding quotes to
+      // mirror WebIDL syntax. ReSpec must still match these dfns to the IDL
+      // enum member (which webidl2 reports without quotes).
+      const body = `
+        <pre class="idl">
+          enum QuotedEnum {
+            "alpha",
+            "beta with spaces"
```
marcoscaceres, line 1060: Although this is legal, this should warn, and is strongly discouraged by the IDL spec.

PR #5305, opened 2026-05-15
fix(core/cddl): use conforming prose dfn attributes

### LINKED ISSUE #5303: bug: core/cddl recognises non-conformant HTML as authoring shortcut
`core/cddl` accepts `<dfn cddl-type>name</dfn>`, `<dfn cddl-key for="rule">key</dfn>`, and `<dfn cddl-value for="rule">"value"</dfn>` as authoring shortcuts. These ask spec authors to write non-conformant HTML: `cddl-type`, `cddl-key`, and `cddl-value` are unprefixed custom attributes (the HTML spec reserves `data-*` for that), and `for=` is only defined on `<label>` and `<output>`, not `<dfn>`.

ReSpec strips the attributes during processing so its output is conformant, but spec source files using these shortcuts won't validate.

Source: [`src/core/cddl.js#L542-L569`](https://github.com/speced/respec/blob/main/src/core/cddl.js#L542-L569). Tests asserting these forms work: [`tests/spec/core/cddl-spec.js#L654-L709`](https://github.com/speced/respec/blob/main/tests/spec/core/cddl-spec.js#L654-L709).

aqilaziz:
Closes #5303.

## Summary
- Use conforming data-dfn-type / data-dfn-for prose definitions for CDDL prose-level dfns.
- Map generated CDDL ids to the actual prose dfn ids so CDDL blocks link to existing prose definitions without creating duplicate block dfns.
- Update CDDL specs and the W3C build artifact.

## Tests
- pnpm exec prettier --check src/core/cddl.js tests/spec/core/cddl-spec.js
- pnpm exec eslint src/core/cddl.js tests/spec/core/cddl-spec.js
- pnpm build:w3c
- $env:BROWSERS='ChromeHeadless'; pnpm exec karma start ./tests/spec/karma.conf.cjs --single-run --grep "Core - CDDL"


marcoscaceres:
@aqilaziz, would you mind removing the build/ directory changes?

aqilaziz:
Removed the `builds/` artifact changes as requested and rebased the branch on current `main`.

The PR diff now only includes:
- `src/core/cddl.js`
- `tests/spec/core/cddl-spec.js`

Verification:
- `git diff --check upstream/main..HEAD`

I did not rerun the full Node lint locally because dependencies are not installed in this checkout and this machine is currently low on disk space.

aqilaziz:
﻿Updated this PR to address the failing checks:

- Replaced the invalid prose-level CDDL example in `examples/basic*.html` with inline CDDL references.
- Updated CDDL processing so prose-level `dfn[data-dfn-type='cddl-*']` definitions are preferred over generated block definitions, avoiding duplicate definitions/link-to-dfn errors.
- Rebuilt `builds/respec-w3c.js`.

Verified locally:

- `pnpm lint`
- `pnpm test:build`
- `pnpm build:w3c`
- `pnpm test:headless`
- `BROWSERS=ChromeHeadless pnpm test:integration -- --grep "Core - CDDL"`
- `node ./tools/respec2html.js examples/basic.built.html <temp> --verbose --timeout 30` + `java -jar vnu-jar <temp>`

marcoscaceres:
@aqilaziz builds/ are back 😊… also, have the agent take into context the original issue. 

The problem is not in the duplicate definitions (duplicates are an authoring error). Definitions should only appear once per document, not that the once in prose are preferred. 

The bug is that we accidentally allowed non-conforming HTML attributes in the original design (e.g, “for=“). 

Does that make sense?

aqilaziz:
@marcoscaceres Thanks for the clarification, and sorry for the churn on `builds/` — I ran `pnpm build:w3c` locally to verify and it looks like that regenerates and stages `builds/` even though it's only meant to be committed at release time (I see it's only touched in the `vX.Y.Z` tag commits, e.g. `3c4c85f`). I'll exclude it from the diff going forward.

Re-reading #5303 more carefully: I understand now — the actual bug is that `normalizeProseDfns()` in `src/core/cddl.js` accepts and silently processes non-conforming shorthand attributes (`cddl-type`, `cddl-key`, `cddl-value`, and `for=` on `<dfn>`) as an "authoring convenience", which means spec source using this pattern doesn't validate as conformant HTML even though ReSpec's *output* does (since the attributes get converted/stripped during processing). My previous fix addressed a symptom (duplicate definitions when both a prose dfn and a generated block dfn exist for the same term) rather than this root design issue, and conflated "prefer prose" with "the duplicate itself is fine" — which you're right isn't the correct framing; duplicates should be flagged as an authoring error, not silently resolved by preference.

Before I push another attempt, I'd rather confirm the intended direction, since this affects the public authoring API for `core/cddl` and there are presumably existing specs using the shorthand:

1. **Remove the shorthand entirely** (breaking change) — authors must write `data-dfn-type`/`data-dfn-for` directly; `normalizeProseDfns()` goes away, and `cddl-type`/`cddl-key`/`cddl-value`/`for=` on `<dfn>` are no longer recognized at all.
2. **Keep the shorthand working but emit a `pub()` lint warning** pointing authors at the conforming form, with a plan to remove it in a future major version.
3. Something else you have in mind — happy to follow whatever approach fits ReSpec's usual deprecation process.

Let me know which direction you'd like and I'll rework the PR accordingly (and this time keep `builds/` out of the diff).

PR #5333, opened 2026-06-08
fix: preserve issueBase when set in config

sanderke:
Setting `issueBase` in config works for issue boxes, but got ignored for the URLs placed in the header of the document.


marcoscaceres:
Would you mind adding a test? 🙏

PR #5344, opened 2026-07-03
feat(core/inlines): render inline code in [[[…|alias]]] link text

### LINKED ISSUE #5343: Support inline code (backticks) in [[[…|alias]]] link text
_Requested by @ZoeBijl._

I'd like to style the visible text of an autolink (`[[[…|text]]]`) as code, the same way `[= `term` =]` already lets me. Right now the backticks are printed literally:

```
[[[css-display#the-display-properties|`display`]]]
```

renders the backticks as-is instead of wrapping `display` in `<code>`.

Run the alias text of `[[[SPEC#id|text]]]` / `[[[#id|text]]]` through the same inline processing that `[= =]` uses, so `` `code` `` in the alias becomes `<code>`. This would make the two inline-link syntaxes consistent — `[= `link` element =]` already produces `<code>link</code>` (via `processInlineContent`), but `[[[…|`text`]]]` does not.

**Additional context**

Note the two `[[[…]]]` forms resolve differently, so the fix isn't uniform:
- same-document `[[[#id|text]]]` inserts the alias verbatim as the link's text content — the easy case.
- cross-spec `[[[SPEC#id|text]]]` uses the alias as `data-lt` (an xref lookup key), and the displayed text ultimately comes from xref resolution, so "linking text" and "display styling" would need to be separated.

Wrapping the whole autolink in `<code>` works today:

```
<code>[[[css-display#the-display-properties|display]]]</code>
```

marcoscaceres:
Closes #5343

Author aliases in the `[[[…|alias]]]` autolink now support inline code, so backticks in the alias render as a `<code>` element instead of appearing literally, matching how the `[= =]` inline-link syntax already does it. A shared `setInlineContent` helper applies the processed alias at the three sites that emit alias display text (same-document, cross-spec section, and cross-spec spec links).

### DIFF
--- a/tests/spec/core/inlines-spec.js
+++ b/tests/spec/core/inlines-spec.js
@@ -631,6 +631,90 @@ describe("Core - Inlines", () => {
     expect(anchor.textContent).toBe("see this section");
   });
 
+  it("processes backticks in [[[#id|`code`]]] alias text", async () => {
+    const body = `
+      <section id="my-section">
+        <h2>My Section Heading</h2>
+      </section>
+      <section>
+        <h2>References</h2>
+        <p id="output">[[[#my-section|the \`display\` property]]]</p>
+      </section>
+    `;
+    const doc = await makeRSDoc(makeStandardOps(null, body));
+    const anchor = doc.querySelector("#output a[href='#my-section']");
+    expect(anchor).toBeTruthy();
+    expect(anchor.querySelector("code").textContent).toBe("display");
+    expect(anchor.textContent).toBe("the display property");
+  });
+
+  it("processes backticks amid surrounding, multi-line text in [[[#id|…]]]", async () => {
+    const body = `
+      <section id="my-section">
+        <h2>My Section Heading</h2>
+      </section>
+      <section>
+        <h2>References</h2>
+        <p id="output">[[[#my-section|foo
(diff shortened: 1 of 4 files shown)


tests/spec/core/inlines-spec.js
```diff
@@ -631,6 +631,90 @@ describe("Core - Inlines", () => {
...
+    `;
+    const doc = await makeRSDoc(makeStandardOps(config, body));
+    const anchor = doc.querySelector("#output a[href]");
+    expect(anchor).toBeTruthy();
+    expect(anchor.href).toBe("https://fetch.spec.whatwg.org/#data-fetch");
+    expect(anchor.querySelector("code").textContent).toBe("fetch");
+    expect(anchor.textContent).toBe("the fetch method");
+  });
+
+  it("processes backticks in [[[SPEC|`code`]]] alias text", async () => {
```
sidvishnoi, line 696: This one feels not needed (use case).

sidvishnoi (approved):

tests/spec/core/inlines-spec.js
marcoscaceres: I can see myself using this... like [[fetch|`fetch()`]]].

PR #5373, opened 2026-07-29
chore(deps-dev): upgrade eslint to 10

### LINKED ISSUE #5369: eslint v9 reaches end of life on 2026-08-06; upgrade blocked by eslint-plugin-import
The eslint docs now carry a notice that v9.x end-of-life is 2026-08-06 and will not be maintained after that. We are on 9.39.4, and npm confirms the shape: eslint's latest is 10.8.0, with 9.39.5 published under a maintenance dist-tag.

We are already on flat config, eslint.config.js, with no .eslintrc anywhere, so that part of the migration is done. The blocker is plugin peer support:

- eslint-config-prettier declares eslint >=7.0.0, fine
- eslint-plugin-prettier declares eslint >=8.0.0, fine
- eslint-plugin-import declares ^2 || ^3 || ^4 || ^5 || ^6 || ^7.2.0 || ^8 || ^9, and 2.32.0 is its newest release, so it caps at eslint 9
- eslint-plugin-jasmine declares no eslint peer range at all, so it is undeclared rather than blocked

eslint-plugin-import-x is the likely path: it declares eslint ^8.57.0 || ^9.0.0 || ^10.0.0 and ships the extensions rule, which is the only import/* rule we enable. That makes it a plugin swap plus a rule-key rename rather than a rewrite, though it has not been tested here.

marcoscaceres:
Closes #5369

Upgrades eslint to 10.8.0 ahead of v9 reaching end of life on 2026-08-06. eslint-plugin-import is replaced with eslint-plugin-import-x, because the former's peer range stops at eslint 9 and 2.32.0 is its newest release; import-x ships the extensions rule, which is the only import rule this config enables.

The pnpm overrides change is a prerequisite, not scope creep: eslint 10 pulls minimatch 10, which needs brace-expansion 5.x's named export, while the previous path-keyed override forced the 1.x default-export shape and made eslint crash on startup. Keying the overrides by version range instead lets both major lines coexist. Four minimatch overrides that resolved nothing are also removed.

eslint 10's updated recommended config reported nine errors, all genuine dead code. Six were assignments never read, including an else branch in dfn-finder.js and a `let lt` in webidl.js that has been unconditionally undefined since the branch assigning it was deleted in 2022, while still being interpolated into a data-lt attribute. Removing it is byte-identical output, since hyperhtml never attaches an attribute whose value is nullish; verified by rendering an IDL fixture before and after and diffing.

The last was a caught error thrown without its cause. Attaching the cause alone would have duplicated the stack, since the message already embedded it, so the message no longer embeds the stack and the CLI's fatal logger now walks the cause chain instead. The trace appears once, and the underlying stack is still reported.


marcoscaceres:
@copilot can you rebase?

src/core/webidl.js
```diff
@@ -62,7 +61,6 @@ const templates = {
       data-link-type="${type === "_IDL_" ? "idl" : type}"
       data-xref-type="${type}"
       data-cite="${cite}"
-      data-lt="${lt}"
```
sidvishnoi, line 65: Good catch!

sidvishnoi (approved):

PR #5374, opened 2026-07-29
Guard caniuse table rendering against missing .head dl element

### LINKED ISSUE #5370: core/caniuse silently drops the browser-support table when .head dl is missing
In `src/core/caniuse.js`, the browser-support table is appended with an optional-chaining call:

```js
const headDlElem = document.querySelector(".head dl");
// ...
headDlElem?.append(...definitionPair.childNodes);
```

If `.head dl` is absent, the `?.` swallows the failure and the module carries on, so the entire caniuse browser-support table is silently dropped. The author gets no warning that the feature they configured produced nothing.

The sibling module handles the same situation correctly. In `src/core/implementation-status.js`:

```js
if (!headDlElem) return;
```

That returns early rather than continuing with work that cannot land anywhere.

Worth making caniuse either report the problem via `showError`/`showWarning` or return early the way implementation-status does, instead of failing silently.

Noted while upgrading TypeScript, which flagged a related unguarded `querySelector` result further down the same function. That guard is being fixed separately; this one is the pre-existing case and is not a regression.

brian-soltani:
run() assumed document.querySelector(".head dl") always finds an element and used optional chaining on headDlElem to avoid crashing when it doesn't. That silently no-ops instead of surfacing the problem, and it still throws later since headDlElem is read unguarded elsewhere in the function.

This adds an early return with a showWarning() call when the .head dl element isn't found, matching the existing pattern used in implementation-status.js, and removes the now-redundant optional chaining on the two remaining headDlElem accesses.

Closes #5370


marcoscaceres (changes requested):
Agree with copilot. This needs a test 🙏 otherwise looks good.

PR #5377, opened 2026-08-04
docs: add an AI policy and agent instructions

marcoscaceres:
Adds an AI policy, and the two files GitHub documents for telling a coding agent how this project works.

The policy asks for disclosure, one concern per pull request, an honest accounting of what an issue asked for, and that a test offered as proof of a fix can actually fail without it. It also says what the maintainers hold themselves to when we generate work ourselves, which is stricter than what we ask of anyone else, and it asks nothing of contributors beyond one model and their own judgment. `CONTRIBUTING.md` points at it, since that is the file an agent reads first.

`.github/copilot-instructions.md` carries the things an agent cannot infer: that karma loads bundles from `builds/` and there is one per profile, so a source change does nothing until it is rebuilt, and that `BROWSERS` must be set or the suite hangs after printing `START:` with nothing explaining why. `.github/workflows/copilot-setup-steps.yml` pre-installs dependencies and keeps puppeteer's browser inside the workspace, since the default cache location is not guaranteed to survive into the agent's session.

Reviewed by Gemini and Copilot as well as by hand. Between them they found a contradiction between two sections of the policy, a labeling rule that required triage access most contributors do not have, a missing privacy rule, and the hang above, which was the worst of them.

Written with AI: Claude drafted the policy and the two configuration files, Gemini and Copilot reviewed them, and I read and ran everything before pushing. Per AI_POLICY.md.


.github/workflows/copilot-setup-steps.yml
```diff
@@ -0,0 +1,54 @@
...
+      - uses: actions/setup-node@v6
+        with: { node-version-file: ".nvmrc", cache: pnpm }
+
+      # Deliberately NOT setting PUPPETEER_SKIP_DOWNLOAD, though the CI workflows
+      # do. CI can skip the download because it points PUPPETEER_EXECUTABLE_PATH at
+      # the runner's Chrome for the same job. Variables set here do not reach the
+      # agent's later commands, so skipping would leave it with neither a bundled
+      # browser nor a path to one. Paying for the download once makes the
+      # environment self-sufficient.
+      - run: pnpm i --frozen-lockfile
```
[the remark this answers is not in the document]
marcoscaceres: Fixed, and you were right about the mechanism. GitHub documents the honored job keys as steps, permissions, runs-on, services, snapshot and timeout-minutes, so a job-level env block is silently dropped when Copilot runs these as setup steps. CI did not catch it because a normal Actions run does honor job env, which made the workflow look correct while the path that actually matters would have been unset.

PUPPETEER_CACHE_DIR is now set on each step that needs it: the install, the explicit browser install, and the check. The job-level env block is gone.

Related, while fixing this: relying on puppeteer's postinstall to provision the browser did not work either. On a warm pnpm store the install finished in about 2.5 seconds and downloaded nothing, because pnpm caches that a build script ran while the browser it fetches lands outside node_modules. There is now an explicit `npx puppeteer browsers install chrome` step.

.github/copilot-instructions.md
```diff
@@ -0,0 +1,107 @@
...
+
+This is the whole sequence. The build is part of it, and so is `BROWSERS`:
+
+```bash
+pnpm i --frozen-lockfile
+pnpm lint                 # tsc -p src/jsconfig.json && eslint .
+pnpm build:w3c && pnpm build:geonovum && pnpm build:aom && pnpm build:dini
+BROWSERS=ChromeHeadless pnpm test        # unit then integration, via karma
+pnpm test:build                          # the builder tool
+pnpm test:headless                       # renders examples through puppeteer
```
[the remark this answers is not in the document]
marcoscaceres: Fixed. The build-and-test block now starts with `export PUPPETEER_CACHE_DIR="$PWD/.cache/puppeteer"`, and there is a paragraph saying to set it for anything that launches a browser, naming `pnpm test:headless` and `tools/respecDocWriter.js` specifically, with the reason: the browser is provisioned in the workspace rather than the home cache, so puppeteer will not find it otherwise.

marcoscaceres:
Also acted on the two suppressed comments, since both were right.

The puppeteer check step now sets `PUPPETEER_CACHE_DIR` explicitly rather than inheriting it. Same root cause as the thread above: job-level env is not honored, so the check would have validated a location nothing had populated.

And `AI_POLICY.md` now reads "If AI generated any of the logic in your contribution", which fixes the ambiguity.

Two things running it afterwards found that the review could not have predicted. Putting the browser inside the workspace means eslint walks it, producing 12 errors from Chrome's own bundled scripts, including `'chrome' is not defined` and `eval can be harmful`. `.gitignore` was not enough on its own, since eslint does not consult it, so `.cache/**` is now in eslint's ignores alongside `builds/**` and `js/**`.

The other is that the workspace cache is about 340 MB and is not shared between checkouts. That is the right trade for an ephemeral agent environment, but wrong for a human with several worktrees, so the instructions now say to skip the export if you already have a browser in puppeteer's default cache.

.github/workflows/copilot-setup-steps.yml
```diff
@@ -0,0 +1,67 @@
...
+      - name: Check puppeteer has a browser
+        env:
+          PUPPETEER_CACHE_DIR: ${{ github.workspace }}/.cache/puppeteer
+        run: |
+          node -e "
+          const fs = require('fs');
+          Promise.resolve(require('puppeteer').executablePath())
+            .then(p => { fs.accessSync(p); console.log('browser ok:', p); })
+            .catch(e => { console.error('no usable browser:', e.message); process.exit(1); })
+          "
```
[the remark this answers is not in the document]
marcoscaceres: Adopted, though the premise does not hold, and it is worth separating the two.

`require('puppeteer')` does not throw here. `.nvmrc` pins Node 24, which supports `require(esm)`, and I ran it: `require works, keys: function` against puppeteer 25.1.0. The CI evidence agrees. In the earlier failing run of this workflow, the probe got as far as reporting `no usable browser: ENOENT ... chrome-linux64/chrome`, which is a filesystem error from `accessSync` after `executablePath()` resolved. It could not have reached that point if the `require` had failed with `ERR_REQUIRE_ESM`.

So this was not a live failure. I have switched to `import('puppeteer')` anyway, because it removes a dependency on `require(esm)` support if the pinned Node version ever moves, and it costs nothing. Verified before pushing: `browser ok: puppeteer/chrome/mac_arm-149.0.7827.22/...`, exit 0.

Also took both alt-text comments. They were right that mine described the scene at length when the caption is the point. Both files now read `alt="Pixel-art robot grinning in front of a burning forest, captioned "AI is welcome here.""`, down from 146 characters to 96. I kept the fire in, because the joke is the fire and a screen reader user should get to be in on it rather than being told only that there is a robot.

.github/copilot-instructions.md
```diff
@@ -0,0 +1,118 @@
...
+**Set `PUPPETEER_CACHE_DIR` to `$PWD/.cache/puppeteer` for anything that launches a
+browser**, which includes `pnpm test:headless` via `tools/respecDocWriter.js`. The
+agent environment provisions the browser there rather than in the default home
+cache, so puppeteer will not find it unless pointed at the same place.
+
+If you are a human with a browser already in puppeteer's default cache, you do not
+need that line, and you should skip it: the workspace copy is around 340 MB and does
+not get shared between checkouts.
+
+Karma reads the bundles in `builds/`, not `src/`. A source change has no effect on
```
sidvishnoi: Note: Except for unit tests - which use src directly. Faster flow.

.github/copilot-instructions.md
```diff
@@ -0,0 +1,118 @@
...
+empty string then flows on into an `html` template and renders as an empty text
+node, so the bug shows up later as missing output rather than as a null failing
+where it happened.
+
+Write en-US English everywhere, including comments and identifiers: behavior,
+color, license, center, analyze, initialize, serialize. Leave existing en-GB
+spelling alone when it is load-bearing, such as an established identifier or a
+verbatim quotation.
+
+Run `npx prettier --write` on changed files. CI fails on unformatted code.
```
sidvishnoi, line 51: `pnpm format --write`

.github/copilot-instructions.md
```diff
@@ -0,0 +1,118 @@
...
+
+The house style is in [AI_POLICY.md](../AI_POLICY.md) under "Write it the way the
+project writes it". In short:
+
+- Pull request descriptions are plain prose. No `## Summary`, no `## Changes`, no
+  `## Test plan`, no emoji, no bold section labels. A closing reference goes on
+  the first line.
+- `Closes #N` only when every ask in the issue is delivered, otherwise `Refs #N`.
+  Confirm the number is an issue and not a pull request before citing it.
+- Commit messages are one imperative subject line, lowercase after any prefix, no
```
sidvishnoi: Also mention the pull request title format.

.github/copilot-instructions.md
```diff
@@ -0,0 +1,118 @@
...
+the reason the issue describes, and pass with the fix. Do not assert a literal the
+implementation just set, and do not exercise a path the issue never mentioned
+while the reported path stays uncovered.
+
+One concern per pull request. A locale addition or a drive-by refactor belongs in
+its own PR even when it is one line and obviously correct.
+
+## Adding a module
+
+A new module under `src/core/` exports `name` and a `run(conf)`. `run` may be
```
sidvishnoi: If a module is meant for only certain profile, add to profile folder instead of core.

.github/copilot-instructions.md
sidvishnoi, line 1: Consider not using hard wrap in markdown. Made it harder to review on phone.

AI_POLICY.md
```diff
@@ -0,0 +1,168 @@
...
+
+Nothing in this section is asked of you. It describes how we run our own AI
+generated work, and it is here so you can see the standard we are applying to
+ourselves rather than only to contributions.
+
+If you are contributing, you need one model and your own judgment. You are never
+expected to own a second AI subscription, or to orchestrate models against each
+other, to send us a patch.
+
+### The test is written by a different model than the fix
```
sidvishnoi, line 78: Me likey

CLAUDE.md
```diff
@@ -0,0 +1,124 @@
+# ReSpec
+
+Browser-based tool that generates W3C specifications. Authors write HTML plus a `<script>` tag; ReSpec handles boilerplate, cross-references, bibliography, and validation. Around half of W3C standards are produced with it, so a regression here reaches a lot of documents.
+
+Build, test, code style, and the house style for pull requests, commits and comments live in the file below, shared with other agents rather than duplicated:
+
+@.github/copilot-instructions.md
```
sidvishnoi, line 7: This can be too much in context. Claude might not even read the full file.

CONTRIBUTING.md
```diff
@@ -1,3 +1,7 @@
-Everyone is welcome to contribute.
+<img src="assets/ai-is-welcome-here.png" alt="Pixel-art robot grinning in front of a burning forest, captioned &quot;AI is welcome here.&quot;" width="220" align="right">
```
sidvishnoi, line 1: Let's not add this image here. It's enough in AI Policy.

CLAUDE.md
```diff
@@ -0,0 +1,124 @@
...
+
+```js
+// Standard test setup
+import {
+  makeRSDoc,
+  makeStandardOps,
+  errorFilters,
+  warningFilters,
+} from "../SpecHelper.js";
+const errors = errorFilters.filter("module/name");
```
sidvishnoi, line 64:
```ts
const pluginErrors = errorFilters.filter("module/name");
// ...
const errors = pluginErrors(doc);
```

CLAUDE.md
```diff
@@ -0,0 +1,124 @@
...
+
+- `ls` new files to confirm they exist (rebase can silently drop additions)
+- `grep` for key additions (imports, function calls) in conflict-resolved files
+- Run `pnpm lint` before pushing — never assume the rebase was clean
+
+### CI lint failures on PRs
+
+Almost always caused by the branch being behind main. Fix:
+
+1. `git rebase main` (resolve conflicts — for `builds/` files, always take `--ours`)
```
sidvishnoi, line 94: Maybe suggest reject any changes to `builds/` and build again, instead of trying to rebase (consumes less tokens, is faster).

CLAUDE.md
```diff
@@ -0,0 +1,124 @@
...
+2. `git diff main --name-only | xargs npx prettier --write`
```
sidvishnoi, line 95:
```suggestion
2. `git diff main --name-only | xargs pnpm format --write`
```

CLAUDE.md
```diff
@@ -0,0 +1,124 @@
...
+3. `git push --force-with-lease`
+
+Do NOT commit `builds/` files — CI rebuilds them. On rebase conflicts in build artifacts, take main's version.
+
+## Pre-PR checklist (learned from Copilot reviews)
```
sidvishnoi, line 100:
```suggestion
## Pre-PR checklist
```

sidvishnoi, line 100: Perhaps don't recommend automatic pushes (keep in CLAUDE.local.md). This will encourage will be send PRs manually, which reduces spam/low quality chances.

CLAUDE.md
```diff
@@ -0,0 +1,124 @@
...
+
+**DOM node reuse:** `html` tagged template literals create DOM nodes. A DOM node can only exist in one place. If an SVG/element is used in a loop (e.g., browser icons), use a factory function (`() => html\`...\``) instead of a shared constant.
+
+**Save/export data flow:** When `removeOnSave` or `beforesave` rewrites content, ensure the rewritten HTML preserves feature-specific URLs and context, not just a generic fallback link.
+
+**Unused properties:** Don't declare properties in constants (e.g., `label` in a Map entry) that are never read. Copilot flags these consistently.
+
+## Cross-spec headings API
+
+`POST /xref/headings` on respec.org looks up section heading text by `{spec, id}`, sourced from w3c/webref `ed/headings/`. This is what lets `[[[SPEC#id]]]` render the actual heading rather than just the spec title.
```
sidvishnoi, line 124: This is more of spec writing, then ReSpec.

.github/copilot-instructions.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Code style
+
+Prefer functional style over imperative loops: `forEach`, `map`, `filter`, `find`, `reduce`, `some`, `every`. Use an early `return` inside `forEach` rather than `continue`. `NodeList`, `Map` and `Set` have `forEach` natively, so prefer `nodeList.forEach()` over `[...nodeList].forEach()` and avoid the extra array.
+
+Every `querySelector`, `closest`, `getElementById` and `getAttribute` result is possibly null; check before use. Do not paper over it with `?? ""`, because the empty string then flows on into an `html` template and renders as an empty text node, so the bug shows up later as missing output rather than as a null failing where it happened.
+
+Write en-US English everywhere, including comments and identifiers: behavior, color, license, center, analyze, initialize, serialize. Leave existing en-GB spelling alone when it is load-bearing, such as an established identifier or a verbatim quotation.
+
+Run `npx prettier --write` on changed files. CI fails on unformatted code.
```
marcoscaceres:
```suggestion
Run `pnpm format --write` on changed files. CI fails on unformatted code.
```

ai-pr-proof-requirement.md
```diff
@@ -0,0 +1,107 @@
...
+3. The proof must use the same repro command, script, test, or steps before and after the fix.
+4. The before-proof must come from the closest reproducible pre-fix state, using one of these modes:
+   - `base-sha`: run the repro on the PR base commit.
+   - `reverted-fix`: run the repro with the fix reverted while keeping any required new test or harness in place.
+   - `captured-failure`: point at a preserved failing run only when the first two modes are genuinely infeasible.
+5. `base-sha` is preferred. If another mode is used, the proof must say why `base-sha` was not valid.
+6. The before-proof must fail for the reason the bug describes.
+7. The after-proof must pass, or otherwise show that the incorrect behavior is gone.
+8. If the bug is visual, UI-based, rendering-based, or output-based, include before and after screenshots or equivalent rendered artifacts.
+9. If no deterministic repro can be produced in any mode, the PR must be labeled speculative and the PR description must say exactly what blocked proof generation.
```
marcoscaceres:
```suggestion
8. If no deterministic repro can be produced in any mode, the PR must be labeled speculative and the PR description must say exactly what blocked proof generation.
9. If the bug is visual, UI-based, rendering-based, or output-based, a human MUST include a before and after screenshots or equivalent rendered artifacts.
```

.github/copilot-instructions.md
marcoscaceres: Already in from the earlier pass: the build and test section now says the unit suite loads `src/` directly, needs no rebuild, and is the faster loop while iterating.

.github/copilot-instructions.md
marcoscaceres: Done. Worth flagging that there was no `format` script on this branch or on main, so `pnpm format --write` would have failed as written — `package.json` now has `"format": "prettier"`, which makes it work. Say if you'd rather it be named or shaped differently.

.github/copilot-instructions.md
marcoscaceres: Added: titles follow conventional commits, `type(scope): summary`, enforced by the Check PR Title workflow.

.github/copilot-instructions.md
marcoscaceres: Added under "Adding a module": `src/core/` only if every profile wants it, otherwise the profile's own folder, as `src/w3c/` already does.

CLAUDE.md
marcoscaceres: Agreed, and fixed at the cause: the `@` import is gone, so those 73 lines are no longer inlined into every session. It is a plain path now, read when a task touches build, test or PR mechanics. I also cut the two spec-authoring sections, so the file is 104 lines with nothing pulled in behind it, down from 124 plus the import.

CONTRIBUTING.md
marcoscaceres: Removed. It only appears in AI_POLICY.md now.

CLAUDE.md
marcoscaceres: Applied. Checked against `tests/spec/core/cddl-spec.js`, which does exactly this: the filter is built once at the top, then called with the doc inside the test.

CLAUDE.md
marcoscaceres: Taken. On a conflict under `builds/` it now says take main's bundles and rebuild only if you need the integration suite locally, rather than resolving a diff of generated code.

CLAUDE.md
marcoscaceres: Taken, with one change I would like your eye on. With a single dot this also lists every file main changed, so on a branch behind main it reformats files the PR never touched: running it as written rewrote nine here, including `pnpm-lock.yaml` and two test fixtures. The doc now uses `git diff main...HEAD --name-only`, which listed 11 files against 94 for the one-dot form.

CLAUDE.md
marcoscaceres: Done.

marcoscaceres: Removed the push step and the "before pushing" framing, so it reads as a checklist to satisfy before opening a PR by hand. Push guidance stays out of the committed file.

CLAUDE.md
marcoscaceres: Agreed, dropped — along with the inline shorthand reference, which is spec-authoring for the same reason.

sidvishnoi (approved):

PR #5378, opened 2026-08-05
fix(w3c/headers): show version URLs as readable IRIs

### LINKED ISSUE #4910: `latestVersion` URLencodes IRIs making them unreadable by people
## Description of problem

https://ofn.gov.cz/aplikace-využívající-otevřená-data/draft/

Latest published version link got URL encoded to:
https://ofn.gov.cz/aplikace-vyu%C5%BE%C3%ADvaj%C3%ADc%C3%AD-otev%C5%99en%C3%A1-data/draft/ making it unreadable

It would stay not encoded, as IRI: https://ofn.gov.cz/aplikace-využívající-otevřená-data/draft/

marcoscaceres:
Closes #4910

The version links in the header now show the human-readable IRI as link text while the `href` keeps the percent-encoded URL, so a spec with non-ASCII characters in its URL reads normally. Applies to all five of the version URLs in that list, not just `latestVersion`: `thisVersion`, `edDraftURI`, `historyURI` and `prevVersion` had the same problem on adjacent lines. Malformed escape sequences are left as-is rather than throwing.


marcoscaceres (commented):
Approving

PR #5379, opened 2026-08-05
fix(core/highlight): use textContent so unrendered code is highlighted

marcoscaceres:
Syntax highlighting used `innerText`, which returns nothing for content a browser has not rendered, so a `<pre>` inside a closed `<details>` was left unhighlighted. It now uses `textContent`, which does not depend on rendering.


marcoscaceres (commented):
approving

PR #5380, opened 2026-08-05
fix(w3c/headers): drop stray brace from TAG finding prevVersion URL

marcoscaceres:
A stray `}` in the template literal meant TAG findings got a previous-version URL ending in a percent-encoded brace, `.../doc/Foo-1977-03-15%7D`. The existing test only asserted `toContain()` on the containing text, so it never saw the trailing character; it now asserts the anchor's full href.

Proof: the same karma run fails in `reverted-fix` mode with `Expected 'https://www.w3.org/2001/tag/doc/Foo-1977-03-15%7D' to be 'https://www.w3.org/2001/tag/doc/Foo-1977-03-15'` and passes on `08ef3ab4` with `TOTAL: 1 SUCCESS`. Reverting only `src/w3c/headers.js` and keeping the test is what produces the before-state, because the test itself is new in this commit. This is a rendering change: the URL is the link's text, so a reader saw the trailing `%7D` in the "Previous version:" row. Verified by probing the rendered DOM in both states.

### DIFF
--- a/src/w3c/headers.js
+++ b/src/w3c/headers.js
@@ -431,7 +431,7 @@ export async function run(conf) {
       const pubDate = ISODate.format(conf.publishDate);
       conf.thisVersion = w3Url(`${latestPath}-${pubDate}`);
       const prevPubDate = ISODate.format(conf.previousPublishDate);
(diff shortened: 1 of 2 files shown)


marcoscaceres:
Approving.

PR #5381, opened 2026-08-06
add integration test for (bug?) nested includes with relative hrefs

gobengo:
Motivation:
* add a test asserting behavior of how nested `data-include` relative hyperlink references are resolved by respec into the actually-fetched URL

Background:
* I used `data-include="includes/section.md"` to include a file that itself had `data-include="example.json"`, thinking this would lead to `includes/example.json` included in my respec document. I was surprised to find my document rendering a 404 error instead, because respec fetched `example.json` not `includes/example.json`.
* [respec docs for data-include](https://respec.org/docs/#data-include) say "Paths are relative to the including document", but this isn't clear for document A that includes a document B that includes a document C. When document B includes document C, is 'the including document' doc B or A?
  * Currently the behavior is untested, but things behave like 'the including document' always means the outer-most including document. The current behavior seems like it may have been a bug/oversight.
  * It seems like `data-include` docs may have been written with the behavior I suspected in mind (paths in nested includes should be resolved relative to the document that includes them, not the top-most include), even if the current implementation behaves differently.
  * I strongly recommend behavior different than the status quo, because that way respec can include documents that themselves include via relative path. That will make it so a file (with nested includes) can be included in more than one respec document, and without having to know about the path of the file that includes it.
 
Questions
* [ ] Is the current behavior a bug or not?
  * If the current behavior is not matching what the docs intended by "Paths are relative to the including document", then I should update this PR so the test asserts not the current behavior, but the intended behavior. If that's appropriate, say so, and I will update the test and make it pass, so that it works like I wanted, and we can consider it a bug fix in the changeling.

Changes
* Adds an integration test for nested `data-include` includes where the included hrefs are relative URIs. Previously, the only test for nested includes used data URIs, which means there is no test asserting behavior of how nested includes build absolute URLs from the relative hrefs.
* Testers can set SPEC_FILE_PATTERN to override the default pattern used by integration tests. This can be used to speed up test runs by only scanning a subset of test file patterns, e.g. `SPEC_FILE_PATTERN='tests/spec/core/data-include-spec.js' pnpm test:integration`

### DIFF
--- a/tests/spec/core/data-include-spec.js
+++ b/tests/spec/core/data-include-spec.js
@@ -133,6 +133,43 @@ describe("Core — Data Include", () => {
     ).toBeTrue();
   });
 
+  it("resolves a nested data-include's relative URL against the top-level document, not the including file's own URL", async () => {
+    // Fixtures live under tests/spec/core/nested-include/:
+    //   outer.html                     <- the document under test
+    //   foo/bar.html                   <- fetched via outer.html's data-include="foo/bar.html"
+    //   includes/data.json             <- resolving "includes/data.json" against outer.html
+    //   foo/includes/data.json         <- resolving "includes/data.json" against foo/bar.html
+    //
+    // foo/bar.html itself has data-include="includes/data.json". This test
+    // asserts which of the two data.json files actually gets fetched, i.e.
+    // whether nested includes resolve relative URLs against the document
+    // that started the include chain, or against the file that declared
+    // the nested data-include.
+    const nestedIncludeUrl = "/tests/spec/core/nested-include/outer.html";
+    const ops = {
+      config: makeBasicConfig(),
+      body: makeDefaultBody(),
+    };
+    const doc = await makeRSDoc(ops, nestedIncludeUrl);
+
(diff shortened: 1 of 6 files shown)


tests/spec/core/data-include-spec.js
```diff
@@ -133,6 +133,43 @@ describe("Core — Data Include", () => {
...
+    expect(outerTarget).toBeTruthy();
+    expect(outerTarget.querySelector("p").textContent).toBe("bar content");
+
+    const innerTarget = doc.querySelector("#inner-target");
+    expect(innerTarget).toBeTruthy();
+    // Current behavior: the nested data-include is resolved relative to
+    // outer.html (the top-level document), *not* relative to foo/bar.html
+    // (the file that declared it) — even though foo/bar.html was itself
+    // fetched from a "foo/" subdirectory.
+    expect(innerTarget.textContent).toContain(
```
gobengo, line 165: If the current behavior is a bug, then I should make this assertion a `not` and remove the `not` from the nest assertion.

PR #5400, opened 2026-08-18
fix(core/worker): resolve highlighter from builds/ in source-module mode

marcoscaceres:
The worker derives the highlighter URL from `import.meta.url`. In the IIFE bundle that resolves to the script src, so `respec-highlight.js` sits next to it and the derivation is right. In source-module mode, which is the dev server and karma, it resolves to `src/core/worker.js`, where no highlighter exists, so the URL points at `src/core/respec-highlight.js` and 404s.

That URL is used twice: for the preload hint, which produces the `404: /base/src/core/respec-highlight.js` warnings in karma output, and for the `fetch()` in `fetchHighlightScript()`. The second one matters more, because the failed fetch falls through to `importScripts()` on the W3C CDN copy. Locally and in CI the worker has therefore been loading the published highlighter rather than the `builds/respec-highlight.js` just built, which makes the highlight tests network-dependent and means a change to the local highlighter is not what gets exercised. That also explains the known CDN flake in the suite.

Deriving against `builds/` in source-module mode fixes both uses at once, and leaves the bundle path untouched since rollup rewrites `import.meta.url` to the script src. The added test asserts the resolved URL is fetchable and same-origin, so a regression fails loudly instead of quietly falling back to the CDN.


marcoscaceres:
AI Generated. Approving. Confirming the 404s for the highlighters are gone from CI.

PR #5403, opened 2026-08-18
chore(deps-dev): upgrade jasmine to 7

marcoscaceres:
Unblocked by #5401. The only thing standing in the way of jasmine 7 was `karma-jasmine-html-reporter`, which assigned to `jasmineStarted` during boot, and jasmine 7 freezes that.

Only the two node-side suites actually run the version being bumped, since `karma-jasmine` depends on its own pinned `jasmine-core@4.6.1` for the browser suites. Neither node suite touches an API jasmine 7 removed. Locally `test:build` passes, lint is clean, and the karma suites are unchanged at 56 unit and 1130 integration.

I could not verify `test:headless` locally: it fails with `listen EPERM` on port 8000, but it fails identically with jasmine 6 installed on this same branch, so it is unrelated to the upgrade. That job running green in CI is the check worth watching here.


marcoscaceres:
Approving. Small fix.

PR #5405, opened 2026-08-19
refactor(core/xref): drop dead-code version-check URL guard, pin construction invariant

anupamme:
## Summary

Clarifies the xref version-check URL construction and adds a regression test.

## Background

This PR originally landed as a security fix (scanner rule `V-002`, flagged as an SSRF at `src/core/xref-db.js:72`). After review — thanks @marcoscaceres — that classification is a **false positive**:

- The version-check URL is built from a **hardcoded constant**: `new URL("meta/version", API_URL)` where `API_URL = "https://respec.org/xref/"` (`src/core/xref.js`). `API_URL` is never reassigned or overridden, so the URL is `https://respec.org/...` **by construction** — it cannot be attacker-influenced.
- There is no `respecConfig` path that changes `API_URL` or the `meta/version` request. The original "if API_URL can be overridden via respecConfig" premise doesn't hold.
- The only author-configurable URL is `conf.xref.url`, which feeds a **different** request (the xref search `POST`) and is an intended, documented, already-tested feature (authors may point xref at their own endpoint). Restricting it would break supported behavior.
- The AWS-metadata scenario doesn't apply: this is a browser-side `fetch()`, not a server-side request.

## Changes

- **Revert** the protocol/hostname guard added to `src/core/xref-db.js` — it was unreachable dead code and needlessly coupled the module to the literal host `respec.org`.
- **Add a regression test** (`tests/spec/core/xref-spec.js`) pinning the real invariant: `API_URL` and the derived `meta/version` URL always resolve to the `https://respec.org` origin.

No security framing and no new URL allowlist. This is a small robustness/documentation change.

### DIFF
--- a/src/core/xref-db.js
+++ b/src/core/xref-db.js
@@ -71,6 +71,10 @@ async function shouldBustCache() {
   }
 
   const url = new URL("meta/version", API_URL).href;
+  const parsedUrl = new URL(url);
+  if (parsedUrl.protocol !== "https:" || parsedUrl.hostname !== "respec.org") {
+    return false;
+  }
   const res = await fetch(url);
   if (!res.ok) return false;
   const lastUpdated = await res.text();
--- a/src/core/xref-db.js
+++ b/src/core/xref-db.js
@@ -71,10 +71,6 @@ async function shouldBustCache() {
   }
 
   const url = new URL("meta/version", API_URL).href;
-  const parsedUrl = new URL(url);
-  if (parsedUrl.protocol !== "https:" || parsedUrl.hostname !== "respec.org") {
-    return false;
-  }
   const res = await fetch(url);
   if (!res.ok) return false;
   const lastUpdated = await res.text();
--- a/tests/spec/core/xref-spec.js
+++ b/tests/spec/core/xref-spec.js
@@ -12,6 +12,7 @@ import {
   makeRSDoc,
   makeStandardOps,
 } from "../SpecHelper.js";
+import { API_URL } from "../../../src/core/xref.js";
 
 describe("Core — xref", () => {
   afterAll(flushIframes);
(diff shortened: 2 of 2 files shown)


marcoscaceres:
I agree with Copilot... this seems to be a false positive and not exploitable.

anupamme:
Thanks for the review. I agree that the current PR doesn’t establish the attacker-controlled data flow needed to call this an exploitable SSRF.

In particular, I made the assumption that API_URL could be overridden through respecConfig, but I haven’t provided a reproducible example showing that this is actually possible in ReSpec. The AWS metadata example also isn’t appropriate without demonstrating a server-side execution path.

I’d like to avoid introducing a restrictive URL allowlist based solely on a scanner finding. Before closing this, would you be open to me reframing the PR around the underlying URL-construction behaviour and adding a regression test, if we can demonstrate a supported configuration path that can influence this request?

If there isn’t such a path, I’m happy to close this as a false positive. Thanks for catching the missing exploitability evidence.

marcoscaceres:
Yeah, go for it!

anupamme:
Thanks again for the review. As discussed, I've reframed this away from the SSRF classification, which was a false positive.

Concretely:

- I've reverted the protocol/hostname guard. The version-check URL is derived from the hardcoded `API_URL` constant (`https://respec.org/xref/`), which isn't overridable via `respecConfig`, so the check was unreachable and it also coupled the module to a literal host.
- I've added a small regression test that pins the actual invariant — `API_URL` and the derived `meta/version` URL always resolve to the `respec.org` origin — so the "not author-configurable" property is documented in the suite.

I've deliberately **not** touched `conf.xref.url`, since that's the intended custom-endpoint feature (and is already covered by the existing test that points xref at a local URL). The PR description has been updated to drop the V-002/SSRF/AWS-metadata framing.

Happy to close instead if you'd prefer not to carry the extra test — but figured pinning the construction invariant was worth keeping.

marcoscaceres:
Yeah, it might be best just to close it. The test doesn't really buy us much here.

PR #5417, opened 2026-08-24
fix(styles): move the self-link text-decoration onto the anchor

### LINKED ISSUE #5415: Section self-link underline: text-decoration: none is on the wrong element
The section self-link (§) beside every heading is underlined in its resting state, and the `text-decoration: none` that is meant to prevent it has no effect because it is declared on the wrong element.

`src/styles/respec.css.js` currently has:

```css
:is(h2, h3, h4, h5, h6) + a.self-link::before {
  content: "§";
  text-decoration: none;
}
```

Text decorations cannot be cancelled by a descendant. CSS Text Decoration 4, [section 2](https://drafts.csswg.org/css-text-decor-4/#line-decoration), states: "They are also not propagated to inline children of inline boxes, although the decoration is applied to such boxes." So the `::before` gets the anchor's underline applied to it while its own `text-decoration` computes to `none` — which is exactly what happens here. Measured on a rendered spec in Safari, the anchor computes to `text-decoration-line: underline` in `rgb(85, 85, 85)`, the `::before` computes to `none`, and the underline is still drawn under the glyph. Declaring `text-decoration: none` on the anchor does remove it.

There is a related inconsistency: `.self-link:hover` already sets `text-decoration: none`, so the § is underlined at rest and loses the underline on hover, which is the reverse of the usual convention.

Suggested fix, one declaration moved rather than added — put it on the existing anchor rule and drop the ineffective one from the `::before`:

```css
:is(h2, h3, h4, h5, h6):not(#toc h2) + a.self-link {
  color: inherit;
  order: -1;
  position: relative;
  left: -1.1em;
  font-size: 1rem;
  opacity: 0.8;
  text-decoration: none;
}
```

Raised by @patrickhlauke while reviewing #5410: https://github.com/speced/respec/pull/5410#issuecomment-5392610142 — "i'd still wonder if it's worth removing the underline (both in dark mode and light mode) for the paragraph symbol, as it just looks a bit ugly, and is now quite faint anyway". Filed separately because he correctly called it scope creep for that PR, which fixes the contrast issue in #5338 and should stay limited to that.

marcoscaceres:
The section self-link's `text-decoration: none` was declared on the `::before` that draws the §, where it has no effect. Per CSS Text Decoration 4, decorations "are also not propagated to inline children of inline boxes, although the decoration is applied to such boxes" — so the anchor's UA underline was still painted under the glyph while the pseudo-element's own `text-decoration` computed to `none`. Moving the same declaration onto the anchor removes it, at rest as well as on hover.

The unit test asserts the anchor's computed `text-decoration-line`, which is the placement that matters; asserting on the pseudo-element would have passed even before this change.

Fixes #5415

### DIFF
--- a/src/styles/respec.css.js
+++ b/src/styles/respec.css.js
@@ -162,11 +162,13 @@ aside.example .marker > a.self-link {
   left: -1.1em;
   font-size: 1rem;
   opacity: 0.8;
+  /* Has to be here, not on the ::before: a decoration is painted across the
+     pseudo-element regardless of what the pseudo-element itself declares. */
+  text-decoration: none;
 }
 
 :is(h2, h3, h4, h5, h6) + a.self-link::before {
   content: "§";
-  text-decoration: none;
 }
 
 :is(h2, h3) + a.self-link {
--- a/tests/unit/core/style-spec.js
+++ b/tests/unit/core/style-spec.js
@@ -57,4 +57,18 @@ describe("Core - style", () => {
     darkToggle.checked = false;
     expectRightSources();
   });
+
+  it("does not underline the section self-link", async () => {
+    // A decoration declared on the anchor is painted across its ::before no
+    // matter what the pseudo-element declares, so it has to be off here.
+    const doc = await makePluginDoc(["/src/core/style.js"], {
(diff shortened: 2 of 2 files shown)


marcoscaceres:
Confirmed visually including dark and light mode. Thanks @patrickhlauke for spotting this.

PR #5420, opened 2026-08-24
fix(styles): stop IDs inside :not() inflating self-link specificity

### LINKED ISSUE #5419: Section self-link hover never becomes fully opaque: :not(#toc h2) outweighs .self-link:hover
Hovering a section self-link is supposed to make it fully opaque, but it never does. `.self-link:hover` loses the cascade to the rule that sets the resting opacity.

In `src/styles/respec.css.js`:

```css
:is(h2, h3, h4, h5, h6):not(#toc h2) + a.self-link {
  /* … */
  opacity: 0.8;
}

.self-link:hover {
  opacity: 1;
  text-decoration: none;
  background-color: transparent;
}
```

`:not()` takes the specificity of its argument, and `#toc h2` contributes an ID, so the first selector scores (1, 1, 3) against the hover rule's (0, 2, 0). The resting `opacity: 0.8` therefore wins even while hovering, and the `opacity: 1` is dead.

Measured with an equal-specificity stand-in for `:hover` (`.self-link.x`, also (0, 2, 0)): computed opacity is `0.8`, i.e. the base rule wins.

The ID weight is accidental — `#toc h2` is there to exclude the table-of-contents heading, not to win specificity battles. Wrapping it in `:where()`, which contributes zero specificity, fixes the cascade without changing what the selector matches:

```css
:is(h2, h3, h4, h5, h6):not(:where(#toc h2)) + a.self-link {
```

Verified in a probe: with that one change the hovered link computes to `1`, and matching is unchanged — a self-link next to an h2 inside `#toc` is still excluded from the rule exactly as before, both with and without hover.

Noticed while working on #5415.

marcoscaceres:
Two selectors in `respec.css.js` carried IDs inside `:not()`, which counts toward specificity, putting both rules far out of reach of ordinary author CSS.

The first was a live bug. `.self-link:hover` sets `opacity: 1`, but it never applied: the ID in `:not(#toc h2)` scored the resting-opacity rule at (1, 1, 3) against the hover rule's (0, 2, 0), so `opacity: 0.8` won even while hovering.

The second has no visible symptom yet. The rule that nudges a self-linked heading left carries four IDs in its `:not()`, also making it (1, 1, 3), so nothing a spec author could reasonably write would override it. Fixed at the same time rather than left as a trap for whoever next adds a competing rule there.

Both are fixed by wrapping the exclusion lists in `:where()`, which contributes no specificity. Verified in a browser probe that matching is unchanged in both cases: headings inside `#toc`, `#abstract`, `#sotd` and `.head` are still excluded, a normal self-linked heading is still offset, and one without a self-link is still untouched.

One caveat on the hover test: `:hover` cannot be synthesised in karma, so it asserts against `.self-link.hover-stand-in`, a selector with the same (0, 2, 0) specificity. That covers the mechanism that was broken, but not the literal hover state. The heading-offset test uses the same technique with a `(0, 2, 0)` stand-in for an author rule.

Fixes #5419

### DIFF
--- a/src/styles/respec.css.js
+++ b/src/styles/respec.css.js
@@ -155,7 +155,10 @@ aside.example .marker > a.self-link {
   left: -.5em;
 }
 
-:is(h2, h3, h4, h5, h6):not(#toc h2) + a.self-link {
+/* The #toc exclusion is wrapped in :where() so it contributes no specificity.
+   As a bare :not(#toc h2) its ID outweighed .self-link:hover and killed that
+   rule's opacity: 1. */
+:is(h2, h3, h4, h5, h6):not(:where(#toc h2)) + a.self-link {
   color: inherit;
   order: -1;
   position: relative;
--- a/src/styles/respec.css.js
+++ b/src/styles/respec.css.js
@@ -150,7 +150,9 @@ aside.example .marker > a.self-link {
   align-items: baseline;
 }
 
-:is(h2, h3, h4, h5, h6):not(#toc > h2, #abstract > h2, #sotd > h2, .head > h2):has(+ a.self-link) {
+/* :where() again, for the same reason as the rule below: these four IDs would
+   otherwise make this (1, 1, 3) and put it out of reach of ordinary author CSS. */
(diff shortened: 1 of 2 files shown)


marcoscaceres:
confirmed visually. Indeed now goes to 1.0 on hover.

PR #5421, opened 2026-08-24
fix(tools): install dependencies before building the release

marcoscaceres:
The release script checks out main, bumps the version, then goes straight to builddeps and builds all four profiles. Nothing installs dependencies in between. Since the bundles pull code directly out of node_modules (src/core/import-maps.js imports node_modules/marked/lib/marked.esm.js, for instance), a release run on a stale node_modules would build the bundles from the old dependency versions while committing the new pnpm-lock.yaml, so the published artifact and its own lockfile would disagree, silently.

This adds a frozen-lockfile install before the build, so the tree is exactly what the lockfile specifies and any drift is a hard failure rather than a quiet one. Verified both ways: the install succeeds on a matching tree, and with a dependency deliberately set to a nonsense version it fails with ERR_PNPM_OUTDATED_LOCKFILE and reports that the specifiers do not match.

Worth noting this is a latent hazard rather than an observed bug. I checked the most recent release by rebuilding its tag from a clean lockfile-exact install and comparing bytes against the bundle that shipped, and they are identical, so that release was built from the right dependencies. It only worked because an install happened to have been run beforehand, which nothing in the script requires.


marcoscaceres:
Confirmed this works locally as expected now. Small fix.

marcoscaceres:
(but also lol... I can't believe we missed this 🤦 🤣 )

PR #5423, opened 2026-08-24
fix(core/biblio-db): await the request and the commit together

### LINKED ISSUE #5422: biblio-db leaves an unhandled rejection when a write transaction aborts
In src/core/biblio-db.js the IndexedDB request and the transaction commit are awaited in sequence, for example `await tx.store.put(details)` followed by `await tx.done`. A transaction has its `done` promise from the moment it is created, and if a request rejects, IndexedDB aborts the transaction and idb rejects `done` as well. The first await has already thrown by then, so the line awaiting `done` is never reached and that rejection has no handler attached.

Three write paths are affected: the expiry sweep in openIdb(), both branches of add() including the expired-entry delete, and clear(). The openIdb() case is the worst of them because it runs at module load while building the `ready` promise, so a failure there surfaces as an unhandled rejection alongside the rejection callers actually see.

Reported by Copilot review on the pull request that introduced the commit awaits. It is a code-level defect rather than one with a convenient reproduction, since it needs a write request to fail against a real IndexedDB.

marcoscaceres:
A transaction has its `done` promise from the moment it is created, so when a request rejects, IndexedDB aborts the transaction and idb rejects `done` too. Awaiting the request and then `done` on the next line means the first await throws, the second line never runs, and that rejection has no handler. In openIdb() this happens at module load, so it surfaces as an unhandled rejection next to the one callers actually see.

All four write paths now await the request and the commit as a pair, for example `await Promise.all([tx.store.delete(id), tx.done])`. The expiry sweep needed its cursor walk wrapped in an async function so the loop and the commit could be awaited together. No sequential commit awaits remain in the file.

Found by Copilot review on the pull request that introduced the commit awaits, where I replied before noticing that pull request had already merged, so the reply landed somewhere that does not contain the fix.

Closes #5422

### DIFF
--- a/src/core/biblio-db.js
+++ b/src/core/biblio-db.js
@@ -52,16 +52,21 @@ async function openIdb() {
   for (const storeName of [...ALLOWED_TYPES]) {
     const tx = db.transaction(storeName, "readwrite");
     const range = IDBKeyRange.lowerBound(now);
-    let result = await tx.store.openCursor(range);
-    while (result?.value) {
-      /** @type {StoredBiblioEntry} */
-      const entry = result.value;
-      if (entry.expires === undefined || entry.expires < now) {
-        await tx.store.delete(entry.id);
+    const sweep = async () => {
+      let result = await tx.store.openCursor(range);
+      while (result?.value) {
+        /** @type {StoredBiblioEntry} */
+        const entry = result.value;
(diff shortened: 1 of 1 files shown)


marcoscaceres:
Ok, yeah, this is much cleaner. I like it.

PR #5424, opened 2026-08-25
fix(tests): bind the headless server to a free port and cover launch time

marcoscaceres:
Three problems in the headless suite, all of which made it look flaky rather than broken.

The server was bound to a hardcoded port 8000 with no error handler, so a bind failure became an uncaught exception attributed to whichever spec happened to be running. Worse, when something else was already serving that port, the specs could pass while exercising that other server files instead of this checkout. That is the dangerous case, because it looks like a green run. It now binds port 0, reads back the assigned port, and a bind failure is reported against the suite with the real error.

The spec timeout was set to the same 30 seconds passed to respec2html as its processing budget. Since browser launch has its own separate budget of 120 seconds, jasmine could kill a child that was still legitimately launching. The spec timeout now covers launch plus processing plus slack.

The server was also never closed; there is now an afterAll for it.

I could not get a green local run to include here: this machine is currently refusing listen on every port including port 0, from every process context, in windows lasting minutes, which is why the failure mode mattered enough to fix. CI is the verification. What I can show is that the new failure output names the cause, "Suite error: listen EPERM ... at tests/headless.cjs:32", instead of blaming an unrelated spec.

### DIFF
--- a/tests/headless.cjs
+++ b/tests/headless.cjs
@@ -4,14 +4,43 @@ const { exec } = require("child_process");
 const http = require("http");
 const serveHandler = require("serve-handler");
 
-const port = 8000;
-const timeout = 30; // seconds
+// Budget respec2html gets for processing the document, passed through as --timeout.
+const processingTimeout = 30; // seconds
+
+// Each spec spawns respec2html, which launches a browser before processing starts.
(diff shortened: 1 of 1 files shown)


marcoscaceres (commented):
Approving... this allows running multiple sessions at the same time locally, which is super useful.

PR #5426, opened 2026-08-25
test: raise the karma jasmine spec timeout above the 5s default

### LINKED ISSUE #5425: Karma suites run under jasmine's 5s default, causing intermittent timeout failures
The karma suites run under jasmine's default 5 second per-spec timeout, and the integration suite renders whole ReSpec documents, some of which fetch over the network. Under CI load that budget is marginal, so specs fail intermittently on timing rather than on behaviour.

Evidence from three consecutive CI runs on one branch: 11 distinct specs failed with no repeats, every one tagged `slow: 5.001 secs`. The same commit both passed and failed. Main's own push run failed one spec in the same window with unrelated code in it. Locally the full integration suite passes 1138/0, so nothing is being masked.

Examples of specs that failed once each and then passed: "displays the Member Submission logo for Member submissions", "doesn't adds a canonical link for drafts", "describes editors and contributors", "has an array of errors and warnings", "doesn't lint non-rec-track docs with status CG-FINAL".

`tests/test-build.cjs` sets `DEFAULT_TIMEOUT_INTERVAL` to 60000 and `tests/headless.cjs` makes it configurable, but the karma configs never set it, so both the unit and integration suites inherit jasmine's 5000.

Raising the karma timeout is mitigation rather than a fix: it lowers the probability of a spurious failure without removing the network dependency. The root fix is local fixtures for the specs that fetch, which is already noted for the xref tests in the repo docs.

Related but distinct from #5424, which fixes the headless suite: a hardcoded port 8000 with no error handler, and a spec timeout smaller than the browser launch budget. This issue is the karma suites under jasmine's 5000 default. The `Headless Tests` check passed in all of the CI runs described above; the failures were in `test:integration`.

marcoscaceres:
The karma suites run under jasmine's default 5 second per-spec timeout. Integration specs render whole ReSpec documents, some of which fetch over the network, and under CI load they do not reliably finish inside that budget.

Over three consecutive CI runs on one branch, 11 distinct specs failed with no repeats, every one tagged `slow: 5.001 secs`, and the same commit both passed and failed. Main's own push run failed one spec in the same window with unrelated code. Locally the full integration suite passes 1138/0, so there is no defect being masked; the timeout is simply marginal.

Only `tests/test-build.cjs` (60000) and `tests/headless.cjs` set `DEFAULT_TIMEOUT_INTERVAL`; the karma configs never did, so both suites inherited 5000.

The accompanying spec asserts the value reaches the browser. A misspelled or misplaced karma `client` option is silently ignored, so without it the suites could quietly fall back to 5000 and the flakiness would return unnoticed.

This is mitigation rather than a root fix: it lowers the probability of a timeout, it does not remove the network dependency. The root fix is local fixtures for the specs that fetch, which is already noted for the xref tests.

Fixes #5425

### DIFF
--- a/tests/karma.conf.base.cjs
+++ b/tests/karma.conf.base.cjs
@@ -83,6 +83,13 @@ module.exports = config => {
     client: {
       // @ts-expect-error
       args: ["--grep", config.grep || ""],
+      jasmine: {
+        // Integration specs render whole ReSpec documents, some of which fetch
+        // over the network, so they do not reliably finish inside jasmine's 5s
+        // default when CI runners are loaded. A hung spec still fails, just
+        // later; browserNoActivityTimeout above remains the outer bound.
+        timeoutInterval: 15000,
+      },
     },
   };
 
--- a/tests/unit/karma-config-spec.js
+++ b/tests/unit/karma-config-spec.js
@@ -1,0 +1,11 @@
+"use strict";
+
+// Guards the karma-jasmine `client.jasmine.timeoutInterval` in
+// tests/karma.conf.base.cjs. A misspelled or misplaced karma client option is
+// silently ignored, so without this the suites would quietly fall back to
+// jasmine's 5s default and the flaky timeouts would return unnoticed.
+describe("Karma config", () => {
+  it("raises the jasmine spec timeout above the 5s default", () => {
+    expect(jasmine.DEFAULT_TIMEOUT_INTERVAL).toBe(15000);
+  });
+});


marcoscaceres:
Closing this in favour of fixing the tests, based on a measurement taken after it was opened.

Raising the timeout is mitigation: it lowers the probability of a spurious failure without removing the cause. The cause turns out to be measurable. Reading the Performance resource timeline out of the rendered iframe for two of the spec shapes that fail intermittently (seo-spec's "REC with group", headers-spec's Member Submission), every rendered W3C document makes 5 to 6 external requests to `www.w3.org`:

```
www.w3.org/StyleSheets/TR/2021/base.css
www.w3.org/StyleSheets/TR/2021/dark.css
www.w3.org/scripts/TR/2021/fixup.js
www.w3.org/StyleSheets/TR/2021/logos/W3C
www.w3.org/StyleSheets/TR/2021/logos/dark.svg
www.w3.org/StyleSheets/TR/2021/W3C-REC     (status-dependent)
```

Durations in that run were 80, 82, 84, 101 and 288ms for a single document. The data-fetching side is already handled: `seedGroupCache` in `tests/spec/respec-cache-helper.js` pre-seeds the Cache API from `tests/data/groups.json`, so group, biblio and xref lookups do not go out. It is the static assets that were never addressed, and they go out on every document render.

That also makes this PR awkward against the repo's own guidance, "Tests: Use local fixtures, never remote endpoints".

Details and the remaining work are in #5425. Two things from here worth keeping when the real fix is written: serving those assets locally rather than raising the budget, and asserting in a spec that whatever test-mode configuration is added actually reaches the browser, since a misspelled karma `client` option is silently ignored.

Caveat on the measurement, so it is not over-trusted: two spec shapes, one browser, one run. The per-document request count is measured; any suite-wide total is extrapolation.

PR #5428, opened 2026-08-25
fix(tests): run the geonovum specs against the geonovum bundle

### LINKED ISSUE #5409: Geonovum spec tests load the w3c build, so geonovum/style.js never runs
The Geonovum spec tests never load the Geonovum build, so `src/geonovum/style.js` never runs and the assertions pass against W3C behaviour instead.

`makeRSDoc` in `tests/spec/SpecHelper.js` picks the bundle from `opts.profile`, defaulting to `"w3c"` and loading `/base/builds/respec-${profile}.js`. `makeStandardAomOps` sets `profile: "aom"`, but `makeStandardGeoOps` returns only `{ body, config }` with no `profile` key, so every test using it silently loads `builds/respec-w3c.js`. The two helpers sit next to each other and are otherwise the same shape, so this reads as an omission rather than a decision.

Two tests in `tests/spec/geonovum/style-spec.js` show the effect. "should include 'fixup.js'" asserts the script src is `https://www.w3.org/scripts/TR/2021/fixup.js`, but Geonovum attaches the 2016 script and 2021 is the W3C one, so that assertion can only hold because the wrong bundle is loaded. "shouldn't include fixup.js when noToc is set" queries the 2016 URL, which never appears in a W3C rendered document, so its `expect(elem).toBeNull()` is true regardless of whether the code under test works. Neither test has been able to fail.

Found by mutation while working on something unrelated: removing the config normalisation from all four profile defaults turned one test red in W3C, one in AOM and one in DINI, while the Geonovum suite stayed fully green.

Adding `profile: "geonovum"` to `makeStandardGeoOps` is the obvious change, and I tried it rather than guessing. It does not come out green: the suite goes to 6 passing and 1 failing, and the failure is `TypeError: Cannot read properties of null (reading 'src')` rather than a tidy assertion failure. Once the right bundle loads, the 2021 script is not in the document and the test dereferences the null result. So that test does not merely encode W3C behaviour, it depends on the W3C bundle being loaded or it throws. Both tests need rewriting against what Geonovum actually emits.

Only `tests/spec/geonovum/style-spec.js` uses this helper, so the change is contained.

marcoscaceres:
The Geonovum specs never passed a profile, so makeRSDoc fell back to its "w3c" default and loaded the W3C bundle, meaning src/geonovum/style.js never ran and the fixup assertion was quietly passing against the W3C 2021 script rather than the 2016 one Geonovum attaches. This also narrows the stylesheet query to rel=stylesheet, because the preload resource hint for base.css carries the same href and made that assertion impossible to fail for specStatus "base". Closes #5409.

### DIFF
--- a/tests/spec/SpecHelper.js
+++ b/tests/spec/SpecHelper.js
@@ -272,6 +272,7 @@ export function makeStandardGeoOps(config = {}, body = makeDefaultBody()) {
   return {
     body,
     config: { ...makeBasicConfig("geonovum"), ...config },
+    profile: "geonovum",
   };
 }
 
--- a/tests/spec/geonovum/style-spec.js
+++ b/tests/spec/geonovum/style-spec.js
@@ -20,10 +20,10 @@ async function loadWithStatus(status, expectedURL) {
     profile: "geonovum",
   };
   const doc = await makeRSDoc(ops);
-  const query = `link[href^='${testedURL}']`;
-  const elem = doc.querySelector(query);
-  expect(elem).toBeTruthy();
-  expect(elem.href).toBe(testedURL);
+  // rel=stylesheet: for specStatus "base" the preload resource hint has the
+  // same href, and querySelector would return that hint instead.
+  const query = `link[rel=stylesheet][href='${testedURL}']`;
+  expect(doc.querySelector(query)).toBeTruthy();
   return doc;
 }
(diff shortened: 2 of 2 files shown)


marcoscaceres (commented):
Approving... tiny change.

PR #5430, opened 2026-08-25
chore: remove the geonovum profile

marcoscaceres:
Geonovum appears to have moved off this profile. Their current document template, Geonovum/NL-ReSpec-template, loads a base build from Logius (respec-nlgov.js) together with their own config scripts and does not reference this profile; the older template that did is one they themselves describe as "Oude, gearchiveerde". Older Geonovum documents still reference respec-geonovum in about a dozen of their repositories, and the one I opened points at their own copy on tools.geostandaarden.nl rather than at ours. That copy reports ReSpec 24.5.2 and still contains the geonovum/headers, informative, inlines, leafletfigures and seo modules this profile has since lost, so it is self-contained and unaffected by this removal.

Meanwhile the in-tree profile has rotted to the point of being unusable. It imports no geonovum/headers.js, so a document rendered with it has no title h1, no div.head and no Abstract or SoTD heading, and ReSpec reports zero errors while producing that. It went unnoticed because the Geonovum tests were loading the W3C bundle, which was #5409, fixed in #5428.

This removes the profile source, its tests and its build wiring, but deliberately leaves the prebuilt `builds/respec-geonovum.js` in place. CI rejects any change under `builds/` in a pull request, and the release tooling only ever writes bundles rather than deleting them, so the artifact stays and `https://speced.github.io/respec/builds/respec-geonovum.js` keeps serving. Anything already pointing at that URL therefore continues to render, which seemed the safer default given I have no data on whether anything still requests it. Removing the stale artifact can be a separate decision.

I have asked Geonovum directly whether they still need it: Geonovum/NL-ReSpec-template#48.

Separately, and not addressed here: a profile finishing without a title while reporting no error looks like a core problem rather than a Geonovum-specific one. dini and aom each import their own headers module today, so they are fine, but nothing in the pipeline would catch it if one of them lost that import the same way.


sidvishnoi (approved):

marcoscaceres:
Going to merge and we can revert if we need to.

Issue #5200, opened 2026-04-15
Theme toggle fails to inject in FireFox


JJdeGroot:
## Description of problem

**URL to affected spec or repo:**

See attached `index.html` file

**What happened (e.g., it crashed)?:**

The theme toggle does not display in FireFox.

**Expected behavior (e.g., it shouldn't crash):**

I expect to see the theme toggle.

**Optional, steps to reproduce:**

1.  Open [index.html](https://github.com/user-attachments/files/26747317/index.html) in FireFox

<img width="1745" height="460" alt="Image" src="https://github.com/user-attachments/assets/26aca394-9cf7-4419-8bd3-925d50c8f87e" />

Multiple errors are logged:
- Cross origin request blocked to load `respect-highlight.js`
- Error in `updateTheme` part of `fixup.js` line 374

Note that no error are logged when the webpage is accessed via: https://w3c.github.io/matf/

I would expect that respect-highlight.js should also be loadable from localhost.

And the updateTheme could potentially also work from localhost?
It does work in Chrome.

<img width="1503" height="250" alt="Image" src="https://github.com/user-attachments/assets/d5bcb9c5-a387-4c0d-9023-220ccde6848d" />

kfranqueiro:
@marcoscaceres as far as I can tell, #5408 does not fully resolve the issue and may be unrelated:
- I was unable to reproduce the issue even when explicitly using a version of respec-w3c.js built from 37.3.2 locally (the document properly includes the `meta` tag that should be opting it in to dark mode support already, and the toggle works for me in Firefox)
- The `respec-highlight.js` request still fails with the same error when running ReSpec client-side on the matf repository (and also others), which might be a red herring as far as the main issue is concerned

More importantly, #5408 has potential to cause accessibility regressions for anything containing custom styles that haven't accounted for dark mode (read: it already caused them in at least one repo I help maintain), and absolutely feels like a breaking change as a result. I would argue that it should be reverted, and that a change that major should be carefully coordinated. (The documentation was also not updated, and still suggests that dark mode support is opt-in.)

(Sorry for edits, I tried to flesh out and organize my thoughts better)

marcoscaceres:
Thanks, @kfranqueiro. I’ll take another look. Ping you to confirm the fix before merging. 🙏

kfranqueiro:
I have further attempted to test this as close to the original conditions of the report as possible:

- Rewound `matf` to `5ae42ae0` (April 15, 2026)
[...]

PR #5436, opened 2026-08-27
fix(w3c/style): put the dark stylesheet last on export so the theme toggle works

marcoscaceres:
Refs #5200, and fixes a regression from #5408. On an exported spec that never opted into dark mode the theme toggle did nothing: `styleMover` was subscribed for the maturity stylesheet unconditionally but for `dark.css` only when the spec had opted in, so on save the maturity sheet moved past it and won the cascade. The dark sheet now moves for every spec, and its `disabled` and `media` state is restored on export, so a document saved while the reader had dark selected no longer ships a permanently dark spec.

Two existing tests asserted the old order and were rewritten.

This makes dark mode reachable where it previously did nothing, which exposes that several of ReSpec's own stylesheets are not dark-ready yet. Those are separate PRs and no release should go out before they land.

Proof: the same source exported with the base build and with this branch. Without the fix `dark.css` lands before `W3C-ED` and enabling it changes nothing; with the fix it lands last and the toggle works. Marcos then confirmed it manually on exported specs: https://github.com/speced/respec/pull/5436#issuecomment-5437023397

### DIFF
--- a/src/w3c/style.js
+++ b/src/w3c/style.js
@@ -101,6 +101,32 @@ function styleMover(linkURL) {
 }
 
 /**
+ * Put the dark stylesheet link back into the state ReSpec chose, because by save
+ * time fixup.js has been driving it live: it sets `media = ""` and toggles
+ * `disabled` to follow the reader's theme, and both are reflected content
+ * attributes. Saving a document while the reader had dark selected would
+ * otherwise export an enabled, unconditional dark sheet, which now sits last in
+ * `head` and so wins the cascade, rendering dark for everyone including readers
+ * without scripting.
+ *
+ * @param {Document} exportDoc
+ * @param {boolean} isDark whether the spec itself opted into dark mode
+ */
+function restoreDarkLinkState(exportDoc, isDark) {
+  const darkLink = exportDoc.querySelector(
+    `head link[rel~="stylesheet"][href="${getStyleUrl("dark.css")}"]`
+  );
+  if (!darkLink) return;
+  if (isDark) {
+    darkLink.setAttribute("media", "(prefers-color-scheme: dark)");
+    darkLink.removeAttribute("disabled");
+  } else {
+    darkLink.removeAttribute("media");
+    darkLink.setAttribute("disabled", "");
+  }
+}
+
+/**
  * @param {Conf} conf
  */
 export function run(conf) {
@@ -142,11 +168,17 @@ export function run(conf) {
   if (!isDark) darkLink.setAttribute("disabled", "");
   document.head.appendChild(darkLink);
   // The dark sheet has to end up last. It and base.css set the same `:root`
-  // custom properties at equal specificity, and the mover above puts base.css
-  // at the end of `head` on export, so without this the toggle enables a sheet
-  // that then loses the cascade and the reader sees nothing change. Also what
-  // W3C Pub Rules require for dark-mode specs.
-  sub("beforesave", styleMover(darkModeStyleURL));
+  // custom properties at equal specificity, and the mover above puts the
+  // maturity-level sheet, which `@import`s base.css, at the end of `head` on
+  // export. Without this the toggle enables a sheet that then loses the cascade
+  // and the reader sees nothing change. Also what W3C Pub Rules require.
+  sub(
+    "beforesave",
+    /** @param {Document} exportDoc */ exportDoc => {
+      styleMover(darkModeStyleURL)(exportDoc);
+      restoreDarkLinkState(exportDoc, isDark);
+    }
+  );
 }
 
 /** @param {Conf} conf */
--- a/src/w3c/style.js
+++ b/src/w3c/style.js
@@ -158,7 +158,7 @@ export function run(conf) {
   }
   // W3C's fixup.js only injects the light/dark/auto toggle when it can find the
   // dark stylesheet link, and it drives that link with `.disabled`. Light-only
-  // specs never had the link, so the toggle silently never appeared (#5200). Add
+  // specs never had the link, so the toggle silently never appeared. Add
   // it either way, switched off when the spec has not opted into dark mode.
   const darkModeStyleURL = getStyleUrl("dark.css");
   const isDark = colorScheme.content.includes("dark");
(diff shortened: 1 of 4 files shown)


marcoscaceres:
Confirmed manually locally that this works as expected on exported specs.

deniak:
@marcoscaceres I understand that PR addresses the issue where the dark mode isn't working as expected of exported documents such as https://w3c.github.io/wcag/guidelines/.

That said, there seems to be a problem with the respec source documents. I tried that PR with the guidelines source document (see https://nix.w3.org/~denis/respec/guidelines/) and while it's working fine on firefox, chrome 151 renders the spec sometimes in light mode and some other times in dark mode (I have the toggle set to auto). There may be some kind of race condition...

marcoscaceres:
Yep, that’s being addressed an a follow up. Coming in after this PR.

kfranqueiro:
I observe the same issue as Denis, specifically when ReSpec is run client-side (not when viewing an already-processed document). Amusingly, it is somehow _always_ wrong:

* When my stored setting is dark or auto (with my system set to dark), it always initially shows light mode
* When my stored setting is light mode, it always shows dark mode.

> Yep, that’s being addressed an a follow up. Coming in after this PR.

Does that branch exist yet so we can get an idea of its complexity? Otherwise it doesn't seem to make much sense to approve this PR that is visibly broken...

Moreover, how many more upcoming PRs will there be? And why are we addressing a regression caused by _a PR that did not fix anything_ by adding more complexity (and maybe more regressions), rather than by reverting the one PR in question?

> This makes dark mode reachable where it previously did nothing, which exposes that several of ReSpec's own stylesheets are not dark-ready yet. Those are separate PRs and no release should go out before they land.

By this logic, 37.3.3+ should not have released with #5408 to begin with. It is also unclear to me, did these issues not already exist in earlier versions when opted into dark mode?

So far, 11 documents have been published to w3.org/TR/ since Monday that exhibit the broken theme toggle (out of 32 total). That number will only increase until it is resolved. This is why I [immediately advocated for reverting #5408](https://github.com/speced/respec/issues/5200#issuecomment-5401354357), which could have limited the number of affected publications to 1 or 2.

There is also still the core issue of #5408 having spontaneously changed default behavior in a patch release, and #5437 plans to move further in that direction. A change this major should only happen _after_ all of the following:

* dark mode styles have been fully vetted, including for accessible color contrast
* advance notice is given to spec editors to adjust custom styles accordingly
  * Although if we give editors this chance, they might as well opt in by themselves via the meta tag to begin with in order to test the styles, which would seemingly obviate any need for these changes?
* docs are updated (currently, respec.org still documents dark mode support as opt-in via `meta` tag)

marcoscaceres:
@deniak @kfranqueiro The race is real, it reproduces on `main`, and it is not caused by
this PR. The fix is #5439: it sets `disabled` on the dark stylesheet link before inserting
the link, rather than assigning the property on the statement after. A write that lands
mid-fetch leaves Chrome applying the sheet anyway and then dropping it on the next write,
so the theme you get depends on whether `dark.css` has arrived. Firefox always has it in
time, which is why it looked Chrome-only. Kenneth, on complexity: the change in
`style.js` is three lines in and six out, and the rest of that PR is the regression test.

On whether the dark styles were already broken before #5408: you are right, they were.
The dark media queries in `var.css.js`, `algorithms.css.js` and `cddl.css.js` predate it,
so any spec that opted in through the meta tag has had them all along. #5408 did not
create those defects, it made them reachable for everyone. That cuts both ways: it is an
argument for fixing them, and it also means reverting #5408 would not fix them for the
specs that already opted in.

The docs point is correct and unfixed too: respec.org still documents dark mode as
opt-in via the meta tag.

On reverting #5408: I would rather fix these. A revert does stop the race, by taking the
link back out again, but it also restores #5200, where the toggle silently never appears
on a light-only spec. And it does nothing for the documents already published: those are
static, so they need republishing either way. That leaves the revert case resting on
limiting future publications, which #5436 and #5439 do as well, without giving up the
toggle.

Your three preconditions are reasonable and I am not going to pretend they were met. The
styles are not fully vetted, editors got no notice, and the docs are stale. I would rather
work through them than revert: the docs I can fix now, the component styles need their own
PR, and #5437 waits until both are done.

deniak:
> @deniak @kfranqueiro The race is real, it reproduces on `main`, and it is not caused by this PR. The fix is #5439: it sets `disabled` on the dark stylesheet link before inserting the link, rather than assigning the property on the statement after. A write that lands mid-fetch leaves Chrome applying the sheet anyway and then dropping it on the next write, so the theme you get depends on whether `dark.css` has arrived. Firefox always has it in time, which is why it looked Chrome-only. Kenneth, on complexity: the change in `style.js` is three lines in and six out, and the rest of that PR is the regression test.
> 
> On whether the dark styles were already broken before #5408: you are right, they were. The dark media queries in `var.css.js`, `algorithms.css.js` and `cddl.css.js` predate it, so any spec that opted in through the meta tag has had them all along. #5408 did not create those defects, it made them reachable for everyone. That cuts both ways: it is an argument for fixing them, and it also means reverting #5408 would not fix them for the specs that already opted in.
> 
> The docs point is correct and unfixed too: respec.org still documents dark mode as opt-in via the meta tag.
> 
> On reverting #5408: I would rather fix these. A revert does stop the race, by taking the link back out again, but it also restores #5200, where the toggle silently never appears on a light-only spec. And it does nothing for the documents already published: those are static, so they need republishing either way. That leaves the revert case resting on limiting future publications, which #5436 and #5439 do as well, without giving up the toggle.
> 

@kfranqueiro and I have reviewed the PRs and confirm #5439 fixes the issue with chrome. Can you merge #5436 and #5439 so we don't end up with a conflict? It will also be easier to follow the changes if there's only one PR.

> Your three preconditions are reasonable and I am not going to pretend they were met. The styles are not fully vetted, editors got no notice, and the docs are stale. I would rather work through them than revert: the docs I can fix now, the component styles need their own PR, and #5437 waits until both are done.

Thank you for addressing my request about allowing the editors to opt-in. I initially thinking it would be easier to keep the old behavior (which is adding the `<meta name="color-scheme" content="light dark”>` to enable dark mode). This way we wouldn't need to edit the current specs that are not compatible with dark mode.

I'm willing to contact the editors to see how they feel about enabling dark mode support for all specs and point them to https://github.com/speced/respec/pull/5437 that adds the new config property.
Enabling dark mode by default would be great and I should probably have pushed more fixing the specs that are compatible yet.

kfranqueiro:
> On reverting #5408: I would rather fix these. A revert does stop the race, by taking the link back out again, but it also restores #5200, where the toggle silently never appears on a light-only spec.

This statement does not accurately describe #5200. The document that it was reported against was _already opted into dark mode using the meta tag_, and the report was that the theme controls did render, but didn't work - which I [could never reproduce](https://github.com/speced/respec/issues/5200#issuecomment-5415092507), even when using the versions of ReSpec and `matf` that were current at the time of the report.

As far as I could tell, there was nothing for #5408 to fix, which is why I suggested reverting it instead of adding more. (I realize there may be other motivations for moving in that direction.)

marcoscaceres:
Specref outage is also blocking us 😢 all the bots are sad.

PR #5437, opened 2026-08-27
feat(w3c/style): let an editor turn dark mode off with darkMode: false

marcoscaceres:
An editor can now set `darkMode: false` and ReSpec stops offering dark mode for that spec: it emits no dark stylesheet, so W3C's fixup.js finds nothing to toggle, and it removes its own `@media (prefers-color-scheme: dark)` blocks so a reader whose operating system prefers dark does not land on a light page with dark ReSpec components in it. This matches Bikeshed's `Dark Mode` metadata, including that second half. Only an explicit `false` disables anything.

The option is transitional and the documentation says so: custom CSS that is not dark-aware is the thing to fix.

Removing those blocks goes through the browser's own CSS parser, `CSSStyleSheet.replaceSync`, rather than scanning for braces. A hand-rolled scanner got four inputs wrong, each of which silently deleted valid CSS: an unterminated block, a dark `@media` opener inside a comment, a brace inside a string, and `not (prefers-color-scheme: dark)`, which is the one construct that forces a page light. The parser handles all four. Two consequences: a stylesheet that has something removed comes back re-serialized, so its formatting is normalized and comments are gone, and one with nothing to remove is returned untouched.

Nothing reaches the author's own CSS. The strip only visits stylesheets that went through `core/insert-style`, which sees only strings ReSpec passed it.

Gemini reviewed the module and its tests and found the unterminated-block case, plus that `disableDarkStyles` had no test at all. Both are fixed here.

Proof: `BROWSERS=ChromeHeadless npx karma start tests/unit/karma.conf.cjs --single-run` covers the strip through the public API, 87 specs. Each guard is mutation-proven rather than merely green: neutering the `not` check fails "keeps `not (prefers-color-scheme: dark)`, which forces a page light", and making the strip a no-op fails six specs including the `disableDarkStyles` lifecycle. The suite was run five times to confirm it is stable under jasmine's random ordering, because `disableDarkStyles` sets a realm-wide flag that nothing resets.

### DIFF
--- a/src/core/insert-style.js
+++ b/src/core/insert-style.js
@@ -2,12 +2,13 @@
 // Module core/insert-style
 // One place where ReSpec adds its own `<style>` elements to the head.
 //
-// Going through here is also what lets `disableDarkStyles()` rewrite ReSpec's CSS without
-// ever reading the spec author's: this module only sees strings ReSpec handed it.
+// Add ReSpec's stylesheets through here and nowhere else: `disableDarkStyles()` rewrites what
+// this module inserted, so anything added directly is beyond its reach, and a spec author's
+// own `<style>` stays out of its reach for the same reason.
 
 export const name = "core/insert-style";
 
-/** The elements this module created, so `disableDarkStyles` can revisit them. */
+/** The stylesheets `disableDarkStyles()` rewrites. */
 const inserted = new Set();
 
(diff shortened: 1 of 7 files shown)


marcoscaceres:
I'm pretty happy where this has ended up and seems to be working as expected now.

PR #5438, opened 2026-08-27
chore: drop geonovum from the copilot setup steps and instructions

marcoscaceres:
Removing the geonovum profile in #5430 left `pnpm build:geonovum` in the Copilot setup steps, so every Copilot code review since then has failed setup and run without the aom and dini bundles or a browser. This drops it from the workflow and from the agent instructions, which still told agents to build that profile and to register new modules in a profile that no longer exists.


marcoscaceres:
Was breaking poor copilot.

PR #5448, opened 2026-09-02
fix(core/biblio): fall back to respec.org when Specref is unreachable

marcoscaceres:
Specref has been offline since 30 August, so references stop resolving. core/biblio now tries it first and falls back to respec.org/bibrefs, so the liveness spec fails until that endpoint is deployed.

Written with Claude: the fallback and its tests are generated logic.


marcoscaceres:
Confirmed fix locally with new endpoint.

PR #5449, opened 2026-09-02
fix(tests): stop a specref outage aborting the whole integration run

marcoscaceres:
`tests/spec/core/biblio-spec.js` probes api.specref.org in a `beforeAll`, but an unreachable host rejects rather than returning a not-ok response, so the rejection aborts the whole integration run instead of just this suite. That defeats the intent already written a few lines below, "This prevents Jasmine from taking down the whole test suite if SpecRef is down", so this wraps the probe in a try/catch and lets the existing `specRefOk` checks report the outage. With specref currently down, the suite goes from 462 completed to 1066, and two consecutive runs now report identical counts instead of varying between 74 and 671.


marcoscaceres:
@copilot resolve the merge conflicts in this pull request

marcoscaceres:
Closing this as superseded, and as it stands it is also mislabeled.

The try/catch this PR describes is no longer in the diff. The service worker work landed on main and rewrote the same beforeAll, replacing the specref fetch with a DOM read, so the conflict resolution here took main's version and dropped this one. What remains is an unrelated path-traversal guard in `tests/darkmode-race.cjs` that arrived with the merge.

Main's approach is the better fix, measured rather than assumed: two consecutive full integration runs on main report 1127 completed and nothing failing, where this branch reported 1127 run with 61 failing. Removing the network dependency beats surviving its absence.

PR #5460, opened 2026-09-11
fix(w3c): Correct capitalization of members

tamsinewing555:
Following W3C style, only capitalize when it is part of a proper name, and not when it is in the plural form.

See: https://w3c.github.io/style-guide/#capitalization


marcoscaceres (approved):

deniak:
Marking the PR as draft because pubrules needs to be updated accordingly

PR #5463, opened 2026-09-11
fix(w3c): Correct capitalization of representatives + a couple more

tamsinewing555:
Following W3C style, only capitalize when it is part of a proper name, and not when it is in the plural form.

See: https://w3c.github.io/style-guide/#capitalization

### DIFF
--- a/src/w3c/templates/sotd.js
+++ b/src/w3c/templates/sotd.js
@@ -201,7 +201,7 @@ function renderNotRec(conf) {
   </p>`;
   if (conf.specStatus === "DISC") {
     updatePolicy = html`<p>
-      Publication as a Discontinued Draft implies that this document is no
+      Publication as a W3C Discontinued Draft implies that this document is no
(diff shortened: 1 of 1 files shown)


marcoscaceres (approved):

marcoscaceres:
@tamsinewing555, @deniak will these still be ok with Pub Rules?

deniak:
> @tamsinewing555, @deniak will these still be ok with Pub Rules?

No, pubrules (https://github.com/w3c/specberus/blob/main/lib/rules-generic-sections.ts#L108 and https://github.com/w3c/specberus/blob/main/lib/rules/sotd/stability.ts#L25) will need to be updated this get deployed, which will also affect the specs generated with bikeshed. @kfranqueiro Can you please coordinate the changes with @tamsinewing555 ?

deniak:
Marking the PR as draft because pubrules needs to be updated accordingly

PR #5469, opened 2026-09-16
fix(w3c): flag a /TR/-track specStatus rendered off w3.org/TR

### LINKED ISSUE #5468: W3C specStatus should raise an error when not on w3.org/TR/

A common mistake that editors make is to set `specStatus=WD` or `specStatus=DNOTE`
on a document that ends up published on github pages.
Not only is this a "lie" (the doc on github pages is not, per the process, a working draft or a draft note),
but it has very practical consequences:
Respec generates a "This version" URL on w3.org/TR/ with the current date, which is broken.

I would like Respec to signal a warning or an error when the current URL is not in w3.org/TR/ and the specStatus indicates a W3C status different from ED.

**Additional context**

How are you currently working around this limitation?

I can try.

pchampin:
fix #5468

- src/w3c/headers.js: new exported pure function isOffTR(specStatus, href) returns true when specStatus is one of the formal /TR/-track statuses (trStatus — Notes, Rec-track, Registry-track) but href isn't under https://www.w3.org/TR/ (or the w3.org alias). A wrapper, validateRenderingLocation, calls this with document.location.href and raises a pill error (via showError) unless: the current frame isn't the top-level browsing context (so test harnesses and embedded preview panes are unaffected), or specStatus was itself overridden via a ?specStatus= query parameter (the existing local-preview mechanism from core/override-configuration.js).

- Statuses like ED, unofficial, base, CG/BG reports, etc. are unaffected — only the statuses that actually generate a w3.org/TR "this version" URL are checked, matching the bug the issue described (a broken "This version" link when e.g. specStatus: WD is deployed on GitHub Pages).

- Added tests/unit/w3c/headers-spec.js with unit tests directly exercising isOffTR (fast path, no DOM/iframe needed, since the integration harness always runs inside an iframe where the new check intentionally no-ops).

- Verified: pnpm lint clean, full unit suite (93 tests) and integration suite (1160 tests) pass with zero failures, and confirmed no example/headless fixtures use a status that would trip the new check.


src/w3c/headers.js
```diff
@@ -657,6 +659,55 @@ export async function run(conf) {
...
+ * to live elsewhere, e.g. on GitHub Pages; claiming any other status while
+ * hosted off /TR/ produces a broken "this version" URL.
+ * @param {string} specStatus
+ * @param {string} href window.location.href of the document being rendered
+ */
+export function isOffTR(specStatus, href) {
+  if (!trStatus.includes(specStatus)) {
+    return false;
+  }
+  let url;
```
marcoscaceres, line 675: Can probably use `URL.canParse(href)` to cut down the five lines...

src/w3c/headers.js
```diff
@@ -657,6 +659,55 @@ export async function run(conf) {
...
+  try {
+    url = new URL(href);
+  } catch {
+    return false;
+  }
+  const isOnTR =
```
marcoscaceres, line 681: This check seems a bit un-elegant to be honest... my (limited) testing seems to suggest that it always redirects to https://www.w3.org/TR/, so maybe always just checks that it `startsWith()`? Or is there cases where https://w3.org/TR/ can occur?

src/w3c/headers.js
```diff
@@ -657,6 +659,55 @@ export async function run(conf) {
...
+/**
+ * True when a document carrying a formal W3C status (a Note, or one on the
+ * Recommendation or Registry track) is rendered somewhere other than
+ * https://www.w3.org/TR/. Only an Editor's Draft (specStatus "ED") is meant
+ * to live elsewhere, e.g. on GitHub Pages; claiming any other status while
+ * hosted off /TR/ produces a broken "this version" URL.
+ * @param {string} specStatus
+ * @param {string} href window.location.href of the document being rendered
+ */
+export function isOffTR(specStatus, href) {
```
marcoscaceres, line 671: exporting this just to test it doesn't seem great, to be honest... what would be better is just create any document (like the rest of the tests do)... they all run on localhost and will violate this policy.

PR #5476, opened 2026-09-22
fix(respec2html): improve logging formatting

sidvishnoi:
1. Remove underlines from inline-code in logs, by turning them into backtick-wrapped cyan color.
2. Indent multiple-lines in message, particularly in `hint`.

| Before | After | 
| --- | --- |
| image | image |


tools/respec2html.js
```diff
@@ -130,6 +130,13 @@ class Logger {
       (!!rsError.cause?.stack || rsError.plugin === "unknown")
     );
   }
+
+  /**
+   * @param {string} text
+   */
+  _reindent(text, padding = 0) {
```
marcoscaceres, line 137: Nit: can we private this instead of `_`?

marcoscaceres (approved):
looks good. Just a nit.

tools/respec2html.js
sidvishnoi: It's JS, not TypeScript.

sidvishnoi: Do you mean `@private`? That won't help much though.

